PR #1, opened 2021-01-13
Walking Skeleton

markspanbroek:
This is an exploration of what the components from the Dagger PoC roadmap would look like when put together. It is a [walking skeleton](https://codeclimate.com/blog/kickstart-your-next-project-with-a-walking-skeleton/), where each component is stripped down to the absolute bare minimum, yet they all work together to fulfill a (very minimal) requirement, which is automatically tested.

Components:
- [ ] DHT
- [x] Merkle DAG
- [x] Chunking
- [x] Repo
- [ ] BitSwap


dagger.nim
```diff
@@ -0,0 +1,22 @@
+import pkg/chronos
+
+type
+  DaggerPeer* = ref object
+
+proc newDaggerPeer*: DaggerPeer =
```
dryajov, line 6: I would keep all the terminology as IPFS in this repo.

It would be helpful to mention that Dagger is an umbrella name for a set of components, one of them is a storage engine, which is based on IPFS, other components such as the incentivization(?) or long term storage system(?), etc, will have their own names.

dagger.nim
```diff
@@ -0,0 +1,22 @@
...
+  DaggerPeer()
+
+proc listen*(peer: DaggerPeer, address: TransportAddress) =
```
dryajov, line 9: We should use the same or as close as possible of an API to existing IPFS implementations - https://github.com/ipfs/js-ipfs/tree/master/docs/core-api.

dagger.nim
markspanbroek: I'll address this in a later commit.

dagger.nim
markspanbroek: I'll address this in a later commit.

dagger.nim
markspanbroek: Should be resolved with the latest couple of commits

dagger.nim
markspanbroek: Should be resolved with the latest couple of commits

ipfs.nimble
```diff
@@ -0,0 +1,10 @@
+version = "0.1.0"
+author = "Dagger Team"
+description = "Decentralized storage in Nim"
+license = "MIT"
+
+requires "nim >= 1.4.2 & < 2.0.0"
```
dryajov, line 6: I think for now, we need to support v1.2.6 at the very least.

ipfs/bitswap/protocol.nim
```diff
@@ -0,0 +1,33 @@
...
+
+export stream except readLoop
+
+const Codec = "/ipfs/bitswap/1.2.0"
+
+type
+  BitswapProtocol* = ref object of LPProtocol
+    connections: AsyncQueue[BitswapStream]
+
+proc new*(t: type BitswapProtocol): BitswapProtocol =
```
dryajov, line 15: by convention, this is usually done in the `init` method that gets called from the constructor.

ipfs/bitswap/stream.nim
```diff
@@ -0,0 +1,38 @@
...
+import pkg/protobuf_serialization
+import pkg/libp2p/stream/connection
+import ./messages
+
+export messages
+
+const MaxMessageSize = 8 * 1024 * 1024
+
+type
+  BitswapStream* = ref object
```
dryajov, line 11: I'm not sure about the need of this type. I think we can get by without it and instead add an RPC handler similar to how its done in `PubSub` - https://github.com/status-im/nim-libp2p/blob/master/libp2p/protocols/pubsub/floodsub.nim#L83-L128. 

It's probably also usefull to addopt a similar structure to pubsub in general, with a "bitswap peer" type that handles each peer individually and has it's own read/write mechanism similar to https://github.com/status-im/nim-libp2p/blob/master/libp2p/protocols/pubsub/pubsubpeer.nim. Note how the peer handles its own connection and has a read loop that parses messages and then invokes the main protocol handler.

ipfs/bitswap/protocol.nim
```diff
@@ -0,0 +1,33 @@
...
+  let connections = newAsyncQueue[BitswapStream](1)
+  proc handle(connection: Connection, proto: string) {.async.} =
+    let stream = BitswapStream.new(connection)
+    await connections.put(stream)
+    await stream.readLoop()
+  BitswapProtocol(connections: connections, codecs: @[Codec], handler: handle)
+
+proc dial*(switch: Switch,
```
dryajov, line 23: It's probably better to use peer event handler here similar to - https://github.com/status-im/nim-libp2p/blob/master/libp2p/protocols/pubsub/pubsub.nim#L538-L545

This will allow to automatically register/unretister a peer when it's connected.

ipfs/ipfsobject.nim
```diff
@@ -0,0 +1,10 @@
+import pkg/libp2p
+
+type
+  IpfsObject* = object
```
dryajov, line 4: Lets get rid of the IPFS naming.

dryajov (changes requested):
I think it would be very helpful to read the https://github.com/ipfs/go-bitswap/blob/master/docs/how-bitswap-works.md, it outlines the algoritm followed by the existing implementations and it's a good starting point to understand the overall flow.

markspanbroek:
Thanks for the review @dryajov, I had a couple more commits lined up that I'm going to push first, and then I'll address your comments.

ipfs.nimble
markspanbroek: I went with the latest Nim, because it's a proof of concept for a new library, and didn't want to take the (albeit small) additional burden of maintaining compatibility with older Nim versions. If however, you find 1.2.6 support important, then I'll add additional CI jobs for this version.

ipfs/bitswap/protocol.nim
markspanbroek: I went with the naming convention for constructors from [the suggested style guide](https://github.com/status-im/nimbus-eth2/blob/a4941105511eeebfc13fad0281457afb86b79fb5/docs/the_style_book/src/02_formatting.md#naming-conventions).

ipfs/ipfsobject.nim
markspanbroek: Calling it simply `Object` felt weird, but should work. Or would you prefer to call it `DaggerObject`?

ipfs.nimble
dryajov: Some of our deps might not be compatible with 1.4.2 or later, libp2p included because our main target it 1.2.6 right now.

ipfs/bitswap/protocol.nim
dryajov: That's not what I meant. The protocol base class defines an init method where the setup of the protocol usually happens, the constructor calls the init method. This is done so that protocols can be inherited from, without having to manually redefine the protocol.

ipfs/ipfsobject.nim
dryajov: Yes, as we agreed, let's drop the ipfs moniker to avoid confusion.

markspanbroek:
As discussed with @dryajov: we'll merge this once the tests are green, and address the comments in later PRs.

ipfs.nimble
markspanbroek: I added support for Nim 1.2.6 in #4.

ipfs/ipfsobject.nim
markspanbroek: Addressed in #6.

PR #5, opened 2021-02-08
Chunking: fix erroneous cast from bytes to string

markspanbroek:
Fixes intermittent test failure on Windows, such as:
https://github.com/status-im/nim-dagger/runs/1855282285?check_suite_focus=true


ipfs/chunking.nim
```diff
@@ -7,4 +7,5 @@ proc createObject*(file: File): IpfsObject =
   IpfsObject(data: cast[seq[byte]](contents))
```
arnetheduck, line 7: there's generally `stew.toBytes` - casting between string and seq should be avoided as it results in invalid strings

PR #24, opened 2021-08-27
Treehash

dryajov:
This adds the treehash algorithm to produce a checksum/merkle root from a stream of chunks


dryajov:
closed as part of #25

PR #26, opened 2021-08-31
initial commit of the Shacham BLS-based and RSA-based public schemes

cskiraly:
Minimal working version with lots of error checks, tests and corrections
still needed.
- using Bearssl RSA code through libp2p
- with selectable BigInt library for experimentation


dagger/storageproofs/bls.nim
```diff
@@ -0,0 +1,257 @@
...
+  return (x, scalar)
+
+proc posKeygen(): (blst_p2, blst_scalar) =
+  rndP2()
+
+proc keygen*(): (PublicKey, SecretKey) =
+  var pk: PublicKey
+  var sk: SecretKey
+  var ikm: array[32, byte]
+  var RNG = initRand(0xFACADE)
```
dryajov, line 116: Just a heads up (and it's probably fine for a first pass PoC), but the std random functions aren't cryptographically secure - we're using `bearssl` as our crypto secure random generator instead. There is `dagger/rng.nim`, which exposes a couple of convenience functions around the bearssl HMAC primitive, preferably, we should be using that and there should be a few example uses in this codebase already.

Saddly, we haven't standardized it across all ouf our projects yet, so you'll fine similar reimplementations across other projects.

cskiraly: Thanks @dryajov. I've used random only for initial testing of the calculation. I will change all random functions in a subsequent patch to crypto secure ones.

dryajov: Cool, yep I figured it was just a first pass, just noting here for clarity 😉

dryajov:
@cskiraly rebasing with `main` now should fix some of the build issues, but there is still a weird one related to our protobuf serialization.

dagger/storageproofs/rsa.nim
```diff
@@ -0,0 +1,195 @@
...
+  result = rsaDecode((hashNameI(t.name, i) * productory) mod N, ssk)
+
+proc st*(ssk: PrivateKey, filename: string): (Tau, seq[BigInt]) =
+  let file = open(filename)
+  let (s, n) = split(file)
+  var t = TauZero(n: n)
+
+  # generate a random name
+  for i in 0 ..< 512 :
+    t.name[i] = rand(byte)
```
dryajov, line 119: we should deff replace this for the `Rng` random.

dagger/storageproofs/bls.nim
```diff
@@ -0,0 +1,390 @@
...
+
+  # generate the coefficient vector for combining sectors of a block: U
+  var ubase: seq[blst_scalar]
+  for i  in 0 ..< s :
+    let (u, ub) = rndP1()
+    t.u.add(u)
+    ubase.add(ub)
+   
+  #TODO: a better bytearray conversion of TauZero for the signature might be needed
+  let signature = sign(ssk.signkey, $t)
```
dryajov, line 263: I don't remember what this mean, why aren't we doing it here?

dagger/storageproofs/bigint/bigints2.nim
```diff
@@ -0,0 +1,46 @@
...
+export bigints
+
+func zero*(T: typedesc): T {.inline.} =
+  bigints.zero
+
+func one*(T: typedesc): T {.inline.} =
+  bigints.one
+
+func mulmod*(a, b, m: BigInt): BigInt =
+  (a * b) mod m
```
mratsim, line 20: I would assume this can get real slow and inefficient with large ints. Is that fine for the POC?

This is my modular arithmetic implementation for quick prototyping: https://github.com/mratsim/number-theory/blob/8c071c89/src/modular_arithmetic.nim#L6-L84

mratsim: Ah just saw that BIgInt is actually stint behind, you can use `mulmod` from here https://github.com/status-im/nim-stint/blob/036c71d06a6b22f8f967ba9d54afd2189c3872ca/stint/modular_arithmetic.nim#L103-L111

dagger/storageproofs/bigint/stint2.nim
```diff
@@ -0,0 +1,22 @@
...
+##  * Apache License, version 2.0, ([LICENSE-APACHE](LICENSE-APACHE))
+##  * MIT license ([LICENSE-MIT](LICENSE-MIT))
+## at your option.
+## This file may not be copied, modified, or distributed except according to
+## those terms.
+
+import stint
+export stint
+
+type BigInt* = StUInt[4096] #TODO: check why 2048 is not enough
```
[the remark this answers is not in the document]
mratsim: yes, need to use https://github.com/status-im/nim-stint/blob/036c71d06a6b22f8f967ba9d54afd2189c3872ca/stint/modular_arithmetic.nim#L103-L111

dagger/storageproofs/bls.nim
```diff
@@ -0,0 +1,390 @@
...
+  signkey: blscurve.SecretKey
+  key: blst_scalar
+
+# public key combining the metadata signing key and the POR validation key
+type PublicKey = object
+  signkey: blscurve.PublicKey
+  key: blst_p2
+
+# POR metadata 
+type TauZero = object
```
mratsim, line 103: Tip: put in comment the name used in the research paper so that we can easily do unicode search 𝜏 or 𝜏₀

dagger/storageproofs/bls.nim
```diff
@@ -0,0 +1,390 @@
...
+#
+# Communicaiton cost in the Setup phase is simply related to the storage cost.
+# The size of the challenge is
+#   q * (8 + 48) bytes
+# The size of the proof is instead
+#   s * 32 + 48 bytes
+
+import blscurve
+import blscurve/blst/blst_abi
+import ../rng
```
mratsim, line 83: Is that CryptoSecure?

dagger/storageproofs/bls.nim
```diff
@@ -0,0 +1,390 @@
...
+  while true:
+    for val in scal.mitems:
+      val = byte Rng.instance.rand(0xFF)
+    scalar.blst_scalar_from_bendian(scal)
+    if blst_scalar_fr_check(scalar).bool:
+      break
+
+  return scalar
+
+proc rndP2(): (blst_p2, blst_scalar) =
```
mratsim, line 147: I strongly suggest using in-place functions instead of returning large arrays or types.
They tend to explode the stack and be poorly optimized by compilers (lots of copies and memset zero).

They're fine for the POC though.

dagger/storageproofs/bls.nim
```diff
@@ -0,0 +1,390 @@
...
+  return n
+
+proc hashToG1(msg: string): blst_p1 =
+  ## Hash to curve with Dagger specific domain separation
+  const dst = "DAGGER-PROOF-OF-CONCEPT"
+  result.blst_hash_to_g1(msg, dst, aug = "")
+
+proc hashNameI(name: openArray[byte], i: int64): blst_p1 =
+  ## Calculate unique filname and block index based hash
+  return hashToG1($name & $i)
```
mratsim, line 198: This will cause issues for specification.

Assuming you have [0xFF, 0xFF, 0xFF] and i = 0
What you will hash is "2552552550" since $0xFF == "255"

Furthermore conversion of integer to 8 bytes to append will need to indicate whether it is done Little or BigEndian. (in spec it's usually I2OS for Integer to Octet String)

dagger/storageproofs/bls.nim
```diff
@@ -0,0 +1,390 @@
...
+  blst_p2_to_affine(aa2, a2)
+  blst_p2_to_affine(bb2, b2)
+
+  blst_miller_loop(loop0, aa2, aa1)
+  blst_miller_loop(loop1, bb2, bb1)
+
+  blst_fp12_mul(gt_point, loop0, loop1)
+  blst_final_exp(gt_point, gt_point)
+
+  return blst_fp12_is_one(gt_point).bool
```
mratsim, line 352: Alternatively the high-level primitives to do this is `blst_core_verify_pk_in_g1` (if pubkey in G1 and signature in G2)

https://github.com/status-im/nim-blscurve/blob/58729c1189871815e8d0c7b96957f0bac3cff9f7/blscurve/blst/blst_min_pubkey_sig_core.nim#L241-L283

dagger/storageproofs/rsa.nim
```diff
@@ -0,0 +1,195 @@
...
+##  * Apache License, version 2.0, ([LICENSE-APACHE](LICENSE-APACHE))
+##  * MIT license ([LICENSE-MIT](LICENSE-MIT))
+## at your option.
+## This file may not be copied, modified, or distributed except according to
+## those terms.
+
+import libp2p/crypto/crypto # for RSA
+import bearssl
+import nimcrypto # for SHA512
+import random
```
mratsim, line 13: unless for generating random tests, we should never import nim std/random in a crypto-sensitive library.

dagger/storageproofs/bls.nim
dryajov: What you mean? Is `CryptoSecure` some specific implementation in this case, or just weird formatting getting in the way? 

We're using the bearssl RNG we're using everywhere else, so I assume it's cryptographically secure in so far as our stack is concerned?

dagger/storageproofs/bls.nim
dryajov: you mean `{.inline.}`?

dagger/storageproofs/bls.nim
dryajov: > This will cause issues for specification.

Hmm, you mean that it will overflow the group if it's not sized with an array type of specific size?

> Furthermore conversion of integer to 8 bytes to append will need to indicate whether it is done Little or BigEndian. (in spec it's usually I2OS for Integer to Octet String)

👍

dagger/storageproofs/bls.nim
mratsim: yes I mean a RNG designed for cryptography.

dagger/storageproofs/bls.nim
mratsim: I mean this

image

What will be hashed is `[255, 255, 255]1`

dagger/storageproofs/rsa.nim
cskiraly: Right, I will remove these using only the bearssl Rng

dagger/storageproofs/bls.nim
cskiraly: I'm using the lower level primitives just before, so it seems reasonable to me to keep it that way also for the final part.

dagger/storageproofs/bls.nim
cskiraly: Usage depends on roles in the system, but in essence, this is signing the metadata (name, size, vector of coefficients) at setup time, such that anyone dealing with it in later steps knows the metadata can be trusted. 
The comment is about the $t conversion, which is not a good conversion for a specification.

dagger/storageproofs/bls.nim
mratsim: Not returning arrays or large types, at least for internal functions.

cskiraly: @mratsim with in-place, which one do you mean, if any?
- the use of var params, as in
`proc rndP2(x: var blst_p2, scalar: var blst_scalar): void =`
- or the use of return through `result`, as in
`proc rndP2(): tuple[x: blst_p2, scalar: blst_scalar]{.noInit.} =`

It is also not exactly clear to me which one you regard as large arrays. Do you refer to the blst_* types, or to the tuple?

dagger/storageproofs/bls.nim
cskiraly: In fact, what will be hashed is the string "[255, 255, 255]1", which is unique. "2552552551" would be more problematic as it could both be 255 255 255 1 as e.g. 25 52 255 25 51. Having said that, I2OS sounds like a good idea for a spec

dagger/storageproofs/bigint/bigints2.nim
cskiraly: I was trying two bigint libraries, `stuint` (which has its own mulmod) and `bigints` for which I had this slow implementation. We could add a better mulmod as you suggest, or use the stuint, but this is all only used in the RSA version. Since the BLS version is much more performant anyway, it seems better to remove the RSA version together with these from the patch.

dagger/storageproofs/bls.nim
mratsim: in-place would be var params.

The Nim definition of large "anything equal or bigger than 24 bytes" is probably suitable. (Any parameter larger than 3*sizeof(float64) is passed by reference unless it's tagged {.bycopy.} at the type level, the corresponding code is a bit hard to find though).

PR #31, opened 2021-11-23
Build system

dryajov:
Moving dagger to nimbus build system to allow for better dependencies control.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,9 +6,6 @@ build/
 # Nimble packages
 /vendor/.nimble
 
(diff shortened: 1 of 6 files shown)


.gitignore
```diff
@@ -1,3 +1,17 @@
-*
```
markspanbroek, line 1: Would be nice if these * lines could remain. They hide the executables that are created by `nim c -r` etc..

markspanbroek (approved):

Issue #19, opened 2021-08-24
A minimal daemon implementation that exposes raw REST endpoints to interact with it


dryajov:
TODO:

- [ ] Should expose REST endpoints to
  - [ ] Upload a file - should return the metadata of the uploaded file
  - [ ] Download a file - should allow streaming of chunks/data as it's being fetched
  - [ ] Connect to a specific peer
  - [ ] Find peers that announce a specific hash or set of hashes
  - [ ] ???

dryajov:
closed by #32

PR #32, opened 2021-11-30
Node setup

### LINKED ISSUE #21: Download chunks
TODO:

- Given a hash of a
  - [ ]  top level chunk locate and download all chunked to by the top level chunk
  - [ ]  terminal chunk download the chunk

dryajov:
This PR attempts to implement a rudimentary version of the dagger node. The purpose of this PR is to create a baseline implementation that can be subsequently evolved. 

This implementation includes:

- Ability to upload and download files over REST endpoints
- Ability to exchange blocks across one or more nodes
- Store dataset blocks and manifest blocks locally

closes #21 #20 #19


tests/dagger/stores/testblockexc.nim
```diff
@@ -118,58 +134,63 @@ suite "BlockExc engine - 2 nodes":
...
 
     # second trigger blockexc to resolve any pending requests
     # for the block
-    blockexc2.putBlocks(@[blk])
+    await blockexc2.putBlock(blk)
 
-    await done
+    # should fail retrieving block from remote
+    check await blockexc1.getBlock(blk.cid)
+      .withTimeout(100.millis) # should succede
```
michaelsbradleyjr, line 163: It's likely related to working my way through a "moving target" (I realize you're actively working on this) but comments on L161 and L163 seem to be at odds, or have I misunderstood something?

tests/dagger/stores/testblockexc.nim
```diff
@@ -118,58 +134,63 @@ suite "BlockExc engine - 2 nodes":
...
-    blockexc1.engine.localStore.addChangeHandler(onBlocks, ChangeType.Added)
-
     # first put the required block in the local store
-    blockexc2.engine.localStore.putBlocks(@[blk])
+    await blockexc2.engine.localStore.putBlock(blk)
 
     # second trigger blockexc to resolve any pending requests
     # for the block
-    blockexc2.putBlocks(@[blk])
+    await blockexc2.putBlock(blk)
```
michaelsbradleyjr, line 159: I may be able to answer my own question with further study of the code, and/or maybe it's something we can discuss later today, but what is the relationship/difference between:
```nim
await blockexc2.engine.localStore.putBlock(blk)
```
and
```nim
await blockexc2.putBlock(blk)
```

tests/dagger/stores/testblockexc.nim
dryajov: Ah, good catch, the comment should be changed. The idea is to first make a request for a non existent block and verify that it fails which is done on line 151, then we add the block to the store and request it again and this time it should actually pass.

tests/dagger/stores/testblockexc.nim
dryajov: They both hold a reference to the same `localStore`, this is so that we can alway check if the block is available locally when requested from the node (for example the rest endpoint) or "put" the block to disk when one arrives from the network.

Blockexc is an implementation of the `BlockStore` interface and so is `localStore`.

dryajov: To expand on this a little bit, the `BlockStore` is the base "interface" used by different components to store and retrieve blocks either from local storage (i.e. disk) or the network. 

Architecturally, there are 2 main BlockStores the `NetworkStore` and the `LocalStore`. The `NetworkStore` is backed by the `BlockExchange` mechanism which as the name suggests implements the exchange of blocks across network peers.

The `LocalStore` abstracts aways storing blocks on disk.

Components then, talk to the `NetworkStore` which encapsulates a `LocalStore` to request or store blocks. The flow is more or less as follows:

Requesting blocks:

- Call `getBlock()` on the `NetworkStore` which will first check the `LocalStore`
  - if a block exist locally then return it
  - if no, dispatch it to the `BlockExchange` engine which will request it from some network peer

Storing blocks:

- Call `putBlock()` on the `NetworkStore`
  - `NetworkStore` calls `putBlock()` on `LocalStore` which will write it to local disk
  - Notify `BlockExchange` that a new block is available so that it can update interested peers with it's availability

dagger/rest/api.nim
```diff
@@ -0,0 +1,199 @@
...
+
+proc encodeString(peerId: PeerID): Result[string, cstring] =
+  ok($peerId)
+
+proc decodeString(T: type PeerID, value: string): Result[PeerID, cstring] =
+  let peer = PeerID.init(value)
+  if peer.isOk:
+    ok(peer.get())
+  else:
+    err(peer.error())
```
Menduist, line 54:
```suggestion
  ? PeerId.init(value)
```

Menduist: Unless it's to convert to cstring, then you could use [mapErr](https://github.com/arnetheduck/nim-result/blob/7c5f2423fe6042adc667ce7630a5b13349a13654/results.nim#L348-L352)

dagger/stores/fsstore.nim
```diff
@@ -0,0 +1,115 @@
...
+  topics = "dagger fsstore"
+
+type
+  FSStore* = ref object of BlockStore
+    cache: BlockStore
+    repoDir: string
+    postfixLen*: int
+
+template blockPath(self: FSStore, cid: Cid): string =
+  self.repoDir / ($cid)[(($cid).len - self.postfixLen)..<($cid).len]
```
Menduist, line 37:
```suggestion
proc blockPath(self: FSStore, cid: Cid): string =
  let sid = $cid
  self.repoDir / sid[^self.postfixLen..^1] / sid
```
And use blockPath.parentDir to get the directory when you want it

Less error prone, imo (eg, delBlock is wrong)

dagger/stores/fsstore.nim
```diff
@@ -0,0 +1,115 @@
...
+
+method getBlock*(
+  self: FSStore,
+  cid: Cid): Future[?Block] {.async.} =
```
Menduist, line 41: Maybe switch to Result instead of Option, and make an enum of possible failures so that caller knows if he can retry etc

dagger/stores/memorystore.nim
```diff
@@ -7,54 +7,72 @@
...
 
 export blockstore
 
+logScope:
+  topics = "dagger memstore"
+
 type
   MemoryStore* = ref object of BlockStore
-    blocks: seq[Block] # TODO: Should be an LRU cache
+    blocks: seq[?Block] # TODO: Should be an LRU cache
```
Menduist, line 29: Why would you store a None block?

Menduist (commented):
Some nitpicks mostly on code style, haven't cloned & looked deeply yet

dagger/manifest.nim
```diff
@@ -0,0 +1,197 @@
...
+  ManifestCodec* = multiCodec("dag-pb")
+
+type
+  BlocksManifest* = object
+    blocks: seq[Cid]
+    htree: ?Cid
+    version*: CidVersion
+    hcodec*: MultiCodec
+    codec*: MultiCodec
+    cidEmptyDigest: MultiHash
```
markspanbroek, line 31: It feels a bit weird to me to have this as an instance variable. I think I would rather like to see a function that returns the empty digest, given a manifest, and drop the instance variable:

```nim
func emptyDigest(manifest: BlocksManifest): ?!MultiHash =
```

dagger/rest/api.nim
```diff
@@ -0,0 +1,199 @@
...
+    MethodGet,
+    "/api/dagger/v1/connect/{peerId}") do (
+      peerId: PeerID,
+      addrs: seq[MultiAddress]) -> RestApiResponse:
+      if peerId.isErr:
+        return RestApiResponse.error(
+          Http400,
+          $peerId.error())
+
+      let addresess = if addrs.isOk and addrs.get().len > 0:
```
markspanbroek, line 78: addresses

dagger/rest/api.nim
```diff
@@ -0,0 +1,199 @@
...
+                "Unable to find Peer!")
+
+            peerRecord.get().addresses.mapIt(
+              it.address
+            )
+
+      await node.connect(peerId.get(), addresess)
+      return RestApiResponse.response("")
+
+  router.api(
```
markspanbroek, line 94: Are there any tests for the REST api?

dagger/stores/memorystore.nim
```diff
@@ -7,54 +7,72 @@
...
+  cid: Cid): Future[?Block] {.async.} =
   ## Get a block from the stores
   ##
 
-  var res: seq[Block]
-  for c in cids:
-    res.add(s.blocks.filterIt( it.cid == c ))
+  trace "Getting block", cid
+  let found = b.blocks.filterIt(
+    (!it).cid == cid
```
markspanbroek, line 39: I'd avoid using `!` in implementation code (it's fine in test code). Perhaps replace it with:
```nim
it.?cid == cid.some
```
Same holds for line 53 and 72.

tests/dagger/blockexc/testblockexc.nim
```diff
@@ -118,69 +130,77 @@ suite "BlockExc engine - 2 nodes":
...
-    blockexc2.engine.localStore.putBlocks(@[blk])
+    check await blockexc2.engine.localStore.putBlock(blk)
 
     # second trigger blockexc to resolve any pending requests
     # for the block
-    blockexc2.putBlocks(@[blk])
+    check await blockexc2.putBlock(blk)
 
-    await done
+    # should succede retrieving block from remote
```
markspanbroek, line 160:
```suggestion
    # should succeed retrieving block from remote
```

markspanbroek (commented):
Nice work @dryajov! Just a few small comments in my review.

dagger.nim
```diff
@@ -0,0 +1,72 @@
...
+import ./dagger/utils/fileutils
+
+when defined(posix):
+  import system/ansi_c
+
+let
+  config = DaggerConf.load()
+
+case config.cmd:
+of StartUpCommand.noCommand:
```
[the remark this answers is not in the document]
dryajov: No, that is not how confutils works... this is the correct way of invoking it https://github.com/status-im/nim-confutils/

dagger/blockexchange/engine.nim
```diff
@@ -68,40 +75,66 @@ proc getPeerCtx*(b: BlockExcEngine, peerId: PeerID): BlockExcPeerCtx =
...
+
+proc blockexcTaskRunner(b: BlockExcEngine): Future[void] {.gcsafe.}
+
+proc start*(b: BlockExcEngine) {.async.} =
+  ## Start the blockexc task
+  ##
+
+  trace "blockexc start"
+
+  if b.blockexcTasks.len > 0:
```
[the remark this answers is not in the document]
dryajov: Ah, good catch I think I copying this around from somewhere else.

dagger/blockexchange/engine.nim
```diff
@@ -68,40 +75,66 @@ proc getPeerCtx*(b: BlockExcEngine, peerId: PeerID): BlockExcPeerCtx =
...
+  trace "NetworkStore stop"
+  if b.blockexcTasks.len <= 0:
+    warn "Stopping blockexc without starting it"
+    return
+
+  b.blockexcRunning = false
+  for t in b.blockexcTasks:
+    if not t.finished:
+      trace "Awaiting task to stop"
+      t.cancel()
```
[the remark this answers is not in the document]
dryajov: Maybe, but I don't think it affects anything here. I like `cancelAndWait` better tho as well.

dagger/blockexchange/engine.nim
```diff
@@ -114,22 +147,23 @@ proc requestBlocks*(
...
     blockPeer.id,
-    cids,
+    @[cid],
     wantType = WantType.wantBlock) # we want this remote to send us a block
 
   if peers.len == 0:
-    return blocks # no peers to send wants to
+    return blk # no peers to send wants to
 
   template sendWants(ctx: BlockExcPeerCtx) =
```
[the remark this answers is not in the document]
dryajov: yeah, leftover from a prev implementatino

dagger/chunker.nim
```diff
@@ -26,118 +28,116 @@ const
...
-    proc(data: var openArray[byte], offset: Natural = 0): int
-    {.gcsafe, closure, raises: [Defect].}
+    proc(data: ChunkBuffer, len: int): Future[int] {.gcsafe, raises: [Defect].}
 
   ChunkerType* {.pure.} = enum
-    SizedChunker
+    FixedChunker
     RabinChunker
 
   Chunker* = ref object of RootObj
```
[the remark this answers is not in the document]
dryajov: I'm not really using inheritance, so I'd rather just make it an `ref obj`.

dagger/manifest.nim
dryajov: The idea is to initialize `emptyDigest` only once. This was going to be originally done at compile time as a `const`, but `Cid.init()` uses `memmove` which breaks compile time initialization, so I had to cache it in the object itself. Maybe we can move it to a `template` instead or worst case scenario I can just add a TODO to mention the reasoning

dagger/manifest.nim
```diff
@@ -0,0 +1,197 @@
...
+      var
+        (b1, b2) = (stack.pop(), stack.pop())
+
+      var
+        digest = MultiHash.digest(
+          $b.hcodec,
+          (b1.hashBytes() & b2.hashBytes()))
+
+      without mh =? digest:
+        return Cid.none
```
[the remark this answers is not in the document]
dryajov: The problem with `?` is that the `Result` types should match, which isn't the case here since `Cid`, `Multihash`, etc, all have their own error types.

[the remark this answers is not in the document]
dryajov: But maybe we can return `Result` instead of `Option`? I'm either way on it, it just didn't seem like it would be relevant to have an error here - you either generated the `Cid` based on the `Tree Hash` or you didn't?

dagger/manifest.nim
markspanbroek: It seems that the empty digest depends on hcodec, so you'd have to find a way to specify that at compile time as well.

dagger/manifest.nim
```diff
@@ -0,0 +1,197 @@
...
+  codec = multiCodec("raw")): ?!T =
+  ## Create a manifest using array of `Cid`s
+  ##
+
+  # TODO: The CIDs should be initialized at compile time,
+  # but the VM fails due to a `memmove` being invoked somewhere
+  let cidEmptyRes = if version == CIDv1:
+      Cid.init("bafybeihdwdcefgh4dqkjv67uzcmw7ojee6xedzdetojuzjevtenxquvyku")
+    else:
+      Cid.init("QmdfTbBqBPQ7VNxZEYEj14VmRuZBkqFbiwReogJgS1zR1n")
```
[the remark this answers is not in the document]
dryajov: Meh... not sure if we want this as a global, you'll get all sorts of issue with threads which we kinda have to use at some point to have async filesystem access.

dagger/rest/api.nim
dryajov: Not yet

dagger.nim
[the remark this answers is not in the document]
Menduist: I meant to put a `when isMainModule:` around this whole block (from line 25 to 72)
It allows dagger to be a library, and is generally good hygiene, imo (eg, if you import dagger from tests, you don't want it to start the node)

dagger/manifest.nim
dryajov: Not really, this strings are self describing and you can initialize a `Cid` with them and it will initialize with the correct type of hash (hcodec) and content it points to, which in this case should be `raw`.

dagger/manifest.nim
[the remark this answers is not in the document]
Menduist: Yeah, it's a bit verbose but it would be better to return a Result and do something like
```nim
let mh = ? MultiHash.digest(xx).mapErr(x => $x) # or any other smart logic in mapErr
```

dagger.nim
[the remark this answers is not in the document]
dryajov: sure, not a bad idea

dagger/stores/memorystore.nim
dryajov: This is a bit of a placeholder and should be replaced with a proper LRU cache, mostly for convenience so I don't have to wrap it in a `some/none` call everywhere.

dagger/stores/memorystore.nim
dryajov: I'm a bit concerned with all the wrapping and unwrapping that goes on here, it definitely feels a bit weird, but for safety you're probably right that it's better style to use `?`.

dagger/manifest.nim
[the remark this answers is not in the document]
dryajov: That said, this is definitely awkward, so I'll think of a way of cleaning it up... probably a template makes more sense here.

dagger/manifest.nim
dryajov: > It seems that the empty digest depends on hcodec, so you'd have to find a way to specify that at compile time as well.

Hmm, you're right that we assume the hash to be `sha-256` and we'd need to be able to generate an empty digest for every possible hash type... thats a bit of a pain, maybe we're better off just restricting everything to `sha-256` for now?

dagger/stores/fsstore.nim
dryajov: I thought about this, but not sure if there is anything specific we want to return here.

dagger/blockexchange/engine.nim
```diff
@@ -68,40 +75,66 @@ proc getPeerCtx*(b: BlockExcEngine, peerId: PeerID): BlockExcPeerCtx =
...
     # TODO: run discovery here to get peers for the block
-    return
-
-  var blocks: seq[Future[bt.Block]]
-  for c in cids:
-    if c notin b.pendingBlocks:
-      # install events to await blocks incoming from different sources
-      blocks.add(
-        b.pendingBlocks.addOrAwait(c).wait(timeout))
+    return blk
```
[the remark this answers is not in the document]
dryajov: No reason to complete it because we can manually connect to a peer that has the block. This is useful for manual testing and such.

dagger/blockexchange/network.nim
```diff
@@ -73,7 +73,7 @@ type
 proc handleWantList(
   b: BlockExcNetwork,
   peer: NetworkPeer,
-  list: WantList) =
+  list: WantList): Future[void] =
```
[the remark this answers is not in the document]
dryajov: Async incurres in a transformation to an internal state machine, unless we need to await there is no reason for that transformation which is costly for many reasons. Usually if the method does something that isn't async and then returns the result of an async call as the last thing it does, there is no need to force the transformation.

dagger/manifest.nim
markspanbroek: Restricting to SHA256 is fine by me.

dagger/manifest.nim
```diff
@@ -0,0 +1,197 @@
...
+              blocks.add(cidRes.get())
+
+      return (cid, blocks).success
+
+func init*(
+  T: type BlocksManifest,
+  blocks: openArray[Cid] = [],
+  version = CIDv1,
+  hcodec = multiCodec("sha2-256"),
+  codec = multiCodec("raw")): ?!T =
```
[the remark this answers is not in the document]
dryajov: The reason we use `CatchableError` is because it's a lot easier to map across different error types and it avoids explicit mapping from one error type to another. It's also just as inspectable as an `enum`, because you can check if the instance is of some specific type with the `is` operator:

```nim
type
  DaggerError                  = object of CatchableError
  NotDaggerError               = object of CatchableError
  CreateEmptyCidFailure        = object of DaggerError

block:
  echo "\n***CreateEmptyCidFailure test***"
  let exc = (ref CreateEmptyCidFailure)(msg: "Unable to get multihash from empty Cid")

  if exc is ref CreateEmptyCidFailure:
    echo exc.msg
  elif exc is ref DaggerError:
    echo "Generic error!"
  elif exc is ref NotDaggerError:
    echo "Should not hit"
  else:
    doAssert false

block:
  echo "\n***DaggerError test***"
  let exc = (ref DaggerError)()

  if exc is ref CreateEmptyCidFailure:
    echo exc.msg
  elif exc is ref DaggerError:
    echo "Generic error!"
  elif exc is ref NotDaggerError:
    echo "Should not hit"
  else:
    doAssert false

block:
  echo "\n***Not DaggerError test***"
  let exc = (ref NotDaggerError)()

  if exc is ref CreateEmptyCidFailure:
    echo exc.msg
  elif exc is ref DaggerError:
    echo "Generic error!"
  elif exc is ref NotDaggerError:
    echo "Not a dagger error!"
  else:
    doAssert false

block:
  echo "\n***Catch All test***"
  let exc = (ref CatchableError)()

  if exc is ref CreateEmptyCidFailure:
    echo exc.msg
  elif exc is ref DaggerError:
    echo "Generic error!"
  elif exc is ref NotDaggerError:
    echo "Not a dagger error!"
  elif exc is ref CatchableError:
    echo "Catch All!"
  else:
    doAssert false
```

If you run the code, it will hit all the proper types depending on the one that got instantiated, but it also allows you to have more generic error handling logic that would otherwise be harder to implement.

dagger/manifest.nim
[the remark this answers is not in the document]
dryajov: Ah, I didn't realized you where suggesting using `once`. I've reworked this section a bit so we only initialize once for all the manifests.

dryajov:
I've reworked the PR to use `Results` everywhere I was previously returning an `Optional` (thanks for the suggestions :-)). 

There are however a couple of issues that make using `Result` a bit clunky, specially with `async`. Specifically the `?` operator is broken in `async` transformations and the fix isn't trivial - it would require ad-hoc patching of either `Chronos` or `Result`.  On the other hand, `Questionable` is quite a bit more ergonomic than pure `Result`, alas it isn't fully integrated with `Results` (as it was originally designed for `Optionals`), specifically there is no way of getting the error when using the `=?` operator. We discussed this during our last call and @markspanbroek  is going to try exposing the error, potentially by returning a tuple. In the meantime,  there are a few places where I use `without a =? res: ...` disregarding the error. This isn't a huge deal as this is done at the edge of the API, where we wouldn't be able to do much with the error than logging it.

I think this is a good starting point and we can keep iterating on it in subsequent PRs. I'd like to merge this to prevent it from bloating even more.

dagger/manifest.nim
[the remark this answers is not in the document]
emizzle: Thanks for all the info. I think this has been resolved now and to clarify iirc, the issue was that everything was `CatchableError` or `DaggerError` but no exceptions were derived past `DaggerError` meaning that it would have been difficult for caller to do what you have illustrated above.

PR #33, opened 2021-12-15
add nimlsp and supporting modules to vendor

michaelsbradleyjr:
Integration of [nimlsp](https://github.com/PMunch/nimlsp) in my editor (Emacs) has been quite the revelation. It's not perfect: crashes every once in awhile, but not often! 

The hover-hints, `M-.` to find a definition, ability to jump to the definition in the sources in `vendor/` (including sources in `vendor/nimbus-build-system/vendor/Nim`), and `M-,` to jump back — wow, huge productivity and learning boost! (for me anyway)

I figured out how to get nimlsp working in the context of a nimbus-build-system project, cf. the changes in this PR. The `make nimlsp` target is entirely optional and shouldn't get in the way of anyone who doesn't have a use for it.

One thing to note: if choosenim is installed, in order for the finding/jumping to use the correct locations (i.e. in nim-dagger's `vendor/`), you'll probably have to disable choosenim while working in the project:
```
$ mv ~/.choosenim ~/.choosenim.disabled
$ mv ~/.nimble ~/.nimble.disabled
```

I start my editor like this:
```
$ cd repos/nim-dagger
$ make nimlsp  # if not already made
$ ./env.sh emacsclient -t -a ''
```

(in both screenshots below I pressed `M-.` to display the full definition)

image

image

Any thoughts or objections re: adding nimlsp and supporting modules to nim-dagger's vendor?

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -118,3 +118,12 @@
 	url = https://github.com/status-im/stint.git
 	ignore = untracked
 	branch = master
+[submodule "vendor/nimlsp"]
+	path = vendor/nimlsp
+	url = https://github.com/PMunch/nimlsp.git
+[submodule "vendor/jsonschema"]
+	path = vendor/jsonschema
+	url = https://github.com/PMunch/jsonschema.git
+[submodule "vendor/ast-pattern-matching"]
+	path = vendor/ast-pattern-matching
+	url = https://github.com/krux02/ast-pattern-matching.git
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,8 @@ LINK_PCRE := 0
 	update \
 	test \
 	clean \
-	libbacktrace
+	libbacktrace \
+	nimlsp
 
 ifeq ($(NIM_PARAMS),)
 # "variables.mk" was not included, so we update the submodules.
@@ -91,6 +92,16 @@ dagger.nims:
 # nim-libbacktrace
 libbacktrace:
 	+ $(MAKE) -C vendor/nim-libbacktrace --no-print-directory BUILD_CXX_LIB=0
+
+# nimlsp
+NIMLSP := vendor/nimlsp/nimlsp
+$(NIMLSP): | deps-common
+	echo -e $(BUILD_MSG) "$@" && cd vendor/nimlsp && \
+		$(ENV_SCRIPT) nim c \
+			-d:danger -d:nimcore -d:nimsuggest --hints:off --opt:speed --threads:on \
+			--verbosity:0 -o:nimlsp src/nimlsp
+
+nimlsp: $(NIMLSP)
 
 # usual cleaning
 clean: | clean-common
--- a/env.sh
+++ b/env.sh
@@ -1,0 +1,7 @@
+#!/usr/bin/env bash
+
+# We use ${BASH_SOURCE[0]} instead of $0 to allow sourcing this file
+# and we fall back to a Zsh-specific special var to also support Zsh.
+REL_PATH="$(dirname ${BASH_SOURCE[0]:-${(%):-%x}})"
+ABS_PATH="$(cd ${REL_PATH}; pwd)"
+source ${ABS_PATH}/vendor/nimbus-build-system/scripts/env.sh
--- a/env.sh
+++ b/env.sh
@@ -4,4 +4,5 @@
 # and we fall back to a Zsh-specific special var to also support Zsh.
 REL_PATH="$(dirname ${BASH_SOURCE[0]:-${(%):-%x}})"
 ABS_PATH="$(cd ${REL_PATH}; pwd)"
+export PATH="${ABS_PATH}/vendor/nimlsp:${PATH}"
 source ${ABS_PATH}/vendor/nimbus-build-system/scripts/env.sh


arnetheduck:
I'm curious why lsp needs to live in the project to make this work..

"goto function" and similar functionality already works out of the box with the vscode plugin and it seems to me that people that want to use other editors should be able to keep their supporting binaries outside the project and just load the right nim environment via `env.sh`.

A downside of keeping it in the project is that it becomes difficult to analyze dependencies for security audits and cruft removal - third-party risk is a thing in most of what we do.

michaelsbradleyjr:
> I'm curious why lsp needs to live in the project to make this work..

I was having trouble getting it to work correctly if it wasn't inside vendor, e.g. if the `nimlsp` executable (symlink to it) was in `~/.nimble/bin`: definitions were failing to resolve or resolving to the wrong `Nim` sources.

> "goto function" and similar functionality already works out of the box with the vscode plugin and it seems to me that people that want to use other editors should be able to keep their supporting binaries outside the project and just load the right nim environment via `env.sh`.

I'll give an alternative approach another shot... I'll try building the `nimlsp` binary so that it's located at e.g. `~/repos/nimlsp/nimlsp`. And then I can invoke the `env.sh` shim like this:
```
$ cd repos/nim-dagger
$ PATH="${HOME}/repos/nimlsp:${PATH}" ./env.sh emacsclient -t -a ''
```

> A downside of keeping it in the project is that it becomes difficult to analyze dependencies for security audits and cruft removal - third-party risk is a thing in most of what we do.

`nimlsp` and its two dependencies (`jsonschema` and `ast-pattern-matching`) in vendor *should* be inert relative to the code compiled into dagger artifacts, since dagger doesn't import them, but it's still a fair point.

zah:
I can explain what has happened. 

`nimlsp` is essentially statically linking a particular version of the Nim compiler by directly importing the source files of Nim (you can think of it as an alternative build of compiler with some minor modifications). This means that a particular build of `nimlsp` will carry a specific version of the Nim compiler and it will use the standard library that is discovered in a relative location to the `nimlsp` binary. 

If that's still not clear, this short code snippet may be worth more than the explanation above:
https://github.com/PMunch/nimlsp/blob/master/src/nimlsppkg/suggestlib.nim#L3-L11

So, when you built `nimlsp` within the `nimbus-build-system` env, you ensured that you are linking exactly our version of Nim into and that the resulting binary is in the right directory to find the standard library that we use.

dryajov:
It's unfortunate... I agree with @arnetheduck that this will complicate security analysis and clutter the view in general, on the other hand it's a shame we can't use the tools that make use the most productive... 

I'm not sure if there is an alternative way of setting nimlsp up so that it doesn't go into our vendor modules but still builds and links against them?

michaelsbradleyjr:
I found a way that works 100%, without having to check nimlsp and its deps into vendor:

(below its assumed that the plain vanilla `env.sh` shim is in the root of nim-dagger; I can open a PR just for that)

```
$ cd repos && \
  git clone https://github.com/PMunch/nimlsp.git nimlsp-dagger && \
  cd nimlsp-dagger

$ ~/repos/nim-dagger/env.sh \
  ~/repos/nim-dagger/vendor/nimbus-build-system/vendor/Nim/bin/nimble \
  --accept --depsOnly --nimbleDir:"${PWD}/.nimble" install

$ ~/repos/nim-dagger/env.sh \
  nim c -d:danger -d:nimcore -d:nimsuggest --opt:speed \
  --path:"${PWD}/$(ls -d .nimble/pkgs/ast_pattern_matching-*)" \
  --path:"${PWD}/$(ls -d .nimble/pkgs/jsonschema-*)" \
  --threads:on -o:nimlsp src/nimlsp
```
Editor invocation:
```
$ cd repos/nim-dagger
$ PATH="${HOME}/repos/nimlsp-dagger:${PATH}" ./env.sh emacsclient -t -a ''
```

**Upsides**
* (obviously) Doesn't involve checking `nimlsp`, `jsonschema`, and `ast-pattern-matching` into vendor.
* Don't have to disable `~/.choosenim` and `~/.nimble`. The `~/repos/nimlsp-dagger/nimlsp` executable doesn't get confused: all definitions are properly resolved within dagger's vendor.

**Downsides**
* Need to setup a `nimlsp-[project]` per nimbus-build-system project, but it's not a big deal (steps above are now worked out and easily repeatable).
* A bit more ceremony re: starting one's editor, but with the help of shell aliases it's not a big deal.

---

Personally I think it's nicer to have nimlsp checked into vendor, but I'm happy to live with this alternative solution (works great) and I definitely appreciate the concerns re: security analysis.

michaelsbradleyjr:
Closing in favor of #34. Can use the approach I described in https://github.com/status-im/nim-dagger/pull/33#issuecomment-996076749 for nimlsp integration.

stefantalpalaru:
Could it be simplified by letting Nimble use its default package dir (and making it a Make target, since you'll re-run it for every new Nim version)?

```make
build-nimlsp:
        ./env.sh vendor/nimbus-build-system/vendor/Nim/bin/nimble \
                --accept install nimlsp@#head
```

Might need to run `nimble uninstall nimlsp || true` first, to force a reinstall.

Such a target could join "ctags" in nimbus-build-system: https://github.com/status-im/nimbus-build-system/blob/25a4c270330026442e09f6cddfb7a944da7cfa4b/makefiles/targets.mk#L174

PR #46, opened 2022-01-20
[rest api]: improve exception handling for `/connect` and add `content-type` header for `/download`

emizzle:
This PR achieves the following:
1. Improves the exception handling when dialling a peer fails or an unknown error occurs.
2. Add a `Content-Type` header to the `/download` endpoint of `application/octet-stream`, which is [defined by MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types#types) as meant to be used for "generic binary data (or binary data whose true type is unknown)". 

Co-authored-by: Michael Bradley <michaelsbradleyjr@gmail.com>

### DIFF
--- a/dagger/rest/api.nim
+++ b/dagger/rest/api.nim
@@ -86,9 +86,14 @@ proc initRestApi*(node: DaggerNodeRef): RestRouter =
             peerRecord.get().addresses.mapIt(
               it.address
             )
+      try:
+        await node.connect(peerId.get(), addresses)
+        return RestApiResponse.response("Successfully connected to peer")
+      except DialFailedError as e:
+        return RestApiResponse.error(Http400, "Unable to dial peer")
+      except CatchableError as e:
+        return RestApiResponse.error(Http400, "Unknown error dialling peer")
 
-      await node.connect(peerId.get(), addresses)
-      return RestApiResponse.response("")
 
   router.api(
     MethodGet,
@@ -113,6 +118,7 @@ proc initRestApi*(node: DaggerNodeRef): RestRouter =
             retr.isErr):
             return RestApiResponse.error(Http404, retr.error.msg)
 
+        resp.addHeader("Content-Type", "application/octet-stream")
         await resp.prepareChunked()
         while not stream.atEof:
           var


dryajov (approved):

dagger/rest/api.nim
```diff
@@ -86,9 +86,14 @@ proc initRestApi*(node: DaggerNodeRef): RestRouter =
             peerRecord.get().addresses.mapIt(
               it.address
             )
+      try:
+        await node.connect(peerId.get(), addresses)
+        return RestApiResponse.response("Successfully connected to peer")
+      except DialFailedError as e:
+        return RestApiResponse.error(Http400, "Unable to dial peer")
+      except CatchableError as e:
```
Menduist, line 94: You need to handle CancelledError like so:
```nim
except CancelledError as e:
  raise e
except CatchableError as e:
###
```

Menduist: Otherwise, the CancelledError will be swallowed, and cancellation will not be handled properly.
(not sure if this procedure can be cancelled though)

dryajov: Hmm, actually I wonder if that is the case since this are essentially callbacks and we don't need to teardown anything in presto itself and the best thing presto can do with the cancellation is return a `500` (or similar)?

dryajov: In fact this is how presto handles this callbacks:

```nim
        let restRes =
          try:
            await route.callback(request, pathParams, queryParams, optBody)
          except HttpCriticalError as exc:
            debug "Critical error occurred while processing a request",
                  meth = $request.meth, peer = $request.remoteAddress(),
                  uri = $request.uri, code = exc.code,
                  path_params = pathParams, query_params = queryParams,
                  content_body = optBody, error_msg = $exc.msg
            return await request.respond(exc.code)
          except CatchableError as exc:
            warn "Unexpected error occured while processing a request",
                  meth = $request.meth, peer = $request.remoteAddress(),
                  uri = $request.uri, path_params = pathParams,
                  query_params = queryParams, content_body = optBody,
                  error_msg = $exc.msg, error_name = $exc.name
            return await request.respond(Http503)
```

Menduist: Ok, good to know, no need to change anything then :)

PR #50, opened 2022-02-09
feat: introduce LRU cache

emizzle:
Replace `MemoryStore` with LRU caching mechanism.

`lrucache` library was forked to https://github.com/status-im/lrucache.nim.

### Note
1. This PR is based off of #49 (which uses the `feat/blockstore-manager` branch).
2. This PR is in draft as we need to allow for `cacheSize` to be specified as a CLI parameter.

Co-authored-by: Eric Mastro <eric.mastro@gmail.com>


Menduist:
Imo there are multiple LRU, at least an in memory one and a on disk one

Maybe we can reuse the same system for the two?

michaelsbradleyjr:
> Imo there are multiple LRU, at least one in memory and one on disk

In what context/s do you think an LRU cache on disk would be helpful? Does it have to do with "free mode" and/or when a node is looking to gain from egress fees for blocks that it's not committed (by contract) to store for a predetermined amount of time?

Maybe we discussed this as a team during one of the weekly meetings, but I don't recall.

michaelsbradleyjr:
I've pushed a commit implementing a `--cache-size` CLI option. If `0` is specified, the cache is disabled (no instance of `CacheStore` is created).

michaelsbradleyjr:
Re: caching on disk, I've started a discussion where we can dig into the idea a bit further: https://github.com/status-im/dagger-research/discussions/65.

emizzle:
Updated to allow for dynamic sized blocks.

emizzle:
Rebased on top of main.

dagger/dagger.nim
```diff
@@ -62,6 +62,13 @@ proc new*(T: type DaggerServer, config: DaggerConf): T =
     .withTcpTransport({ServerFlags.ReuseAddr})
     .build()
 
+  # Most Local > Most Remote: order is important!
+  var stores: seq[BlockStore]
```
dryajov, line 66: Is this still needed?

michaelsbradleyjr: Probably not, will take another look.

emizzle: Updated so that the init block instantiates the size of the cache based on cli option.

emizzle:
> LGTM, except for that initialization block I commented about.

Should be g2g now.

dagger/dagger.nim
```diff
@@ -62,10 +61,16 @@ proc new*(T: type DaggerServer, config: DaggerConf): T =
     .withTcpTransport({ServerFlags.ReuseAddr})
     .build()
 
+  let cache = block:
```
dryajov, line 64: You don't need the `block`, `if` can be used as an expression as well. So 

```nim
let cache = if config.cacheSize > 0:
      CacheStore.new(cacheSize = config.cacheSize * MiB)
    else:
      CacheStore.new()
```

Works just as well.

emizzle: I find the block easier to read personally, but will update it.

emizzle: Removed the `block:`.

emizzle:
Fixed the `block`, should be g2g now.

dryajov (approved):

PR #51, opened 2022-03-02
Nim wrapper for Leopard-RS

michaelsbradleyjr:
See: https://github.com/catid/leopard

This PR is not intended to come out of draft. Most likely, the wrapper will be extracted into another repo (`status-im/nim-leopard`).

My goal is to invite feedback and discussion; and I need some help as well, because it's not fully working yet.

This is my first attempt at wrapping a C++ library, though I have previous experience wrapping C libs. So there was some guesswork involved and probably mistakes. Thanks in advance for helping me sort it out!

I did the prototyping in this repo because it allows me to easily check that once libleopard is statically linked to e.g. nim-dagger's `testAll` binary, and having engaged Nim's C++ backend via the leopard wrapper (I think it does, not 100% sure), that our non-leopard Nim code still works as expected. Also, it allowed me to avoid wrestling up-front with how to wire-up a vendored Makefile and/or compiler directives in nim-leopard's `.nim` sources in a way that's flexible enough for dagger's purposes but not "hardwired" to them.

---

First, a jumble of notes about the wrapper impl and supporting changes in this PR. When I describe something that needs to be done/changed, see the changes in this PR for how I'm currently doing it.

The clang that ship's with Apple's Xcode doesn't support `-fopenmp`. Leopard's cmake config automatically checks for OpenMP support: it's detected on e.g. Ubuntu 20.04 and in a MinGW context on Windows (as seen in GHA). LLVM/clang installed on macOS with e.g. Homebrew *does* support `-fopenmp`, but for that to work some env variables need to be setup, `PATH` needs to adjusted, and some extra flags need to be passed to `cmake` and the Nim compiler.

Flags passed to `cmake` vary by platform and 64 vs. 32 bit arch. 

nim-dagger has another vendored dependency that uses `cmake`. It seems to need some changes in order for Homebrew clang/++ to be used instead of Xcode's clang on macOS:

https://github.com/status-im/nim-libbacktrace/blob/master/Makefile#L40-L41

Can probably be changed like this, but I haven't tried yet:
```Makefile
CC ?= gcc
CXX ?= g++
```
I'm also not sure if it matters that different `clang` versions are used when building `.a` that will be statically linked to the final executable.

The wrapper is fully contained in `dagger/leopard.nim`. The exported procs match up with the API presented in Leopard-RS' [readme](https://github.com/catid/leopard#encoder-api), plus `leo_result_string`:
* `leo_init`
* `leo_result_string`
* `leo_encode_work_count`
* `leo_encode`
* `leo_decode_work_count`
* `leo_decode`

In `tests/dagger/testleopard.nim` I did a rough and *minimal* port of Leopard-RS' [`tests/benchmark.cpp`](https://github.com/catid/leopard/blob/master/tests/benchmark.cpp). The only goal was/is (for now) to ensure that all procs exported from `dagger/leopard.nim` are called at least once and work as expected.

In `dagger/leopard.nim` it was necessary to use `importCpp` for `leo_init`, but for the other procs it was necessary to use `importc` instead, else the compiler fails with an error when compiling code that involves calling the proc. I don't understand why.

---

`tests/dagger/testleopard.nim` [runs successfully](https://github.com/status-im/nim-dagger/actions?query=branch%3Aleopard) on each platform/arch that we're currently testing in GHA. At the bottom of the output of `build/testAll` this is printed:
```
Parameters: [original count=100] [recovery count=10] [buffer bytes=64000] [loss count=32768] [random seed=2]
encode_work_count: 32
decode_work_count: 128
total_bytes: 6400000
encodeResult: Operation succeeded
decodeResult: Operation succeeded
```

However, the reports of `Operation succeeded` are probably misleading.

When coding the port of Leopard-RS' [`tests/benchmark.cpp`](https://github.com/catid/leopard/blob/master/tests/benchmark.cpp), I wasn't sure how to "translate" use of `std::vector` into Nim. I was also unsure about the importance of `leopard::SIMDSafeAllocate` and `leopard::SIMDSafeFree`.

I did some searching re: both, but it wasn't clear how to proceed, so I gave it a shot with Nim's `newSeqWith` and pointers obtained via `addr`. I also went with empty data (still the case in `tests/dagger/testleopard.nim` on this branch). It was encouraging to get `encodeResult: Operation succeeded` and `decodeResult: Operation succeeded`!

However, yesterday I broadened the experiment, trying to lightly replicate the recovery verification in [`tests/benchmark.cpp`](https://github.com/catid/leopard/blob/master/tests/benchmark.cpp) with non-empty `original_data` and data losses. @emizzle gave me a hand later in the day. The efforts did not succeed. No matter what was tried (with seqs and pointers) it didn't seem that `original_data` was being recovered, despite `decodeResult: Operation succeeded`.

At this point, I think it may be necessary to wrap `std::vector`, `leopard::SIMDSafeAllocate` and `leopard::SIMDSafeFree` and use them when moving data through Leopard-RS' wrapped API.

Or there may just be some misunderstanding on my part how to properly use Nim's `seq` with the API, and/or misunderstanding of how to verify data recovery / extract recovered data.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -99,6 +99,8 @@ jobs:
         run: |
           brew install libomp
           echo "$(brew --prefix)/opt/llvm/bin" >> $GITHUB_PATH
+          echo "CC=$(brew --prefix)/opt/llvm/bin/clang" >> $GITHUB_ENV
+          echo "CXX=$(brew --prefix)/opt/llvm/bin/clang++" >> $GITHUB_ENV
           echo "LDFLAGS=-L$(brew --prefix)/opt/libomp/lib -L$(brew --prefix)/opt/llvm/lib -Wl,-rpath,$(brew --prefix)/opt/llvm/lib" >> $GITHUB_ENV
           echo "LIBLEOPARD_CMAKE_FLAGS=-DCMAKE_BUILD_TYPE=Release -DCMAKE_C_COMPILER=$(brew --prefix)/opt/llvm/bin/clang -DCMAKE_CXX_COMPILER=$(brew --prefix)/opt/llvm/bin/clang++" >> $GITHUB_ENV
           echo "NIM_EXTRA_PARAMS=--passL:-fopenmp --passL:-L$(brew --prefix)/opt/libomp/lib" >> $GITHUB_ENV
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,6 @@ jobs:
       fail-fast: false
       max-parallel: 20
       matrix:
-        branch: [v1.4.6]
         target:
           # Unit tests
           - os: linux
@@ -35,7 +34,7 @@ jobs:
           - target:
               os: windows
             builder: windows-2019
-    name: '${{ matrix.target.os }}-${{ matrix.target.cpu }} (${{ matrix.branch }})'
+    name: '${{ matrix.target.os }}-${{ matrix.target.cpu }}'
     runs-on: ${{ matrix.builder }}
     steps:
       - name: Checkout nim-dagger
@@ -93,6 +92,7 @@ jobs:
           EOF
           chmod 755 external/bin/gcc external/bin/g++
           echo '${{ github.workspace }}/external/bin' >> $GITHUB_PATH
+          echo "LIBLEOPARD_CMAKE_FLAGS=-DCMAKE_BUILD_TYPE=Release -DCMAKE_C_FLAGS=-m32 -DCMAKE_CXX_FLAGS=-m32" >> $GITHUB_ENV
 
       - name: Restore MinGW-W64 (Windows) from cache
         if: runner.os == 'Windows'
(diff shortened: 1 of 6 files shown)


michaelsbradleyjr:
I've just pushed a commit with changes in `tests/dagger/testleopard.nim`.

@emizzle and I continued our experiments today. We realized that the Nim type we needed to use is `seq[pointer]` instead of `seq[seq[byte]]`. After making appropriate changes related to `seq[pointer]` we've reached a point where we felt we knew when/why encode/decode woud succeed or fail and could then test our predictions with mock data.

So the wrapper does seem to be working but there are still some open questions:
* When there is data loss in `original_data` but enough erasure code data to decode successfully, how can the missing original data be extracted? It seems leopard does not provide a facility for that (`leo_decode` does *not* do that exactly), but maybe something has been overlooked.
* Will the wrapper behave and perform as expected in real world usage if we're not wrapping and using C++`std::vector` , `leopard::SIMDSafeAllocate` and `leopard::SIMDSafeFree`?

michaelsbradleyjr:
This PR has served it's purpose. Closing in favor of integrating [nim-leopard](https://github.com/status-im/nim-leopard) into nim-dagger in a future PR.

There is already a [`nim_leopard`](https://github.com/status-im/nim-dagger/tree/nim_leopard) branch in this repo for checking that integration won't break existing code/tests.

PR #63, opened 2022-03-30
Sales

markspanbroek:
Adds a simple sales algorithm that watches for storage request, and offers storage if it has matching available storage.

Does not yet include persistent state, so it won't be able to recover from a node shutdown in the middle of a sale.

### DIFF
--- a/dagger/sales.nim
+++ b/dagger/sales.nim
@@ -16,7 +16,7 @@ type
     subscription: ?Subscription
     available*: seq[Availability]
     offerExpiryInterval*: UInt256
-    onSale*: OnSale
+    onSale: ?OnSale
   Availability* = object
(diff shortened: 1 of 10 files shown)


dagger/sales.nim
```diff
@@ -0,0 +1,74 @@
...
+  Availability* = object
+    id*: array[32, byte]
+    size*: uint64
+    duration*: uint64
+    minPrice*: UInt256
+
+func new*(_: type Sales, market: Market): Sales =
+  Sales(market: market)
+
+proc new*(_: type Availability,
```
Menduist, line 25: Nitpick, but per status style guide, we should use `init` when it's an object (and `new` for `ref object`s)

PR #75, opened 2022-04-06
Add DHT

Menduist:
- [X] Persistent key system for bootstrap nodes
- [X] Add config options for discovery
- [X] DaggerNode.findPeer (used by the REST API)
- [X] Block advertisement
- [X] Block discovery in block exchange
- [x] Cleanup (move everything to a mockable discovery.nim)
- [x] Tests

As can be seen by the amount of TODOs, this is a first pass
Relies on some custom branches of libp2p, libp2p-dht, lrucache.nim


Menduist:
@dryajov @michaelsbradleyjr 
I got most of this rolling, the only missing part is the providing advertisement

Could you confirm to me how you see this working:
- We advertise every block in every stores (that seems a bit weird at first, but why not)
- So everytime we call any `putBlock`, we'll advertise the new block(s)
- And we have a heartbeat which will advertise every block in every store periodically

dagger/blockexchange/engine.nim
```diff
@@ -112,62 +125,126 @@ proc stop*(b: BlockExcEngine) {.async.} =
 
   trace "NetworkStore stopped"
 
+proc discoverOnDht(b: BlockExcEngine, bd: BlockDiscovery) {.async.} =
+  bd.lastDhtQuery = Moment.fromNow(10.hours)
+  defer: bd.lastDhtQuery = Moment.now()
+
+  let
+    blockId = bd.toDiscover.toNodeId()
```
dryajov, line 133: Maybe something like to `toHandle` or `toHash` is more appropriate here, `toNodeId` is a bit confusing :)

dagger/conf.nim
```diff
@@ -47,17 +49,38 @@ type
...
+      netPrivKeyFile* {.
+        desc: "Source of network (secp256k1) private key file (random|<path>)"
+        defaultValue: "random"
+        name: "net-privkey" }: string
 
       bootstrapNodes* {.
         desc: "Specifies one or more bootstrap nodes to use when connecting to the network."
         abbr: "b"
-        name: "bootstrap-nodes" }: seq[MultiAddress]
+        name: "bootstrap-nodes" }: seq[string]
```
dryajov, line 83: Why string here? The serialization knows how to convert between strings and multiaddrs.

Menduist: It's a signed peer record, not a MA
We could use a MA, it wouldn't be less safe, I think

dryajov: Is it serialized? In that case we should add serialize/deserialize for those and use the proper signed record types instead.

dagger/blockexchange/engine.nim
```diff
@@ -112,62 +143,153 @@ proc stop*(b: BlockExcEngine) {.async.} =
 
   trace "NetworkStore stopped"
 
+proc toDiscoveryId*(cid: Cid): NodeId =
+  ## To discovery id
+  readUintBE[256](keccak256.digest(cid.data.buffer).data)
```
dryajov, line 148: It would be much better, if we can avoid the double hashing here. The `Cid` already contains a hash - we're using `sha-256`...

dagger/blockexchange/engine.nim
```diff
@@ -55,6 +70,12 @@ type
     peersPerRequest: int                          # max number of peers to request from
     wallet*: WalletRef                            # nitro wallet for micropayments
     pricing*: ?Pricing                            # optional bandwidth pricing
+    advertisedBlocks: seq[Cid]
```
dryajov, line 73: A set is probably more appropriate?

dagger/conf.nim
```diff
@@ -66,17 +68,38 @@ type
...
-        defaultValue: @[MultiAddress.init("/ip4/0.0.0.0/tcp/0").tryGet()]
-        defaultValueDesc: "/ip4/0.0.0.0/tcp/0"
-        abbr: "a"
-        name: "listen-addrs" }: seq[MultiAddress]
+      listenPorts* {.
+        desc: "Specifies one or more listening ports for the node to listen on."
+        defaultValue: @[Port(0)]
+        defaultValueDesc: "0"
+        abbr: "l"
+        name: "listen-ports" }: seq[Port]
```
dryajov, line 76: Why do we need more than one port to listen on?

Menduist: Because we can :) Before this PR, multiples MAs were accepted, now I accept multiple port, not sure if it's interesting to anyone, but it's cheap to do so..

dagger/blockexchange/engine.nim
Menduist: Yeah, would be ideal, but I think we need to modify libp2p to expose the hash without multihash headers
I can do it in a follow up PR, not urgent, I think

dryajov: It's already exposed, albeit a bit chunkily, this is how you access it - `mh.data.buffer[mh.dpos..(mh.dpos + mh.size - 1)]`.

dryajov: > Yeah, would be ideal, but I think we need to modify libp2p to expose the hash without multihash headers I can do it in a follow up PR, not urgent, I think

I would agree that we probably ok with the way things are for now, on a second thought, this might be the desired way of doing it either way...

dagger/blockexchange/engine.nim
dryajov: Probably an ordered hashset is both faster and safer here in that case.

PR #76, opened 2022-04-06
Storageproofs constantine

cskiraly:
WIP: use Constantine for proof generation and verification instead of BLST

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -178,3 +178,6 @@
 [submodule "vendor/nim-leopard"]
 	path = vendor/nim-leopard
 	url = https://github.com/status-im/nim-leopard.git
+[submodule "vendor/constantine"]
+	path = vendor/constantine
+	url = https://github.com/mratsim/constantine
--- a/dagger/storageproofs/README.md
+++ b/dagger/storageproofs/README.md
@@ -1,1 +1,9 @@
 Nim implementation of Proof of Storage related schemes
+
+Implementation of the BLS-based public PoS scheme from
+Shacham H., Waters B., "Compact Proofs of Retrievability"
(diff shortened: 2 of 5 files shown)


dryajov:
Very cool! Should we rebase with `main`? Also, how close to working is this?

arnetheduck:
constantine is not audited, and neither do we plan on doing so any time soon - it would be better to stick with blst in general, unless there are strong reasons not to

cskiraly:
Good point about auditing, thanks!

mratsim:
Agree on auditing.

Regarding backends, one of the plans is to check if the Ethereum BN curve (called alt_bn128 or bn256 or BN254 or BN254_Snarks depending on context) is suitable for Codex/Dagger:
- It's 33% smaller and would lead to faster checks.
- It would allow on-chain verification as the BLS12-381 EIP is stuck in limbo at the moment (EIP 2537 discussion: https://ethereum-magicians.org/t/eip-2537-bls12-precompile-discussion-thread/4187)

Unfortunately, there is no easy to use backend that has been audited:
- https://github.com/status-im/nim-bncurve is a direct port from the (unaudited) Zcash implementation. It is unaudited as well and significantly slower than the Cloudflare or Google backend used in Geth: https://hackmd.io/@gnark/eccbench 
image _Note that since then there has been +30% perf improvement from compiler fixes on parameter passing and algorithmic improvements_

gmega:
@cskiraly @dryajov: guys, this PR has been open for a while. Is this still relevant at all or has it been superseded by more recent work?

dryajov:
Closing as this approach is scrapped for now.

PR #80, opened 2022-04-12
Nimble lock file

zah:
The workflow based on Nimble lock files will be documented soon in the https://github.com/status-im/nim-workspace repo. For now, I plan to merge this as a helpful step for testing the new [Nim language server](https://github.com/nim-lang/langserver) in the repo .

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,10 @@ build/
 # Nimble packages
 /vendor/.nimble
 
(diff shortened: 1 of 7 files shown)


Menduist:
The CI failure seem to come from this PR, because we reused the `config.nims` from the NBS, which `import os` and makes everything fail

PR #84, opened 2022-04-13
Contract interactions

markspanbroek:


dagger/contracts/interactions.nim
```diff
@@ -0,0 +1,35 @@
...
+  let market = OnChainMarket.new(contract)
+  let proofs = OnChainProofs.new(contract)
+  ContractInteractions(
+    purchasing: Purchasing.new(market),
+    sales: Sales.new(market),
+    proving: Proving.new(proofs)
+  )
+
+proc new*(_: type ContractInteractions): ContractInteractions =
+  let provider = JsonRpcProvider.new("ws://localhost:8545")
```
dryajov, line 33: Not a biggie, but we can add the provider url to the config object.

markspanbroek: Fixed

PR #87, opened 2022-04-20
Cleanup engine and rework discovery

dryajov:


Menduist:
While a cleanup was definitively needed, I think this is a step in the wrong direction

The previous system was made with block pricing in mind, this one would be sub ideal:
- If we connect to multiple peers providing a block at the same time, we would request a block to each of them, so we would pay multiple times the same block
- The old system allowed to continue discovery until we find a reasonable price, this one doesn't, it just ask the block every time

Regarding advertisement, AFAICT, this system just shoots as fast as it can advertisements. That's not ideal, because, let's say you have a TTL of 30 minutes and 100 ms / block:
- If we have less than (30*60*1000 / 100) 18000 blocks, we'll advertise them too fast, wasting bandwidth, and putting more load on the network for nothing
- If we have more than 18k blocks, we'll advertise them too slowly

Also, we rely on the order returned by `listBlocks` to choose the advertisement order, but since this eventually comes from the filesystem, we can't rely on it

Also, the BlockDiscovery mechanism of the previous system was quite nice, IMO, since it tracked, amongst other things, the number of inflight iwants, allowing more advanced logic (only query DHT if we have little amount of iwants, etc)

But all of that would be good enough for the PoC, however is commit is just plain broken, as can be seen by tweaking this test:
```diff
--- a/tests/dagger/blockexc/discovery/testdiscovery.nim
+++ b/tests/dagger/blockexc/discovery/testdiscovery.nim
@@ -175,6 +175,8 @@ suite "E2E - Multiple Nodes Discovery":
         if cid in advertised:
           result.add(advertised[cid])
 
+    discard await blockexc[0].engine.requestBlock(blocks[0].cid)
+
     let futs = collect(newSeq):
       for b in blocks:
         blockexc[0].engine.requestBlock(b.cid)
```
Once you're connected to a peer, you'll never try to dial new peers

Unrelated, but you broke the `--log-format` option by removing the part in conf.nims. I agree it wasn't pretty, but it was needed for devops stuff

Menduist:
Nevermind, the entire test is broken, here is the fixed one:
```nim
--- a/tests/dagger/blockexc/discovery/testdiscovery.nim
+++ b/tests/dagger/blockexc/discovery/testdiscovery.nim
@@ -117,7 +117,7 @@ suite "E2E - Multiple Nodes Discovery":
         discovery = MockDiscovery.new(s.peerInfo, 0.Port)
         wallet = WalletRef.example
         network = BlockExcNetwork.new(s)
-        localStore = CacheStore.new(blocks.mapIt( it ))
+        localStore = CacheStore.new()
         engine = BlockExcEngine.new(localStore, wallet, network, discovery)
         networkStore = NetworkStore.new(engine, localStore)
 
@@ -175,15 +175,19 @@ suite "E2E - Multiple Nodes Discovery":
         if cid in advertised:
           result.add(advertised[cid])
 
-    let futs = collect(newSeq):
-      for b in blocks:
-        blockexc[0].engine.requestBlock(b.cid)
-
     await allFuturesThrowing(
       switch.mapIt( it.start() ) &
       blockexc.mapIt( it.engine.start() )
     )
 
+    # dial node 1 & 2, basically
+    discard await blockexc[0].engine.requestBlock(blocks[0].cid)
+    discard await blockexc[0].engine.requestBlock(blocks[7].cid)
+
+    let futs = collect(newSeq):
+      for b in blocks:
+        blockexc[0].engine.requestBlock(b.cid)
+
     await allFutures(futs)
 
     await allFuturesThrowing(
```

With the corrected test, it does pass, but it takes 30s (until the discovery kicks in)

PR #88, opened 2022-04-20
Restore "Cleanup engine and rework discovery (#87)"

Menduist:
Which was reverted because it broke discovery & logs, see https://github.com/status-im/nim-codex/pull/87#issuecomment-1103613281


dryajov:
closing in favor of #89

PR #89, opened 2022-04-20
Engine

dryajov:


Menduist:
(continuing discussion from #87)

> The problem with the previous implementation was that it completely obviated the structure and flow of the system

Indeed, my first pass was pretty convoluted, I opened a branch where I cleanup a fair share of complexity: https://github.com/status-im/nim-codex/tree/cleanupdisc

My point is just that while my first version was pretty ugly, it was functionally closer to the end goal, imo. And while your PR was cleaner, it was lacking functionality


Just to clarify, because I'm not sure we agree on this at this point: **if a block is not in your IWANTs, you don't know if a peer has it or not**
Sources:
- https://github.com/ipfs/go-bitswap/blob/master/docs/how-bitswap-works.md#processing-incoming-blocks
- https://github.com/status-im/nim-codex/blob/ce59dbd4a21ffe7d0eb9e1fc81c2d9d7cbc0fb30/dagger/blockexchange/engine.nim#L323
- https://github.com/status-im/nim-codex/blob/ce59dbd4a21ffe7d0eb9e1fc81c2d9d7cbc0fb30/dagger/blockexchange/engine.nim#L532-L538

And that's a very good thing: you don't want to send the full list of blocks you hold to your peers, only the one they are interested in


So, the process to get a block is:
- (check if a peer already sent you an IHAVE, but that shouldn't happen in normal circumstances)
- Send IWANTs to every peer you are connected to, + every new peer
- Wait for them to reply, or a timeout to pass
- Start DHT queries to get new peers. When you connect to them, the full IWANT list will be sent to them and they'll tell if you they have a block you want

When a peer sends you an IHAVE, request the block from him, or add it to the list of providers and do something smarter here

So, if we look at this branch:
https://github.com/status-im/nim-codex/blob/8afad9843f867e3e29f0516d14239daf3ec5779a/dagger/blockexchange/engine.nim#L249-L253
Shouldn't happen in normal circumstances

https://github.com/status-im/nim-codex/blob/8afad9843f867e3e29f0516d14239daf3ec5779a/dagger/blockexchange/engine.nim#L256-L257
This is very wrong, and the reason why most tests are passing at the moment
This says "If no peer has a block, ask the block to our first peer". Indeed, in most test, we have a single peer, so we'll ask him a block and it will work. But in reality, your first peer is very unlikely to have the block you want

https://github.com/status-im/nim-codex/blob/8afad9843f867e3e29f0516d14239daf3ec5779a/dagger/blockexchange/engine.nim#L277-L285
This is what most, if not all block, will hit. This sends IWANTs to your connected peers, which is good, but then the process finishes. If one of your peer has it, great, otherwise you'll have to wait for the 30s loop of discovery to kick in
Which is why the test I tweaked in #87 takes 30s to finish, and that's IMO the big issue with this version


> This isn't accurate. The current implementation uses an asyncqueue for backpressure, the queue will hold up to 100 elements and there are 10 concurrent tasks reading from the queue. This will only fire up to 10 concurrent requests and preload up to 100 blocks to discover or advertise.

That wasn't my point. My point is that we need to control the rate of advertisement, to advertise every block once in each TTL of the DHT (more or less)
You're system just shoot advertisement as fast as the concurrent task allows, without any regard to the TTL timing.


Unrelated, found an issue in my original test: https://github.com/status-im/nim-codex/commit/9eb3b61c6e01874d4e87067a826c943a2a9c5824#diff-faa7d8ea42e04c3c85781a925751d801dbf71237e3693c74f48eac5c633b1133R302


Anyway, if you manage to pass the test instantly-ish, and leave the logs working, I'm not against merging this PR, since most of my other remarks are about the block pricing, which is in the future anyway


And just to clarify, my system works like that:
- Send IWANT to every connected peer, waiting for their response
- When we have less than 3 inflight IWANT, start the DHT, and shoot requests every 30s + request time
  - 3 is a magic number, but basically my thinking is that most dataset should be on at least 3 nodes
- When a peer has the block, add them to the list, and pause the discovery process

As you mentioned, it doesn't have any limits in terms of concurrent request, but that should be easy to add with a semaphore, and actually more straightforward than workers. But IMO that's not a priority: if a block is being downloaded, it's because the user asked for it, and it should be done as fast as possible

dryajov:
> Just to clarify, because I'm not sure we agree on this at this point: if a block is not in your IWANTs, you don't know if a peer has it or not

Yes, definitely your IHAVE list can be pretty big, in the order of hundreds of MBs, there is no need to ask for it until you know exactly what you need. 

As far as I can tell from your outline, your interpretation seems correct.

> My point is just that while my first version was pretty ugly, it was functionally closer to the end goal, imo. And while your PR was cleaner, it was lacking functionality

It probably was and I don't necessarily claim it was wrong, the problem was that it changed how the engine operated in a manner that became intractable (at least for me). My gut feeling with this particular set of functionality is that it is _subtle_ and it's easy to get wrong, thus I strived to make it as straightforward as possible and have always chosen the simplest solution at hand and avoided overcomplicating as much as possible.

> This is very wrong, and the reason why most tests are passing at the moment

I think there is some confusion, this is sending a request for a block optimistically, in the case we failed to identify a peer with an explicit have for that block, we're picking the first one for lack of any better metrics at this point and (arguably), once we have peer scores in place, this should go the the peer with the higher score. As far as tests passing, we've got explicit tests for this particular scenario. I also think that you might have ran with an older version of the tests, because we aren't anymore awaiting `30s` for the discovery tests to pass, there was a bug in the previous version that I think is fixed here, but if you're still running into the same issue it's worth taking another look - in any case, it should not wait for 30 seconds, because we're firing discovery explicitly in the `requestBlocks` call, the `30s` timeout is a timeout in between iwant/ihave DHT requests, nothing more.

> If one of your peer has it, great, otherwise you'll have to wait for the 30s loop of discovery to kick in
Which is why the test I tweaked in https://github.com/status-im/nim-codex/pull/87 takes 30s to finish, and that's IMO the big issue with this version

I partially addressed it in the previous paragraph, but I'll point it out again, we're only awaiting in between iwant/ihave requests, in the `requestBlock` call we're scheduling a DHT request explicitly, which bypasses the `30s` sleep.

> Anyway, if you manage to pass the test instantly-ish, and leave the logs working, I'm not against merging this PR, since most of my other remarks are about the block pricing, which is in the future anyway

As far as I can tell, the tests are passing instantly right now, but it's worth a check, maybe it's something particular to your system, if so it needs fixing obviously. As far as the logs go, if you managed to make the dynamic syncs work as expected, then I can revert my change here - LMK.

> but that should be easy to add with a semaphore, and actually more straightforward than workers.

Heh, my first attempt was with the semaphore, I also thought it would be more straightforward ;)

Menduist:
Pushed my failing test, and corrected the other one.

PR #91, opened 2022-04-25
[contracts] Make contract interactions optional in the node

markspanbroek:
Adds --eth-deployment parameter to specify where the deployment json file is located.
Ensures that the node does not crash when deployment json is missing or incorrect.

### DIFF
--- a/dagger/conf.nim
+++ b/dagger/conf.nim
@@ -154,6 +154,12 @@ type
         name: "eth-account"
       .}: EthAddress
 
+      ethDeployment* {.
+        desc: "The json file describing the contract deployment"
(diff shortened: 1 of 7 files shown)


dagger/conf.nim
```diff
@@ -154,6 +154,12 @@ type
         name: "eth-account"
       .}: EthAddress
 
+      ethDeployment* {.
+        desc: "The json file describing the contract deployment"
+        defaultValue: string.default
+        name: "eth-deployment"
+      .}: string
```
Menduist, line 161: Confutils does work with optionals

Menduist (approved):

PR #107, opened 2022-05-23
[chore] bump nim-ethers to 0.1.7

emizzle:
Bump nim-ethers to v0.1.7 which includes the ability to wait for transaction confirmations.


dryajov:
@emizzle I think this needs some adjustment, any chance we can get this building?

emizzle:
Should be all fixed up now @dryajov.

PR #125, opened 2022-06-27
WIP: Harmonize BlockStore API

Bulat-Ziganshin:
Test PR from a newhire

### DIFF
--- a/codex/stores/cachestore.nim
+++ b/codex/stores/cachestore.nim
@@ -109,23 +109,20 @@ method putBlock*(
 
 method delBlock*(
   self: CacheStore,
-  cid: Cid): Future[bool] {.async.} =
-  ## delete a block/s from the block store
+  cid: Cid): Future[?!void] {.async.} =
+  ## Delete a block from the blockstore
   ##
 
   trace "Deleting block from cache", cid
   if cid.isEmpty:
     trace "Empty block, ignoring"
-    return true
+    return success()
 
-  try:
-    let removed = self.cache.del(cid)
-    if removed.isSome:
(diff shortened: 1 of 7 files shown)


codex/stores/cachestore.nim
```diff
@@ -109,23 +109,20 @@ method putBlock*(
...
 
-  try:
-    let removed = self.cache.del(cid)
-    if removed.isSome:
-      self.currentSize -= removed.get.data.len
-      return true
-    return false
-  except EmptyLruCacheError:
-    return false
+  let removed = self.cache.del(cid)
```
dryajov, line 121: This should use `.catch` to convert the exception to a result, otherwise this will propagate the exception up the stack without much compile time checking.

Bulat-Ziganshin: The current implementation of cache.del function doesn't raise EmptyLruCacheError: https://github.com/status-im/lrucache.nim/blob/master/src/lrucache.nim#L75

Overall, it seems that, in this implementation, only functions explicitly dealing with LRU entries, such as removeLru, can raise this exception

dryajov (approved):
Looks good, I would merge this and continue with the remaining of the "harmonization" in other PRs.

Bulat-Ziganshin:
> Looks good, I would merge this and continue with the remaining of the "harmonization" in other PRs.

ok

dryajov:
> > Looks good, I would merge this and continue with the remaining of the "harmonization" in other PRs.
> 
> ok. should I merge it myself?

Yes, usually we let the PR author do the merge.

PR #128, opened 2022-06-30
windows: workaround to prevent random segfaults

michaelsbradleyjr:


Bulat-Ziganshin:
I would be happy to see it merged

PR #133, opened 2022-07-04
Revert "Bump libbacktrace and secp256k1"

yyoncho:
Reverts status-im/nim-codex#126

Seems like this is breaking the CI.


michaelsbradleyjr:
It's not necessary to revert #126. The macOS builds are failing for an unrelated reason. There is a fix for that problem included in #129, see [L150](https://github.com/status-im/nim-codex/blob/ci/msys2/.github/workflows/ci.yml#L150).

yyoncho:
@michaelsbradleyjr thank you

Issue #137, opened 2022-07-12
Multitier BlockStore


Bulat-Ziganshin:
Implement BlockStore that combines [multiple levels](https://github.com/status-im/nim-datastore/blob/master/datastore/tiered_datastore.nim) and implements some common functionality (f.e. processing of empty blocks) that's now repeated in every BlockStore implementation.

Issue #131, opened 2022-07-01
SQLite store performance tests


Bulat-Ziganshin:
Once SQLite store will be available, we will need to have a benchmark measuring its performance for our tasks, and ideally - comparison with the FileStore backend.

Bulat-Ziganshin:
Testing scenarios can be complex. I propose to develop testing API first. Since ATM we never delete blocks, it just have to test reading and writing in different proportions:
- `write_blocks(N): seq[Cid]`:
  - generate N random blocks
  - optionally compute their CIDs (and return them from the function)
  - optionally write blocks to DB
- `read_blocks(seq[Cid])`
- `list_blocks(): seq[Cid]`

And auxiliary functions:
- `write_list(filename, seq[Cid])` - store Cids to the file
- `read_list(filename): seq[Cid]`
- `shuffle(var seq[auto])` - reshuflle list
- `random_extract(seq[auto], N)` - extract N random elements from the list, possible repeated
- reverse, sort, take(N) - should be in sequtils?

We can also make an advanced function, mixing writing and reading blocks in given proportion:
- read_write_blocks(init: seq[Cid], R, W): seq[Cid]` - given a database with `init` records, make R reading and W writing requests in random order and return resulting list of blocks

Bulat-Ziganshin:
The next question is how to implement scripts. AFAIK, using Nim for scripting will mean that the entire codebase will be recompiled each time we change a script. Also, we need a system for fast experiments.

So I propose to start with very simple cmdline API: `w150` means write 150 blocks, `r` means read all blocks, and `r150` means read 150 random blocks.

emizzle:
Discussion needed.

dryajov:
> Testing scenarios can be complex.

I would keep things as simple as possible without overcomplicating anything. 

We already have some basic unit tests for our datastores https://github.com/status-im/nim-codex/blob/main/tests/codex/stores/, lets take this as a baseline and create a parametrizable benchmark that we can launch with a combination of parameters. 

In particular I would like to see a comparison between the FS and SQL stores. As for the actual benchmark instrumentation, we can use something extremely simple as depicted here - https://stackoverflow.com/questions/36577570/how-to-benchmark-few-lines-of-code-in-nim. 

We don't have to recompile the entire project, we can compile and run the benchmark with a simple `nim c -r benchmark/storebench.nim -s=<sie-of-block> -n<num-of-blocks> ...`

So to summarize this is what I would suggest:

[...]

PR #160, opened 2022-07-27
add SQLiteStore backend

### LINKED ISSUE #138: Add SQLite BlockStore backend
Make BlockStore backend that uses [Nim-datastore](https://github.com/status-im/nim-datastore)

- [ ] https://github.com/status-im/nim-datastore/milestone/1

michaelsbradleyjr:
This PR adds an implementation of `SQLiteStore` that can function as a drop-in replacement for `FSStore`, and I've made that replacement in `codex/codex.nim`.

During the process I looked for opportunities to integrate nim-datastore more directly, i.e. as opposed to wrapping `type SQliteDatastore` inside of this PR's `type SQLiteStore`. That should be possible, but will be a more significant set of changes and there should be additional discussion. I think a good context to consider such changes is #137.

Also, the drop-in approach will allow for a more direct performance comparison per #131.

Closes #138.

### DIFF
--- a/codex/codex.nim
+++ b/codex/codex.nim
@@ -118,7 +118,7 @@ proc new*(T: type CodexServer, config: CodexConf): T =
 
     wallet = WalletRef.new(EthPrivateKey.random())
     network = BlockExcNetwork.new(switch)
-    localStore = FSStore.new(config.dataDir / "repo", cache = cache)
+    localStore = SQLiteStore.new(config.dataDir / "repo", cache = cache)
     peerStore = PeerCtxStore.new()
     pendingBlocks = PendingBlocksManager.new()
     discovery = DiscoveryEngine.new(localStore, peerStore, network, blockDiscovery, pendingBlocks)
--- a/codex/stores/blockstore.nim
+++ b/codex/stores/blockstore.nim
@@ -54,6 +54,13 @@ method listBlocks*(self: BlockStore, onBlock: OnBlock): Future[?!void] {.base.} 
 
   raiseAssert("Not implemented!")
 
+method close*(self: Blockstore): Future[void] {.base.} =
+  ## Close the blockstore, cleaning up resources managed by it.
+  ## For some implementations this may be a no-op
+  ##
+
+  raiseAssert("Not implemented!")
+
 proc contains*(self: BlockStore, blk: Cid): Future[bool] {.async.} =
   ## Check if the block exists in the blockstore.
   ## Return false if error encountered
--- a/codex/stores/sqlitestore.nim
+++ b/codex/stores/sqlitestore.nim
@@ -1,0 +1,209 @@
+## Nim-Codex
+## Copyright (c) 2021 Status Research & Development GmbH
+## Licensed under either of
+##  * Apache License, version 2.0, ([LICENSE-APACHE](LICENSE-APACHE))
+##  * MIT license ([LICENSE-MIT](LICENSE-MIT))
+## at your option.
+## This file may not be copied, modified, or distributed except according to
+## those terms.
+
+import pkg/upraises
+
+push: {.upraises: [].}
+
+import std/options
+
+import pkg/chronos
+import pkg/chronicles
+import pkg/datastore/sqlite_datastore
+import pkg/libp2p
+import pkg/questionable
+import pkg/questionable/results
+
+import ./blockstore
+import ./cachestore
+
+export blockstore, sqlite_datastore
+
+logScope:
+  topics = "codex sqlitestore"
+
+type
+  SQLiteStore* = ref object of BlockStore
+    cache: BlockStore
+    datastore: SQLiteDatastore
+
+const
+  allBlocks = when (let keyRes = Key.init("*"); true):
+    if keyRes.isOk: Query.init(keyRes.get)
+    else: raise (ref Defect)(msg: keyRes.error.msg)
+
+proc new*(
+  T: type SQLiteStore,
+  repoDir: string,
+  cache: BlockStore = CacheStore.new()): T =
+
+  let
+    datastoreRes = SQLiteDatastore.new(repoDir)
+
+  if datastoreRes.isErr:
+    raise (ref Defect)(msg: datastoreRes.error.msg)
+
+  T(cache: cache, datastore: datastoreRes.get)
+
+proc datastore*(self: SQLiteStore): SQLiteDatastore =
+  self.datastore
+
+proc blockKey*(blockCid: Cid): ?!Key =
+  let
+    keyRes = Key.init($blockCid)
+
+  if keyRes.isErr:
+    trace "Unable to construct CID from key", cid = blockCid, error = keyRes.error.msg
+
+  keyRes
+
+method getBlock*(
+  self: SQLiteStore,
+  cid: Cid): Future[?!(?Block)] {.async.} =
+  ## Get a block from the cache or database.
+  ## Save a copy to the cache if present in the database but not in the cache
+  ##
+
+  trace "Getting block from cache or database", cid
+
+  if cid.isEmpty:
+    trace "Empty block, ignoring"
+    return success cid.emptyBlock.some
+
+  without cachedBlkOpt =? await self.cache.getBlock(cid), error:
+    trace "Unable to read block from cache", cid, error = error.msg
+
+  if cachedBlkOpt.isSome:
+    return success cachedBlkOpt
+
+  without blkKey =? blockKey(cid), error:
+    return failure error
+
+  without dataOpt =? await self.datastore.get(blkKey), error:
+    trace "Unable to read block from database", key = blkKey.id, error = error.msg
+    return failure error
+
+  without data =? dataOpt:
+    return success Block.none
+
+  without blk =? Block.new(cid, data), error:
+    trace "Unable to construct block from data", cid, error = error.msg
+    return failure error
+
+  let
+    putCachedRes = await self.cache.putBlock(blk)
+
+  if putCachedRes.isErr:
+    trace "Unable to store block in cache", cid, error = putCachedRes.error.msg
+
+  return success blk.some
+
+method putBlock*(
+  self: SQLiteStore,
+  blk: Block): Future[?!void] {.async.} =
+  ## Write a block's contents to the database with key based on blk.cid.
+  ## Save a copy to the cache
+  ##
+
+  trace "Putting block into database and cache", cid = blk.cid
+
+  if blk.isEmpty:
+    trace "Empty block, ignoring"
+    return success()
+
+  without blkKey =? blockKey(blk.cid), error:
+    return failure error
+
+  let
+    putRes = await self.datastore.put(blkKey, blk.data)
+
+  if putRes.isErr:
+    trace "Unable to store block in database", key = blkKey.id, error = putRes.error.msg
+    return failure putRes.error
+
+  let
+    putCachedRes = await self.cache.putBlock(blk)
+
+  if putCachedRes.isErr:
+    trace "Unable to store block in cache", cid = blk.cid, error = putCachedRes.error.msg
+
+  return success()
+
+method delBlock*(
+  self: SQLiteStore,
+  cid: Cid): Future[?!void] {.async.} =
+  ## Delete a block from the database and cache
+  ##
+
+  trace "Deleting block from cache and database", cid
+
+  if cid.isEmpty:
+    trace "Empty block, ignoring"
+    return success()
+
+  let
+    delCachedRes = await self.cache.delBlock(cid)
+
+  if delCachedRes.isErr:
+    trace "Unable to delete block from cache", cid, error = delCachedRes.error.msg
+
+  without blkKey =? blockKey(cid), error:
+    return failure error
+
+  let
+    delRes = await self.datastore.delete(blkKey)
+
+  if delRes.isErr:
+    trace "Unable to delete block from database", key = blkKey.id, error = delRes.error.msg
+    return failure delRes.error
+
+  return success()
+
+method hasBlock*(
+  self: SQLiteStore,
+  cid: Cid): Future[?!bool] {.async.} =
+  ## Check if a block exists in the database
+  ##
+
+  trace "Checking database for block existence", cid
+
+  if cid.isEmpty:
+    trace "Empty block, ignoring"
+    return true.success
+
+  without blkKey =? blockKey(cid), error:
+    return failure error
+
+  return await self.datastore.contains(blkKey)
+
+method listBlocks*(
+  self: SQLiteStore,
+  onBlock: OnBlock): Future[?!void] {.async.} =
+  ## Process list of all blocks in the database via callback.
+  ## This is an intensive operation
+  ##
+
+  for kd in self.datastore.query(allBlocks):
+    let
+      (key, _) = await kd
+      cidRes = Cid.init(key.name)
+
+    if cidRes.isOk:
+      await onBlock(cidRes.get)
+    else:
+      trace "Unable to construct CID from key", key = key.id, error = $cidRes.error
+
+  return success()
+
+method close*(self: SQLiteStore): Future[void] {.async.} =
+  ## Close the underlying cache and SQLite datastore
+  ##
+
+  if not self.cache.isNil: await self.cache.close
+  if not self.datastore.isNil: self.datastore.close
--- a/tests/codex/stores/testsqlitestore.nim
+++ b/tests/codex/stores/testsqlitestore.nim
@@ -1,0 +1,225 @@
+import std/oids
+import std/options
+import std/os
+import std/random
+import std/sequtils
+import std/sets
+
+import pkg/asynctest
+import pkg/chronos
+import pkg/stew/byteutils
+
+import pkg/codex/chunker
+import pkg/codex/blocktype as bt
+import pkg/codex/stores
+
+import ../helpers
+
+suite "SQLite Store":
+  randomize()
+
+  var
+    store: SQLiteStore
+
+  let
+    repoDir = getAppDir() / "repo"
+
+  proc randomBlock(): bt.Block =
+    let
+      blockRes = bt.Block.new(($genOid()).toBytes)
+
+    require(blockRes.isOk)
+    blockRes.get
+
+  var
+    newBlock: bt.Block
+
+  setup:
+    removeDir(repoDir)
+    require(not dirExists(repoDir))
+    store = SQLiteStore.new(repoDir)
+    newBlock = randomBlock()
+
+  teardown:
+    if not store.isNil: await store.close
+    store = nil
+    removeDir(repoDir)
+    require(not dirExists(repoDir))
+
+  test "putBlock":
+    let
+      blkKeyRes = blockKey(newBlock.cid)
+
+    assert blkKeyRes.isOk
+
+    let
+      blkKey = blkKeyRes.get
+
+    var
+      # bypass cache
+      containsRes = await store.datastore.contains(blkKey)
+
+    assert containsRes.isOk
+    assert not containsRes.get
+
+    let
+      putRes = await store.putBlock(newBlock)
+
+    check: putRes.isOk
+
+    # bypass cache
+    containsRes = await store.datastore.contains(blkKey)
+
+    assert containsRes.isOk
+
+    check: containsRes.get
+
+  test "getBlock":
+    var
+      r = rand(100)
+
+    # put `r` number of random blocks before putting newBlock
+    if r > 0:
+      for _ in 0..r:
+        let
+          b = randomBlock()
+          kRes = blockKey(b.cid)
+
+        assert kRes.isOk
+
+        let
+          # bypass cache
+          pRes = await store.datastore.put(kRes.get, b.data)
+
+        assert pRes.isOk
+
+    let
+      blkKeyRes = blockKey(newBlock.cid)
+
+    assert blkKeyRes.isOk
+
+    var
+      # bypass cache
+      putRes = await store.datastore.put(blkKeyRes.get, newBlock.data)
+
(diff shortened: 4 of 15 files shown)


codex/stores/sqlitestore.nim
```diff
@@ -0,0 +1,204 @@
...
+  trace "Getting block from database", cid
+  if cid.isEmpty:
+    trace "Empty block, ignoring"
+    return success cid.emptyBlock.some
+
+  let
+    getCachedRes = await self.cache.getBlock(cid)
+
+  if getCachedRes.isErr:
+    trace "Unable to read block from cache", cid, error = getCachedRes.error.msg
```
michaelsbradleyjr, line 68: I made this symmetric with the "trace but don't fail" logic of L95-L96.

For `FSStore` when there's an error reading from the cache it does `return failure`

https://github.com/status-im/nim-codex/blob/7cfb157d327624d848a09c2df4eda20ea677a8e4/codex/stores/fsstore.nim#L49-L51

And when there's an error writing to the cache it only traces

https://github.com/status-im/nim-codex/blob/7cfb157d327624d848a09c2df4eda20ea677a8e4/codex/stores/fsstore.nim#L101-L102

I think it would be best to be consistent re: errors when accessing the cache, but is it better to only trace or trace + `return failure`?

codex/stores/sqlitestore.nim
```diff
@@ -0,0 +1,204 @@
...
+  if delRes.isErr:
+    trace "Unable to delete block", key = blkKey.id, error = delRes.error.msg
+    return failure "Unable to delete block"
+
+  let
+    delCachedRes = await self.cache.delBlock(cid)
+
+  if delCachedRes.isErr:
+    trace "Unable to delete block from cache", cid,
+      error = delCachedRes.error.msg
```
michaelsbradleyjr, line 152: Here I also only trace, but `FSSstore` returns the failure if there was an error deleting from the cache, which is at odds with the behavior of `putBlock` for `FSStore`

https://github.com/status-im/nim-codex/blob/7cfb157d327624d848a09c2df4eda20ea677a8e4/codex/stores/fsstore.nim#L124

codex/stores/sqlitestore.nim
```diff
@@ -0,0 +1,204 @@
...
+    let
+      (key, _) = await kd
+      cidRes = Cid.init(key.name)
+
+    if cidRes.isOk:
+      let
+        cid = cidRes.get
+
+      try:
+        await onBlock(cid)
```
michaelsbradleyjr, line 186: `type OnBlock` is marked with `upraises: []`

https://github.com/status-im/nim-codex/blob/7cfb157d327624d848a09c2df4eda20ea677a8e4/codex/stores/blockstore.nim#L24

So do we really need to `try/except` here?

I copied that logic from `FSSstore`, but locally confirmed that it compiles and runs correctly when `await onBlock(cid)` is not wrapped in `try`.

Bulat-Ziganshin:
Answering your questions - we don't have consistent logic. We need to decide which logic we should use.

So I'm OK with any reasonable behavior for now.

The code is good, but I'm not happy with tests:
1. since we have always-enabled cache, it's hard to test DB itself
2. most tests put just one block to Store, I think that more reasonable test should put multiple blocks in order to check whether DB extracts correct block by key and not just the last block pushed to DB.

Yes, I know that other Stores weren't tested that hard, but this one will become our primary store. So, if you need to merge it faster, we can implement this in later PR. If you have time now, it will be useful improvement.

Can we f.e. instead of using 3 fixed blocks, generate 100 blocks programmatically?

codex/stores/sqlitestore.nim
```diff
@@ -0,0 +1,204 @@
...
+logScope:
+  topics = "codex sqlitestore"
+
+type
+  SQLiteStore* = ref object of BlockStore
+    cache: BlockStore
+    datastore: SQLiteDatastore
+
+const
+  allBlocks = Query.init(Key.init("*").expect("should not fail"))
```
dryajov, line 37: I'm really not a fan of this type of error handling, with `expect`, this is extremely fragile.

codex/stores/sqlitestore.nim
```diff
@@ -0,0 +1,204 @@
...
+  let
+    blkKey = blockKey(cid)
+    delRes = await self.datastore.delete(blkKey)
+
+  if delRes.isErr:
+    trace "Unable to delete block", key = blkKey.id, error = delRes.error.msg
+    return failure "Unable to delete block"
+
+  let
+    delCachedRes = await self.cache.delBlock(cid)
```
dryajov, line 148: I would do this first, if we fail dropping the record from the sql store, we should still drop if from the cache.

codex/stores/sqlitestore.nim
dryajov: I think we should be fine removing the try/except from both

tests/codex/stores/testsqlitestore.nim
```diff
@@ -0,0 +1,150 @@
...
+    store: SQLiteStore
+
+  let
+    newBlock = bt.Block.new("New Block".toBytes()).expect("should not fail")
+    repoDir = getAppDir() / "repo"
+
+  setup:
+    removeDir(repoDir)
+    require(not dirExists(repoDir))
+    store = SQLiteStore.new(repoDir)
```
dryajov, line 26: I don't think that the `SQLiteStore` should be messing around with creating directories, I would move that out.

dryajov (commented):
LGTM except for the comments I left - specially the error handling with `expect`.

I would also rework `SQLiteStore` to not create directories and similar and instead take a path (and check that it exists, but no more than that) - it really isn't something that it should be concerned with.

codex/stores/sqlitestore.nim
Bulat-Ziganshin: @dryajov what you propose to use instead? Michael repalced `expect` with explicit rasising of Defect, which is the same action, just with longer code. Overall, we have a lot of `raise Defect` in nim-datastore package since we don't yet decided how to deal with all those situations

May be keep it awhile so we can start performance testing, and return back in a few weeks?

codex/stores/sqlitestore.nim
michaelsbradleyjr: Done.

codex/stores/sqlitestore.nim
michaelsbradleyjr: I've removed the usage of `except` I introduced in this module and in the tests. In the case of `proc new*(T: type SQLiteStore, ...): T`, I now raise the error if construction of `SQLiteDatastore` fails.

https://github.com/status-im/nim-codex/blob/3ddcb3b9f422cd663fe3ee3d0c23753dd76f1f7f/codex/stores/sqlitestore.nim#L46-L50

We could instead make the return type `?!T` but then result handling will ripple outward, working against this initial version of the `SQLiteStore` backend being a drop-in replacement for `FSSstore`. However, I'll be happy to make those additional changes if that's preferred.

codex/stores/sqlitestore.nim
michaelsbradleyjr: Changed.

tests/codex/stores/testsqlitestore.nim
michaelsbradleyjr: `SQLiteStore` does not create a directory, at least not directly; rather, that's handle by nim-datastore's `SQLiteDatastore` constructor:

https://github.com/status-im/nim-datastore/blob/master/datastore/sqlite_datastore.nim#L201-L211

We can change the behavior of nim-datastore, of course, but maybe it can be considered in separate issue/PR?

codex/stores/sqlitestore.nim
michaelsbradleyjr: > explicit rasising of Defect, which is the same action

It is the same action effectively, but a couple of things to consider:

Only in the case of defining the constant `allBlocks` and in `SQLiteStore` construction do I raise. In both cases that results in the original error message being displayed (the former at compile-time, latter at run-time) instead of the less helpful informative `should not fail` message.

In the other cases I added additional use of `trace` and return any failure for `Key.init`, also with the goal of making it more clear what went wrong.

michaelsbradleyjr:
> 1. since we have always-enabled cache, it's hard to test DB itself
> 2. most tests put just one block to Store, I think that more reasonable test should put multiple blocks in order to check whether DB extracts correct block by key and not just the last block pushed to DB.
> 
> Yes, I know that other Stores weren't tested that hard, but this one will become our primary store. So, if you need to merge it faster, we can implement this in later PR. If you have time now, it will be useful improvement.
> 
> Can we f.e. instead of using 3 fixed blocks, generate 100 blocks programmatically?

@Bulat-Ziganshin I have introduced randomness into the tests, as suggested. Where appropriate, I've also bypassed the cache by calling methods on `store.datastore` directly, which should increase confidence that we're testing the DB itself.

Bulat-Ziganshin:
@dryajov @michaelsbradleyjr Let's discuss the entire expect/Defect topic. 

First, if you apply `expect` to ?!T, the final error message includes both strings. So, the sole difference between `expect` and manual `raise Defect` is the longer code. Here's its definition:

```Nim
func expect*[T, E](self: Result[T, E], m: string): T =
  ## Return value of Result, or raise a `Defect` with the given message - use
  ## this helper to extract the value when an error is not expected, for example
  ## because the program logic dictates that the operation should never fail
  ##
  ## ```nim
  ## let r = Result[int, int].ok(42)
  ## # Put here a helpful comment why you think this won't fail
  ## echo r.expect("r was just set to ok(42)")
  ## ```
  if not self.o:
    when E isnot void:
      raiseResultDefect(m, self.e)
    else:
      raiseResultDefect(m)
  when T isnot void:
    self.v

func raiseResultDefect(m: string, v: auto) {.noreturn, noinline.} =
  mixin `$`
  when compiles($v): raise (ref ResultDefect)(msg: m & ": " & $v)    <== note this line
  else: raise (ref ResultDefect)(msg: m)
```


### SQLiteStore.new 

I think that the proper approach is to return ?!SQLiteStore and raise exception for it in the caller (codex.nim). Since error on DB open is absolutely possible runtime error, but Codex can't function without blockstore. Or we can raise exception in SQLiteStore.new itself, as we don't plan to export this code to other projects.


### blockKey function

Michael, you made blockKey return `?!Key` and then repeated the same `if blkKeyRes.isErr` block 3 times. Why not do it once in blockKey itself and return just `Key`?

Moreover, when blockKey can't construct proper key, this mean error in our program logic that results in unparseable keys: https://github.com/status-im/nim-datastore/blob/master/datastore/key.nim#L170-L194

This means that the entire program is broken and hardly can function at all. Looks like a proper candidate for Defect rather that modest `failure`.


### Use of =? constructs

Muchael, f.e. this code
```Nim
    getRes = await self.datastore.get(blkKey)

  if getRes.isErr:
    trace "Unable to read block", key = blkKey.id, error = getRes.error.msg
    return failure "Unable to read block"

  let
    dataOpt = getRes.get
```

can be replaced with
```Nim
  without dataOpt =? await self.datastore.get(blkKey), error:
    trace "Unable to read block", key = blkKey.id, error = error.msg
    return failure "Unable to read block"
```

Just look for `isErr`, in most cases it can be replaced with shorter code.

michaelsbradleyjr:
@Bulat-Ziganshin
> First, if you apply `expect` to ?!T, the final error message includes both strings. So, the sole difference between `expect` and manual `raise Defect` is the longer code. Here's its definition: ... <== note this line

I see what you're saying, and I can look into why it's not happening that way, but consider the compile-time stack trace we get with `expect` like this:

```nim
const
  allBlocks = Query.init(Key.init(":::").expect("should not fail"))
```

> stack trace: (most recent call last)
/path/to/repos/nim-codex/codex/stores/sqlitestore.nim(42, 41)
/path/to/repos/nim-codex/vendor/nim-stew/stew/results.nim(749, 24) expect
/path/to/repos/nim-codex/vendor/nim-stew/stew/results.nim(345, 9) raiseResultDefect
/path/to/repos/nim-codex/vendor/nim-stew/stew/results.nim(345, 9) ***Error: unhandled exception: should not fail [ResultDefect]***

But when manually raising `Defect`:

```nim
const
  allBlocks = when (let keyRes = Key.init(":::"); true):
    if keyRes.isOk: Query.init(keyRes.get)
    else: raise (ref Defect)(msg: keyRes.error.msg)
```

> stack trace: (most recent call last)
/path/to/repos/nim-codex/codex/stores/sqlitestore.nim(39, 11)
/path/to/repos/nim-codex/codex/stores/sqlitestore.nim(39, 11) ***Error: unhandled exception: id string contains an invalid Namespace: id string must not contain more than one delimiter ":" [Defect]***

The same thing plays out at runtime if I force an error re: `new` or `blockKey`: with `expect` the stack trace only has e.g. *"should not fail"*.

> I think that the proper approach is to return ?!SQLiteStore and raise exception for it in the caller (codex.nim). Since error on DB open is absolutely possible runtime error, but Codex can't function without blockstore. Or we can raise exception in SQLiteStore.new itself, as we don't plan to export this code to other projects.

`?!SQliteStore` may be the better/proper approach. By raising instead of propagating the value/error wrapped in `Result`, I kept it to a one-line swap in `codex.nim`, but we can pretty easily make the change to `?!`.

> Michael, you made blockKey return ?!Key and then repeated the same if blkKeyRes.isErr block 3 times. Why not do it once in blockKey itself and return just Key?

The usage of `trace` at least can be moved inside `blockKey` and it will cut down on repetition. We could swallow the error after `trace` (not a good idea) or raise inside the body of `blockKey` to further reduce the code. But if we're going to stick with `?!Key` as the return and we don't want to make `blockKey` a template instead of a proc, then we need to check `isErr` and return the failure in the body of the caller, though as you suggest later in your comment, there are some opportunities to cleanup using `without`, which I will attempt now.

> This means that the entire program is broken and hardly can function at all. Looks like a proper candidate for Defect rather that modest failure.

That was actually my motivation for using `expect` previously, though after seeing that `expect` swallows the original error message, I would do this:

```nim
proc blockKey*(blockCid: Cid): Key =
  without key =? Key.init($blockCid), error:
    raise (ref Defect)(msg: error.msg)

  key
```

Bulat-Ziganshin:
>consider the compile-time stack trace we get with expect

you are right - I didn't checked whether my idea works, but why you stopped here? why `expect` swallows failure message, what can we do to fix that?

> if we're going to stick with ?!Key as the return

blockKey is a helper function we are using only inside this module. So the best approach is to put extra functionality here rather than copy it 3 times. If we will ever need variant of blockKey returning ?!Key, we can add it at this time. But making 3 copies of the same code only to make blockKey more universal that we currently need is a kind of overengineering.

codex/stores/blockstore.nim
```diff
@@ -54,6 +54,13 @@ method listBlocks*(self: BlockStore, onBlock: OnBlock): Future[?!void] {.base.}
 
   raiseAssert("Not implemented!")
 
+method close*(self: Blockstore) {.base.} =
```
dryajov, line 57: We probably want `close` to be `async` as well.

codex/stores/sqlitestore.nim
dryajov: We probably want the same behaviour in FSStore either way

codex/codex.nim
```diff
@@ -118,7 +118,7 @@ proc new*(T: type CodexServer, config: CodexConf): T =
 
     wallet = WalletRef.new(EthPrivateKey.random())
     network = BlockExcNetwork.new(switch)
-    localStore = FSStore.new(config.dataDir / "repo", cache = cache)
+    localStore = SQLiteStore.new(config.dataDir / "repo", cache = cache)
```
dryajov, line 121: Maybe we should allow making this parametrizable and be able to select FS vs SQL, but we don't have to do this as part of this PR.

tests/codex/stores/testsqlitestore.nim
dryajov: We can address it as part of a separate PR, that should be fine, but this is definitely something to fix.

michaelsbradleyjr:
@Bulat-Ziganshin 
> blockKey is a helper function we are using only inside this module. So the best approach is to put extra functionality here rather than copy it 3 times

I moved the repeated `trace` inside `proc blockKey` in d85b3f9. And in 2220719 I used `without` / `=?` to make it a little less verbose.

So now we have this repeated a few times in the module:
```nim
without blkKey =? blockKey(cid), error:
  return failure error
```

It can be reduced further by making `proc blockKey` into a template or by returning `Key` instead of `?!Key` and raising Defect inside `proc blockKey` in the case of an error.

Or should we stick with `?!Key` and live with repeated `without...`?

Bulat-Ziganshin:
@michaelsbradleyjr I think that incorrect Key should be a Defect, because it means we incorrectly build keys and thus the entire block store layer fails to work.

Also, I think we should define `$(ResultFailure e) = e.msg` in our code till it will be resolved by upstream. It will allow to return back to `expect` unless @dryajov has some objections.

dryajov:
> I think that the proper approach is to return ?!SQLiteStore and raise exception for it in the caller (codex.nim). Since error on DB open is absolutely possible runtime error, but Codex can't function without blockstore. Or we can raise exception in SQLiteStore.new itself, as we don't plan to export this code to other projects.

I wouldn't be opposed to doing `new(...): ?!SQLiteStore` if needed, I agree that the store can fail on construction (connection error, etc...), but I don't think it is up to the store to "crash" the process with a `Defect` - I would allow the calling code to decide wether we want to stop or continue execution so returning a `Result` as we now commonly do is IMO a better approach.

This flexibility is particularly useful when testing as it allows simulating different scenarios, but it's not limited to that.

> This means that the entire program is broken and hardly can function at all. Looks like a proper candidate for Defect rather that modest failure.

We don't always control where the key comes from, for example, a `Cid` sent from the network might be malformed. We should handle that accordingly (handle error, log, etc...), but raising a `Defect` due to that would allow anyone to crash our application by simply sending us a malformed `Cid`s, probably not what we want. 

The example is a bit contrived, but in the context of network bound applications we should be more defensive than with user facing applications, so defaulting to the most severe error path is not always the best approach.

> Just look for isErr, in most cases it can be replaced with shorter code.

I agree, lets try to streamline this code using `without` (or similar).

Bulat-Ziganshin:
>a Cid sent from the network might be malformed

I just didn't thought about the network! So I agree to keep the current code awhile, but in long-term it raises two questions:
1. We give malicious user full access to our internal [Key-building machinery](https://github.com/status-im/nim-datastore/blob/master/datastore/key.nim#L181), thus he can build Key from any Namespace. Instead, we should use some Key.initStr that **checks** that string doesn't contain any special symbols at all. Just now we have a leaking abstraction.
2. It will be nice to check for malformed Keys right at receiving msg from external world. Well, proper BlockStore API would allow BlockStore operations only using special BlockKey datatype and protect the path "string -> BlockKey".

codex/stores/blockstore.nim
Bulat-Ziganshin: it's not a library, just an internal API. We don't have any async open, and we don't need async close **ATM**

Bulat-Ziganshin:
Michael, can you please modify SQLiteDatastore to support self.cache=nil (and use nil as default cache in `new`)? I think it may be useful for performance testing in order to exclude overheads of extra caching layer. The data are anyway are cached by OS in the filecache, so it's not necessarily will reduce the speed.

tests/codex/stores/testsqlitestore.nim
michaelsbradleyjr: Tracking in https://github.com/status-im/nim-codex/issues/178 and https://github.com/status-im/nim-datastore/issues/21.

codex/stores/blockstore.nim
michaelsbradleyjr: I had already made the change as suggested, but hadn't seen @Bulat-Ziganshin's comment until after I pushed the change some time later. It's self-contained in a `[wip]` commit so I can drop it, or we can keep it.

codex/codex.nim
michaelsbradleyjr: Tracking in #179.

michaelsbradleyjr:
@Bulat-Ziganshin 
> Instead, we should use some Key.initStr that checks that string doesn't contain any special symbols at all. Just now we have a leaking abstraction.

Tracking in https://github.com/status-im/nim-datastore/issues/22.

michaelsbradleyjr:
@Bulat-Ziganshin 
> can you please modify SQLiteDatastore to support self.cache=nil

Tracking in #180.

dryajov:
> We give malicious user full access to our internal [Key-building machinery](https://github.com/status-im/nim-datastore/blob/master/datastore/key.nim#L181), thus he can build Key from any Namespace. Instead, we should use some Key.initStr that checks that string doesn't contain any special symbols at all. Just now we have a leaking abstraction.

Great idea, we should definitely sanitize the keys before doing anything. 

> On the second thought, we already have this protection in some form - Cid can't be formed from malformed string, isn't it? But we don't enforce it yet through the type system, what we need is to allow only Cid (and in future other "good types") be used for BlockStore operations:

Our `BlockType` is either the `Block` or the `Cid` itself, both are safe types that can only be formed if all checks are passed. The `Cid` is checked here https://github.com/status-im/nim-libp2p/blob/master/libp2p/cid.nim#L74-L168 (thats why I said it was a contrived example) and the `Block` here https://github.com/status-im/nim-codex/blob/main/codex/blocktype.nim#L139-L156.

dryajov:
> PS: sending malformed requests makes a great idea for our testsuite too.

Yes, this is usually done with fuzzing, but we should definitely add it to our test suite at some point.

tests/codex/stores/testsqlitestore.nim
michaelsbradleyjr: See https://github.com/status-im/nim-datastore/pull/23.

PR #161, opened 2022-07-28
Contract store

dryajov:
Integrate block fetching with contracts

### DIFF
--- a/codex/node.nim
+++ b/codex/node.nim
@@ -302,21 +302,25 @@ proc start*(node: CodexNodeRef) {.async.} =
   if contracts =? node.contracts:
     # TODO: remove Sales callbacks, pass BlockStore and StorageProofs instead
     contracts.sales.onStore = proc(cid: string, _: Availability) {.async.} =
-      # store data in local storage
+      ## store data in local storage
+      ##
+
       without cid =? Cid.init(cid):
         trace "Unable to parse Cid", cid
-        return
+        raise newException(CodexError, "Unable to parse Cid")
 
-      without manifest =? await node.fetchManifest(cid):
+      without manifest =? await node.fetchManifest(cid), error:
         trace "Unable to fetch manifest for cid", cid
-        return
+        raise error
 
       trace "Fetching block for cid", cid
       let batch = max(1, manifest.blocks.len div FetchBatch)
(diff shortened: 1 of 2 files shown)


codex/node.nim
```diff
@@ -327,13 +310,25 @@ proc start*(node: CodexNodeRef) {.async.} =
     # TODO: remove Sales callbacks, pass BlockStore and StorageProofs instead
     contracts.sales.onStore = proc(cid: string, _: Availability) {.async.} =
       # store data in local storage
-      (await node.store(Cid.init(cid).tryGet())).tryGet()
+      without cid =? Cid.init(cid):
+        trace "Unable to parse Cid", cid
+        return
```
markspanbroek, line 308: The onStore callback needs to fail with an exception here, otherwise the sales agent will happily go along with trying to fill a slot, even though it doesn't have the data to generate a proof.

codex/node.nim
```diff
@@ -327,13 +310,25 @@ proc start*(node: CodexNodeRef) {.async.} =
...
+
+      without manifest =? await node.fetchManifest(cid):
+        trace "Unable to fetch manifest for cid", cid
```
markspanbroek, line 311: Same here

codex/node.nim
```diff
@@ -60,10 +60,29 @@ proc connect*(
...
+
+  try:
+    let
+      batch = max(1, manifest.blocks.len div batch)
+    trace "Prefetching in batches of", FetchBatch
+    for blks in manifest.blocks.distribute(batch, true):
+      discard await allFinished(
+        blks.mapIt( node.blockStore.getBlock( it ) ))
+  except CatchableError as exc:
+    trace "Exception prefetching blocks", exc = exc.msg
```
markspanbroek: The onStore callback needs to be able to see whether the fetching of the blocks failed or succeeded (see reasoning below).

markspanbroek:
Nice, I like the split between manifest and block fetching.

Bulat-Ziganshin:
be careful - first two commits picked from my PR and they shouldn't be here since I merged my PR to main

dryajov:
I've fixed the raise related issues, I think this is ready for review. I still need to simplify the handling of protected vs non-protected manifests, but I'll do that in a different PR.

PR #164, opened 2022-07-28
Multiple slots in a storage contract

markspanbroek:
Allows multiple slots in a storage contract, so that more than one host can contribute.

A step towards the design in https://github.com/status-im/codex-research/pull/83.
Uses the smart contract update from https://github.com/status-im/dagger-contracts/pull/13.

### DIFF
--- a/codex/contracts/market.nim
+++ b/codex/contracts/market.nim
@@ -42,18 +42,20 @@ method getRequest(market: OnChainMarket,
   else:
     return none StorageRequest
 
-method getHost(market: OnChainMarket,
-               id: array[32, byte]): Future[?Address] {.async.} =
-  let address = await market.contract.getHost(id)
-  if address != Address.default:
-    return some address
-  else:
-    return none Address
-
 method fulfillRequest(market: OnChainMarket,
                       requestId: array[32, byte],
                       proof: seq[byte]) {.async.} =
   await market.contract.fulfillRequest(requestId, proof)
+
+method getHost(market: OnChainMarket,
+               requestId: array[32, byte],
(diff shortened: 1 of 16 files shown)


codex/contracts/market.nim
```diff
@@ -43,17 +43,20 @@ method getRequest(market: OnChainMarket,
     return none StorageRequest
 
 method getHost(market: OnChainMarket,
-               id: array[32, byte]): Future[?Address] {.async.} =
-  let address = await market.contract.getHost(id)
+               requestId: array[32, byte],
```
dryajov, line 46: Should we use `distinct` for this datatypes?

codex/purchasing.nim
```diff
@@ -77,11 +76,8 @@ proc run(purchase: Purchase) {.async.} =
       done.complete()
     let request = purchase.request
     let subscription = await market.subscribeFulfillment(request.id, callback)
-    try:
-      await done
-      purchase.selected = await market.getHost(request.id)
-    finally:
-      await subscription.unsubscribe()
+    await done
```
dryajov, line 79: what if `done` throws, try/finally seems safer here?

dryajov (approved):
LGTM, except for the comments.

codex/purchasing.nim
markspanbroek: `done` is a variable local to this proc, that is only completed on line 76. It won't throw.

codex/contracts/market.nim
markspanbroek: Yes, that is a good idea. I'd like to address that in a separate PR, if that's ok.

PR #170, opened 2022-07-29
Better batching

dryajov:


Bulat-Ziganshin:
Comment: code refactoring + tests that size of each fetched blocks batch indeed <= batchSize

PR #181, opened 2022-08-04
Better command line options for Ethereum

### LINKED ISSUE #163: Remove hard requirement of reachable Ethereum provider
Owing to recent changes on `main`, when following the instructions in the Codex readme and running e.g. `./build/codex --data-dir="$(pwd)/Codex1" -i=127.0.0.1` the process crashes almost immediately if the default `--eth-provider` is not reachable.

It seems there's presently a hard requirement that an Ethereum RPC endpoint is reachable, the default endpoint being `ws://localhost:8545`.

The availability of an Ethereum provider should be opt-in.

markspanbroek:
All interactions with the Ethereum node are disabled by default. Only when the ~~`--enable-eth`~~ `--persistence` parameter is set, do we start interactions with the Ethereum node.

Includes improved error messages when parameters are missing, or the Ethereum node is not reachable.

fixes #163

### DIFF
--- a/codex/codex.nim
+++ b/codex/codex.nim
@@ -53,13 +53,13 @@ proc stop*(s: CodexServer) {.async.} =
   s.runHandle.complete()
 
 proc new(_: type ContractInteractions, config: CodexConf): ?ContractInteractions =
-  if not config.ethEnabled:
+  if not config.persistence:
     if config.ethAccount.isSome:
-      warn "Ethereum account was set, but Ethereum interaction are not enabled"
+      warn "Ethereum account was set, but persistence is not enabled"
     return
 
   without account =? config.ethAccount:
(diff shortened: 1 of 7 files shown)


dryajov:
LGTM, but maybe it would be better to make the flag more generic and perhaps turn it into a command? For example `persistence` and put all the purchasing/selling related flags behind it?

markspanbroek:
Thanks! I would not go for a command instead of a flag, because the command is more or less "start the node" and that doesn't change. We could indeed make the name of the flag more generic, for instance `--enable-persistence` or just plain `--persistence`.

dryajov:
> "start the node" and that doesn't change

Ah, your right, lets stick with a flag then.

PR #200, opened 2022-08-05
support self.cache=nil in SQLiteStore

### LINKED ISSUE #180: Support self.cache=nil in SQLiteStore
See https://github.com/status-im/nim-codex/pull/160#issuecomment-1201863491.

michaelsbradleyjr:
Also fix a discrepancy where cli option `--cache-size` is documented as `0 disables the cache` in `codex/conf.nim`, but previously the value `0` would result in a cache being constructed with default parameter values for `CacheStore.new()`

Closes #180.

### DIFF
--- a/codex/codex.nim
+++ b/codex/codex.nim
@@ -102,11 +102,11 @@ proc new*(T: type CodexServer, config: CodexConf): T =
     .withTcpTransport({ServerFlags.ReuseAddr})
     .build()
 
-  let cache =
-    if config.cacheSize > 0:
-      CacheStore.new(cacheSize = config.cacheSize * MiB)
-    else:
-      CacheStore.new()
+  var
+    cache: CacheStore
+
+  if config.cacheSize > 0:
+    cache = CacheStore.new(cacheSize = config.cacheSize * MiB)
 
   let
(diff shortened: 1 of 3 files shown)


michaelsbradleyjr:
At first I tried to do this:

```nim
type
  SQLiteStore* = ref object of BlockStore
    cache: ?BlockStore
    datastore: SQLiteDatastore

proc new*(
  T: type SQLiteStore,
  repoDir: string,
  cache: ?BlockStore = CacheStore.new()): T =
```

The idea being it would be nice to be able to disable the cache explicitly by passing `cache = CacheStore.none`, and `isSome` checks would take the place of `not isNil`.

However, `Option[BlockStore] isnot Option[CacheStore]` and I couldn't find a clean way of working around that, so I went with the original idea of `cache = nil` representing a disabled cache.

michaelsbradleyjr:
These changes can quickly be extended to `FSStore` as well, but I wanted to check the approach is what we want before doing that.

Bulat-Ziganshin:
I think this PR is also OK.

Bulat-Ziganshin (approved):

PR #201, opened 2022-08-07
custom query for listBlocks

michaelsbradleyjr:
I was reviewing some notes from my discussion last week with @dryajov and noted his observation that `SQLiteStore` seemed about 2x slower for reads compared to `FSStore`. On Friday I spent some time with nim-datastore re: https://github.com/status-im/questionable/issues/19, so it was fresh in my mind. I realized something that is probably impacting the read performance of `SQLiteStore` (cf. #160), which wraps nim-datastore's `SQLiteDatastore`.

nim-datastore's [`iterator query`](https://github.com/status-im/nim-datastore/blob/master/datastore/sqlite_datastore.nim#L352) for `SQLiteDatastore`  yields a `(key, data)` tuple, i.e. both the `id` and `data` columns are read for each matching row.

However, `method listBlocks` only needs each `Cid`, i.e. it only needs to read from the `id` column in the database and reading/processing `data` for each matching row is unnecessary. At the time I implemented `SQLiteStore` I was aware `data` wasn't needed, so I wrote this code:

https://github.com/status-im/nim-codex/blob/959d025f9bc6e50952dae4f581a293c7e1d280c8/codex/stores/sqlitestore.nim#L192-L194

We weren't benchmarking at the time, so I left it as something to possibly optimize in the future. But based on what Dmitriy said, I think it is affecting read performance.

In this PR I've introduced a custom iterator/query used by `method listBlocks` for `SQLiteStore` that only reads the `id` column for each matching row.

I think the behavior of `method query` for `SQLiteDatastore` (yielding `(key, data)`) makes sense for the general case, but the design of `nim-datastore` was/is purposely amenable to specializations by users; this PR is an example of that.

Related to #131.

### DIFF
--- a/codex/stores/sqlitestore.nim
+++ b/codex/stores/sqlitestore.nim
@@ -19,6 +19,7 @@ import pkg/datastore/sqlite_datastore
 import pkg/libp2p
 import pkg/questionable
 import pkg/questionable/results
+import pkg/sqlite3_abi
 
 import ./blockstore
 import ./cachestore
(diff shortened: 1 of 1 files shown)


Bulat-Ziganshin:
For efficiency, listBlocksQuery should return plain string rather than Future[Key]. It's internal code not supposed to be used from other places, there is no any need to make it universal, but we want to make it faster.

Also, remove the "where" part - it will make code faster and smaller.

michaelsbradleyjr:
> For efficiency, listBlocksQuery should return plain string rather than Future[Key]...

Good points. Since it's a custom query, it's not necessary to use `type Key` from nim-datastore, and because we're listing all the blocks there's no need for `WHERE...GLOB` matching.

Changes made.

Bulat-Ziganshin:
@michaelsbradleyjr listBlocksQuery don't need to be async. it's our internal iterator, so we can make it asynchronous anytime we will really need that

michaelsbradleyjr:
> listBlocksQuery don't need to be async

Change made.

Bulat-Ziganshin (approved):
For me, it's fine and ready for merge

PR #204, opened 2022-08-09
[build] update nim-datastore

### LINKED ISSUE #178: Update vendor/nim-datastore when "don't create directory" behavior is changed
See https://github.com/status-im/nim-datastore/issues/21.

michaelsbradleyjr:
Also make changes to `codex/codex.nim` and tests for `SQLiteStore` re: changes in nim-datastore.

Closes #178.

### DIFF
--- a/codex/codex.nim
+++ b/codex/codex.nim
@@ -133,7 +133,15 @@ proc new*(T: type CodexServer, config: CodexConf): T =
 
     wallet = WalletRef.new(EthPrivateKey.random())
     network = BlockExcNetwork.new(switch)
-    localStore = SQLiteStore.new(config.dataDir / "repo", cache = cache)
+    repoDir = config.dataDir / "repo"
+
+  if io2.createPath(repoDir).isErr:
+    trace "Unable to create data directory for block store", dataDir = repoDir
+    raise (ref Defect)(
+      msg: "Unable to create data directory for block store: " & repoDir)
+
+  let
+    localStore = SQLiteStore.new(repoDir, cache = cache)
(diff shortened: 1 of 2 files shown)


codex/codex.nim
```diff
@@ -133,7 +133,15 @@ proc new*(T: type CodexServer, config: CodexConf): T =
...
-    localStore = SQLiteStore.new(config.dataDir / "repo", cache = cache)
+    repoDir = config.dataDir / "repo"
+
+  if io2.createPath(repoDir).isErr:
+    trace "Unable to create data directory for block store", dataDir = repoDir
+    raise (ref Defect)(
+      msg: "Unable to create data directory for block store: " & repoDir)
+
+  let
+    localStore = SQLiteStore.new(repoDir, cache = cache)
```
michaelsbradleyjr, line 144: Since `SQLiteStore` (nor `SQLiteDatastore`) creates the directory where the SQLite database will be stored, it's necessary for the executable to do this.

It wasn't necessary to do this previously because `FSStore` implicitly handles directory creation
https://github.com/status-im/nim-codex/blob/a7fdb975df56b5be1e6f724389b182e31e7ccfc3/codex/stores/fsstore.nim#L89-L93

We could modify `FSStore.new` to check if `repoDir` exists and fail if it does not.

Bulat-Ziganshin:
Sorry, I can't review since I don't watch this discussion. So, up to Dmitry.

codex/codex.nim
dryajov: It's OK for now, but we'll probably move this to our repo abstraction once we have one.

PR #205, opened 2022-08-09
[stores] update getBlock return type

### LINKED ISSUE #177: Consistent behavior for cache-related errors in BlockStore impls

### LINKED ISSUE #182: Update getBlock return type

> getBlock: Future[?!Block]

michaelsbradleyjr:
Change return type for `method getBlock` from `Future[?!(?Block)]` to `Future[?!Block]`.

Closes #177.
Closes #182.


Bulat-Ziganshin:
I realized that there is a way to have both simple type for getBlock and recognition of "no block" situation. For that, we just need to have special exception type and return it when block wasn't found instead of "Block not found" string. What you think about it?

michaelsbradleyjr:
> we just need to have a special exception type and return it when a block wasn't found instead of the "Block not in ..." string. What do you think about it?

We could do that, or we could create error type/s that derive from `CatchableError` ***and*** do the equivalent with `stew/results` + `questionable` with the help of `when ... is`

See these [doc comments in `stew/results`](https://github.com/status-im/nim-stew/blob/master/stew/results.nim#L154-L167)
>  ##
    ## Considerations for the error type
    ##
    ## * Use a `string` or a `cstring` if you want to provide a diagnostic for
    ##   the caller without an expectation that they will differentiate between
    ##   different errors. Callers should never parse the given string!
    ## * Use an `enum` to provide in-depth errors where the caller is expected
    ##   to have different logic for different errors
    ## * Use a complex type to include error-specific meta-data - or make the
    ##   meta-data collection a visible part of your API in another way - this
    ##   way it remains discoverable by the caller!
    ##
    ## A natural "error API" progression is starting with `Opt[T]`, then
    ## `Result[T, cstring]`, `Result[T, enum]` and `Result[T, object]` in
    ## escalating order of complexity.

Just as the docs suggest, I started with `Option[Block]` (`?Block` inside `?!`) because it seemed useful to distinguish *"not found"* from error states.

However, per discussion under https://github.com/status-im/nim-codex/pull/130#discussion_r925990621, the goal of #182 was/is to remove the distinction.

michaelsbradleyjr:
> We could do that

Sorry, I read what you wrote too quickly and understood as

> have a special exception type and ***raise*** it when a block wasn't found

In effect, we're saying the same thing, and it's what the `stew/results` docs recommend, though those docs also recommend starting with `Option[T]` before moving on to something more complex.

Bulat-Ziganshin:
`?!T` == `Result[T, ref CatchableError]`. I propose to make special subtype of CatchableError - it will allow caller to distinguish whether we really have a problem or just there is no block with such Id.

and yes, it makes half step back - but actually allows to combine benefits of both approaches:
- code that don't need to distinguish NOT_FOUND and ERROR, will be simpler because it doesn't need to unpack two layers of indirection
- code that needs to disttinguish them, can do that by checking the exception type

I'm sorry that it didn't found this approach earlier. But I think that it's the best we can do to combine benefits of both ways ("?! ?" for clarity, and "?!" for simplicity).

michaelsbradleyjr:
Okay, so I propose to implement (will do it now)
```nim
type
  BlockNotFoundError = object of CatchableError
  BlockReadError = object of CatchableError
```
And refine caller behavior with respect to those types.

Bulat-Ziganshin:
Agree

dryajov:
> Okay, so I propose to implement (will do it now)
> 
> ```nim
> type
>   BlockNotFoundError = object of CatchableError
>   BlockReadError = object of CatchableError
> ```
> 
> And refine caller/callee behavior with respect to those types.

We actually already have a base type for this in https://github.com/status-im/nim-codex/blob/178_update_vendor_datastore/codex/errors.nim. I like this approach better than all others as well, lets go with that.

michaelsbradleyjr:
Closing in favor of #219.

PR #209, opened 2022-08-12
ALT PR: update getBlock return type

### LINKED ISSUE #177: Consistent behavior for cache-related errors in BlockStore impls

### LINKED ISSUE #182: Update getBlock return type

michaelsbradleyjr:
Change return type for `method getBlock` from `Future[?!(?Block)]` to `Future[BlockResult]`.

Closes #177.
Closes #182.

Alternative to #205.

---

This PR is an alternative to #205 per discussion under https://github.com/status-im/nim-codex/pull/205#issuecomment-1210864052.

I ran into some difficulties while working on this. My initial idea was to use `when...is` to discriminate return types, but I wasn't able to get that to work. I may have missed/misunderstood something re: what Nim allows for, so what I write below can possibly be disregarded, and/or the changes in this PR can possibly be reworked into something more desirable. Also, what I observed may be reducible to *"that's how Nim's type system works"*, but it wasn't obvious to me before I started experimenting, in which case it's a good learning experience, though in hindsight I saw something similar in #200 where I [noted](https://github.com/status-im/nim-codex/pull/200#issuecomment-1205899757) `Option[BlockStore] isnot Option[CacheStore]`.

My approach was to take the changes in #205 and rework them so that different error types can represent different conditions, and thereby callers can decide to behave differently depending on the error type.

That proved to be problematic for return types involving `Result[T, E]`, which are precisely what we're working with. Given types `BlockNotFoundError = ref object of BlockError` and `BlockReadError = ref object of BlockError`, both descending from `BlockError = ref object of CodexError`, it wasn't possible to use `when...is` in the caller to distinguish the error types given a return type of `Result[Block, ref BlockError]`.

I tried using type constructs such as `BlockNotFoundError | BlockReadError`, but in the context of `Result[...]` that doesn't work either.

What I ultimately did was introduce a `kind` field on `type BlockError`, which leverages `type BlockErrorKind = enum ...` so that errors can be discriminated at runtime.

That works! But it has a number of implications:

We need several `enum` members to distinguish different kinds of errors, versus using two distinctions (`some` or `none`) for different kinds of values in `Result[...]`

I spec'd these:
```nim
type
  BlockErrorKind* = enum
    BlockConstructErr,
    BlockKeyErr,
    BlockNetReqErr,
    BlockNotFoundErr,
    BlockReadErr
```

For our purposes at present, those could be reduced to:
```nim
type
  BlockErrorKind* = enum
    BlockNotFoundErr,
    BlockOtherErr,
    BlockReadErr
```

But adding an `enum` member seems cheap, i.e. compared to the other trade-offs of this PR vs. #205.

Our preferred error-handling library [`status-im/questionable`](https://github.com/status-im/questionable) makes some (pretty natural) assumptions such that return type `BlockResult = Result[Block, ref BlockError]` becomes problematic. The `without` helper becomes unusable for such return types, and it's necessary to use `ok` from `stew/results` instead of questionable's `success` because of a problem (undiagnosed) in combination with nim-chronos' `complete`. It's also sometimes necessary to be more verbose when using questionable's `failure`, i.e. `failure (ref BlockError)(kind: ..., msg: ...)`.

Those incompatibilities ripple outward, so more changes have to be made in this PR vs. #205.

It may be possible to resolve all these problems in an elegant manner, and as I suggested at the start, some of the problems may owe to my not considering other possibilities.

I look forward to discussion, but my current impression is that we should close this PR and stick with #205, or close #205 ***and*** this PR, sticking with `Future[?!(?Block)]`, though in the latter case there are some changes in this PR and #205 re: #177 that are worthwhile and can/should be extracted into a future PR.

### DIFF
--- a/codex/node.nim
+++ b/codex/node.nim
@@ -73,14 +73,13 @@ proc fetchManifest*(
     return failure "CID has invalid content type for manifest"
 
   trace "Received retrieval request", cid
-  without blkOrNone =? await node.blockStore.getBlock(cid), error:
-    return failure(error)
 
-  without blk =? blkOrNone:
-    trace "Block not found", cid
-    return failure("Block not found")
+  let
+    getRes = await node.blockStore.getBlock(cid)
 
-  without manifest =? Manifest.decode(blk):
+  if getRes.isErr: return failure getRes.error
+
+  without manifest =? Manifest.decode(getRes.get):
     return failure(
       newException(CodexError, "Unable to decode as manifest"))
 
@@ -107,7 +106,7 @@ proc fetchBatched*(
 
       await allFuturesThrowing(allFinished(blocks))
       if not onBatch.isNil:
-        await onBatch(blocks.mapIt( it.read.get.get ))
+        await onBatch(blocks.mapIt( it.read.get ))
     except CancelledError as exc:
       raise exc
     except CatchableError as exc:
--- a/codex/blockexchange/engine/engine.nim
+++ b/codex/blockexchange/engine/engine.nim
@@ -379,8 +379,8 @@ proc taskHandler*(b: BlockExcEngine, task: BlockExcPeerCtx) {.gcsafe, async.} =
 
     # Extract succesfully received blocks
     let blocks = blockFuts
-      .filterIt(it.completed and it.read.isOk and it.read.get.isSome)
-      .mapIt(it.read.get.get)
+      .filterIt(it.completed and it.read.isOk)
+      .mapIt(it.read.get)
 
     if blocks.len > 0:
       trace "Sending blocks to peer", peer = task.id, blocks = blocks.len
--- a/codex/blocktype.nim
+++ b/codex/blocktype.nim
@@ -29,6 +29,18 @@ type
     cid*: Cid
     data*: seq[byte]
 
+  BlockErrorKind* = enum
+    BlockConstructErr,
+    BlockKeyErr,
+    BlockNetReqErr,
+    BlockNotFoundErr,
+    BlockReadErr
+
+  BlockError* = object of CodexError
+    kind*: BlockErrorKind
+
+  BlockResult* = Result[Block, ref BlockError]
+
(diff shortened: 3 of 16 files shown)


codex/node.nim
```diff
@@ -73,14 +73,13 @@ proc fetchManifest*(
...
-  without blkOrNone =? await node.blockStore.getBlock(cid), error:
-    return failure(error)
 
-  without blk =? blkOrNone:
-    trace "Block not found", cid
-    return failure("Block not found")
+  let
+    getRes = await node.blockStore.getBlock(cid)
+
+  if getRes.isErr: return failure getRes.error
```
Bulat-Ziganshin, line 81:
```
without blk =? await node.blockStore.getBlock(cid), error:
  return failure(error)
```

michaelsbradleyjr:
```
/path/to/nim-codex/codex/node.nim(67, 35) template/generic instantiation of `async` from here
/path/to/nim-codex/codex/node.nim(77, 11) template/generic instantiation of `without` from here
/path/to/nim-codex/vendor/questionable/questionable/withoutresult.nim(37, 23) template/generic instantiation of `replaceInfix` from here
/path/to/nim-codex/codex/node.nim(77, 15) template/generic instantiation of `override=?`gensym236685202` from here
/path/to/nim-codex/vendor/questionable/questionable/withoutresult.nim(30, 18) Error: type mismatch: got <BlockResult>
but expected one of: 
proc isFailure[T](value: ?!T): bool
  first type mismatch at position: 1
  required type for value: Result[isFailure.T, ref CatchableError]
  but expression 'optional`gensym236685205`gensym236700001' is of type: BlockResult

expression: isFailure(optional`gensym236685205`gensym236700001)
stack trace: (most recent call last)
```

Bulat-Ziganshin: sorry for the noise

michaelsbradleyjr:
Closing in favor of #219.

PR #215, opened 2022-08-17
[chore] clean up `array[32, byte]` types

### LINKED ISSUE #184: Add distinct type for `array[32, byte]`

### LINKED ISSUE #214: Clean up `array[32, byte]` types
Clean up the variety of `array[32, byte]`'s to make the code a bit more descriptive and avoid confusion between types like `RequestId` and `SlotId`.

emizzle:
Cleaning up the variety of `array[32, byte]`'s to make the code a bit more descriptive and avoid confusion between types like `RequestId` and `SlotId`.
- rename `ContractId` to `SlotId`
- add `RequestId`, `PurchaseId`, `Nonce` types as distinct aliases of `array[32, byte]`
- rename `Proving.contracts` to `Proving.slots`
- Add/modify conversions to support the distinct type (ABI encoding/decoding, JSON encoding, REST decoding).
- Update tests 

Closes #214
Closes #184 (duplicate)

### NOTE
Updates nim-ethers dependency to PR branch https://github.com/status-im/nim-ethers/pull/31. Probably best to merge the nim-ethers PR first, then update the commit hash to point to main.

### DIFF
--- a/codex/contracts/market.nim
+++ b/codex/contracts/market.nim
@@ -35,7 +35,7 @@ method requestStorage(market: OnChainMarket,
   return request
 
 method getRequest(market: OnChainMarket,
-                  id: array[32, byte]): Future[?StorageRequest] {.async.} =
+                  id: RequestId): Future[?StorageRequest] {.async.} =
   let request = await market.contract.getRequest(id)
   if request != StorageRequest.default:
     return some request
@@ -43,7 +43,7 @@ method getRequest(market: OnChainMarket,
     return none StorageRequest
 
 method getHost(market: OnChainMarket,
-               requestId: array[32, byte],
+               requestId: RequestId,
                slotIndex: UInt256): Future[?Address] {.async.} =
   let slotId = slotId(requestId, slotIndex)
   let address = await market.contract.getHost(slotId)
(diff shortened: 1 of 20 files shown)


emizzle:
There is still an option to go further by making these types distinct, however the benefit would likely only be compiler restricting which types can be passed in specific situations. The downside may include the need to convert the types `array[32, byte]` and/or create conversion methods for each of the types (eg json rpc conversions). Moving to distinct types is tracked in #184.

emizzle:
These types are now distinct in commit https://github.com/status-im/nim-codex/pull/215/commits/c634edeb16ffdcc870541e080fc4ac311d8afc77.

dryajov:
> There is still an option to go further by making these types distinct, however the benefit would likely only be compiler restricting which types can be passed in specific situations. The downside may include the need to convert the types `array[32, byte]` and/or create conversion methods for each of the types (eg json rpc conversions). Moving to distinct types is tracked in #184.

This is actually the exact reason you'd want to use `distinct` to enable compiler checks and avoid accidentally mixing up different types/instances.

dryajov (approved):
LGTM - good stuff  👍

emizzle:
I would merge this, however it would change the `nim-ethers` dependency to an open PR branch https://github.com/status-im/nim-ethers/pull/31. We could merge https://github.com/status-im/nim-ethers/pull/31, however it would diverge nim-ethers' main branch in to two, one with a nim-eth dependency (and wallet support), and the other without. 

We could merge this as is, leaving the dependency on a branch of nim-ethers for now until the nim-eth dependency issue is solved. Thoughts?

emizzle:
I'm going to merge this for now as we have an open issue that specifically asked to deal with this situation #220.

PR #219, opened 2022-08-19
update getBlock return type

### LINKED ISSUE #177: Consistent behavior for cache-related errors in BlockStore impls

### LINKED ISSUE #182: Update getBlock return type

### LINKED ISSUE #210: getBlock: return BlockNotFound exception
Modify #205 in the following way:
- add exception BlockNotFound as subtype of CodexError
- return it from getBlock when appropriate
- modify places where getBlock is called, to distinguish BlockNotFound from other errors - only when it makes sense. When you are in doubt, we may discuss concrete places

Note - all other exceptions should be processed in the same way as in #205 or prior to it. We just want to modify BlockNotFound handling, so it combines the benefits of #205 and prior code.

michaelsbradleyjr:
Change return type for `method getBlock` from `Future[?!(?Block)]` to `Future[?!Block]`.

Use `type BlockNotFoundError = object of CodexError` to differentiate between *"block not found in the store"* and other errors.

Closes #177.
Closes #182.
Closes #210.

Alternative to #205, #209.

---

What I was missing previously was understanding of how at runtime to use Nim's [`of` operator](https://nim-lang.org/docs/manual.html#types-tuples-and-object-types) to distinguish error types inheriting from `CatchableError`

> Objects provide many features that tuples do not. Objects provide inheritance and the ability to hide fields from other modules. Objects with inheritance enabled have information about their type at runtime so that the `of` operator can be used to determine the object's type. The `of` operator is similar to the `instanceof` operator in Java.

I wish I'd known about / understood it beforehand, but I didn't; definitely a valuable "tool in the toolbox".

### DIFF
--- a/codex/erasure/erasure.nim
+++ b/codex/erasure/erasure.nim
@@ -114,13 +114,9 @@ proc encode*(
       for j in 0..<blocks:
         let idx = blockIdx[j]
         if idx < manifest.len:
-          without blkOrNone =? await dataBlocks[j], error:
+          without blk =? (await dataBlocks[j]), error:
             trace "Unable to retrieve block", error = error.msg
-            return error.failure
-
-          without blk =? blkOrNone:
-            trace "Block not found", cid = encoded[idx]
-            return failure("Block not found")
+            return failure error
 
           trace "Encoding block", cid = blk.cid, pos = idx
           shallowCopy(data[j], blk.data)
@@ -211,12 +207,8 @@ proc decode*(
 
         idxPendingBlocks.del(idxPendingBlocks.find(done))
 
-        without blkOrNone =? (await done), error:
-          trace "Failed retrieving block", exc = error.msg
-          return error.failure
-
-        without blk =? blkOrNone:
-          trace "Block not found"
+        without blk =? (await done), error:
+          trace "Failed retrieving block", error = error.msg
           continue
 
         if idx >= encoded.K:
--- a/codex/stores/networkstore.nim
+++ b/codex/stores/networkstore.nim
@@ -31,28 +31,20 @@ type
     engine*: BlockExcEngine # blockexc decision engine
     localStore*: BlockStore # local block store
 
-method getBlock*(self: NetworkStore, cid: Cid): Future[?! (? bt.Block)] {.async.} =
+method getBlock*(self: NetworkStore, cid: Cid): Future[?!bt.Block] {.async.} =
   ## Get a block from a remote peer
   ##
 
-  trace "Getting block from network store", cid
+  trace "Getting block from local store or network", cid
 
-  let blk = await self.localStore.getBlock(cid)
-  if blk.isErr:
-    return blk
-  if blk.get.isSome:
-    trace "Retrieved block from local store", cid
-    return blk
-
-  trace "Block not found in local store", cid
-  try:
+  without blk =? await self.localStore.getBlock(cid), error:
+    if not (error of BlockNotFoundError): return failure error
+    trace "Block not in local store", cid
     # TODO: What if block isn't available in the engine too?
-    let blk = await self.engine.requestBlock(cid)
-    # TODO: add block to the local store
-    return blk.some.success
-  except CatchableError as exc:
-    trace "Exception requesting block", cid, exc = exc.msg
-    return failure(exc)
+    # TODO: add retrieved block to the local store
+    return (await self.engine.requestBlock(cid)).catch
+
+  return success blk
 
 method putBlock*(self: NetworkStore, blk: bt.Block): Future[?!void] {.async.} =
   ## Store block locally and notify the network
--- a/codex/stores/sqlitestore.nim
+++ b/codex/stores/sqlitestore.nim
@@ -11,8 +11,6 @@ import pkg/upraises
 
 push: {.upraises: [].}
 
-import std/options
-
 import pkg/chronos
 import pkg/chronicles
 import pkg/datastore/sqlite_datastore
@@ -72,7 +70,7 @@ proc blockKey*(blockCid: Cid): ?!Key =
 
 method getBlock*(
   self: SQLiteStore,
-  cid: Cid): Future[?!(?Block)] {.async.} =
+  cid: Cid): Future[?!Block] {.async.} =
   ## Get a block from the cache or database.
   ## Save a copy to the cache if present in the database but not in the cache
   ##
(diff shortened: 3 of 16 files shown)


codex/erasure/erasure.nim
```diff
@@ -114,13 +114,9 @@ proc encode*(
       for j in 0..<blocks:
         let idx = blockIdx[j]
         if idx < manifest.len:
-          without blkOrNone =? await dataBlocks[j], error:
+          without blk =? dataBlocks[j].read, error:
```
dryajov, line 117: This wont work... What you are doing is essentially bypassing the async mechanism and attempting to read a future directly, if the future is pending or failed this will raise (potentially a defect in the future).

michaelsbradleyjr: I thought it is sure to work because of https://github.com/status-im/nim-codex/blob/0b3084c379fbb86893fd7688996f10ae893f638f/codex/erasure/erasure.nim#L106-L107

codex/stores/networkstore.nim
```diff
@@ -31,28 +31,27 @@ type
...
+
+  let
+    getRes = await self.localStore.getBlock(cid)
+
+  if getRes.isErr:
+    if not (getRes.error of BlockNotFoundError): return getRes
+    trace "Block not in local store", cid
+    try:
+      # TODO: What if block isn't available in the engine too?
+      return ok await self.engine.requestBlock(cid)
```
dryajov, line 48: you can probably do an `(await self.engine.requestBlock(cid)).catch`, which would convert to a questionable result.

codex/erasure/erasure.nim
michaelsbradleyjr: So if I understand you correctly, by using `await` instead of `.read` we'd properly end up in an `except` but with a `.read` it might trigger a Defect and crash?

dryajov: Yes, it's safer to call an `await` than a pure `read`. It will for example handle cancelations properly, propagating it as required, this is probably the most important point now that I think about it.

Here is the definition of `read`:

```nim
proc internalCheckComplete*(fut: FutureBase) {.
     raises: [Defect, CatchableError].} =
  # For internal use only. Used in asyncmacro
  if not(isNil(fut.error)):
    when defined(chronosStackTrace):
      injectStacktrace(fut)
    raise fut.error

proc internalRead*[T](fut: Future[T]): T {.inline.} =
  # For internal use only. Used in asyncmacro
  when T isnot void:
    return fut.value

proc read*[T](future: Future[T] ): T {.
     raises: [Defect, CatchableError].} =
  ## Retrieves the value of ``future``. Future must be finished otherwise
  ## this function will fail with a ``ValueError`` exception.
  ##
  ## If the result of the future is an error then that error will be raised.
  if future.finished():
    internalCheckComplete(future)
    internalRead(future)
  else:
    # TODO: Make a custom exception type for this?
    raise newException(ValueError, "Future still in progress.")
```

and this is `await`:

```nim
template await*[T](f: Future[T]): untyped =
  when declared(chronosInternalRetFuture):
    #work around https://github.com/nim-lang/Nim/issues/19193
    when not declaredInScope(chronosInternalTmpFuture):
      var chronosInternalTmpFuture {.inject.}: FutureBase = f
    else:
      chronosInternalTmpFuture = f
    chronosInternalRetFuture.child = chronosInternalTmpFuture

    # This "yield" is meant for a closure iterator in the caller.
    yield chronosInternalTmpFuture

    # By the time we get control back here, we're guaranteed that the Future we
    # just yielded has been completed (success, failure or cancellation),
    # through a very complicated mechanism in which the caller proc (a regular
    # closure) adds itself as a callback to chronosInternalTmpFuture.
    #
    # Callbacks are called only after completion and a copy of the closure
    # iterator that calls this template is still in that callback's closure
    # environment. That's where control actually gets back to us.

    chronosInternalRetFuture.child = nil
    if chronosInternalRetFuture.mustCancel:
      raise newCancelledError()
    chronosInternalTmpFuture.internalCheckComplete()
    when T isnot void:
      cast[type(f)](chronosInternalTmpFuture).internalRead()
  else:
    unsupported "await is only available within {.async.}"
```

michaelsbradleyjr: I changed it to use `await`.

codex/stores/networkstore.nim
michaelsbradleyjr: I reworked it using `.catch` and `without`.

codex/stores/networkstore.nim
```diff
@@ -37,21 +37,18 @@ method getBlock*(self: NetworkStore, cid: Cid): Future[?!bt.Block] {.async.} =
...
-  let
-    getRes = await self.localStore.getBlock(cid)
-
-  if getRes.isErr:
-    if not (getRes.error of BlockNotFoundError): return getRes
+  without blk =? await self.localStore.getBlock(cid), error:
+    if not (error of BlockNotFoundError): return failure error
     trace "Block not in local store", cid
-    try:
+    without blk =? (await self.engine.requestBlock(cid)).catch, error:
```
dryajov, line 43: I think we can just return the result from `requestBlock` directly, all we're doing is passing converting a `Result` to another `Result`?

michaelsbradleyjr: We can. It seemed a little more linear to not do that, e.g. with respect to the `# TODO comments` which this PR inherited, i.e. leaving room for refinement of the logic there.

michaelsbradleyjr: Though... maybe there isn't too much room for refinement in that case, at present.

codex/stores/sqlitestore.nim
```diff
@@ -84,24 +82,26 @@ method getBlock*(
...
 
   if not self.cache.isNil:
-    without cachedBlkOpt =? await self.cache.getBlock(cid), error:
-      trace "Unable to read block from cache", cid, error = error.msg
+    let
+      cachedBlockRes = await self.cache.getBlock(cid)
 
-    if cachedBlkOpt.isSome:
-      return success cachedBlkOpt
+    if not cachedBlockRes.isErr:
```
Bulat-Ziganshin, line 91: if cachedBlockRes.isOk:

codex/erasure/erasure.nim
```diff
@@ -209,12 +205,8 @@ proc decode*(
 
         idxPendingBlocks.del(idxPendingBlocks.find(done))
 
-        without blkOrNone =? (await done), error:
-          trace "Failed retrieving block", exc = error.msg
-          return error.failure
-
-        without blk =? blkOrNone:
-          trace "Block not found"
+        without blk =? await done, error:
```
Bulat-Ziganshin, line 208: for me, "(await done)" with parentheses makes it more readable

codex/erasure/erasure.nim
```diff
@@ -114,13 +114,9 @@ proc encode*(
       for j in 0..<blocks:
         let idx = blockIdx[j]
         if idx < manifest.len:
-          without blkOrNone =? await dataBlocks[j], error:
+          without blk =? await dataBlocks[j], error:
```
Bulat-Ziganshin, line 117: also "(await dataBlocks[j])"

codex/stores/networkstore.nim
michaelsbradleyjr: Change made.

codex/stores/sqlitestore.nim
michaelsbradleyjr: Change made.

codex/erasure/erasure.nim
michaelsbradleyjr: Change made.

codex/erasure/erasure.nim
michaelsbradleyjr: Change made.

dryajov:
I think this looks pretty good, anything else missing here?

michaelsbradleyjr:
I think it's good to go, just need to squash down the `[wip]` commits.

dryajov (approved):

PR #226, opened 2022-08-22
Create memory-leak detecting test suite

dryajov:
Expected to fail for now.

### DIFF
--- a/codex/streams/storestream.nim
+++ b/codex/streams/storestream.nim
@@ -29,6 +29,9 @@ export stores, blocktype, manifest, chronos
 logScope:
   topics = "codex storestream"
 
+const
+  StoreStreamTrackerName* = "StoreStream"
+
 type
   # Make SeekableStream from a sequence of blocks stored in Manifest
   # (only original file data - see StoreStream.size)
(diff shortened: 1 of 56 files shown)


Bulat-Ziganshin:
My task is to use these new checks to find memory leaks in stream usage

dryajov:
We're also missing tests for `AsyncStreamWrapper`

benbierens:
Renamed from 'Add chronos trackers'.

These changes now roll out the `checksuite`. This test suite automatically adds leak checking with Chronos to the tests. It verifies this by adding a tracker for `StoreStream`.
I would like to make new PRs for adding trackers to other modules in the future. New trackers should be added to `tests/helpers/trackers.nim` and they will be automatically checked by all the checksuites.

markspanbroek (approved):
:pray: Nice, thanks!

Not too sure about the name `checksuite`, perhaps use `trackersuite` or `tracksuite`? :runner:

benbierens:
Merged by Dmitriy on 22 Jun 2023 20:01:21 +02:00

PR #228, opened 2022-08-24
REST: added /upload?chunk=uint parameter

### LINKED ISSUE #225: Add BlockSize parameter to the upload REST API
Plus add BlocksSize parameter to node.store (and rename it to node.upload?).

It will allow us to easily experiment with different block sizes, and allow users to choose BlockSize for each dataset individually.

And yeah, it can reveal a ton of bugs - be ready! :)

Bulat-Ziganshin:
Uploaded file split into blocks of requested size when being stored in BlockStore, that allows user to control efficiency of storage, and provide us simple facility to test performance with various block sizes.

To do:
- [x] make REST parameter optional, rename it to "chunk"
- [x] fix memleak by closing AsyncStreamWrapper (and look for similar problems in other program parts)

Closes #225

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,14 +96,15 @@ Example:
  curl -vvv "127.0.0.1:8080/api/codex/v1/download/<Cid of the content>" --output <name of output file>
  ```
 
-### `/api/codex/v1/upload`
+### `/api/codex/v1/upload[?chunk=N]`
 
 Upload a file, upon success returns the `Cid` of the uploaded file.
+The optional 'chunk' parameter specifies Manifest.blockSize.
 
 Example:
 
 ```bash
-curl -vvv -H "content-type: application/octet-stream" -H Expect: -T "<path to file>" "127.0.0.1:8080/api/codex/v1/upload" -X POST
+curl -vvv -H "content-type: application/octet-stream" -H Expect: -T "127.0.0.1:8080/api/codex/v1/upload?chunk=65536" -X POST "<path to file>"
 ```
 
 ### `/api/codex/v1/debug/info`
--- a/codex/rest/api.nim
+++ b/codex/rest/api.nim
@@ -169,7 +169,7 @@ proc initRestApi*(node: CodexNodeRef, conf: CodexConf): RestRouter =
       ##
 
       trace "Handling file upload"
-      var bodyReader = request.getBodyReader()
+      let bodyReader = request.getBodyReader()
       if bodyReader.isErr():
         return RestApiResponse.error(Http500, bodyReader.error)
 
@@ -177,11 +177,11 @@ proc initRestApi*(node: CodexNodeRef, conf: CodexConf): RestRouter =
       # wait 1000ms before giving up
       await request.handleExpect()
 
-      try:
-        let
-          bodyStream = AsyncStreamWrapper.new(reader = AsyncStreamReader(bodyReader.get))
-          blockSize = (blockSize.get |? BlockSize).int
+      let
+        bodyStream = AsyncStreamWrapper.new(reader = AsyncStreamReader(bodyReader.get))
+        blockSize = (blockSize.get |? BlockSize).int
 
+      try:
         without cid =? (await node.store(bodyStream, blockSize)), error:
           trace "Error uploading file", exc = error.msg
           return RestApiResponse.error(Http500, error.msg)
@@ -193,7 +193,7 @@ proc initRestApi*(node: CodexNodeRef, conf: CodexConf): RestRouter =
       except AsyncStreamError:
         return RestApiResponse.error(Http500)
       finally:
-        await bodyReader.get.closeWait()
+        await bodyStream.closeImpl()
 
       # if we got here something went wrong?
       return RestApiResponse.error(Http500)
--- a/codex/rest/api.nim
+++ b/codex/rest/api.nim
@@ -161,10 +161,10 @@ proc initRestApi*(node: CodexNodeRef, conf: CodexConf): RestRouter =
   router.rawApi(
     MethodPost,
     "/api/codex/v1/upload") do (
-      blockSize: Option[uint]
+      chunk: Option[uint]
     ) -> RestApiResponse:
       ## Upload a file in a streamming manner,
-      ## split it into blocks of blockSize,
+      ## split it into blocks of `chunk` bytes,
       ## and save to the node's BlockStore
       ##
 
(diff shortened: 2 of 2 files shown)


codex/rest/api.nim
```diff
@@ -218,27 +218,32 @@ proc initRestApi*(node: CodexNodeRef, conf: CodexConf): RestRouter =
...
 
       try:
-        without cid =? (
-          await node.store(AsyncStreamWrapper.new(reader = AsyncStreamReader(reader)))), error:
+        let runUpload =
+          if (chunkOpt =? chunk) and (blockSize =? chunkOpt) and (blockSize <= MaxBlockSize):
+            node.store(bodyStream, blockSize = blockSize.int)
+          else:
+            node.store(bodyStream)
+
```
Bulat-Ziganshin, line 245: Current code just ignores any errors in syntax of `?chunk` REST parameter and goes into "else" part i.e. default chunk size in these cases. We may return parsing error message instead, I don't know whether it makes sense?

codex/rest/api.nim
```diff
@@ -218,27 +218,32 @@ proc initRestApi*(node: CodexNodeRef, conf: CodexConf): RestRouter =
...
-      # some clients (curl), wait 1000ms
-      # before giving up
-      #
+      # Attempt to handle `Expect` header some clients (curl),
+      # wait 1000ms before giving up
       await request.handleExpect()
 
-      let
-        reader = bodyReader.get()
+      let bodyStream = AsyncStreamWrapper.new(reader = AsyncStreamReader(bodyReader.get))
```
Bulat-Ziganshin, line 237: is there a possibility that AsyncStreamWrapper.new will raise an exception? I ask that because in prior code, this operation was inside `try`, and now I moved it our in order to call bodyStream.closeImpl in `finally`. 

If we need to catch exceptions in AsyncStreamWrapper.new too, I will write a bit more complex code.

README.md
```diff
@@ -119,14 +119,15 @@ Example:
...
 
 Upload a file, upon success returns the `Cid` of the uploaded file.
+The optional 'chunk' parameter specifies Manifest.blockSize.
 
 Example:
 
 ```bash
-curl -vvv -H "content-type: application/octet-stream" -H Expect: -T "127.0.0.1:8080/api/codex/v1/upload" -X POST "<path to file>"
+curl -vvv -H "content-type: application/octet-stream" -H Expect: -T "127.0.0.1:8080/api/codex/v1/upload?chunk=65536" -X POST "<path to file>"
 ```
```
Bulat-Ziganshin, line 132: Please make sure that my English is correct here. Also, you may have better wording for this, or have a better name instead of "chunk" for the parameter.

codex/rest/api.nim
dryajov: Yes, it could because `new` will call `initStream`, which does some additional work that could raise. I would move the assignement into the `try` but leave the declaration as a `var bodyStream`... Not ideal, but not much we can do.

codex/rest/api.nim
```diff
@@ -218,27 +218,32 @@ proc initRestApi*(node: CodexNodeRef, conf: CodexConf): RestRouter =
...
 
       try:
-        without cid =? (
-          await node.store(AsyncStreamWrapper.new(reader = AsyncStreamReader(reader)))), error:
+        let runUpload =
+          if (chunkOpt =? chunk) and (blockSize =? chunkOpt) and (blockSize <= MaxBlockSize):
+            node.store(bodyStream, blockSize = blockSize.int)
```
dryajov, line 242: Why the double layered `=?`? Is `chunk` is already optional?

codex/rest/api.nim
dryajov: If you look at the top of the file, there is a `validate` function that is currently empty which `presto` uses to validate all requests. I believe we can use it to validate the syntax of `chunk`.

README.md
```diff
@@ -119,14 +119,15 @@ Example:
  curl -vvv "127.0.0.1:8080/api/codex/v1/download/<Cid of the content>" --output <name of output file>
  ```
 
-### `/api/codex/v1/upload`
+### `/api/codex/v1/upload[?chunk=N]`
 
 Upload a file, upon success returns the `Cid` of the uploaded file.
+The optional 'chunk' parameter specifies Manifest.blockSize.
```
dryajov, line 125: I would not mention the `Manifest`, I would just say that chunks/blocks are going to of the specified size and padded up to `N`.

README.md
dryajov: `chunk` should be fine.

Bulat-Ziganshin: Please check entire changes in README, GitHub GUI shows only few last lines before the comment

codex/rest/api.nim
Bulat-Ziganshin: yeah, I printed chunk variable contents with echo:
```
None[Result[system.uint, system.cstring]]      # no ?chunk parameter at all
------------------
Some(err(Non-decimal character encountered))   # ?chunk=-2
------------------
Some(ok(1000000000))           # ?chunk=1000000000
```

The type is `Opt[ Result[uint,cstring]]`, so external Opt is set to None when there is no "?chunk" in URL at all. And internal Result is used to return the parsing error message.

README.md
Bulat-Ziganshin: Padding is our internal machinery. User of REST API don't see it since download now returns non-padded file

dryajov: `Manifest.blockSize` is also internal machinery and there is no way as of now to get a manifest from codex as of now, tho that might change. 

I would keep `Manifest.blockSize` out just say that chunks will be of a maximum size of `N`.

dryajov, line 125:
```suggestion
The optional `chunk` parameter allows specifying a custom chunk size. 

For large files (several gigabytes and up), specifying a chunk size too small might lead to a large manifest file since each hash will occupy some number of bytes. For example, sha2 needs 32 bytes, plus a few bytes for additional information packed in the [Cid](https://github.com/multiformats/cid) (multihash id, version, etc...). Hence, the resulting manifest will be of size `((file_size / chunk) * ~32)`. A manifest with a million blocks (around ~7gb with an 8K chunk size), would result in a manifest file of around ~30mb.
```

codex/rest/api.nim
dryajov: We should move `AsyncStreamWrapper.new` inside the try...

codex/rest/api.nim
dryajov: We should add a validator for the `chunk` parsing.

codex/rest/api.nim
```diff
@@ -249,7 +254,7 @@ proc initRestApi*(node: CodexNodeRef, conf: CodexConf): RestRouter =
       except AsyncStreamError:
         return RestApiResponse.error(Http500)
       finally:
-        await reader.closeWait()
+        await bodyStream.closeImpl()
```
dryajov, line 257: `closeImpl()` is an internal/private method and shouldn't be called directly, you should instead call `close()`.

dryajov (changes requested):

codex/rest/api.nim
Bulat-Ziganshin: We can use this function to check that `chunk` has a reasonable value, say between 512 and 1 GB. Does it seem reasonable to you?

dryajov: Yes, that sounds reasonable 👍 

We could potentially (not now, in the future), add specifiers for the size, like `mb`, `kb`, `gb`, etc... Just a thought...

dryajov:
@benbierens this is probably a good PR to pick up and finish.

PR #229, opened 2022-08-25
[marketplace] support freeing of slots and failed state

emizzle:
Support storage contract parameters to free slots after too many proofs missed, as well as fail a contract after too many slots have been freed.

NOTE:
1. This PR relies on #248, due to a bump in `nim-ethers`, which requires `bearssl`, and #248 does the necessary dependency updates.
1. This PR relies on https://github.com/status-im/dagger-contracts/pull/16, which should likely be merged first, then the submodule dependency commit hash updated in this PR before merging.

### DIFF
--- a/codex/contracts/market.nim
+++ b/codex/contracts/market.nim
@@ -1,4 +1,6 @@
+import std/strutils
 import pkg/ethers
+import pkg/ethers/testing
 import pkg/upraises
 import pkg/questionable
 import ../market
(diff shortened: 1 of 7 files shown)


markspanbroek (commented):
Thanks @emizzle!

emizzle:
@markspanbroek this one should be good to go as well 👍

emizzle:
Will cherry pick the commits from this PR into #211 and close this PR for simplicity.

PR #230, opened 2022-08-25
Copy all fields when [un]protect Manifest

### LINKED ISSUE #224: Copy all fields when [un]protect Manifest
Now we have a bug (blockSize and other fields aren't preserved) in https://github.com/status-im/nim-codex/blob/7ec039f822d4133255334ff3be9d79f91f1c3545/codex/erasure/erasure.nim#L265

And on opposite conversion, all fields are copied manually. And there is no Nim function for shallow copying all fields: https://forum.nim-lang.org/t/5539#34522

But we can use code from https://nim-lang.github.io/Nim/iterators.html :
```Nim
for name, v1, v2 in fieldPairs(a1, a2):
  when name != "blocks": 
    v2 = v1
```

EDIT: `Error: parallel 'fields' iterator does not work for 'case' objects`. So we can use `deepcopy`, but it will involve extra realloc.

Bulat-Ziganshin:
Manifest has quite a lot of fields and it's important to copy all of them when we create a new Manifest from the old one. Now we do it only in two situations - when we add ECC info to Manifest, and when we drop it back. So, I created a helper function that copies all Manifest fields except for `blocks` and changed all code making a new Manifest from the old one to call this function.

Change list:
- [x] fixed bug in erasure.decode - decoded manifest didn't copy fields of original, in particular blockSize
- [x] introduced manifest.addProtection and manifest.removeProtection functions specifically to convert manifest between protected and unprotected forms
- [x] added test of manifest.removeProtection
- [x] added test of erasure.encode and erasure.decode with non-default blockSize
- [x] added comment to erasure.encode

Closes #224

### DIFF
--- a/codex/erasure/erasure.nim
+++ b/codex/erasure/erasure.nim
@@ -156,8 +156,7 @@ proc encode*(
 proc decode*(
   self: Erasure,
   encoded: Manifest): Future[?!Manifest] {.async.} =
-  ## Decode a protected manifest into it's original
-  ## manifest
+  ## Decode a protected manifest into its original manifest
   ##
   ## `encoded` - the encoded (protected) manifest to
   ##             be recovered
@@ -254,10 +253,10 @@ proc decode*(
   finally:
     decoder.release()
 
-  without decoded =? Manifest.new(blocks = encoded.blocks[0..<encoded.originalLen]), error:
-    return error.failure
+  without decoded =? encoded.unprotect(blocks = encoded.blocks[0..<encoded.originalLen]), error:
+    return failure error
 
-  return decoded.success
+  return success decoded
 
 proc start*(self: Erasure) {.async.} =
   return
--- a/codex/erasure/erasure.nim
+++ b/codex/erasure/erasure.nim
@@ -82,7 +82,7 @@ proc encode*(
     parity       = parity
 
   trace "Erasure coding manifest", blocks, parity
-  without var encoded =? manifest.addProtection(blocks, parity), error:
+  without var encoded =? manifest.protect(blocks, parity), error:
     trace "Unable to create manifest", msg = error.msg
     return error.failure
 
@@ -262,7 +262,7 @@ proc decode*(
   finally:
     decoder.release()
 
-  without decoded =? encoded.removeProtection(), error:
+  without decoded =? encoded.unprotect(), error:
     return failure error
 
   return success decoded
--- a/codex/erasure/erasure.nim
+++ b/codex/erasure/erasure.nim
@@ -82,7 +82,7 @@ proc encode*(
     parity       = parity
 
   trace "Erasure coding manifest", blocks, parity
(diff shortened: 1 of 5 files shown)


codex/manifest/manifest.nim
```diff
@@ -153,6 +153,37 @@ proc cid*(self: Manifest): ?!Cid =
 # Constructors
 ############################################################
 
+func copyAllScalarFields(
```
dryajov, line 156: I'm not entirely sure what the benefit of doing this is, looks like a non-standard way of initializing an object - this isn't very nim'ish...

I think a better approach would be to instead call the default `Manifest.new()` from the protected `new` and then set the remaining protected fields.

codex/erasure/erasure.nim
```diff
@@ -82,7 +82,7 @@ proc encode*(
     parity       = parity
 
   trace "Erasure coding manifest", blocks, parity
-  without var encoded =? Manifest.new(manifest, blocks, parity), error:
+  without var encoded =? manifest.addProtection(blocks, parity), error:
```
dryajov, line 85: Not sure about `addProtection/removeProtection`, perhaps `protect/unprotect`?

dryajov (changes requested):
I'm not a big fan of the `copyAllScalarFields` approach, I think it adds unnecessary noise and its not very nim'ish in general, looks pretty foreign.

codex/manifest/manifest.nim
Bulat-Ziganshin: I don't understand your idea pleas give me a sketch.

We have two operations, protect and unprotect, that essentially copy contructors. They copy most fields and change a few. But I can't use things like shallowCopy because
- copying entire `blocks` arrays is inefficient sincew we anyway will replace it immediately
- we shouldn't copy `protected` field because we need to change it, and Nim doesn't allow changing discriminator fields

so, both `protect` and `unprotect` has to copy all fields except `blocks` and give new value to `protected` at the costruction time. I just factotred out the common code in the single procedure. And I don't like even this solution because new fields added to Manifest will not appear here magically

dryajov: I was talking about doing something like:

```nim
func protect*(self: Manifest, K, M: int) =
  var protected = Manifest.new(self.rootHash,
    self.originalBytes,
    self.blockSize,
    @[],
    self.version,
    self.hcodec,
    self.codec)

  var protected.protected = true
  protected.originalCid = self.cid
  protected.originalLen = self.len
  protected.K = K
  protected.M = M

  ...
  protected
```

Bulat-Ziganshin: current code avoids duplication of common code (Manifest.new with the long field list) in `protect` and `unprotect` by moving it into copyAllScalarFields. So you propose to remove copyAllScalarFields and put this code directly into `protect` and `unprotect`? I tried to avoid code duplication, especially since it depends on the field list, so when new fields are added, one will have to modify both functions.

PR #232, opened 2022-08-27
pass NIM_PARAMS to nim compiler in exec target of Makefile

michaelsbradleyjr:
With this change `make exec` runs a release build instead of a debug build.


Bulat-Ziganshin:
Michael, please merge. Dmitry confirmed that it was just a mistake.

PR #236, opened 2022-09-05
bump nim-ethers to 0.2.2, add nim-eth dependency

emizzle:
Support latest version of nim-ethers which relies on nim-eth.

NOTE: This commit points to PR #31 which is a branch of nim-ethers. That branch should probably be merged first, and this PR updated to point to the commit on main.


dryajov:
Ah, realized that this would be better left open for now, but should be no biggie to do another PR pointing to the correct main commit.

PR #240, opened 2022-09-08
Update curl upload command

dryajov:


c-blake:
It was working for me without the empty `Expect` header.  Not sure if that requirement is vestigial.

dryajov:
> It was working for me without the empty `Expect` header. Not sure if that requirement is vestigial.

It isn't, but we do respect it, so there is no harm in sending it. I think the order of the path and the url, which was switched, was the issue - unless the previous command worked for you?

c-blake:
You are right as to the order problem.  Simplest possible examples are usually best unless further down the road people will need the complexity for something.

PR #248, opened 2022-09-14
track latest nim-libp2p's master branch

### LINKED ISSUE #252: Track nim-libp2p's unstable branch

michaelsbradleyjr:
The goal of this PR is to have nim-codex start tracking nim-libp2p's `unstable` branch.

To make that work, a few other dependencies need to be bumped.

It's also necessary to make some changes to nim-libp2p-dht so that it works with nim-libp2p's `unstable` branch and latest nim-bearssl. See https://github.com/status-im/nim-libp2p-dht/pull/47.

The changes to nim-libp2p-dht were successful: all tests passing for https://github.com/status-im/nim-libp2p-dht/pull/47.

~However, in nim-codex the result is mixed at time of writing:~
~* `make exec` compiles successfully and `build/codex` seems to work as expected.~
~* `task testIntegration` compiles successfully and all tests pass.~
~* `task testContracts` compiles successfully and all tests pass.~
~* `task textCodex` (in `make test` and `make testAll`) compiles successfully but multiple tests are failing and `[Suite] Test PoR store` hangs indefinitely.~

`make testAll` builds successfully and all tests are passing. `make exec` builds successfully and `build/codex` is working as expected.

---

Closes #252.

### DIFF
--- a/codex/blockexchange/network/networkpeer.nim
+++ b/codex/blockexchange/network/networkpeer.nim
@@ -7,6 +7,9 @@
 ## This file may not be copied, modified, or distributed except according to
 ## those terms.
 
+import pkg/upraises
+push: {.upraises: [].}
+
 import pkg/chronos
 import pkg/chronicles
 import pkg/libp2p
@@ -21,6 +24,8 @@ const
   MaxMessageSize = 100 * 1 shl 20 # manifest files can be big
 
 type
+  ConnProvider* = proc(): Future[Connection] {.gcsafe, closure.}
+
   RPCHandler* = proc(peer: NetworkPeer, msg: Message): Future[void] {.gcsafe.}
 
   NetworkPeer* = ref object of RootObj
--- a/codex/rng.nim
+++ b/codex/rng.nim
@@ -12,11 +12,11 @@ import pkg/upraises
 push: {.upraises: [].}
 
 import pkg/libp2p/crypto/crypto
-import pkg/bearssl
+import pkg/bearssl/rand
 
 type
   RngSampleError = object of CatchableError
-  Rng* = ref BrHmacDrbgContext
+  Rng* = ref HmacDrbgContext
 
 var rng {.threadvar.}: Rng
 
@@ -25,16 +25,15 @@ proc instance*(t: type Rng): Rng =
     rng = newRng()
   rng
 
-# Random helpers: similar as in stdlib, but with BrHmacDrbgContext rng
+# Random helpers: similar as in stdlib, but with HmacDrbgContext rng
 # TODO: Move these somewhere else?
 const randMax = 18_446_744_073_709_551_615'u64
 
 proc rand*(rng: Rng, max: Natural): int =
   if max == 0: return 0
 
-  var x: uint64
   while true:
-    brHmacDrbgGenerate(addr rng[], addr x, csize_t(sizeof(x)))
+    let x = rng[].generate(uint64)
     if x < randMax - (randMax mod (uint64(max) + 1'u64)): # against modulo bias
       return int(x mod (uint64(max) + 1'u64))
 
--- a/tests/codex/helpers.nim
+++ b/tests/codex/helpers.nim
@@ -37,12 +37,11 @@ proc corruptBlocks*(
   manifest: Manifest,
   blks, bytes: int): Future[seq[int]] {.async.} =
   var pos: seq[int]
-  while true:
-    if pos.len >= blks:
-      break
 
-    var i = -1
-    if (i = Rng.instance.rand(manifest.len - 1); pos.find(i) >= 0):
+  doAssert blks < manifest.len
+  while pos.len < blks:
+    let i = Rng.instance.rand(manifest.len - 1)
+    if pos.find(i) >= 0:
       continue
 
     pos.add(i)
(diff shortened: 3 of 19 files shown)


codex/blockexchange/network/networkpeer.nim
```diff
@@ -21,6 +24,8 @@ const
   MaxMessageSize = 100 * 1 shl 20 # manifest files can be big
 
 type
+  ConnProvider* = proc(): Future[Connection] {.gcsafe, closure.}
```
michaelsbradleyjr, line 27: `ConnProvider` was removed in https://github.com/status-im/nim-libp2p/pull/749.

By adding it here, the types line up and compilation succeeds. However, with the behavioral changes in https://github.com/status-im/nim-libp2p/pull/749 adding the type here is probably not the thing to do, and may be responsible for failures in network related tests when running `make test`.

I spent some time looking at https://github.com/status-im/nim-libp2p/pull/749 but it wasn't entirely clear to me (yet) what the proper changes would be in nim-codex. Hopefully @Menduist can provide some insights.

codex/blockexchange/engine/pendingblocks.nim
```diff
@@ -39,12 +39,12 @@ proc getWantHandle*(
...
   if cid notin p.blocks:
      p.blocks[cid] = newFuture[Block]().wait(timeout)
-     trace "Adding pending future for block", cid
+     trace "Adding pending future for block", cid = $cid
 
   try:
     return await p.blocks[cid]
   except CancelledError as exc:
-    trace "Blocks cancelled", exc = exc.msg, cid
+    trace "Blocks cancelled", exc = exc.msg, cid = $cid
```
michaelsbradleyjr, line 47: For a reason I don't understand yet, passing `cid` to chronicle's trace started causing compilation errors, with json_serialization complaining about MultiCodec.

I tried bumping the serialization libraries and chronicles, but it didn't solve the problem. As a quick workaround, I introduced changes such as `cid = $cid` throughout the codebase.

codex/blockexchange/network/networkpeer.nim
Menduist: My bad, I didn't realized this type was used outside of libp2p, should have checked
However your fix seems correct to me
The tests are probably failing for another reason. I'll take a look

tests/codex/helpers.nim
```diff
@@ -37,28 +37,24 @@ proc corruptBlocks*(
   manifest: Manifest,
   blks, bytes: int): Future[seq[int]] {.async.} =
   var pos: seq[int]
-  while true:
-    if pos.len >= blks:
-      break
 
-    var i = -1
-    if (i = Rng.instance.rand(manifest.len - 1); pos.find(i) >= 0):
+  doAssert blks < manifest.len
```
Menduist, line 41: The changes in this file are not necessary (just some cleanup I did while troubleshooting), but still nice IMO

michaelsbradleyjr: 👍 that's how I understood them.

codex/rng.nim
```diff
@@ -32,10 +32,8 @@ const randMax = 18_446_744_073_709_551_615'u64
 proc rand*(rng: Rng, max: Natural): int =
   if max == 0: return 0
 
-  var x: uint64
   while true:
-    var rng = rng[]
-    hmacDrbgGenerate(rng, addr x, csize_t(sizeof(x)))
+    let x = rng[].generate(uint64)
```
michaelsbradleyjr, line 36: Just wanting to understand this better. Looking at `func generate` in [bearssl/rand](https://github.com/status-im/nim-bearssl/blob/master/bearssl/rand.nim), the first parameter is `var HmacDrbgContext` for all impls.

So when caller passes `rng[]` it's effectively like passing a `var`? It's probably an aspect of the calling conventions I didn't fully appreciate.

The reason I ask is that `type Rng* = ref HmacDrbgContext` so I was thinking of `rng[]` as deref'ing from `ref HmacDrbgContext` to just `HmacDrbgContext`.

Menduist: `var HmacDrbgContext` = a HmacDrbgContext you can modify
`ref HmacDrbgContext` = a managed pointer to a HmacDrbgContext you can modify

When you use `[]`, you just dereference that pointer to a value you can modify
_However_, if you assign it to a variable it will be copied to that variable (no longer a "shared reference")

`var` parameters have a different semantic, a `var` parameter is actually a non-nil `ref` under the hood

codex/blockexchange/engine/pendingblocks.nim
dryajov: This requires chronicles to know how to serialize it properly, it would be nice if we could add a `formatIt()` for `Cid` in libp2p, to avoid re-defining it everywhere.

michaelsbradleyjr: It was working previously, though. It only started being a problem owing to one/more of the deps that I needed to bump to get the tests compiling with nim-libp2p#unstable. When I got close to submitting the changes as a draft PR, I tried `make exec` and ran into this problem.

Menduist: Feel free to open PRs in libp2p :)

dryajov: @michaelsbradleyjr if you look through the codebase, there should be plenty of examples of how to add a custom formatter, just search for `formatIt`.

Actually, I think adding the formatter to codex directly is probably the best way to do it, after all libp2p shouldn't care about that.

dryajov:
Since this changes break protobuf serialization and we're currently tracking https://github.com/status-im/nim-protobuf-serialization/pull/36 in this branch, which doesn't have an ETA the sanest thing to do is to replace protobuf serialization with minprotobuf. I suggest we do that and revert to protobuf serialization once it's ready for prime time.

PR #250, opened 2022-09-14
codex nim-libp2p unstable auto bump

Menduist:
This PR just exists to allow CI to run, should not be merged / closed


Menduist:
This PR should remain open to test every change we do on libp2p against codex, or is it not possible? @dryajov

dryajov:
> This PR should remain open to test every change we do on libp2p against codex, or is it not possible? @dryajov

Ah, that should be fine.

Issue #237, opened 2022-09-08
Document that Windows Defender may delete suspicious files from FSStore


Bulat-Ziganshin:
It failed when I tried to upload and then download an EXE file using FSStore. I found that Windows Defender considers some of those 64 KB chunk files as viruses and moves them to quarantine (i.e. its own hidden directory). It was probably the first block containing EXE file header - for WD it looks like a weird executable file.

We may add this info to a "Troubleshooting" guide. Or transform data (like `xor FF`) before saving them.

dryajov:
> It failed when I tried to upload and then download an EXE file using FSStore. I found that Windows Defender considers some of those 64 KB chunk files as viruses and moves them to quarantine (i.e. its own hidden directory). It was probably the first block containing EXE file header - for WD it looks like a weird executable file.
> 
> We may add this info to a "Troubleshooting" guide. Or transform data (like `xor FF`) before saving them.

I think there is a way to white list directories in windows defender - https://support.microsoft.com/en-us/windows/add-an-exclusion-to-windows-security-811816c0-4dfd-af4a-47e4-c301afe13b26.

Bulat-Ziganshin:
Now I think that we should modify blocks as they are written to disk. Users don't read docs for issues like that.

Issue #159, opened 2022-07-26
Implement storage limits for local repository


dryajov:
Currently, the local repository doesn't implement any storage limits and it's possible to fill up the entirety of the users drive.

We need to add:

- [ ] Some sort of storage quota mechanism
  - [ ] Percentage based - user designates a `%` of bytes available to codex
  - [ ] Exact amount - user designates the exact number of bytes available to codex
- [ ] Pruning of old chunks from the system when some threshold has been reached
  - [ ] This requires running a periodic task that cleans up old chunks
- [ ] More??

Bulat-Ziganshin:
while working on the BlockStore API, I found that current code doesn't use delBlock at all, meaning that local store can only grow :)  So I think it's also important feature for any practical usage, no less that SQLite backend.

dryajov:
Yes, this is definitely an important feature for a usable product, so this should be relatively high on our priority list.

Bulat-Ziganshin:
Yes, all issues that currently in our priority list are assigned to me or @michaelsbradleyjr, so we will take care of it.

If this simple approach will go out of control, we can create milestone for them.

michaelsbradleyjr:
I have started researching some possibilities.

For `SQLiteStore` introduced in #160 we could make use of SQLite's [DBSTAT Virtual Table](https://www.sqlite.org/dbstat.html).

The [SQLITE_ENABLE_DBSTAT_VTAB](https://www.sqlite.org/compile.html#enable_dbstat_vtab) compile-time option seems to already be enabled in our builds via nim-sqlite3-abi, e.g. this query works for a database created by an instance of `type SQLiteStore`

```
SELECT pgsize FROM dbstat WHERE name='Store' AND aggregate=TRUE;
```

The result is *"the total amount of disk space used"* in bytes. Note that "total" doesn't include *"Freelist pages, pointer-map pages, [or] the lock page"*, so numbers reported by e.g. `ls` may include some overhead.

We could periodically query the `dbstat` info and if the database is over quota, we could sum the data size for the oldest `N` rows (100 or 1000 or whatever) using the timestamp column that's already part of the `Store` table created by nim-datastore. Repeat `M` times until we've reached the amount of space we want to clear within some delta, then delete the oldest `N * M` rows.

That would require some custom queries, but nim-datastore exports the necessary fields and helpers so that it's entirely doable.

[...]

PR #251, opened 2022-09-15
Add a run-time start-up switch -s/--store-backend to choose the data store backend

c-blake:
With current default chunk sizes, a `d:release` build & big upload, SQLite uses 20% more space & 30% more time than files on a Linux tmpfs/Intel.  So, SQLite may not be the hoped for performance win (and there may be other workloads where the perf delta is much worse).

More generally, different host devices/FSes will always have different ideal back-ends.  So, some kind of choice makes baseline sense. (Indeed, any user choice raises the idea of a translation engine to re-format a repo (FS <-> SQLite) when re-hosting to a new device, but that is unlikely to be a near term concern.)

If the idea of a DataStore is switching backends behind its abstraction, this also lets us exercise said switchability.  It also lets us more easily test any "above DataStore" quota abstraction -- e.g., some generic [pre-space-allocating solution](https://github.com/status-im/nim-codex/issues/159#issuecomment-1245141769) with multiple backends.

(I realize this needs some hook into our testing framework which I am still figuring out, but I wanted to start the PR for any design feedback.)

EDIT: Related https://github.com/status-im/nim-codex/issues/179 and https://github.com/status-im/nim-codex/pull/160#pullrequestreview-1057936220

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ The following options are available:
      --metrics-address      Listening address of the metrics server [=127.0.0.1].
      --metrics-port         Listening HTTP port of the metrics server [=8008].
  -d, --data-dir             The directory where codex will store configuration and data..
- -s, --store-backend        The data store backend [=StoreKind.FS].
+ -s, --store-backend        data store backend: fs, sqlite [=fs].
  -l, --listen-port          Specifies one or more listening ports for the node to listen on. [=0].
  -i, --listen-ip            The public IP [=0.0.0.0].
      --udp-port             Specify the discovery (UDP) port [=8090].
--- a/README.md
+++ b/README.md
@@ -45,6 +45,7 @@ The following options are available:
      --metrics-address      Listening address of the metrics server [=127.0.0.1].
      --metrics-port         Listening HTTP port of the metrics server [=8008].
  -d, --data-dir             The directory where codex will store configuration and data..
+ -s, --store-backend        The data store backend [=StoreKind.FS].
  -l, --listen-port          Specifies one or more listening ports for the node to listen on. [=0].
  -i, --listen-ip            The public IP [=0.0.0.0].
      --udp-port             Specify the discovery (UDP) port [=8090].
--- a/codex/codex.nim
+++ b/codex/codex.nim
@@ -140,8 +140,15 @@ proc new*(T: type CodexServer, config: CodexConf): T =
     raise (ref Defect)(
       msg: "Unable to create data directory for block store: " & repoDir)
 
+  let localStore =
+    case config.storeBackend
+    of FS:
+      notice "Using FS backend data store"
+      FSStore.new(repoDir, cache = cache)
+    of SQLite:
+      notice "Using SQLite backend data store"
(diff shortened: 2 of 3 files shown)


michaelsbradleyjr:
Nice work! I'll try it out with a local build in a little while.

Bulat-Ziganshin:
can you please add 3rd variant - "ram" that just assigns "localStore = cache"? it will useful for benchmarking the network part, without the slowdown caused by the disk storage.

please note that later you will have to add commit for README since it documents all switches

c-blake:
I think this is a great idea (I did a RAM fs on Linux myself), but should probably be a new DataStore fork maybe with just a Nim `Table` and then be "routed" to that backend via a new CLI enum value..maybe named "volatile" or "non-persistent" or something.

Bulat-Ziganshin:
For me, SQLite backend is still the best choice since it doesn't use too many files and the SQLite engine ensures and can check the integrity of DB. I feel it is a bit more reliable. Also, FS backend has a small problem on Windows (#237).

For me, 20-30% performance loss (with large SQLite pagesize) is manageable for the initial production version. In the long term, we may explore other DB engines, but first we need to establish requirements for the blockstore (i.e. implement other scenarios of its use, in particular usage quotas #159).

Bulat-Ziganshin:
@c-blake no, just "localStore = cache" will do it. The CacheStore by itself implements the BlockStore API.

c-blake:
This `of RAM: cache` branch approach does work, but it is really fragile since the user must provide a big enough `--cache` size parameter.

Linux has `tmpfs` (or just `/dev/shm`) and Mac OSX has `ramfs`.  [This](https://beebom.com/create-ram-disk-windows-10-super-fast-read-write-speeds/) suggests it is not so hard on Windows.  Maybe there is an easier way?  This RAM disk way works with *any* back end and also allows one to kill & restart the server and so on.  It seems a bit nicer overall.  Others should maybe weigh in here in light of the fragility.

It sounds like we will still would want this CLI parameter (as well as a `--quota`) but that maybe Dmitriy's new `Repo` abstraction will take the `StoreKind` as a parameter to dispatch to various backends instead of my `let..case`.  I agree that higher layer is the natural next step to enforce a quota (as mentioned in the top of this PR).

c-blake:
Trade-offs.  More set-up (|not if the expert user has it ready to go - on Linux `/dev/shm` is always already there), maybe a tiny amount of overhead (but likely small compared to all else we do), but extra fragility/size anticipation & bad failure modes (hanging on dload) if wrong.

I pushed commit, but I feel at least 1 other person should vote before I update the README with the new option and its valid values.  Or `-s=cached` could also stay a "secret" value (i.e. only document FS & SQLite).

dryajov:
Thanks @c-blake, I'll abstain from merging this right now, but will integrate it with my work on repo limits. The reason is that there might be changes around the current stores that might affect this. Overall, great work tho!

dryajov:
I think this flag is still useful for experimentation purposes and we could expose either the FS datastore or the SQLite datastore, this would helping this backends.

PR #261, opened 2022-09-29
[contracts] Fix intermittently failing clock test

markspanbroek:
Use a sub-second offset to avoid off-by-one errors in the test at the second boundary.

Co-authored-by: @emizzle


emizzle (approved):
LGTM!

PR #265, opened 2022-09-29
Changes required for reasonable upload/download speed

### LINKED ISSUE #263: Changes required for reasonable upload/download speed
Changes that are required to run upload/download at 100+ MB/s, which should be enough for the first production version:
- [ ] BlockSize=62 KB
- [ ] disable prefetchBlocks() - it may help in downloading from other nodes, but locally, it just reads the data from disk twice (when LRUCache is smaller than filesize)

Later we will add proper prefetching as described in #171.

Bulat-Ziganshin:
These changes increased upload/download speed on my Win10 Zen3 box to 50/150 MB/s (with FSStore):
* Increase BlockSize to 62 KiB
* Remove useless unbounded prefetching on downloading

Closes  #263


c-blake:
I can confirm that the performance impact of these two changes seems to be substantial on 127.0.0.1 throughput (which probably translates to CPU usage by early adopters on a real network).  What is best long-term depends upon many things, but I could see some early adopters being turned off by high CPU load running a node.

Bulat-Ziganshin:
@markspanbroek increasing BlockSize to 62 KiB makes integration tests fail. Can we fix it?

62 KB block size makes upload/download 5x faster than 8 KB used now, see https://github.com/status-im/nim-codex/issues/239#issuecomment-1240714497

markspanbroek:
> increasing BlockSize to 62 KiB makes integration tests fail. Can we fix it?

Yes, we can! :smile: The integration test checks whether a sale can happen between a host and a client. In the test we make 0x1F00 bytes available for sale, and the 62 KiB doesn't fit. So if you change 0x1F00 to 0xFFFFF [in  testIntegration](https://github.com/status-im/nim-codex/blob/8d315b38b8eec185f219caf8c62229054652dbdd/tests/testIntegration.nim#L83), then the test will pass.

Bulat-Ziganshin:
Increase blocksize from ~8KiB to ~64KiB and remove useless unbounded prefetching on download.

It increased upload/download speed on my Win10 Zen3 box to 50/150 MB/s (with FSStore). 
It also makes manifest files 8x smaller.
Closes #263.

PR #271, opened 2022-10-05
Replace protobuf serialization for Block exchange with minprotobuf

### LINKED ISSUE #264: Replace protobuf serialization with minprotobuf
Protobuf serialization has a number of issues that make it unsuitable for production use for the foreseeable feature, these shortcomings where fine while prototyping, but are an issue for production oriented development - for example it's not possible to use it with the latest version of libp2p due to dependencies and such.  

We already use minprotobuf in codex to serialize the manifest, we should also rewrite the https://github.com/status-im/nim-codex/blob/main/codex/blockexchange/protobuf/message.proto with minprotobuf. 

We could also rewrite https://github.com/status-im/nim-codex/blob/main/codex/storageproofs/por/serialization/por.proto, but this is going to be replaced with the new snark based approach which would not require serializing authenticators, so at least for now it's not a priority.

Bulat-Ziganshin:
- [x] minprotobuf serialization for Block exchange
- [x] fix `Message.ProtobufDecode(data).get` in codex\blockexchange\network\networkpeer.nim

Closes #264

### DIFF
--- a/codex/blockexchange/protobuf/message.nim
+++ b/codex/blockexchange/protobuf/message.nim
@@ -1,0 +1,194 @@
+# Protocol of data exchange between Codex nodes
+# and Protobuf encoder/decoder for these messages.
+#
+# Eventually all this code should be auto-generated from message.proto.
+
+import pkg/libp2p/protobuf/minprotobuf
+
+
+type
+  WantType* = enum
+    wantBlock = 0,
+    wantHave = 1
+
+  Entry* = object
+    `block`*: seq[byte]     # The block cid
+    priority*: int32        # The priority (normalized). default to 1
+    cancel*: bool           # Whether this revokes an entry
+    wantType*: WantType     # Note: defaults to enum 0, ie Block
+    sendDontHave*: bool     # Note: defaults to false
+
+  Wantlist* = object
+    entries*: seq[Entry]    # A list of wantlist entries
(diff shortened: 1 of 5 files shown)


dryajov:
This should unblock #248

codex/blockexchange/protobuf/message.nim
```diff
@@ -0,0 +1,194 @@
...
+
+#
+# Decoding Message from seq[byte] in Protobuf format
+#
+
+proc decode*(_: type Entry, pb: ProtoBuffer): ProtoResult[Entry] =
+  var
+    value = Entry()
+    field: uint64
+  discard ? pb.getField(1, value.`block`)
```
dryajov, line 123: Why are we discarding the result here?

Bulat-Ziganshin: let's check an example you provided: https://github.com/status-im/nim-libp2p/blob/unstable/libp2p/protocols/pubsub/rpc/protobuf.nim#L243

and the definition: https://github.com/status-im/nim-libp2p/blob/unstable/libp2p/protobuf/minprotobuf.nim#L444

IF getField finds field with given number in the buffer, it assigns decoded value to the variable, otherwise it is kept unchanged. And then it return ?!bool to signal a) decoding error , b) whether it found&assigned this field

PR #273, opened 2022-10-09
WIP: fix crash when can't find a CID and timeout hits

cskiraly:
This is to fix a crash if a CID is not found (after 10 minutes default)

WIP as it still needs tests.

Signed-off-by: Csaba Kiraly <csaba.kiraly@gmail.com>


dryajov:
Closing in favor of https://github.com/status-im/nim-codex/pull/315

PR #283, opened 2022-10-24
Load purchase state from chain

markspanbroek:
Depends on #211.


codex/purchasing/states/started.nim
```diff
@@ -1,9 +1,20 @@
...
     raiseAssert "invalid state"
 
-  purchase.future.complete()
+  let clock = purchase.clock
+  let market = purchase.market
+  let request = purchase.request
+
+  try:
+    await clock.waitUntil(await market.getRequestEnd(request.id))
+  except CatchableError as error:
```
emizzle, line 17: Is it possible to narrow down this Error type a bit more? We may get a `JsonRpcProviderError` from ethers. Any others?

codex/purchasing/states/unknown.nim
```diff
@@ -20,8 +22,8 @@ method enterAsync(state: PurchaseUnknown) {.async.} =
       of RequestState.Cancelled:
         state.switch(PurchaseCancelled())
       of RequestState.Finished:
-        discard # TODO
+        state.switch(PurchaseFinished())
       of RequestState.Failed:
-        discard # TODO
+        state.switch(PurchaseFailed())
   except CatchableError as error:
```
emizzle, line 28: Could we narrow the error type down a bit more? `JsonRpcProviderError` is an option, any others?

codex/purchasing/states/started.nim
markspanbroek: Hard to say, I don't think that `nim-ethers` wraps all exceptions from its underlying libraries into `EthersError`s. Though, no matter what kind of error occurs here, we can handle them all the same: move to the `Error` state, so that the REST API can report the error to the user.

codex/purchasing/states/descriptions.nim
```diff
@@ -0,0 +1,27 @@
...
+  elif state of PurchaseFinished:
+    "finished"
+  elif state of PurchasePending:
+    "pending"
+  elif state of PurchaseStarted:
+    "started"
+  elif state of PurchaseSubmitted:
+    "submitted"
+  else:
+    "unknown"
```
emizzle, line 26: It might be a bit cleaner to create a `description` pragma which accepts a description and if none provided, defaults to the name of the of the type.
Something like this:
```nim
template describe*(name: string) {.pragma.}
    ## Specifies the description for the object

template description*(obj: auto | typedesc): string =
    when macros.hasCustomPragma(obj, describe):
        macros.getCustomPragmaVal(obj, describe)
    else:
        typetraits.name(obj.type).toLower

### Usage:

type
  PurchaseStarted* {.describe("started").} = object of PurchaseState
  PurchaseNone* = object of PurchaseState

proc test() =
  check PurchaseStarted.description == "started"
  check PurchaseNone.description == "purchasenone"
```

codex/purchasing/states/failed.nim
```diff
@@ -0,0 +1,9 @@
+import ../statemachine
+import ./error
+
+type
+  PurchaseFailed* = ref object of PurchaseState
+
+method enter*(state: PurchaseFailed) =
+  let error = newException(PurchaseError, "Purchase failed")
```
emizzle, line 8: It might be helpful to add another layer of information here by describing the error type:
```suggestion
  type PurchaseFailedError* = ref object of of PurchaseError
  let error = newException(PurchaseFailedError, "Purchase failed")
```

tests/codex/testproving.nim
```diff
@@ -49,7 +49,7 @@ suite "Proving":
     proving.onProofRequired = onProofRequired
     proofs.setProofRequired(id, true)
     await proofs.advanceToNextPeriod()
-    check called
+    check eventually called
```
emizzle, line 52: Out of curiosity, what does adding the `eventually` solve? Why do we sometimes need to wait 5 seconds?

tests/contracts/testContracts.nim
```diff
@@ -73,7 +77,9 @@ ethersuite "Storage contracts":
 
   test "can be payed out at the end":
```
emizzle, line 78:
```suggestion
  test "can be paid out at the end":
```

tests/contracts/testMarket.nim
```diff
@@ -168,26 +173,66 @@ ethersuite "On-Chain Market":
...
+      await market.fillSlot(request.id, slotIndex.u256, proof)
+    for slotIndex in 0..request.ask.maxSlotLoss:
+      let slotId = request.slotId(slotIndex.u256)
+      while true:
+        try:
+          await waitUntilProofRequired(slotId)
+          let missingPeriod = periodicity.periodOf(await provider.currentTime())
+          await provider.advanceTime(periodicity.seconds)
+          await storage.markProofAsMissing(slotId, missingPeriod)
+        except JsonRpcProviderError as e:
```
emizzle, line 195: Thinking about this again, maybe it's more prudent to use `ProviderError` in case we decide to change which type of provider we use for tests in the future.

```suggestion
        except ProviderError as e:
```

emizzle (dismissed):
Looks great, good job!

codex/purchasing/states/descriptions.nim
markspanbroek: Good idea. I don't think we can use custom pragmas because the actual subtype of PurchaseState is not known until runtime. I have used a `method` instead, so that we can still define the description in the same module as the type.

codex/purchasing/states/failed.nim
markspanbroek: Seems a bit of overkill to me for now, but if we need it we can add it later.

tests/codex/testproving.nim
markspanbroek: The call above to `advanceToNextPeriod()` used to include a 2 second delay. I removed that delay, so that we can now wait the exact time that it takes for the callback to be called.

tests/contracts/testContracts.nim
markspanbroek: Good catch, thanks.

tests/contracts/testMarket.nim
markspanbroek: Agreed, I changed it.

markspanbroek:
CI fails on code coverage upload. This is unrelated to this PR, so I'm going to merge it.

PR #303, opened 2022-11-08
[WIP] Attempt to speed up development

markspanbroek:


dryajov:
Not sure if we made the decision to switch to 1.6 completely? Not opposed to, but we should make sure that all the make file commands default to 1.6 instead of 1.2 without having to set the `NIM_COMMIT` flag or similar.

markspanbroek:
> Not sure if we made the decision to switch to 1.6 completely? Not opposed to, but we should make sure that all the make file commands default to 1.6 instead of 1.2 without having to set the `NIM_COMMIT` flag or similar.

I figured that we could gain some speed by not running everything on both 1.2 and 1.6, so I'm now running all tests on 1.6 and only the unit tests on 1.2. Could have been the other way around, but seeing as nimbus-build-system has recently switched to 1.6 I thought we could give it a try.

I integrated the latest version of nimbus-build-system, so there will be no need to set NIM_COMMIT in our development environments.

dryajov:
> > Not sure if we made the decision to switch to 1.6 completely? Not opposed to, but we should make sure that all the make file commands default to 1.6 instead of 1.2 without having to set the `NIM_COMMIT` flag or similar.
> 
> I figured that we could gain some speed by not running everything on both 1.2 and 1.6, so I'm now running all tests on 1.6 and only the unit tests on 1.2. Could have been the other way around, but seeing as nimbus-build-system has recently switched to 1.6 I thought we could give it a try.
> 
> I integrated the latest version of nimbus-build-system, so there will be no need to set NIM_COMMIT in our development environments.

Running everything on 1.6 makes sense and I think it should be fine to run unit tests on 1.2 - perhaps its even worth considering sunsetting its support, given that at this point it's really ancient, but either way.

tests/codex/storageproofs/testpor.nim
```diff
@@ -14,6 +14,7 @@ import pkg/codex/blocktype as bt
 import ../helpers
 
 const
+  BlockSize = 62
```
dryajov, line 17: This should be at least `31 * 4`

tests/codex/testerasure.nim
```diff
@@ -16,6 +16,7 @@ import ./helpers
 
 suite "Erasure encode/decode":
 
+  const BlockSize = 64
```
dryajov, line 19: This is way too small to do anything meaningful, at least 1KB should be used.

Menduist:
When you'll be done, will be interested to have a small write-up of what you learned (best gains, what was useless, etc) :)

tests/codex/storageproofs/testpor.nim
markspanbroek: Thanks! I updated it to `31 * 4`.

tests/codex/testerasure.nim
markspanbroek: Thanks! I updated it to `1024`.

cskiraly:
When building Leopard:
- I seem to remember we are also building test apps that take time
- I see `make[1]: warning: jobserver unavailable: using -j1.  Add '+' to parent make rule.` in logs. I'm not sure it builds with multiple jobs, but worth a try

markspanbroek:
> * I seem to remember we are also building test apps that take time

Thanks! I opened a PR here: https://github.com/status-im/nim-leopard/pull/9

markspanbroek:
> When you'll be done, will be interested to have a small write-up of what you learned (best gains, what was useless, etc) :)

I wrote down some some of the lessons learned here @Menduist: https://github.com/status-im/nim-codex/pull/311

PR #305, opened 2022-11-10
[WIP] Fix nimble

markspanbroek:
Depends on #303


codex.nimble
```diff
@@ -25,7 +25,7 @@ requires "nim >= 1.2.0",
          "secp256k1",
          "stew",
          "upraises >= 0.1.0 & < 0.2.0",
-         "lrucache",
+         "https://github.com/status-im/lrucache.nim#daggerintegration",
```
dryajov, line 28: Any reason not to merge `daggerintegration` to master/upstream?

markspanbroek: I don't know. @Menduist you created this branch in `lrucache.nim`, do you know whether it can be merged?

Menduist: I didn't open a PR back then because it was missing tests
Adding tests & opening PR now

Menduist: I pushed to our fork master & published a new release.
However, we've fallen behind on upstream and are now conflicting. I think the best way forward would be to open two PRs for our respective changes, and once they are merged, sync with upstream
wdyt @emizzle?

emizzle: SGTM

dryajov: I don't mind if we keep our fork separate. Also, we have a couple of our own LRU implementations laying around that could potentially replace it if we move them into their own package/repo. Like this one https://github.com/status-im/nim-libp2p-dht/blob/main/libp2pdht/private/eth/p2p/discoveryv5/lru.nim.

markspanbroek: I set the requirement to point to our fork for now

PR #306, opened 2022-11-10
[marketplace] Load sales state from chain

### LINKED ISSUE #298: Modify the entire Sales process to use a state machine

emizzle:
Closes #298 

Loads the sales state from the chain. 

The state machine used is asynchronous, and allows for asynchronous events (eg contract events) to interrupt any ongoing state by cancelling the active `enterAsync` future.

~~NOTE: This PR relies on a draft dagger-contracts PR https://github.com/status-im/dagger-contracts/pull/23.~~


codex/sales/salesagent.nim
```diff
@@ -0,0 +1,85 @@
...
+
+# fwd declarations
+proc subscribeCancellation*(agent: SalesAgent): Future[void] {.gcsafe.}
+proc subscribeFailure*(agent: SalesAgent): Future[void] {.gcsafe.}
+
+proc populateRequest(agent: SalesAgent) {.async.} =
+  if agent.request.isNone:
+    agent.request = await agent.sales.market.getRequest(agent.requestId)
+
+proc init*(agent: SalesAgent, numSlots: uint64) {.async.} =
```
markspanbroek, line 27: Can we name this `start` (and rename `deinit` to `stop`)?

codex/sales/salesagent.nim
```diff
@@ -0,0 +1,85 @@
...
+    sales: sales,
+    requestId: requestId,
+    availability: availability,
+    request: request)
+
+# fwd declarations
+proc subscribeCancellation*(agent: SalesAgent): Future[void] {.gcsafe.}
+proc subscribeFailure*(agent: SalesAgent): Future[void] {.gcsafe.}
+
+proc populateRequest(agent: SalesAgent) {.async.} =
```
markspanbroek, line 23: I would call this `retrieveRequest`, because the name `populateRequest` is already used elsewhere with a different meaning (populating the fields of a storage request object)

codex/sales/salesagent.nim
```diff
@@ -0,0 +1,85 @@
...
+
+proc subscribeFailure*(agent: SalesAgent) {.async.} =
+  let market = agent.sales.market
+
+  proc onFailed(_: RequestId) {.async.} =
+    without request =? agent.request:
+      return
+
+    without state =? (agent.state as SaleState):
+      return
```
markspanbroek, line 80: Can be a bit shorter, if you like:

```suggestion
    without request =? agent.request and
            state =? (agent.state as SaleState):
      return
```

codex/sales/states/filled.nim
```diff
@@ -0,0 +1,46 @@
...
+    raiseAssert "invalid state"
+
+  try:
+    let market = agent.sales.market
+
+    without slotIndex =? agent.slotIndex:
+      raiseAssert "no slot selected"
+
+    if availability =? agent.availability:
+      agent.sales.remove(availability)
```
markspanbroek, line 31: Availability should be removed much earlier; no two agents should start with the same availability otherwise you might end up selling your available store more than once :)

codex/sales/states/filled.nim
```diff
@@ -0,0 +1,46 @@
...
+
+    let host = await market.getHost(agent.requestId, slotIndex)
+    let me = await market.getSigner()
+    if host == me.some:
+      await state.switchAsync(SaleFinished())
+    else:
+      let error = newException(SaleFilledError, "Sale host mismatch")
```
markspanbroek, line 38: This is not really an error as such? We simply lost the slot to another host that was faster.

codex/utils/statemachine.nim
```diff
@@ -100,3 +107,24 @@ method enter(state: AsyncState) =
...
+    trace "Switching sales state", `from`=state, to=newState
+    if activeTransition =? state.activeTransition and
+        not activeTransition.completed:
+      await activeTransition.cancelAndWait()
+    # should wait for exit before switch. could add a transition option during
+    # switch if we don't need to wait
+    await state.exitAsync()
+    state.context = none StateMachine
+  else:
+    trace "Switching sales state", `from`="no state", to=newState
```
markspanbroek, line 122: This seems overly specific for the generic state machine

tests/codex/testsales.nim
```diff
@@ -1,9 +1,13 @@
 import std/sets
+import std/sequtils
+import std/sugar
+import std/times
 import pkg/asynctest
 import pkg/chronos
-import pkg/codex/contracts/requests
-import pkg/codex/proving
+# import pkg/codex/contracts/requests
```
markspanbroek, line 7:
```suggestion

```

markspanbroek (approved):
Nice! Just a few small comments from my side.

codex/sales/salesagent.nim
emizzle: Done

codex/sales/salesagent.nim
emizzle: Done

codex/sales/salesagent.nim
```diff
@@ -0,0 +1,85 @@
...
+proc subscribeCancellation*(agent: SalesAgent): Future[void] {.gcsafe.}
+proc subscribeFailure*(agent: SalesAgent): Future[void] {.gcsafe.}
+
+proc populateRequest(agent: SalesAgent) {.async.} =
+  if agent.request.isNone:
+    agent.request = await agent.sales.market.getRequest(agent.requestId)
+
+proc init*(agent: SalesAgent, numSlots: uint64) {.async.} =
+  let rng = Rng.instance
+  let slotIndex = rng.rand(numSlots - 1)
```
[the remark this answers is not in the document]
emizzle: Done

codex/sales/salesagent.nim
emizzle: Done

codex/sales/states/filled.nim
emizzle: Moved this to the downloading state. This should work across node restarts once the availability state is persisted on disk.

codex/sales/states/filled.nim
emizzle: I agree that an exception may be too harsh for this situation, but it is still an graceful error of sorts as it is an unexpected outcome. Maybe we could create a new property on the `SaleErrored` state that indicates a graceful error?

codex/utils/statemachine.nim
emizzle: Removed "sales ".

codex/sales/salesagent.nim
```diff
@@ -0,0 +1,103 @@
...
+                    slotIndex: ?UInt256,
+                    request: ?StorageRequest): SalesAgent =
+  SalesAgent(
+    sales: sales,
+    requestId: requestId,
+    availability: availability,
+    slotIndex: slotIndex,
+    request: request)
+
+# fwd declarations
```
markspanbroek, line 19: This comment doesn't really add anything for me, I would leave it out.

codex/sales/states/unknown.nim
```diff
@@ -0,0 +1,47 @@
...
+
+method enterAsync(state: SaleUnknown) {.async.} =
+  without agent =? (state.context as SalesAgent):
+    raiseAssert "invalid state"
+
+  let market = agent.sales.market
+
+  try:
+    without requestState =? await market.requestState(agent.requestId):
+      raiseAssert "state unknown"
```
markspanbroek, line 28: This would kill the node when the state cannot be retrieved from the contract. Seems a bit harsh :sweat_smile:

codex/sales/states/unknown.nim
```diff
@@ -0,0 +1,47 @@
...
+  await state.switchAsync(SaleFailed())
+
+method enterAsync(state: SaleUnknown) {.async.} =
+  without agent =? (state.context as SalesAgent):
+    raiseAssert "invalid state"
+
+  let market = agent.sales.market
+
+  try:
+    without requestState =? await market.requestState(agent.requestId):
```
markspanbroek, line 27: Now that we have a slot state, perhaps we can use that instead?

codex/contracts/market.nim
```diff
@@ -80,7 +96,7 @@ method subscribeRequests(market: OnChainMarket,
                          callback: OnRequest):
                         Future[MarketSubscription] {.async.} =
   proc onEvent(event: StorageRequested) {.upraises:[].} =
-    callback(event.requestId, event.ask)
+    asyncSpawn callback(event.requestId, event.ask)
```
markspanbroek, line 99: I would avoid `asyncSpawn` on callbacks. We cannot yet specify that async callbacks are not allowed to raise exceptions, and if they do happen to raise an exception, then the node crashes.

Ideally we'd keep the callbacks non-async. When we really to need to call async code in a callback, then we can do the `asyncSpawn` inside the callback; it's then easier to check that the async code doesn't raise any exceptions.

markspanbroek (commented):
Great work on the contract updates @emizzle! I left a few comments.

I was wondering why nimbus-build-system is updated in this PR?

codex/sales/states/filled.nim
markspanbroek: You're right, let's keep it like it is.

codex/contracts/market.nim
```diff
@@ -91,7 +107,7 @@ method subscribeSlotFilled*(market: OnChainMarket,
                            Future[MarketSubscription] {.async.} =
   proc onEvent(event: SlotFilled) {.upraises:[].} =
     if event.requestId == requestId and event.slotIndex == slotIndex:
-      callback(event.requestId, event.slotIndex)
+      asyncSpawn callback(event.requestId, event.slotIndex)
```
dryajov, line 110: What happens if the `callback` fails? `asyncSpawn` is usually the wrong primitive to use in this construct, maybe `await callback`?

codex/contracts/market.nim
dryajov: why not just `await` the callback?

emizzle: > why not just `await` the callback?

`onEvent` is not async. One option would be to allow async event subscriptions in ethers. Something like this: https://github.com/status-im/nim-ethers/pull/35

emizzle: I pushed a commit that supports this: https://github.com/status-im/nim-codex/pull/306/commits/c2a8cea91b117d298b44cf1c0e8ff3728ec22cc9.

codex/sales/states/unknown.nim
emizzle: Updated.

codex/contracts/market.nim
markspanbroek: That's is one way of solving the problem, but the disadvantage is that it keeps spreading `{.async.}` further and further (even into the nim-ethers library). I'd like to suggest an alternative approach: make sure that the state machine can handle normal (sync?) event handlers while allowing async behavior in the states. I have an idea on how to do that and will work this out shortly, so that we can compare approaches.

emizzle: What are the disadvantages of spreading `{.async.}` further and into ethers (which already uses async handlers for other event, eg new blocks)?

emizzle: > make sure that the state machine can handle normal (sync?) event handlers while allowing async behavior in the states

Do you mean having states that support both `enter` and `enterAsync` for example?

I can agree with that, but I believe in this particular situation, the way the async state machine is setup in this PR, we need to support async callbacks. Here's my logic... It's handling an event subscription in a contract. With the suggested changes in the ethers PR, we have the option to use sync or async event handlers, depending on what is required. In this particular case, the event handler callback enters a new state asynchronously (hence the need for an async callback). The async state transition is required so that the current state can be cancelled, allowing new contract events to come in and change the state. Using an `asyncSpawn` in the callback would not work in this case as the state machine would not have a handle on the current state and thus would not be able to cancel it.  If everything was synchronous, the entire `enter` state transition will need to complete its run *before* any new contract event handlers can run (which may, in turn, transition the state machine to a new state). We already saw that this can be a problem if there's any kind of network or IO interactions in a state. 

I'm very curious to see the solution you are working on. Happy to jump on a call with you this evening to discuss futher 👍

markspanbroek: > What are the disadvantages of spreading `{.async.}` further and into ethers (which already uses async handlers for other event, eg new blocks)?

Async code adds complexity and is more difficult to reason about than sync code, so as a general rule I try to avoid it except where necessary. When we run into a boundary between async and sync code (as we do with the synchronous event handlers that need to perform an async cancelAndWait in the state machine),  we can either make the calling code async, or carefully design the boundary to support the use case. 

If we make the calling code async, this may require more code to become async, which requires more code to become async, etc. Until we run into a dependency that we do not control and therefore cannot make async. At which point we will be forced to design the sync / async boundary. I prefer to contain the async code, and design the sync / async boundary early on.

I'll try to code out an alternative to the state machine with a carefully designed sync/async boundary soon. Let's jump on that call, and discuss it :smile:

markspanbroek: Added a couple of commits that introduce the new state machine design from #344 into this PR, which allows for synchronous callbacks.

codex/contracts/market.nim
markspanbroek: Should be resolved with the new async state machine design.

PR #309, opened 2022-11-14
CI with nimble

markspanbroek:
Runs continuous integration with nimble instead of nimbus build system. Both simplifies the CI configuration and speeds up the build, but it's probably not wise to merge this until a version of Nimble that supports lock files has been released with Nim.

Depends on #305.


gmega:
Incredibly I think this may start making sense to consider again once nimble integrates the SAT solver that's been in the works for Atlas, assuming package maintainers also start issuing proper releases. But I'd close this issue for now as I don't think it's anything we'll be doing short-term, what do you think @markspanbroek?

markspanbroek:
Yeah, let's close this for now.

PR #325, opened 2022-12-19
Bump nim-sqlite3-abi

yyoncho:


dryajov:
@benbierens I think this is a good bump, the failure comes from coverage upload which is a frequent failure due to the service limitations. Worth a smoke test and a merge...

dryajov:
This is quite old, closing.

PR #340, opened 2023-02-03
[marketplace] Add Reservations Module

### LINKED ISSUE #296: Persist availability on disk

### LINKED ISSUE #318: Restore slotIndex from chain instead of reassigning a random index

emizzle:
Closes #296.
Closes #318.

Add a reservations module to persist storage availabilities across node restarts, as per [the design document](https://github.com/status-im/codex-research/blob/main/design/sales.md). The bytes made available for storage by the user are reserved in the repo, and the serialised availability is persisted in the repo meta datastore.

### NOTE:
The [`codex-contracts-eth` branch](https://github.com/status-im/codex-contracts-eth/tree/get-active-slot-no-slot-collateral) is a detached head. It contains changes needed to get this PR to work. Those changes are also included in https://github.com/status-im/codex-contracts-eth/pull/51, however that PR also contains slot collateral contract changes, which would require further client changes to support, and this PR is big enough already. Client updates to support slot collateral can be handled in a future PR, and the [`get-active-slot-no-slot-collateral` branch](https://github.com/status-im/codex-contracts-eth/tree/get-active-slot-no-slot-collateral) can be deleted once the future PR is merged.


codex/contracts/interactions/clientinteractions.nim
```diff
@@ -0,0 +1,55 @@
...
+
+type
+  ClientInteractions* = ref object of ContractInteractions
+    purchasing*: Purchasing
+
+proc new*(_: type ClientInteractions,
+          signer: Signer,
+          deployment: Deployment): ?ClientInteractions =
+
+  without address =? deployment.address(Marketplace):
```
markspanbroek, line 23: Would be nice if we could remove this remaining duplication between `ClientInteractions.new()` and `HostInteractions.new()`. Perhaps introduce a `ContractInteractions.init()` that is called from both of them?

codex/sales.nim
```diff
@@ -143,17 +102,15 @@ proc handleRequest(sales: Sales,
...
   for slotId in slotIds:
-    # TODO: this needs to be optimised
-    if request =? await market.getRequestFromSlotId(slotId):
-      let availability = sales.findAvailability(request.ask)
-      without slotIndex =? findSlotIndex(request.ask.slots,
-                                          request.id,
-                                          slotId):
-        raiseAssert "could not find slot index"
+    if (request, slotIndex) =? (await market.getActiveSlot(slotId)):
+      let availability = await sales.context.reservations.find(
```
markspanbroek, line 109: Why don't we look up the reservation based on the slot id? Couldn't this go wrong when there are multiple reservations with of the same size, duration and price?

codex/sales/reservations.nim
```diff
@@ -0,0 +1,314 @@
...
+      let rollbackErr = rollbackInnerErr.toErr(AvailabilityDeleteFailedError,
+        "Failed to delete persisted availability during rollback")
+      rollbackInnerErr.parent = reserveErr
+      return failure(rollbackErr)
+
+    return failure(reserveErr)
+
+  return success()
+
+# TODO: call site not yet determined. Perhaps reuse of Availabilty should be set
```
markspanbroek, line 205: The "list of availability" model is starting to show its cracks here. We should probably move to a different model soon, whereby we allow the user to simply indicate once how much storage they want to make available in total, and for which duration and price they'd like to engage into contracts. It would be up to the sales agent to sell off pieces of the total available storage as it sees fit.

codex/sales/states/downloading.nim
```diff
@@ -38,5 +41,24 @@ method run*(state: SaleDownloading, machine: Machine): Future[?State] {.async.}
...
+      pricePerSlot = request.ask.pricePerSlot,
+      used = false
+    return
+  
+  data.availability = some availability
+
+  if err =? (await agent.context.reservations.markUsed(
+    availability,
+    request.slotId(data.slotIndex))).errorOption:
+    raiseAssert "failed to mark availability as used, error: " & err.msg
```
markspanbroek, line 61: This would crash the node, we should probably move to the error state instead.

tests/codex/sales/helpers.nim
```diff
@@ -0,0 +1,16 @@
...
+import pkg/codex/sales/reservations
+
+proc allAvailabilities*(r: Reservations): Future[seq[Availability]] {.async.} =
+  var ret: seq[Availability] = @[]
+  without availabilities =? (await r.availabilities), err:
+    raiseAssert "failed to get availabilities, error: " & err.msg
+  for a in availabilities:
+    if availability =? (await a):
+      ret.add availability
+  return ret.reversed()
```
markspanbroek, line 16: Why reversed?

markspanbroek (commented):
Thanks Eric, I'm happy to see the reservations module take shape. I left a few comments. We should probably get the test failure fixed as well ;)

codex/sales/reservations.nim
dryajov: This is indeed how I envisioned it and how it got implemented in the RepoStore.

codex/sales/reservations.nim
```diff
@@ -0,0 +1,314 @@
...
+  availability.slotId.isSome
+
+func key(id: AvailabilityId): ?!Key =
+  (ReservationsKey / id.toArray.toHex)
+
+func key*(availability: Availability): ?!Key =
+  return availability.id.key
+
+func available*(self: Reservations): uint =
+  return self.repo.quotaMaxBytes - self.repo.totalUsed
```
dryajov, line 114: This should probably be moved to the repo.

codex/sales/reservations.nim
```diff
@@ -0,0 +1,314 @@
...
+  without key =? availability.key, err:
+    return failure(err)
+
+  if err =? (await self.repo.metaDs.put(
+    key,
+    @(availability.toJson.toBytes))).errorOption:
+    return failure(err)
+
+  # TODO: reconcile data sizes -- availability uses UInt256 and RepoStore
+  # uses uint, thus the need to truncate
```
dryajov, line 187: I thoug about changing this to UIn256 in the RepoStore, but 2^64 can represent 16 exabytes, which seems enough for the short term ;-)

dryajov:
Another question, if the local db gets deleted and/or corrupted, would all the availabilities and sales be lost or is it possible to recover them with on-chain data?

codex/sales/reservations.nim
emizzle: > The "list of availability" model is starting to show its cracks here. We should probably move to a different model soon, whereby we allow the user to simply indicate once how much storage they want to make available in total, and for which duration and price they'd like to engage into contracts. It would be up to the sales agent to sell off pieces of the total available storage as it sees fit.

Should we handle this as a future PR so that we can move forward with user testing?

> This is indeed how I envisioned it and how it got implemented in the RepoStore.

Agreed. We discussed moving to a different model, but wanted to retrofit this to the old availability model so we could enable user testing, with plans to update the model and api.

tests/codex/sales/helpers.nim
emizzle: It was an attempt to retrieve availability in the order added, to make the tests prettier :) But I don't think order is preserved in the repo, so likely this is not needed and can be removed.

emizzle:
> From what I can tell, it is possible to create different availabilities for purchase with different parametters, such as price and duration, is that correct? An explanation of the PR would be super helpful.

Yes, this is how the existing availabilities api works. This PR simply persists the availabilities and does not change the api.

emizzle:
> Another question, if the local db gets deleted and/or corrupted, would all the availabilities and sales be lost or is it possible to recover them with on-chain data?

This is a great question. The availabilities would indeed be lost if the db gets deleted/corrupted. The host would need to re-add availabilities before being able to resume their role in filling a slot. This brings up a good question though -- if deletion/corruption of the db happens, and the host re-adds availability, they'd have to restart their node to resume their filled slot role (due to the order of node start > check active slots > match to availabilities > resume role in slot), so perhaps we should consider an alternative solution. This could tie in well with a new availability model discussed in a comment above.

codex/sales.nim
```diff
@@ -35,13 +34,12 @@ import ./sales/states/unknown
 ##     |                          | ---- storage proof ---> |
 
 export stint
-export availability
+export reservations
```
AuHau, line 37: Should we also replace "availability" with "reservations" in the comment describing the Sales mechanism above?

codex/sales/reservations.nim
AuHau: Agree, what was the reasoning for having this parameter mutable and exposed over REST? IMHO I would expect having something like @markspanbroek is describing and having it in CLI/config parameter...

AuHau (commented):
I have more general comments/questions to understand Nim better :-) But also few high-level comments and suggestions ;-)

dryajov:
> > Another question, if the local db gets deleted and/or corrupted, would all the availabilities and sales be lost or is it possible to recover them with on-chain data?
> 
> This is a great question. The availabilities would indeed be lost if the db gets deleted/corrupted. The host would need to re-add availabilities before being able to resume their role in filling a slot. This brings up a good question though -- if deletion/corruption of the db happens, and the host re-adds availability, they'd have to restart their node to resume their filled slot role (due to the order of node start > check active slots > match to availabilities > resume role in slot), so perhaps we should consider an alternative solution. This could tie in well with a new availability model discussed in a comment above.

Indeed this is a big issue, we cannot rely on the local database for anything but transient/ephemeral state, everything else needs to be persisted on chain or done in a way that will no be affected by local state.

If by "availabilities" we mean, chunks of space available for sale at specific prices and durations, then loosing them isn't a big deal, in other words, if a user configured his node to match with offers against storage requests and that is lost, that should be fine, since this can be recreated without much issues.

If on the other hand, if this affects ongoing slot fillings then it is a big deal breaker - it should not, under any circumstances, ongoing slot fillings. I would expect that information about a particular ongoing slot filling is on chain and can be retrieved without the need of any local state. 

Is that the case right now or not?

dryajov:
This is the way I've been thinking about availabilities so far:

- A provider creates an availability with these characteristics - "I want to sell up to 10Gb of storage for up to 3 months at >= $0.004 per GB"
- A user creates a request saying "I want to buy 5Gb for 2 months at <= $0.005 per GB"

When a request comes in, the engine checks its local availabilities and against the request and decides wether it wants to enter into the contract or not.

Availabilities are purely local, and are not persisted on chain, so loosing them is not a big deal. However, once a provider has committed to a slot, this information should be global and survive restarts and local database corruption. In fact, this should probably not be even stored locally.

codex/contracts/interactions/hostinteractions.nim
```diff
@@ -0,0 +1,67 @@
...
+    proving*: Proving
+
+proc new*(_: type HostInteractions,
+          signer: Signer,
+          deployment: Deployment,
+          repoStore: RepoStore): ?HostInteractions =
+
+  without address =? deployment.address(Marketplace):
+    error "Unable to determine address of the Marketplace smart contract"
+    return none HostInteractions
```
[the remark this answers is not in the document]
dryajov: Yes, constructors should return `Result` and if they failed to construct the object properly they should error out, returning an `Option` from a constructor is almost always wrong.

codex/node.nim
```diff
@@ -43,14 +43,18 @@ type
 
   CodexError = object of CatchableError
 
+  Contracts* = tuple
+    client: ?ClientInteractions
+    host: ?HostInteractions
```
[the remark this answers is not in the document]
dryajov: Another thing to keep in mind is that Nim always initialized object types, so there is no way of telling wether they were set or not, in this cases optionals are used to signal that. With ref types this isn't an issue because they are nillable, but pure object types aren't.

In other words, if you have code like this:

```nim
type
  Obj1 = object
     val: int

  Obj2 = object
     val: string
  
  Obj3 = object
    o1: Obj1
    o2: Obj2

let obj3 = Obj3(); # you can't tell wether o1 and o2 where initialized or not

# try doing something if o1 or o2 are set - you can't, o1 > 0 and o2.len > 0, don't really signal wether they were set

  Obj4 = object
    o1: ?Obj1
    o2: ?Obj2


let obj4 = Obj4(o1: 0.some, 02: "".some) 

  # you can now do 
  if obj4.o1.isSome: 
    echo "Yay, o1 was set!"
  else:
    echo "Yaiks, o1 was not set"
```

codex/sales.nim
dryajov: Whats the relationship between the slot and the reservation? Should there be one? The repostore currently simply counts used bytes vs reserved bytes, once a slot has been allocated and being stored, the "reserved" is simply counted as "used".

codex/sales/reservations.nim
emizzle: > having something like @markspanbroek is describing and having it in CLI/config parameter...

I agree with this, it always felt odd being a REST endpoint. Perhaps it made more sense to have a REST endpoint in the old model as it contained a list of availabilities. The other benefit of having a REST endpoint would be modifying the availability without restarting the process. 

Moving forward with the new model as described it would make a lot of sense to include available bytes, duration, and min cost as CLI parameters, and then later if we want the ability to modify these values dynamically, without process restart, we can add that it in. Wdyt?

codex/contracts/interactions/hostinteractions.nim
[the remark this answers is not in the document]
emizzle: I definitely agree it's an odd construct. 

Contextually, this PR split `ContractInteractions` into `HostInteractions` and `ClientInteractions`, in an attempt to prevent passing a RepoStore to constructors that don't need it in the constructor. Effectively it is defining two roles for a node: client and host. There could be cli options passed that allow a node to operate only as a host or only as a node, if desired.

I can modify the constructor return types to return a Result as dmitriy suggested.

emizzle:
> I would expect that information about a particular ongoing slot filling is on chain and can be retrieved without the need of any local state.
Is that the case right now or not?

Yes, that is the case right now. This was the focus of the work we did with restoring state using the state machine.

> However, once a provider has committed to a slot, this information should be global and survive restarts and local database corruption. In fact, this should probably not be even stored locally.

By "this information", are you referring to the availabilities, or which slots a host has filled? This PR handles persisting availabilities *on disk* using the *old model* of availabilities, so that a node can survive restarts, but not necessarily catastrophic loss. The flow is like this: the host hits a REST endpoint and specifies a list of its availabilities. When a request comes in, the parameters of the request are matched with the host's availabilities. If there's a match, the host attempts to fill the slot. Once filled, the slot is recorded as filled *on chain* and its host is recorded as the address of the host filling the slot. The availability, currently, is not persisted on chain, as it can be easily recreated locally using the REST api in the old model or perhaps cli params (discussed above) in the new model.

Now, if a host suffers a *catastrophic* loss, the locally persisted availabilities will be lost. When that host comes back up, the availability will need to be recreated by the host (using whatever model we decide) BEFORE the state restoration routine runs that queries the onchain active slots for the hosts and matches those slots to local availabilities. As a side note, this is another reason I believe the cli parameters would be a good option (along with the "new" model of specifying storage size, duration, min cost) as it will allow the availabilities to be recreated prior to the state restoration routine.

Please be aware that we aimed the scope of this PR not as a full-fledged catastrophic recovery solution, but as a way to allow node restarts in order to enable us to begin user testing. All the above discussion is super important and needs to be had, but just so we are aware, if decide we want to scope all of that into this PR, we will be delaying user testing. Let's make a decision one way or the other. Perhaps we need to discuss on a call?

codex/sales.nim
emizzle: An availability is marked as used by assigning the availability a slotId, which is stored in the repo:
https://github.com/status-im/nim-codex/blob/288c8cdcd881c122bd172beccef9f97525f59a6f/codex/sales/reservations.nim#L243-L248

A "used" availability is simply an availability that has a slotId assigned:
https://github.com/status-im/nim-codex/blob/288c8cdcd881c122bd172beccef9f97525f59a6f/codex/sales/reservations.nim#L103-L104

codex/contracts/interactions/clientinteractions.nim
emizzle: What is left that is duplicated? The main difference between the two is that `HostInteractions` takes a RepoStore in its constructor while `ClientInteractions` does not.

dryajov:
> > I would expect that information about a particular ongoing slot filling is on chain and can be retrieved without the need of any local state.
> > Is that the case right now or not?
> 
> Yes, that is the case right now. This was the focus of the work we did with restoring state using the state machine.
> 
> > However, once a provider has committed to a slot, this information should be global and survive restarts and local database corruption. In fact, this should probably not be even stored locally.
> 
> By "this information", are you referring to the availabilities, or which slots a host has filled? This PR handles persisting availabilities _on disk_ using the _old model_ of availabilities, so that a node can survive restarts, but not necessarily catastrophic loss. The flow is like this: the host hits a REST endpoint and specifies a list of its availabilities. When a request comes in, the parameters of the request are matched with the host's availabilities. If there's a match, the host attempts to fill the slot. Once filled, the slot is recorded as filled _on chain_ and its host is recorded as the address of the host filling the slot. The availability, currently, is not persisted on chain, as it can be easily recreated locally using the REST api in the old model or perhaps cli params (discussed above) in the new model.
> 
> Now, if a host suffers a _catastrophic_ loss, the locally persisted availabilities will be lost. When that host comes back up, the availability will need to be recreated by the host (using whatever model we decide) **_before_** the state restoration routine runs that queries the onchain active slots for the hosts and matches those slots to local availabilities. As a side note, this is another reason I believe the cli parameters would be a good option (along with the "new" model of specifying storage size, duration, min cost) as it will allow the availabilities to be recreated prior to the state restoration routine.
> 
> Please be aware that we aimed the scope of this PR not as a full-fledged catastrophic recovery solution, but as a way to allow node restarts in order to enable us to begin user testing. All the above discussion is super important and needs to be had, but just so we are aware, if decide we want to scope all of that into this PR, we will be delaying user testing. Let's make a decision one way or the other. Perhaps we need to discuss on a call?

I meant specifically the filled slots, in which case it seems like we're on the right track (great!). In other words, we don't care, as you point out, about persisting availabilities across catastrophic failures, after all, the provider can easily recreate them. Filled slots on the other hand do need to be on chain and that appears to be the case after your clarification. So looks like we're good.

With regards to how involved this availabilities flow/handling should be, I would say KISS. We probably don't care about availabilities after they were filled, so matching them to slots is probably no necessary. The information for the duration, size, etc, for filled slots is already on chain, so it should be possible to just use that, trying to synchronize the two, local and on-chain information, is almost certainly an unnecessary complication. 

I would  think of availabilities strictly as pre-contract "offers for sale", in other words, this is local information that can be altered at any point by the storage node without affecting any global state and they get discarded and/or adjusted once a slot has been occupied. Discarded if an availability gets matched to a request 1:1, altered if a request uses part of an availability, in which case the available storage in the availability gets adjusted to reflect the matched request.

dryajov:
> Please be aware that we aimed the scope of this PR not as a full-fledged catastrophic recovery solution, but as a way to allow node restarts in order to enable us to begin user testing. All the above discussion is super important and needs to be had, but just so we are aware, if decide we want to scope all of that into this PR, we will be delaying user testing. Let's make a decision one way or the other. Perhaps we need to discuss on a call?

I think this is totally fine and as I explained in my other comment, probably sufficient overall. Lets keep this PR moving forward and figure out if we need something else in subsequent PRs/iterations.

codex/contracts/interactions/clientinteractions.nim
markspanbroek: It's not a biggie, but this code seems to be the same across `HostInteractions` and `ClientInteractions`:

```nim
  without address =? deployment.address(Marketplace):
    let err = newException(ContractAddressError,
      "Unable to determine address of the Marketplace smart contract")
    return failure(err)

  let contract = Marketplace.new(address, signer)
  let market = OnChainMarket.new(contract)
  let clock = OnChainClock.new(signer.provider)
```

codex/sales/reservations.nim
markspanbroek: > Moving forward with the new model as described it would make a lot of sense to include available bytes, duration, and min cost as CLI parameters, and then later if we want the ability to modify these values dynamically, without process restart, we can add that it in. Wdyt?

Yes, I agree. The REST endpoint made sense to me for the old model, but it no longer makes sense in the new model. When we implement the new model (in a different PR, I'd say), we should move to CLI/config parameters.

codex/node.nim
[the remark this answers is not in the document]
markspanbroek: > Out of curiosity, why do you define `client` & `host` as optional? Is there a possible case where one will be defined and the other not? From top of my head, I can't think of any.

A codex node can play different roles in the network, `host` and `client` being two of them. In the `host` role, the node will offer storage for sale, and only a small portion of nodes in the network will do that. In the `client` role nodes will buy storage, which we expect a larger portion of the nodes in the network to do. But the vast majority of nodes will probably only download content from the network, which requires neither the `host` nor the `client` role.

codex/codex.nim
```diff
@@ -106,10 +109,16 @@ proc new(_: type ContractInteractions, config: CodexConf): ?ContractInteractions
...
+  var client: ?ClientInteractions
+  var host: ?HostInteractions
   if deployment =? config.ethDeployment:
-    ContractInteractions.new(config.ethProvider, account, deployment)
+    client = ClientInteractions.new(config.ethProvider, account, deployment)
+    host = HostInteractions.new(config.ethProvider, account, repo, deployment)
   else:
-    ContractInteractions.new(config.ethProvider, account)
+    client = ClientInteractions.new(config.ethProvider, account)
+    host = HostInteractions.new(config.ethProvider, account, repo)
```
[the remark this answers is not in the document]
emizzle: Good catch, updated.

codex/sales/reservations.nim
emizzle: Let's keep the availabilities via REST for now as discussed in https://github.com/status-im/codex-research/pull/161.

emizzle:
After the above discussion and the discussion in https://github.com/status-im/codex-research/pull/161, the plan for this PR is as follows:
- [ ] fix/update some low-hanging PR comments
- [ ] reserve space as availabilities come in
- [ ] release space as a request completes (finished/failed/error)
- [ ] when a request is matched to an availability, flag availability as `inUse` (in memory only), then delete the availability once slot is filled
- [ ] remove all markUsed/markUnused machinery, as we don't need to persist this info
- [ ] on node restart, for active slots, do not find availability

In follow up PRs, we can add:
- [ ] CLI parameter that specifies an initial % of the host's disk to reserve (TBD, discussion ongoing in https://github.com/status-im/codex-research/pull/161, waiting to understand how we'll do this without distinguishing how much space each module has reserved)
- [ ] ... anything I'm missing?

dryajov:
> After the above discussion and the discussion in [status-im/codex-research#161](https://github.com/status-im/codex-research/pull/161), the plan for this PR is as follows:
> 
> * [ ]  fix/update some low-hanging PR comments
> * [ ]  reserve space as availabilities come in
> * [ ]  release space as a request completes (finished/failed/error)
> * [ ]  when a request is matched to an availability, flag availability as `inUse` (in memory only), then delete the availability once slot is filled
> * [ ]  remove all markUsed/markUnused machinery, as we don't need to persist this info
> * [ ]  on node restart, for active slots, do not find availability
> 
> In follow up PRs, we can add:
> 
> * [ ]  CLI parameter that specifies an initial % of the host's disk to reserve (TBD, discussion ongoing in [Storage space reservations for the sales process codex-research#161](https://github.com/status-im/codex-research/pull/161), waiting to understand how we'll do this without distinguishing how much space each module has reserved)
> * [ ]  ... anything I'm missing?

LGTM, except:

> * [ ]  reserve space as availabilities come in
> * [ ]  release space as a request completes (finished/failed/error)

As discussed in https://github.com/status-im/codex-research/pull/161, we want to reserve a large chunk and release piecemeal as blocks come in.

emizzle:
> As discussed in https://github.com/status-im/codex-research/pull/161, we want to reserve a large chunk and release piecemeal as blocks come in.

My goal was to not put too much scope into this PR and was thinking it would be easiest to implement reserving/releasing as availabilities come in, then later add in the large chunk reservation at startup, which is why I put the first task in the follow up section as:
> CLI parameter that specifies an initial % of the host's disk to reserve (TBD, discussion ongoing in https://github.com/status-im/codex-research/pull/161, waiting to understand how we'll do this without distinguishing how much space each module has reserved)

dryajov:
> 

I see, I would have imagined that allocating upfront is actually simpler, but YMMV.

codex/codex.nim
```diff
@@ -96,7 +96,10 @@ proc stop*(s: CodexServer) {.async.} =
 
   s.runHandle.complete()
 
-proc new(_: type ContractInteractions, config: CodexConf): ?ContractInteractions =
+proc new(_: type ContractInteractions,
```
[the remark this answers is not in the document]
emizzle: Updated.

codex/contracts/interactions/hostinteractions.nim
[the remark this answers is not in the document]
emizzle: Updated.

codex/sales/reservations.nim
emizzle: done

tests/codex/sales/helpers.nim
emizzle: removed `.reversed` as it likely doesn't make a difference

codex/sales/states/downloading.nim
emizzle: done

codex/contracts/interactions/clientinteractions.nim
emizzle: done

codex/sales/states/errored.nim
```diff
@@ -21,14 +23,14 @@ method run*(state: SaleErrored, machine: Machine): Future[?State] {.async.} =
...
-     availability =? data.availability:
-    onSaleErrored(availability)
+  if availability =? data.availability:
+    let reservations = context.reservations
+    if (exists =? await reservations.exists(availability.id)) and
+      exists == true:
 
-    await agent.unsubscribe()
+      if err =? (await reservations.markUnused(availability)).errorOption:
+        raiseAssert "Failed to mark availability unused, error: " & err.msg
```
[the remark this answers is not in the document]
emizzle: done

codex/rest/api.nim
```diff
@@ -248,10 +248,13 @@ proc initRestApi*(node: CodexNodeRef, conf: CodexConf): RestRouter =
     "/api/codex/v1/sales/availability") do () -> RestApiResponse:
       ## Returns storage that is for sale
 
-      without contracts =? node.contracts:
+      without contracts =? node.contracts.host:
         return RestApiResponse.error(Http503, "Sales unavailable")
 
-      let json = %contracts.sales.available
+      without unused =? (await contracts.sales.reservations.unused), err:
+        return RestApiResponse.error(Http500, err.msg)
```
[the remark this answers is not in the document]
emizzle: done

codex/rest/api.nim
```diff
@@ -268,15 +271,19 @@ proc initRestApi*(node: CodexNodeRef, conf: CodexConf): RestRouter =
...
 
       without availability =? Availability.fromJson(body), error:
         return RestApiResponse.error(Http400, error.msg)
 
-      contracts.sales.add(availability)
+      if not contracts.sales.reservations.available(availability.size.truncate(uint)):
+        return RestApiResponse.error(Http422, "Not enough storage quota")
+
+      if err =? (await contracts.sales.reservations.reserve(availability)).errorOption:
+        return RestApiResponse.error(Http500, err.msg)
```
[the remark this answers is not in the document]
emizzle: done

codex/node.nim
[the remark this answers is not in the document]
AuHau: @markspanbroek makes sense, but currently, there is no way how to specify if the node wants to be host or client right? (That is why I have pointed this out) So this is envisioned to be configurable in the future?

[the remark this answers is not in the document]
markspanbroek: Good point @AuHau, you currently cannot specify this on the command line. I do want to keep the option open in the architecture though. We're currently starting things like `Proving` and `Sales`, even when we don't have any storage available to sell. Which is fine for now, it's harmless. But I'd like to have the option in the future to only start this machinery when needed.

emizzle:
I will fix the conflicts first thing tomorrow, but I think this can get another review in the meantime.

codex/contracts/deployment.nim
```diff
@@ -12,8 +12,8 @@ const defaultFile = "vendor" / "codex-contracts-eth" / "deployment-localhost.jso
 ## been exported with Hardhat deploy.
 ## See also:
 ## https://github.com/wighawag/hardhat-deploy/tree/master#6-hardhat-export
-proc deployment*(file = defaultFile): Deployment =
-  Deployment(json: parseFile(file))
+proc deployment*(file: ?string = some defaultFile): Deployment =
```
markspanbroek, line 15: This looks a bit weird to me: an optional parameter with a default value.
Perhaps we can avoid this change; see comments below.

codex/contracts/interactions/clientinteractions.nim
```diff
@@ -0,0 +1,52 @@
...
+export purchasing
+export chronicles
+
+type
+  ClientInteractions* = ref object of ContractInteractions
+    purchasing*: Purchasing
+
+proc new*(_: type ClientInteractions,
+          signer: Signer,
+          deployment: Deployment): ?!ClientInteractions =
```
markspanbroek, line 23: I'd avoid propagating the `Deployment` interface further into the code. We'll probably want to get rid of it (#372). Perhaps we can replace the `deployment` parameter with a `marketplaceAddress` parameter?

codex/contracts/interactions/hostinteractions.nim
```diff
@@ -0,0 +1,62 @@
...
+export chronicles
+
+type
+  HostInteractions* = ref object of ContractInteractions
+    sales*: Sales
+    proving*: Proving
+
+proc new*(_: type HostInteractions,
+          signer: Signer,
+          deployment: Deployment,
```
markspanbroek, line 24: Same as for ClientInteractions, I'd avoid using `Deployment` here, and perhaps use `marketplaceAddress` instead.

codex/sales/reservations.nim
```diff
@@ -0,0 +1,349 @@
...
+  x.size == y.size and
+  x.duration == y.duration and
+  x.minPrice == y.minPrice
+
+proc `$`*(id: AvailabilityId): string = id.toArray.toHex
+
+proc toErr[E1: ref CatchableError, E2: AvailabilityError](
+  e1: E1,
+  _: type E2,
+  msg: string = "see inner exception"): ref E2 =
```
markspanbroek, line 91: I don't think that Nim by default prints the inner exception, so perhaps it's better to use the original msg?

```suggestion
  msg: string = e1.msg): ref E2 =
```

codex/sales/salescontext.nim
```diff
@@ -11,15 +14,18 @@ type
     onStore*: ?OnStore
     onClear*: ?OnClear
     onSale*: ?OnSale
-    onSaleErrored*: ?OnSaleErrored
     proving*: Proving
+    reservations*: Reservations
+  # TODO: do not declare BatchProc as it's declared in node, but causes a
+  # circular dep if node is imported
+  BatchProc* = proc(blocks: seq[bt.Block]): Future[void] {.gcsafe, upraises:[].}
```
markspanbroek, line 21: Perhaps we can move this definition into a separate module, that is imported in both `node.nim` and `salescontext.nim`?

codex/sales/states/downloading.nim
```diff
@@ -38,5 +46,49 @@ method run*(state: SaleDownloading, machine: Machine): Future[?State] {.async.}
...
+      request.ask.slotSize,
+      request.ask.duration,
+      request.ask.pricePerSlot,
+      used = false):
+    info "no availability found for request, ignoring",
+      slotSize = request.ask.slotSize,
+      duration = request.ask.duration,
+      pricePerSlot = request.ask.pricePerSlot,
+      used = false
+    return some State(SaleFinished())
```
markspanbroek, line 59: When there is no availability, we shouldn't move to the "finished" state. The finished state signals that the sale was successful, and that we should start providing storage proofs.

codex/stores/repostore.nim
```diff
@@ -64,6 +64,12 @@ iterator items*(q: BlockExpirationIter): Future[?BlockExpiration] =
 func totalUsed*(self: RepoStore): uint =
   (self.quotaUsedBytes + self.quotaReservedBytes)
 
+func available*(self: RepoStore): uint =
+  return self.quotaMaxBytes - self.totalUsed
+
+func available*(self: RepoStore, bytes: uint): bool =
```
markspanbroek, line 70: I would probably call this `hasAvailable()`, to differentiate from the `available` property above.

tests/codex/sales/testsales.nim
```diff
@@ -4,157 +4,192 @@ import std/sugar
...
     request.expiry = (clock.now() + 42).u256
 
   teardown:
+    await repo.stop()
     await sales.stop()
 
-  test "has no availability initially":
-    check sales.available.len == 0
+  proc getAvailability: ?!Availability =
+    waitFor reservations.get(availability.id)
```
markspanbroek, line 78: I'd avoid using `waitFor` here. We're already in an async context, so we can use `await`.

tests/contracts/testInteractions.nim
```diff
@@ -1,32 +1,64 @@
 import std/os
-import codex/contracts
+import pkg/datastore
+import pkg/codex/contracts
+import pkg/codex/stores
 import ../ethertest
 import ./examples
 
-ethersuite "Marketplace Contract Interactions":
+# suite "Marketplace Contract Interactions":
```
markspanbroek, line 8:
```suggestion
```

markspanbroek (commented):
Nice work @emizzle. Looks really good! I left a few small comments.

codex/sales/states/downloading.nim
emizzle: True, we should perhaps define a start where eg `SaleIgnored` such that the state of an agent doesn't remain on `SaleDownloading` when it has been ignored.

tests/codex/sales/testsales.nim
emizzle: This was done for a few reasons:
1. In the tests, we're not going to maximise our use of the async event loop
2. The test check is cleaner when `getAvailability` is sync
3. Questionable doesn't seem to support chaining an async proc, eg

This works:
```nim
proc getAvailability: ?!Availability =
  waitFor reservations.get(availability.id)

check getAvailability().?used == success false
```

This doesn't work:
```nim
proc getAvailability: Future[?!Availability] {.async.} =
  return await reservations.get(availability.id)

check (await getAvailability()).?used == success false
```
Errors with:
```shell
/Users/egonat/repos/status-im/nim-codex/vendor/questionable/questionable/chaining.nim(21, 20) Error: undeclared field: 'unsafeGet' for type asyncloop.Future [type declared in /Users/egonat/repos/status-im/nim-codex/vendor/nim-chronos/chronos/asyncfutures2.nim(56, 3)]
```

codex/contracts/deployment.nim
emizzle: Updated the `HostInteractions`, `ClientInteractions`, and `Deployment` constructors.

codex/contracts/interactions/clientinteractions.nim
emizzle: Updated.

codex/sales/states/downloading.nim
emizzle: Perhaps we'd also like to remove the sales agent from the list of agents in the sales module once Finished/Ignored?

codex/sales/salescontext.nim
emizzle: Done.

codex/sales/reservations.nim
emizzle: Good call.

codex/sales.nim
emizzle: I think availability is still accurate here as availability is added via the REST api as a first step.

markspanbroek (approved):

PR #346, opened 2023-02-15
Add a new backports module for mainline Nim APIs "from the future".

c-blake:
For a variety of understandable reasons, Status tends to stay on lagged versions of Nim.  Sometimes new APIs useful "today" (and much more broadly) get added to Nim main.  This PR creates a new `utils/backports.nim` module to give a home for small APIs like this.  This API in particular was developed in https://github.com/nim-lang/Nim/pull/21349.

It is conceivable the granularity should be a whole directory (e.g. `backports/parseutils.nim`), but it seemed reasonable to wait until we had >1 exported proc to worry about the next level up of modularity / this sort of organization.

### DIFF
--- a/codex/utils/backports.nim
+++ b/codex/utils/backports.nim
@@ -1,0 +1,73 @@
+## A module for backports of small APIs *already included* in mainline Nim
+## (unlike, say the "stew" package) BUT in (possibly) later versions than in use
+## at Status/NBS.  Formatting is slightly unorthodox to keep "diffs" between a
+## version here and the one in mainline almost nil (E.g. up to `when`s & func
+## header indents).
+
+import std/parseutils
+
+when not declared(parseSize):
+ func toLowerAscii(c: char): char =
+  if c in {'A'..'Z'}: char(uint8(c) xor 0b0010_0000'u8) else: c
+
+ func parseSize*(s: openArray[char], size: var int64, alwaysBin=false): int =
+  ## Parse a size qualified by binary or metric units into `size`.  This format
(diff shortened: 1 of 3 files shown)


c-blake:
Unit tests have been written (and pass the Nim CI) here: https://github.com/nim-lang/Nim/blob/devel/tests/stdlib/tparseutils.nim#L62

How these should be represented in our own unit tests struck me as an open question.  I was kind of imagining this one module having probably less than 5..10 calls in it -- maybe ever.  So, we could probably just add a parallel `tests/code/utils/testbackports.nim` module.

dryajov:
> Unit tests have been written (and pass the Nim CI) here: https://github.com/nim-lang/Nim/blob/devel/tests/stdlib/tparseutils.nim#L62
> 
> How these should be represented in our own unit tests struck me as an open question. I was kind of imagining this one module having probably less than 5..10 calls in it -- maybe ever. So, we could probably just add a parallel `tests/code/utils/testbackports.nim` module.

Yep, placing the test in `tests/code/utils/testbackports.nim` sounds like the right way to go about it for now.

arnetheduck:
fyi we generally put stuff like this in https://github.com/status-im/nim-stew/tree/master/stew/shims

dryajov:
> fyi we generally put stuff like this in https://github.com/status-im/nim-stew/tree/master/stew/shims

Not opposed, we can move it to stew.

c-blake:
This is in stew/shims/parseutils now.  Closing this.

PR #347, opened 2023-02-16
Blockstore maintenance

benbierens:


codex/stores/maintenance.nim
```diff
@@ -0,0 +1,93 @@
...
+    repoStore: repoStore,
+    interval: interval,
+    timer: timer,
+    clock: clock,
+    numberOfBlocksPerInterval: numberOfBlocksPerInterval,
+    offset: 0
+  )
+
+proc runBlockCheck(self: BlockMaintainer): Future[void] {.async.} =
+  var blockCidsToDelete = newSeq[Cid](0)
```
dryajov, line 49: I'm not sure if this makes much sence, let me explain - the point of an iterator is that we don't have to consume all items coming from the source before having to process them, essentially, an `iterator` is a streaming primmitive, what's being done here invalidates the purpose of the `iterator`. I would suggest we rework this so that we delete expired records as they are being read from the `iterator`.

codex/utils/timer.nim
```diff
@@ -0,0 +1,54 @@
...
+    name: timerName,
+    loopFuture: Future[void].none
+  )
+
+proc timerLoop(timer: Timer) {.async.} =
+  try:
+    while true:
+      await sleepAsync(timer.interval)
+      await timer.callback()
+  except Exception as exc:
```
dryajov, line 37: `Exception` is the root of the exception hierarchy in Nim, including fatal errors (ouch!), usually anything except `CatchableError` should be left alone.

dryajov (commented):
Nice job! 

I know it's still a wip, but it's a great start, with the exception of a few gotchas and observations, which I left comments for. 

Namelly:

- the timer primitives which should be kept as `chronos` 
- the use of `Exception` and
- the iterator related changes

codex/stores/blockstore.nim
```diff
@@ -50,7 +48,7 @@ method getBlock*(self: BlockStore, cid: Cid): Future[?!Block] {.base.} =
 method putBlock*(
   self: BlockStore,
   blk: Block,
-  ttl = Duration.none): Future[?!void] {.base.} =
+  ttl = times.Duration.none): Future[?!void] {.base.} =
```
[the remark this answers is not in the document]
benbierens: I chose not to use chronos' Duration or Moment for this use-case. It uses an int64 to represent time in nano-seconds. I remember figuring out that using this for timestamps would have us run out of bits some time in the 2040s. I checked just now, and it seem we wouldn't run out until 2262. Perhaps you could check? Mark's been using a SecondsSince1970 type as an int64 and I kind of want to use that. It's easy to understand and definitely future-proof.

I saw the Duration and Moment from chronos were definitely designed to run with its high-speed timer code. Its largest supported time constant (nim-chronos/chronos/timer.nim line 298) is Week.

codex/utils/timer.nim
```diff
@@ -0,0 +1,54 @@
...
+
+method start*(timer: Timer, callback: TimerCallback, interval: Duration) {.base.} =
+  if timer.loopFuture.isSome():
+    return
+  trace "Timer starting: ", timer.name
+  timer.callback = callback
+  timer.interval = interval
+  let future = timerLoop(timer)
+  timer.loopFuture = future.some
+  asyncSpawn future
```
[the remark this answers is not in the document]
benbierens: Removed it, still works 👌

codex/stores/blockstore.nim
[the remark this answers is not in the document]
dryajov: I wasn under the impression we were mixing chronos and std timers, it appears that we aren't, which is good. Using epoch here actually makes sence since we're simply recording it in the store.

codex/stores/maintenance.nim
```diff
@@ -0,0 +1,96 @@
...
+import pkg/questionable
+import pkg/questionable/results
+
+import ./repostore
+import ../utils/timer
+import ../clock
+import ../systemclock
+
+const
+  DefaultBlockMaintenanceIntervalSeconds* = 10 * 60
```
dryajov, line 24: `std/times` also has human readable units, we should probably just use `10.minutes`.

codex/stores/repostore.nim
```diff
@@ -41,11 +45,9 @@ const
 
   BlocksTtlKey* = Key.init(CodexBlocksTtlNamespace).tryGet
 
-  DefaultBlockTtl* = 24.hours
+  DefaultBlockTtlSeconds* = 24 * 60 * 60
```
dryajov, line 48: same, should be possible to do `24.hours` using `std/times`.

codex/codex.nim
```diff
@@ -130,6 +130,7 @@ proc new*(T: type CodexServer, config: CodexConf, privateKey: CodexPrivateKey):
 
   if config.cacheSize > 0:
     cache = CacheStore.new(cacheSize = config.cacheSize * MiB)
+    ## Is unused?
```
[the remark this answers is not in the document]
benbierens: If you can explain the plan for this cacheStore, I'll happily take a stab at it.

codex/stores/repostore.nim
```diff
@@ -222,6 +243,41 @@ method listBlocks*(
   iter.next = next
   return success iter
 
+proc createBlockExpirationQuery(maxNumber: int, offset: int): ?!Query =
+  let queryString = ? (BlocksTtlKey / "*")
+  let queryKey = ? Key.init(queryString)
+  success Query.init(queryKey, offset = offset, limit = maxNumber)
+
+method getBlockExpirations*(self: RepoStore, maxNumber: int, offset: int): Future[?!BlockExpirationIter] {.async, base.} =
```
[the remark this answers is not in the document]
benbierens: In the current arrangement, the maintenance module does not interact with any datastores, only with the repoStore.
If we move getBlockExpirations to maintenance, then it will need to know about the metadata Sqllite store, and both it and the repostore will be using it. I figured I'd keep the knowledge of the datastore and how to format keys for it in one place, but I do agree getBlockExpirations feels a bit out of place here.
Suppose I could move the creating and retrieving of the expiration metadata both to the maintenance module? Then repostore uses the metadata store for the quote key, and maintenance uses it for expiration.... Do you like this plan? [Y/N]

codex/stores/repostore.nim
```diff
@@ -93,10 +107,24 @@ method getBlock*(self: RepoStore, cid: Cid): Future[?!Block] {.async.} =
   trace "Got block for cid", cid
   return Block.new(cid, data)
 
+proc getBlockExpirationTimestamp(self: RepoStore, ttl: ?times.Duration): SecondsSince1970 =
```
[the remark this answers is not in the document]
benbierens: Did you mean 
```
proc createBlockExpirationMetadataKey(self: RepoStore, cid: Cid): ?!Key =
```
?

codex/stores/maintenance.nim
benbierens: I understand that using the seq[Cid] goes against the way iterators work. I used it here to avoid the classic 'modifying a collection during iteration'. I would imagine the underlying datastore at the moment can handle it, but I didn't want to build this assumption into the code.
I considered the max memory use of seq[Cid] and I figured it is probably harmless. What do you think?

codex/utils/timer.nim
benbierens: Changed it to CatchableError.
Is it true that because this is running in a Future, anything not caught here will cause the application to crash?

codex/stores/blockstore.nim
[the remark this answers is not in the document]
benbierens: Well, your impression is partially right: I am using both chronos and std/times Duration types. I need to use chronos for the asyncSleep method. And if I only use chronos then I have to convert seconds into nanoseconds, and then back again for no reason. I don't understand why chronos introduces its own Duration type. The std/times one seems to do everything the chronos one does.
So, the way I see it:
A) Use chronos in both cases and do unnecessary unit conversion
B) Use chronos for the timer duration and std/times for the blockTTL duration
I don't really like either option. What do you think?

codex/stores/cachestore.nim
```diff
@@ -156,7 +157,7 @@ func putBlockSync(self: CacheStore, blk: Block): bool =
 method putBlock*(
   self: CacheStore,
   blk: Block,
-  ttl = Duration.none): Future[?!void] {.async.} =
+  ttl = times.Duration.none): Future[?!void] {.async.} =
```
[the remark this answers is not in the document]
benbierens: Will update if necessary following from the previous comment thread

codex/stores/networkstore.nim
```diff
@@ -49,7 +50,7 @@ method getBlock*(self: NetworkStore, cid: Cid): Future[?!bt.Block] {.async.} =
 method putBlock*(
   self: NetworkStore,
   blk: bt.Block,
-  ttl = Duration.none): Future[?!void] {.async.} =
+  ttl = times.Duration.none): Future[?!void] {.async.} =
```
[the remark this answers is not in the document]
benbierens: Nice improvement, thanks!

codex/stores/blockstore.nim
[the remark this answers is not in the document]
dryajov: So `chronos` has it's own functions for epoch:

```Nim
func epochSeconds*(moment: Moment): int64 =
  moment.value div Second.value

func epochNanoSeconds*(moment: Moment): int64 =
  moment.value
```

We want to stick with `chronos` because we're heavily dependent on it across the stack. Check out the `chronos.timer` module, specially the `Moment.now`, `Moment.init` and epoch methods. 

My approach with the `ZeroMoment` was based on an older version of `chronos` that didn't expose the epoch methods, which this, the systems times module should not be necessary anymore.

codex/stores/repostore.nim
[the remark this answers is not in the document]
dryajov: Yeah, moving the key helpers to their own module.

codex/stores/maintenance.nim
dryajov: The underlying store should definitely support this functionality, sqlite does and most other DB or KV stores support it as well, the only other thing could be the `fsstore`, but it is implemented to be safe there as well. In other words, we should probably make it a requirement for stores to support retrive-modify operations with cursors/iterators.

codex/utils/timer.nim
dryajov: > Is it true that because this is running in a Future, anything not caught here will cause the application to crash?

It depends. 

If you `asyncSpawn` the future any exception is propagated freely without any chance of being caught, so any spawned proc should guard itself with a `try/except CatchableError ...`. 

Futures that are awaited or stored in a variable, can be handled normally with a `try/except/finally` as you would expect.

so:

```nim
proc throws(): Future[int] {.async.} = 
  raise newException(CatchableError, "Error!")

# 1) handling exceptions and propagating cancellations
try:
  discard await throws()
except CatchableError as exc:
  # do something
except CancelledError as exc:
   raise  exc

# 2) store future in variable
let fut = throws() 

# same as `1`
try:
  discard await fut
except CatchableError as exc:
  # do something
except CancelledError as exc:
   raise  exc

# same as `1` but reading the value from the future, will reaise the stored exception if it failed (same as `await`)
try:
  discard fut.read()
except CatchableError as exc:
  # do something
except CancelledError as exc:
   raise  exc

# 3) `asyncSpawn` that crashes the app, nothing the catch the exception
asynSpawn throws()

# 4) `asyncSpawn` that doesn't crash (should not propagate cancellations!)
proc safeSpawn() {.async.} =
  try:
    discard await throws()
  except CatchableError as exc:
    # do something without crashing

asyncSpawn safeSpawn()

```

Cancellation should almost always be propagated, unless they are in an async spawn (or similar code).

benbierens: This makes sense. Thanks. It's a good reason to prefer storing futures in variables.

codex/stores/maintenance.nim
benbierens: All right, easy fix :D

codex/stores/blockstore.nim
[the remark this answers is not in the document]
benbierens: I've removed std/times and used chronos for Durations everywhere. And I'm using SecondsSince1970 for timestamps. This gives a conversion overhead, but it's not a problem and it does clean up some code.

codex/codex.nim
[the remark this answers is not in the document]
dryajov: The cache store is just an LRU in memory store, which we use as a cache. It was introduced before we had the `Datastore` abstraction, so it might be a little out of place right now. We can still be used but we might just want to rewrite it as a `Datastore`, which makes it more generic. 

On the other hand, the block store abstraction is just as good, so it might make sense to continue using as is - I'm not sure, certainly the `Datastore` route is more work because we don't have it implemented.

In any case, this would required either explicit calls to the cache, before falling back to the network or local store, or use something like a `TieredDatastore` to dispatch the calls for us.

dryajov (changes requested):
I left a few more minor comments, but we're getting closer :)

codex/stores/repostore.nim
```diff
@@ -93,6 +103,17 @@ method getBlock*(self: RepoStore, cid: Cid): Future[?!Block] {.async.} =
   trace "Got block for cid", cid
   return Block.new(cid, data)
 
+proc getBlockExpirationTimestamp(self: RepoStore, ttl: ?Duration): SecondsSince1970 =
+  let duration = ttl |? self.blockTtl
+  self.clock.now() + duration.seconds
```
dryajov, line 108:
```suggestion
  Moment.fromNow(duration).epochSeconds()
```

codex/stores/repostore.nim
```diff
@@ -54,12 +54,25 @@ type
     postFixLen*: int
     repoDs*: Datastore
     metaDs*: Datastore
+    clock: Clock
```
dryajov, line 57: This should be removed...

codex/stores/maintenance.nim
```diff
@@ -0,0 +1,96 @@
...
+    return numberReceived
+
+  let expirations = await self.repoStore.getBlockExpirations(
+    maxNumber = self.numberOfBlocksPerInterval,
+    offset = self.offset
+  )
+
+  without iter =? expirations, err:
+    trace "Unable to obtain blockExpirations iterator from repoStore"
+    return
```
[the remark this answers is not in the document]
benbierens: I moved two of three nested procs out. There was no reason why couldn't stand alone. (I didn't know there'd be memory overhead for nesting them if I didn't use any of the outer-scoped variables.) I've changed the iterator usage to a for-each. runBlockCheck is now a comfortable 20 lines.

codex/utils/keyutils.nim
```diff
@@ -44,3 +47,10 @@ proc setupKey*(path: string): ?!PrivateKey =
   return PrivateKey.init(
     ? path.readAllBytes().mapFailure(CodexKeyError))
     .mapFailure(CodexKeyError)
+
+proc createBlockExpirationMetadataKey*(cid: Cid): ?!Key =
```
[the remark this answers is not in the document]
benbierens: This makes more sense, yes :D I shall find them and move them.

codex/utils/timer.nim
```diff
@@ -0,0 +1,50 @@
...
+    return
+  trace "Timer starting: ", name=timer.name
+  timer.callback = callback
+  timer.interval = interval
+  timer.loopFuture = timerLoop(timer)
+
+method stop*(timer: Timer) {.async, base.} =
+  if timer.loopFuture != nil:
+    trace "Timer stopping: ", name=timer.name
+    await timer.loopFuture.cancelAndWait()
```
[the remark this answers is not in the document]
benbierens: I checked by changing the while true in the timerLoop to something that finishes. Then I used a test-case to check. cancelAndWait() does not throw when the timerLoop has already returned. I would like to make a unit test for this, but the while-true prevents this, except for the case of an exception which breaks out of the loop. But this case is already tested in "Exceptions raised in timer callback are handled". (It calls timer.stop and expects no exceptions)

codex/stores/repostore.nim
benbierens: I would not like to remove the clock member from the repostore. The ability to inject a test-mock is just too convenient. Not directly referencing static library methods from business-logic modules is perhaps a good idea in general. However, I did try and use Moment.now().epochSeconds() instead of std/times in systemclock.nim. As this is the place where I want to wrap the system time. Unfortunately, it seems to be incorrect by a sizable margin. At the moment of running the test for systemclock:
std/times - now.utc as Unix yields 1677060610
chronos - Moment.now().epochSeconds() yields 180628
If the difference was mere hours, timezones could be to blame, but 180628 is January 3 in 1970. I had a look at the chronos code to see what's happening, but I'm afraid the platform-specific details happening there are a bit out of my league.

codex/codex.nim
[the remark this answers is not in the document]
benbierens: I understand what the CacheStore does. I don't understand why it does it. Can you explain to me a use-case where this would be used?

I think it would be pretty nice to implement a tiered-date-store module: When you call it, it'll pass the call to any number of other datastores in order. In practice that could be cache, local, network. Although, I imagine this makes more sense of getting blocks than it does for putting blocks.

Either way, this should probably be a new branch and a new PR, I'd say. Let's wrap this one up first, yeah? :D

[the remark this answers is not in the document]
dryajov: The `CacheStore` stores unserialized blocks in memory, it helps when the block needs to be read from a slow medium, like a spinning drive. You probably won't notice it's absence with a fast SSD tho, thats why it should be optional.

codex/stores/repostore.nim
dryajov: > I did try and use Moment.now().epochSeconds() instead of std/times in systemclock.nim. As this is the place where I want to wrap the system time. Unfortunately, it seems to be incorrect by a sizable margin. At the moment of running the test for systemclock:

Yes, it is indeed broken and the `epoch` methods should have never been exported as the values are opaque and not supposed to be portable... Pity because I hate mixing clocks in this case, it makes things error prone.

> The ability to inject a test-mock is just too convenient.

Mocking _is_ convenient and often overused. In general, mocking makes sense when fine grained control of the different layers is required, otherwise, it's most likely not testing much. 

For example, in the case of the clock, testing for expirations can be done by simply doing an `await sleepAsync(1.seconds)` or similar, and in the case of the repostore, instead of mocking, its possible to initialize the store with in-memory sqlite DBs like this - `SQLiteDatastore.new(Memory).tryGet()` - take a look at how it's done here https://github.com/status-im/nim-datastore/blob/master/tests/datastore/sql/testsqliteds.nim#L18. 

It isn't a big deal, but "two method" classes and other glue should be avoided because it adds unnecessary clutter to the codebase...

markspanbroek: > or example, in the case of the clock, testing for expirations can be done by simply doing an `await sleepAsync(1.seconds)` or similar,

I agree that mocking is often overused, and using in-memory databases and http servers is often a great way to avoid mocks. But with clocks I don't agree. If you start adding sleeps to your tests, then they all add up when you do a full test run. This slows down the feedback cycle, which slows down development for the entire team.

dryajov: This sleeps can literally be milliseconds, which adds close to nothing to the duration of the test...

benbierens: I think I disagree with both of you.... :D
When a unit-test fails, I want to know for certain that the error is in the module that the test is for. Using real external dependencies (timers, in-memory databases, etc) adds code/behavior to the test. Not only do you have to keep track of this while writing the test, (writing around the behavior of these modules to get them to act how you want for your test) but changes to modules _other_ than the unit under test can cause a unit test to fail. Worse still: If a faulty change is made to a fundamental module of the system that is never mocked out in testing, many or all tests can fail. This removes the use of tests for locating faults in the codebase.
Now you might say the in-memory-DB, Key.init, Cid.init, etc are external libraries and so they are unlikely to cause trouble in this way. Fair enough. But we are building direct dependencies here, and that will make the code (test-code or not) more fragile.
In unit tests, ideally, I want everything _except_ the unit under test to be fake, completely under control of the test, and in no way responsive to changes elsewhere in the codebase. (Date-types not withstanding.)

I know some people object to this one-to-one unit testing approach by saying tests of this sort will do nothing to safeguard the functionality of the total system. This is true, but that's what integration tests are for.
</rant>

I am perfectly happy to set my opinions aside and rolling with the codebase as it's been going. (As you see, I've used the in-memoryDB myself in this changeset.) But here you have my take on the topic.

dryajov: Thanks for sharing your point of view! I'd say your reasoning is uncontroversial and pretty common, if anything, Mark's and my takes are probably the controversial ones. Never the less, let me provide some counterpoints for contrast.

> When a unit-test fails, I want to know for certain that the error is in the module that the test is for.

Yes, but it is a balancing act between "coercing your code into doing something" vs "actually making sure that the code does what you want". Presumably, what we want is the latter.

> Using real external dependencies (timers, in-memory databases, etc) adds code/behavior to the test. Not only do you have to keep track of this while writing the test, (writing around the behavior of these modules to get them to act how you want for your test) but changes to modules _other_ than the unit under test can cause a unit test to fail.

What your describing here, is closer to how the application is going to "exist" in the wild vs our idealized view of that. 

But closer to your point, yes, mocking is usefull when you don't have well behaved modules or when you need more control than you otherwise get from these modules. In other words, mocking is a sort of necessary evil, even by your own account - "writing around the behavior of these modules to get them to act how you want for your test".

But more broadly, mocking is a usefull tool, but as every tool it has a time and a place.

> Worse still: If a faulty change is made to a fundamental module of the system that is never mocked out in testing, many or all tests can fail.

For all intents and purposes, this is good! You want your test to fail when the underlying functionality has changed in some fundamental way.

> This removes the use of tests for locating faults in the codebase.

I would argue that this extends beyond the surface of the codebase and improves the reach of testing.

> Now you might say the in-memory-DB, Key.init, Cid.init, etc are external libraries and so they are unlikely to cause trouble in this way.

Thats never the point, the point is to be able to locate and detect errors, be it in code directly under the control of the project or any third party modules it relies on. This is a fundamental point, and I'm willing to concede that what we do is **not** unit testing, in the strict cense of what unitesting means, but I would also argue, that the usefulness of unitesting is quite limited if looked at objectively.

> But we are building direct dependencies here, and that will make the code (test-code or not) more fragile.

In a way we are and it is intentionall. Partly because of the nature of the software we're building and partly due to the very well understood limitations of unittests and their innadequacy in many scenarios.

So what do we actually do? What is our (so far informal) testing approach? I would characterize it as mostly bottom-up testing vs top-down testing. With top-down testing being more akin to unittesting.

It mostly relates to where the complexity of the software lies and what is the most convenient way to build it? Is it top down or bottom up? 

If the complexity is mostly at the surface, most likely it would start being build top down. This is where mocking is usually prevalent, a mock is built for the components bellow the ones being tested as you keep descending down the component tree. Unsurprisingly this is usually software with complex user interactions, often complicated UIs or similar - sales, trading, medical, etc...

Codex on the other hand, has very little complexity at the top, most of the core functionality sits behind two calls/methods - `upload` and `download`. As you descend, you get more and more complexity. How you deal with that depends, but its common to start either somewhere in the middle of the component tree or at the very bottom... Once you start building it from the bottom up, most of the components naturally complement each other as you keep ascending and, there is rarely a need for mocking, except for a few narrow cases...

> I know some people object to this one-to-one unit testing approach by saying tests of this sort will do nothing to safeguard the functionality of the total system.

I've seen this, but I've never heard convincing enough arguments for why that is, I think unittesting is perfectly fine, but it cannot be treated as the one and only way of testing correctness and complettness in isolation. I think it all depends on the nature of the software, where complexity lies and how it gets built - as per the arguments above.

> that's what integration tests are for.

Indeed, bottom-up testing is a sort of integration testing, because it allows you to integrate the system as the components are being build from the bottom up, but it is fine grained enough that it allows you to do that with enough control over the flow that it suffices to catch and isolate issues without too much hassle.

Our other integration tests, test the entire system as a whole, with all the components plugged in and an environment as close to a real setup as possible.

I hope this gives a good enough rationale for why things are the way they are :)

codex/codex.nim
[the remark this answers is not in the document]
dryajov: > Either way, this should probably be a new branch and a new PR, I'd say. Let's wrap this one up first, yeah? :D

Totally, it needs its own PR.

tests/testIntegration.nim
```diff
@@ -9,6 +9,7 @@ import ./ethertest
 import ./contracts/time
 import ./integration/nodes
 import ./integration/tokens
+import ./integration/testblockexpiration
```
dryajov, line 12: Maybe it's time split this out into it's own module? One for contract negotiation and another for block maintenance? We usually have the `testSomething.nim` reserved for calling the individual test units.

dryajov: Don't have to be in this PR btw...

dryajov (approved):
Looks good, besides a few nit-pics. The mocking piece is an interesting discussion, but we can probably move forward with it for now.

dryajov:
Lets get this merged and create a couple of new PRs:

- Split `testIntegration.nim` into two or more modules and use `testIntegration.nim` as the executor module (the module that includes other integration tests), put integration tests into their own folders for more clarity and consistency.
- New PR to wire the in memory cache into codex
  - We probably want to do it with the datastore abstraction as discussed previously, but not critical for now, we could just reuse `CacheStore` as is

codex/stores/repostore.nim
benbierens: I understand, thank you. I don't think we're going to agree on this point, but luckily, we don't have to in order to work well together. When I'm writing code, I always ask myself how much knowledge is in the function or module I'm working on. If it has a lot of knowledge about the layout of many parts of the system, the code's going to be very fragile and will need constant tinkering to keep up with changes elsewhere in the codebase. I'm always looking for way to hide away and abstract out knowledge from my modules, to make them make sense in an isolated way, so to speak. I know I have had some respectful disagreements on where to draw this line with.... most of my team members in the past. But as I said, I shall in no way insist on doing things my way.

PR #349, opened 2023-03-01
[sales statemachine] add async state machine error handling

emizzle:
- add error handling to catch errors during `state.run`
- Sales: create `ErrorHandlingState` to use a base class for handling the `onError` callback, which is called when there is an error in the state's run method.
- Sales: `onError` can be overridden by individual states if needed.
- Because there is already a state named SaleFailed which is meant to react to an onchain `RequestFailed` event and also because this callback is called from `SaleErrored`, renaming to `onSaleErrored` prevents ambiguity and confusion as to what has happened at the callback callsite.
### NOTE
- base branch is #306

### DIFF
--- a/codex/sales/states/errorhandling.nim
+++ b/codex/sales/states/errorhandling.nim
@@ -1,0 +1,9 @@
+import pkg/questionable
+import ../statemachine
+import ./errored
+
+type
+  ErrorHandlingState* = ref object of SaleState
+
+method onError*(state: ErrorHandlingState, error: ref CatchableError): ?State =
+  some State(SaleErrored(error: error))
--- a/tests/codex/helpers/mockmarket.nim
+++ b/tests/codex/helpers/mockmarket.nim
@@ -115,7 +115,11 @@ method requestState*(market: MockMarket,
 
 method slotState*(market: MockMarket,
                   slotId: SlotId): Future[SlotState] {.async.} =
-  return market.slotState[slotId]
+  if market.slotState.hasKey(slotId):
+    return market.slotState[slotId]
+  else:
+    return SlotState.Free
+
 
 method getRequestEnd*(market: MockMarket,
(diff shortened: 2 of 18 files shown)


markspanbroek (approved):
Thanks Eric, this is a nice improvement on the state machine, which collapses a lot of duplicated error handling code!

Instead of adding review comments, I have added a few commits to your branch with things that I thought would be nice improvements. Feel free to revert them if they don't make sense.

tests/codex/helpers/mockmarket.nim
```diff
@@ -115,7 +115,11 @@ method requestState*(market: MockMarket,
 
 method slotState*(market: MockMarket,
                   slotId: SlotId): Future[SlotState] {.async.} =
-  return market.slotState[slotId]
+  if market.slotState.hasKey(slotId):
+    return market.slotState[slotId]
+  else:
+    return SlotState.Free
```
emizzle, line 121: This would implicit, but definitely helps readability when explicit 👍

codex/sales/states/errorhandling.nim
```diff
@@ -0,0 +1,9 @@
+import pkg/questionable
+import ../statemachine
+import ./errored
+
+type
+  ErrorHandlingState* = ref object of SaleState
+
+method onError*(state: ErrorHandlingState, error: ref CatchableError): ?State =
```
emizzle, line 8: This should annotated with `{.base.}`?

emizzle:
This is a much better way to get around the circular dependency that I struggled with 👏 It would be great if we could test that overriding `onError` inside of a `ErrorHandlingState` (in an integration test capacity) as we are creating a new inheritance pattern in the sales state machine that isn't covered by the asyncstatemachine tests. But testing this doesn't really seem possible and I guess we are relying on the inheritance properties of the nim compiler, which should be tested in their own right.

PR #351, opened 2023-03-02
[sales statemachine] prevent memory leaks and add salesagent tests

emizzle:
1. prevent memory leaks and nil access defects by:
  - allowing multiple subscribe/unsubscribes of salesagent
  - disallowing individual salesagent subscription calls to be made externally (requires the .subscribed check)
  - allowing mutiple start/stops of asyncstatemachine
  - disregard asyncstatemachine schedules if machine not yet started
2. add salesagent tests:
  - test multiple subscribe/unsubscribes
  - test scheduling machine without being started
  - test subscriptions are working correctly with external events
  - test errors can be overridden at the state level for ErrorHandlingStates.

### DIFF
--- a/codex/sales/salesagent.nim
+++ b/codex/sales/salesagent.nim
@@ -11,6 +11,7 @@ import ./availability
 type SalesAgent* = ref object of Machine
   context*: SalesContext
   data*: SalesData
+  subscribed: bool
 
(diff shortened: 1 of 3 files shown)


markspanbroek (approved):
Nice, thanks @emizzle!

PR #353, opened 2023-03-06
Enable stylecheck

benbierens:


codex/blocktype.nim
```diff
@@ -35,7 +35,7 @@ type
     cid*: Cid
     data*: seq[byte]
 
-template EmptyCid*: untyped =
+template emptyCid*: untyped =
```
dryajov, line 38: Hmm, this is weird, but we're kinda faking types here, so it would be convenient to leave it capitalized. Any way we could ignore some symbols like this?

codex/manifest/coders.nim
```diff
@@ -158,8 +158,8 @@ func decode*(_: DagPBCoder, data: openArray[byte]): ?!Manifest =
       protected: pbErasureInfo.buffer.len > 0)
 
   if self.protected:
-    self.K = K.int
-    self.M = M.int
+    self.ecK = ecK.int
+    self.ecM= ecM.int
```
dryajov, line 162: missing space

dryajov (changes requested):
Not a big deal obviously, but it would be convenient if we could continue using the capitalized versions of `EmptyCid` and `EmptyDigest`.

codex/blocktype.nim
benbierens: I agree. Will apply Tanguy's suggestion.

codex/manifest/coders.nim
benbierens: I wonder why this wasn't caught by the style check. It would be very nice if we could convince it to...

benbierens:
I've switched to `usages` but I do not trust it. I intentionally messed up some casings and alignment in our own code and it does not detect it.
image
`styleCheck=error` does detect the incorrect casings, but is still quite happy with my disgusting whitespace use. I don't really want to use error, because of the `emptyCid` situation mentioned above.
So, what's the next move here?

benbierens:
Ah yes! In the config.nims I put:
```
--styleCheck:usages
--styleCheck:error
```
And that does indeed work... with one small problem: It's still finding style errors in vendor modules. Namely, these: nim-blscurve, nim-datastore, nim-ethers, nim-leopard, nim-taskpools.

Menduist:
blscurve needs a bump, the others are not used by nimbus/waku AFAIK, so probably need fixing

markspanbroek (dismissed):

dryajov (approved):
LGTM!

PR #356, opened 2023-03-09
running `make` should build exec

dryajov:


emizzle:
I think we might need to update codex-infra as well: https://consul.infra.status.im/ui/he-eu-hel1/services/build-nim-codex-01-timer/instances/metal-01.he-eu-hel1.codex.test/build-nim-codex-01-timer/health-checks

PR #360, opened 2023-03-12
Adds linux/arm/v7

benbierens:


benbierens:
I'm trying to set up a little upload/download long-running test on my raspberry pi with this docker image. The pi needs linux/arm/v7.

c-blake (approved):

cskiraly:
Did you manage to test it? I seem to remember i have ported Leopard to ARM64 only, not 32-bit. I think I've only tested on RPis over 64-bit OSes.

PR #366, opened 2023-03-14
Memory store and lru cache

benbierens:


codex/conf.nim
```diff
@@ -188,7 +188,7 @@ type
         name: "block-mn" }: int
 
       cacheSize* {.
-        desc: "The size in MiB of the block cache, 0 disables the cache - might help on slow hardrives"
+        desc: "Size in MiB of the in-RAM block cache, 0 disables the cache - might help on slow hardrives"
```
c-blake, line 191: Even though it seems NBS has not quite moved to it yet, since nim-1.6.12 dropped 3 days ago with it contained (and it is also in `stew/shims/parseutils` for older Nim), we could/should soon start using `std/parseutils.parseSize`.

The documentation for things like this is a little tricky.  It's probably best to be explicit with a suggestive example like `e.g. "512 MiB, 4G, etc."`.

codex/stores/memorystore.nim
```diff
@@ -0,0 +1,186 @@
+## Nim-Codex
```
c-blake, line 1: If there is a compatibility problem with just using `vendor/lrucache` then it should be clearly documented here (or at the use site).  OTOH, if there is no problem, just using it directly may be better.

c-blake: Ok.  Maybe mentioning "Replacement policy may be other than LRU" in a comment is enough, if that is the crux.  Just something for casual readers later.  Just a thought.

dryajov: > could also be used as a third option from the CLI

Is this really a practical use case? Not sure if I see one here.

codex/stores/memorystore.nim
```diff
@@ -0,0 +1,182 @@
...
+
+logScope:
+  topics = "codex memorystore"
+
+type
+  MemoryStoreNode = ref object
+    key: Cid
+    val: Block
+
+  MemoryStore* = ref object of BlockStore
```
dryajov, line 39: In general, I think the `CacheStore` already did this, what is the reason to re-implement it?

dryajov: Furthermore, I think if we're going for something more generic, I would implement this as a datastore and use it with the tiered datastore abstraction to enable caching at any layer.

codex/stores/blockstoremanager.nim
```diff
@@ -0,0 +1,84 @@
...
+  BlockMaintainer.new(
+    repoStore,
+    interval = config.blockMaintenanceIntervalSeconds.seconds,
+    numberOfBlocksPerInterval = config.blockMaintenanceNumberOfBlocks)
+
+proc getBlockStore(repoStore: RepoStore, config: BlockStoreManagerConfig): BlockStore =
+  if config.cacheSize > 0:
+    return CacheStore.new(backingStore = repoStore, cacheSize = config.cacheSize * MiB)
+  return repoStore
+
```
dryajov, line 74: IMO, this create methods and the constructor, isn't doing any good at all. For starters we're moving away from injecting dependencies during construction to handling construction in the module itself, which is never gonna make things easier/better in the long run and as in our previous convo about mocks, will probably lead to more mocks to overcome this limitation. Perhaps, just leaving construction out of the manager is a sounder approach?

codex/stores/blockstoremanager.nim
```diff
@@ -0,0 +1,84 @@
...
+proc new*(T: type BlockStoreManager, config: BlockStoreManagerConfig): T =
```
dryajov, line 75: I'm not sure if the manager has a place in the codebase given how things are right now. In general, the idea of having a consolidated store manager have popped up a few times, but every time we try it ends up bringing more complexity than necessary.

dryajov (changes requested):
Overall, I'm not 100% convinced with this approach, let me summarize what I see so far:

- The MemoryStore, duplicates most of the functionality of the CacheStore but with a home bread LRU (or am I missing something?), which I'm not sure makes any sense?
- The Manager moves object construction into the module itself, which departs from dependency injection and increases testing complexity

I think the better approach might be:

- Move the MemoryStore to a datastore abstraction and use it in conjunction with a TieredStore abstraction
- Rething the Manager approach - I would like to see a more consolidated way of handling stores, but so far it doesn't seem to be adding anything

Offcourse, the even simpler approach is to just plug back the `CacheStore` and be done with it. Any reason not to do this?

markspanbroek (approved):

codex/codex.nim
```diff
@@ -188,29 +193,27 @@ proc new*(T: type CodexServer, config: CodexConf, privateKey: CodexPrivateKey):
...
 
-    repoStore = RepoStore.new(
-      repoDs = repoData,
-      metaDs = SQLiteDatastore.new(config.dataDir / CodexMetaNamespace)
-        .expect("Should create meta data store!"),
-      quotaMaxBytes = config.storageQuota.uint,
-      blockTtl = config.blockTtlSeconds.seconds)
+    blockMaintainer = BlockMaintainer.new(
+        repoStore,
+        interval = config.blockMaintenanceIntervalSeconds.seconds,
```
dryajov, line 207: is this shorter than 80 columns 🫣

PR #380, opened 2023-03-22
[rest api] fix purchases documentation

emizzle:
Documentation in the REST OpenAPI spec were showing incorrect information about Purchasing endpoints.

Update the open api spec to reflect information about purchases, not requests.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -91,18 +91,6 @@ components:
           type: string
           description: Signed Peer Record to advertise DHT connection information
 
-    Purchase:
-      type: object
-        properties:
-          state:
-            type: string
-            description: Description of the Request's state
-          error:
-            type: string
-            description: If Request failed, then here is presented the error message
-          request:
-            $ref: "#/components/schemas/StorageRequest"
-
     SalesAvailability:
       type: object
       required:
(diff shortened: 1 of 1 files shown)


openapi.yaml
```diff
@@ -91,6 +91,18 @@ components:
...
+        properties:
+          state:
+            type: string
+            description: Description of the Request's state
+          error:
+            type: string
+            description: If Request failed, then here is presented the error message
+          request:
+            $ref: "#/components/schemas/StorageRequest"
+
```
AuHau, line 105: I think this is duplicate as you have renamed `StorageRequestState` into `Purchase` [here](https://github.com/status-im/nim-codex/pull/380/files#diff-d910ba2ef878f7db0223a966b81c8b3f3b65027bb39e4431bb05140171eece39R180)?

AuHau (commented):
Thanks a lot! 🙏 Generally LGTM, but I think there is one component duplicated...

markspanbroek (dismissed):

openapi.yaml
emizzle: Good catch! Duplicate removed.

AuHau (approved):
LGTM, thanks!

PR #387, opened 2023-04-03
Validator

markspanbroek:
Introduces a very simple validator that monitors slots on the blockchain, and invokes `markProofAsMissing()` on the marketplace contract when a host failed to provide a storage proof.

Whether or not a host really missed a proof (and should therefore be punished) is judged by the smart contract, the validator only ensures a timely trigger of the smart contract logic.

Adds a `--validator` option to the codex node, which starts the validator process.
Adds a `--validator-max-slots` option, that indicates how many slots the validator will monitor at most.

~~Depends on changes in https://github.com/status-im/nim-ethers/pull/40, it should be merged first.~~


codex/codex.nim
```diff
@@ -123,17 +124,25 @@ proc new(_: type Contracts,
...
-                                  repo,
-                                  marketplaceAddress)
-
-  (client.option, host.option)
+  let provider = JsonRpcProvider.new(config.ethProvider)
+  let signer = provider.getSigner(account)
+  let marketplace = Marketplace.new(marketplaceAddress, signer)
+  let market = OnChainMarket.new(marketplace)
+  let clock = OnChainClock.new(provider)
+  let purchasing = Purchasing.new(market, clock)
```
emizzle, line 132: In the future, when we may not necessarily want to run a client, the `purchasing` module will not need to be instantiated. The same is true for the `proving` module if we're in client-only mode. This is one of the reasons I kept the construction of these modules inside of `ClientInteractions.new` and `HostInteractions.new`.

codex/codex.nim
```diff
@@ -123,17 +124,25 @@ proc new(_: type Contracts,
...
+  let proofs = OnChainProofs.new(marketplace)
+  let proving = Proving.new(proofs, clock)
+  let sales = Sales.new(market, clock, proving, repo)
+
+  let client = some ClientInteractions.new(clock, purchasing)
+  let host = some HostInteractions.new(clock, sales, proving)
```
emizzle, line 138: In the future, when we may not necessarily want to run a client, the `purchasing` module will not need to be instantiated. The same is true for the `proving` module if we're in client-only mode. This is one of the reasons I kept the construction of these modules inside of `ClientInteractions.new` and `HostInteractions.new`.

codex/contracts/market.nim
```diff
@@ -109,16 +112,31 @@ method subscribeRequests(market: OnChainMarket,
...
+    callback(event.requestId, event.slotIndex)
   let subscription = await market.contract.subscribe(SlotFilled, onEvent)
   return OnChainMarketSubscription(eventSubscription: subscription)
 
+method subscribeSlotFilled*(market: OnChainMarket,
+                            requestId: RequestId,
+                            slotIndex: UInt256,
+                            callback: OnSlotFilled):
+                           Future[MarketSubscription] {.async.} =
+  proc onSlotFilled(eventRequestId: RequestId, eventSlotIndex: UInt256) =
```
emizzle, line 127: Why not use the same callback parameters `event: SlotFilled`?

codex/node.nim
```diff
@@ -387,6 +388,13 @@ proc start*(node: CodexNodeRef) {.async.} =
       error "Unable to start client contract interactions: ", error=error.msg
       node.contracts.client = ClientInteractions.none
 
+  if validatorContracts =? node.contracts.validator:
+    try:
+      await validatorContracts.start()
+    except CatchableError as error:
```
emizzle, line 394: Are there specific error types that we could catch that may give better error info in the log and provide a better debugging story?

codex/validation.nim
```diff
@@ -0,0 +1,57 @@
...
+      for slotId in validation.slots:
+        let state = await validation.market.slotState(slotId)
+        if state != SlotState.Filled:
+          ended.incl(slotId)
+      validation.slots.excl(ended)
+      await sleepAsync(1.seconds) # TODO: wait for next period
+  except CancelledError:
+    discard
+  except CatchableError as e:
+    error "Validation failed", msg = e.msg
```
emizzle, line 46: Are there specific error types that we could catch that may give better error info in the log and provide a better debugging story?

As an example,
```nim
try:
  await validatorContracts.start()
except JsonRpcProviderError as e:
  error "RPC provider error, validation failed", msg = e.msg
except CatchableError as e: # catchall
  error "Validation failed", msg = e.msg
```

codex/codex.nim
markspanbroek: Thanks Eric, I addressed this in later commits.

codex/contracts/market.nim
markspanbroek: Because that is something from the nim-ethers library, and the Market abstraction is meant to be agnostic to its backend.

codex/validation.nim
markspanbroek: I don't think that that's a good idea; JsonRpcProviderError is something very deep inside the nim-ethers library, and I don't want the validator to know about it. That's why we have the Market abstraction.

codex/node.nim
markspanbroek: See https://github.com/status-im/nim-codex/pull/387#discussion_r1165421772

tests/codex/testvalidation.nim
```diff
@@ -0,0 +1,63 @@
...
+  test "the list of slots that it's monitoring is empty initially":
+    check validation.slots.len == 0
+
+  test "when a slot is filled on chain, it is added to the list":
+    await market.fillSlot(slot.request.id, slot.slotIndex, @[])
+    check validation.slots == @[slot.id]
+
+  test "when slot state changes, it is removed from the list":
+    await market.fillSlot(slot.request.id, slot.slotIndex, @[])
+    market.slotState[slot.id] = SlotState.Finished
```
AuHau, line 41: Can you please also cover `SlotState.Failed`?

codex/validation.nim
```diff
@@ -0,0 +1,98 @@
...
+        validation.slots[slotId] = ProofRequirements()
+  let subscription = await validation.market.subscribeSlotFilled(onSlotFilled)
+  validation.subscriptions.add(subscription)
+
+proc removeSlotsThatHaveEnded(validation: Validation) {.async.} =
+  var ended: HashSet[SlotId]
+  for slotId in validation.slots.keys:
+    let state = await validation.market.slotState(slotId)
+    if state != SlotState.Filled:
+      ended.incl(slotId)
```
AuHau, line 57: Honestly, not a big fan of this approach 😅. It seems rather wasteful on the number of RPC calls. Instead, I suggest subscribing to events and using that to remove the slots. However, that would require adding the ﻿`SlotPaid` event into the contract as that case is not covered in the ﻿`freeSlot()` method. Or use the `SlotFreed` event in that case as well.

codex/validation.nim
AuHau, line 1: Please include some info/notice/debug logging. Like we are waiting for period, slots are being added/removed etc. IMHO we should step up the logging as that will be crucial especially when we enter the multi-node (distributed) testing, where it will be harder to understand the big picture of the events happening across the cluster.

AuHau (commented):
Did a first pass of the PR. Generally looks good, but IMHO, this seems to me like a first iteration and sort of naive implementation. 

I still need some design rationales. Like what is the motivation to run a validator? As I see it now, it just burns node's funds on gas fees, but there is no reward/compensation. I know there is a repair reward planned, but now I don't recall any reward for validators.

Another big-picture question for me is how to ensure that the "slot-space" is covered? Or that every slot is monitored at least by one validator? Currently, the first-come-first-served method IMHO might very easily miss some of the slots.

AuHau:
And I like the refactor of the interactions! Very nice simplification.

codex/validation.nim
emizzle: Ok, `JsonRpcProviderError` is not the right type then, but we may be able to scratch out some better error info with an error type indicating there was an error coming back from the RPC interface. Perhaps `ProviderError` or `EthersError`?

markspanbroek: `ProviderError` and `EthersError` are also nim-ethers specific errors, which would tie the validator to a specific implementation of Market. If we really want to have a more specific log message, and we don't want to break the Market abstraction, then we'd have to wrap every method in Market into a `try:except` that catches the underlying error and wraps it in a `MarketError`. But I'm not sure how useful that would be.

emizzle: > But I'm not sure how useful that would be.

It would allow us to recover from certain errors that are maybe not necessarily errors. For example, assuming we were to wrap just the market-specific call in a try/except, we could catch reverts without completely failing the validation run. I'm seeing a lot of "Validation timed out" reverts, which wouldn't necessarily mean we want to stop validating, but instead it would be a good error to log for debug purposes.

codex/codex.nim
```diff
@@ -123,17 +127,26 @@ proc new(_: type Contracts,
...
-  let client = ClientInteractions.new(config.ethProvider,
-                                      account,
-                                      marketplaceAddress)
-  let host = HostInteractions.new(config.ethProvider,
-                                  account,
-                                  repo,
-                                  marketplaceAddress)
-
-  (client.option, host.option)
+  let provider = JsonRpcProvider.new(config.ethProvider)
```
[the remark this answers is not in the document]
markspanbroek: Not immediately, because `new` is a synchronous proc. But quite soon afterwards, because any call to the provider will fail.

codex/contracts/market.nim
```diff
@@ -34,6 +34,15 @@ method approveFunds*(market: OnChainMarket, amount: UInt256) {.async.} =
 method getSigner*(market: OnChainMarket): Future[Address] {.async.} =
   return await market.signer.getAddress()
 
+method periodicity*(market: OnChainMarket): Future[Periodicity] {.async.} =
+  let config = await market.contract.config()
```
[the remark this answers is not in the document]
markspanbroek: Yes, that is a good optimisation.

codex/validation.nim
markspanbroek: Good point, it is rather wasteful. The events are tricky though. We'd not only have to subscribe to the events related to the slot, but also the events for the request. As soon as a request has failed, for instance, we'd like to stop monitoring its slots. We don't want to wait until a host has called `freeSlot()`, because it might choose not to.
I opted to accept the extra RPC calls for now, in exchange for simpler code that we can more easily adapt to changes in the design. But there is a lot of room for optimisation, that's for sure :D

codex/validation.nim
markspanbroek: Good point. I always find it very hard to choose what to log, because you just don't know what you're going to need. But waiting for the next period, adding/removing slots seems like a good start. I'll add those.

codex/validation.nim
emizzle: Ignore this for now, it's more distracting than it needs to be for the bigger picture :)

codex/validation.nim
AuHau: Hmm you are right about listening to the failed requests, but IMHO that is like extra 30 lines of code, don't really see it that problematic. 

Just want to raise the issue of the number of RPC calls, while it should not be something we should super optimize as there are more urgent things to solve, it would be IMHO good idea to keep it in mind and write "low-number RPC-call code" if it is possible. The reason why I raise this issue is that some providers offer free tiers where you don't have to pay for access until X amount of RPC calls. It would be good for easier adoption not to cross that line if we want to make it easy for small operators to join the network.

codex/validation.nim
AuHau: From my early days, I still remember my "ex-team lead" telling me to log every function entry and exit. While that is I guess, the most fool-proof way, I find that bit over the top. But basically, imagine that somebody else runs your code and runs into a bug. You don't have access to their machine and they can only send you logs. When I am thinking about what to log, I log things which then helps me to reconstruct what steps the program did before failing.
But generally, log more than less ;-) If it feels excessive, then use `debug`/`trace` levels which are directly meant for that...

AuHau:
> Good observation, it is a first iteration and very naive. But that was on purpose. We need a validator now to do end-to-end testing of the marketplace, and we know that there are most likely going to be big changes in the role of the validator. So I went for something very simple that we could easily change when the proof requirements need to change, or we add proof aggregation, or...

Got it, then LGTM 👍

>There is a reward for validators planned. It is paid out of the slashed collateral.

Alright, that makes sense.

> That is a good question. How do you coordinate a network of untrusted validators? The incentives are there to guide the validators into the correct behaviour. Because only the first validator to mark a proof as missing gets rewarded, and the others lose gas fees, the incentives should guide the validators to non-covered slot space. But having the incentives right doesn't give us an implementation yet. Something to be explored.

Yup, I agree. It is OK for user testing, but then we will have to develop something more feature-proof.

codex/validation.nim
markspanbroek: Allright, added logging. I didn't want to go overboard with logging, because too many log statements can hurt readability of the code. But I think I struck a reasonable balance.

codex/validation.nim
markspanbroek: Agreed, it is good to be aware of the number of RPC calls. I don't want to spend the time to rewrite the code to listen to events in this PR, because there's higher priority work. But definitely something to tackle in a follow-up PR.

PR #388, opened 2023-04-04
[marketplace] check randomly assigned slot index is not already filled

### LINKED ISSUE #371: [marketplace] ensure random slot index is not already taken
When storage is requested, and that event is handled on a host, we assign a random slot index and kick off the sales state machine process. We may need to perform a check to ensure that the random slot is not already taken by another host.

In addition, later in the state machine process after we've downloaded the data and generated a proof, we attempt to fill the slot by calling `fillSlot` with the proof and the slot index randomly assigned. Perhaps at that point, that particular slot index may already be filled (`fillSlot` would return a `"Slot is not free"` error), at which point we could attempt to reassign a random slot index (check that it's not already in use), and call `fillSlot` again. Repeat until all possible slots have been tried or the slot filling was successful.

emizzle:
Closes #371 

When a storage request is handled by the sales module, a slot index was randomly assigned and then the slot was filled.

Now, the random slot index is checked that its state is `SlotState.Free` before continuing with the download process.

An additional sales agent state was added, `SalePreparing`, that handles this step of assigning a random slot index. All state machine setup, such as retrieving the request and subscribing to events that were previously in the `SaleDownloading` state have been moved to `SalePreparing`. If all indices of the request have been filled, the state is changed to `SaleIgnored`.

### DIFF
--- a/codex/rng.nim
+++ b/codex/rng.nim
@@ -15,7 +15,7 @@ import pkg/libp2p/crypto/crypto
 import pkg/bearssl/rand
 
 type
-  RngSampleError = object of CatchableError
+  RngSampleError* = object of CatchableError
   Rng* = ref HmacDrbgContext
 
 var rng {.threadvar.}: Rng
--- a/codex/sales.nim
+++ b/codex/sales.nim
@@ -4,7 +4,6 @@ import pkg/upraises
 import pkg/stint
 import pkg/chronicles
 import pkg/datastore
-import ./rng
 import ./market
 import ./clock
 import ./proving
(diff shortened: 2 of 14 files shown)


markspanbroek:
I'm not entirely sure that we need this. A host will try to fill a slot only when a new request is seen on the blockchain. At that point in time no slots will be filled, so checking it doesn't seem useful?

emizzle:
If a host randomly selects the same slot index as another host, the slowest host will eventually lose out on filling the slot. The host will then need to go back to the beginning of the state machine and start again with a random slot, at which point there may be many filled slot indices. The restart work has been done in https://github.com/codex-storage/nim-codex/pull/413, which is based on this PR.

emizzle:
Closing in favour of #455.

PR #393, opened 2023-04-13
[marketplace] Simulate invalid proof submissions

### LINKED ISSUE #392: Add configuration flag to simulate proof failures once every N proofs

emizzle:
Closes #392.

Create a simulated prover that sends an invalid proof (seq[byte] of length 0) every `N` proofs, where `N` is defined by a new cli param `--simulate-failed-proofs`.


dryajov:
Can we put this functionality behind a compile flag as well?

emizzle:
> Can we put this functionality behind a compile flag as well?

I like that idea better than a cli flag, as it's really not meant for end user use, but for testing. Do we absolutely need the cli flag @markspanbroek?

emizzle:
Actually, i was thinking about this further, and using only compile flags may make testing much slower as we'd have to compile a binary for each variation we'd be testing. @dryajov, I'm curious what the advantages of compile flag would be for testing?

AuHau:
Yeah, I like the compile flag idea as well. 

> Actually, i was thinking about this further, and using only compile flags may make testing much slower as we'd have to compile a binary for each variation we'd be testing. @dryajov, I'm curious what the advantages of compile flag would be for testing?

Maybe @dryajov meant compile flag that adds the `--simulate-failed-proofs` CLI flag? In this way, we have to compile only one binary and then use normal CLI flags for the specification of the proofs failure rate

emizzle:
> Maybe @dryajov meant compile flag that adds the `--simulate-failed-proofs` CLI flag? In this way, we have to compile only one binary and then use normal CLI flags for the specification of the proofs failure rate

Ah yes, good idea, could very well be what he meant.

markspanbroek:
> compile flag that adds the `--simulate-failed-proofs` CLI flag? In this way, we have to compile only one binary and then use normal CLI flags for the specification of the proofs failure rate

Yes, I'd go with this.

dryajov:
> Yeah, I like the compile flag idea as well.
> 
> > Actually, i was thinking about this further, and using only compile flags may make testing much slower as we'd have to compile a binary for each variation we'd be testing. @dryajov, I'm curious what the advantages of compile flag would be for testing?
> 
> Maybe @dryajov meant compile flag that adds the `--simulate-failed-proofs` CLI flag? In this way, we have to compile only one binary and then use normal CLI flags for the specification of the proofs failure rate

Yep, thats what I mean 👍

markspanbroek (approved):
Nice, thanks @emizzle!

tests/integration/multinodes.nim
```diff
@@ -0,0 +1,123 @@
...
+import pkg/chronicles
+import ../ethertest
+import ./codexclient
+import ./nodes
+
+export ethertest
+export codexclient
+export nodes
+
+template multinodesuite*(name: string,
```
AuHau, line 14: Nice 👍

tests/integration/testIntegration.nim
```diff
@@ -87,7 +87,7 @@ twonodessuite "Integration tests", debug1 = false, debug2 = false:
     check newSize > 0 and newSize < size.u256
 
   test "node slots gets paid out":
-    let marketplace = Marketplace.new(Marketplace.address, provider.getSigner())
+    var marketplace = Marketplace.new(Marketplace.address, provider.getSigner())
```
AuHau, line 90: I guess you don't need this?

codex/proving/simulated.nim
```diff
@@ -0,0 +1,54 @@
...
+            market: Market,
+            clock: Clock,
+            failEveryNProofs: uint): SimulatedProving =
+
+    let p = SimulatedProving.new(market, clock)
+    p.failEveryNProofs = failEveryNProofs
+    return p
+
+  method init(proving: SimulatedProving) {.async.} =
+    if proving.failEveryNProofs > 0'u and await proving.market.isMainnet():
```
AuHau, line 30: I am not sure about this. IMHO I would remove the `isMainnet` check. First of all, we don't know if we gonna even run on Mainnet (most probably not). I guess you are trying to be on the safe side in case somebody finds binary compiled with the simulated-proving enabled and tried to run it on Mainnet, right? IMHO it is a low probability as they would also have to specify the config parameter to enable this and if they would really want to run this on mainnet then they can always compile their own version to work around this restriction...

AuHau (commented):
Generally, LGTM! Especially love the testing work here!

I am just not so sure if it is a good idea to have the `isMainnet` check.

codex/conf.nim
```diff
@@ -237,6 +240,13 @@ type
         name: "validator-max-slots"
       .}: int
 
+      simulateProofFailures* {.
+          desc: "Simulates proof failures once every N proofs. 0 = disabled."
+          defaultValue: 0
+          name: "simulate-proof-failures"
+          hidden
+        .}: uint
+
```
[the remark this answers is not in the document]
emizzle: I tried, but it would not cooperate 😅  The best I could do was add the `hidden` option to the compile flag, which hides from it from cli help.

codex/proving/simulated.nim
emizzle: Yes, this is mostly an artifact leftover from before adding the compilation flag. Will remove.

AuHau (approved):
LGTM, thanks! Great work 👌

PR #413, opened 2023-05-09
[marketplace] restart sales state machine and memory leak prevention

emizzle:
### restart sales state machine if slot is filled by other host
When a slot is filled by other host, the state machine should not park in the `SaleErrored` state, and instead moves to the `SaleRestart` state, where it calls `onSaleRestart`. The sales module handles `onSaleRestart`, restarting the state machine, getting a new slot index while ignoring the previously attempting slot index.

### [statemachine] prevent memory leak by cleaning up ended agents
Once agents have parked in an end state (eg ignored, cancelled, finished, failed, errored), they were not getting cleaned up and the sales module was keeping a handle on their reference. This could prvent a memory leak if the number of requests coming in is high.

Add an `onCleanUp` that stops the agents and removes them from the sales module agent list. `onCleanUp` is called from sales end states (eg ignored, cancelled, finished, failed, errored).

### DIFF
--- a/codex/contracts/requests.nim
+++ b/codex/contracts/requests.nim
@@ -3,6 +3,8 @@ import pkg/contractabi
 import pkg/nimcrypto
 import pkg/ethers/fields
 import pkg/questionable/results
+import pkg/json_serialization
+import pkg/upraises
 
 export contractabi
 
@@ -193,3 +195,17 @@ func price*(request: StorageRequest): UInt256 =
 
 func size*(ask: StorageAsk): UInt256 =
   ask.slots.u256 * ask.slotSize
+
+proc writeValue*(
+  writer: var JsonWriter,
+  value: SlotId | RequestId) {.upraises:[IOError].} =
+
+  mixin writeValue
+  writer.writeValue value.toArray
+
+proc readValue*[T: SlotId | RequestId](
+  reader: var JsonReader,
+  value: var T) {.upraises: [SerializationError, IOError].} =
+
+  mixin readValue
+  value = T reader.readValue(T.distinctBase)
--- a/codex/sales.nim
+++ b/codex/sales.nim
@@ -74,7 +74,7 @@ func new*(_: type Sales,
   ))
 
 proc remove(sales: Sales, agent: SalesAgent): OnCleanUp =
-  proc: Future[void] {.gcsafe, upraises:[], async.} =
+  proc(): Future[void] {.gcsafe, upraises:[], async.} =
     await agent.stop()
     sales.agents.keepItIf(it != agent)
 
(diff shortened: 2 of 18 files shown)


markspanbroek:
I don't think that it makes sense for a host to go back and try again with a different slot for the same request. The chances of the host filling the slot are very slim, because it already spent a lot of time filling another slot, therefore it will be at a disadvantage to other hosts that started earlier. It would make more sense for the host to bail on the request and start filling a slot on a newly submitted request.

emizzle:
In a busy network with many hosts and a wide range of availabilities, you are probably right. But what happens when the request parameters are tight and there are not a lot of matching hosts? It would be a shame if a request doesn’t start successfully if there were hosts in the network willing to participate but they collided on their assigned slot index. The probability of this happening is higher the smaller the number of slots in the request are. If other parameters are restrictive, eg total payout, there may not be many willing participants.

Restarting the process doesn’t present a downside to the host apart from a few network requests checking slot state, as they were already going to provide storage in the first place.

markspanbroek:
If the request parameters are tight, and there are not a lot of matching hosts, then the request wasn't very attractive. The client should try again with a higher payout for hosts, to make it more attractive :smile:. This is how prices are determined in the market.

> Restarting the process doesn’t present a downside to the host apart from a few network requests checking slot state, as they were already going to provide storage in the first place.

I know that we haven't implemented this yet, but the idea is for hosts to only download the data for the slot that they are trying to fill, not for the entire dataset that a request entails. So you wouldn't be able to fill a different slot for the same request without the need to download additional data, and recalculating a proof.

emizzle:
> If the request parameters are tight, and there are not a lot of matching hosts, then the request wasn't very attractive. The client should try again with a higher payout for hosts, to make it more attractive 😄. This is how prices are determined in the market.

Yes, but there WERE matching hosts at that price point that were willing to participate given the storage requirements, fulfilling the market demand. But simply because the client got assigned the same slot index as another, faster host, shouldn't mean that we stand in the way of allowing them to fill another slot *if it is free*. If there are hosts willing to provide storage, shouldn't we try to get the storage request filled?

An illustration of this can be seen in the distributed network tests with only two hosts and a request specifying two slots. The chance of this happening is 50%. That can be extrapolated to conditions with tight storage requirements.

> I know that we haven't implemented this yet, but the idea is for hosts to only download the data for the slot that they are trying to fill, not for the entire dataset that a request entails. So you wouldn't be able to fill a different slot for the same request without the need to download additional data, and recalculating a proof.

Fair enough, the client will also need to re-download the slot-specific data, however before doing so, it will have checked that the slot is indeed free, so the chances of that slot already being filled again are much smaller, due to the time lapsed between the first and second attempt. It could also happen that the host attempts to retry, but all the slots are filled, so the only downside are the network requests checking the slot state.

emizzle:
Closing in favour of #455

Issue #415, opened 2023-05-10
Expose information about hosted requests


AuHau:
Currently, there is no endpoint that exposes information about what hosts store, for how long, how much money they will get for it, the state, etc.

Issue #414, opened 2023-05-10
Merge proving and sales state machines


AuHau:
The storage host has two parts: a state machine for setting up the Request (e.g. downloading the info, proving, filling the slot, etc.) which completes with a filled slot, and then the Slot/Request is passed to the proving module. This module handles proving, collecting the reward, and cleaning up. These two parts should be merged.

PR #416, opened 2023-05-10
feat: storage host collects reward

AuHau:
This PR adds hotfix-style support for collecting a host's reward. It is plugged into the proving component, but it will be refactored later, as described in #414. 

The test case is also temporary, as we currently do not have any endpoint that would expose the state of a host's requests that he or she hosts. This will be improved once #415 is fixed. 

As I had a hard time making the test pass, I have improved the logging support for the Sales State Machine, integration tests, and some documentation around this. Only then I have realized that the proving component requires new blocks to be mined in order for the proving cycle to kick in, so I added `provider.advanceTime()` and then all worked 🙄

### DIFF
--- a/tests/integration/twonodes.nim
+++ b/tests/integration/twonodes.nim
@@ -10,14 +10,8 @@ export ethertest
 export codexclient
 export nodes
 
-template twonodessuite*(name: string, debug1, debug2: bool, body) =
+template twonodessuite*(name: string, debug1, debug2: bool | string, body) =
   twonodessuite(name, $debug1, $debug2, body)
-
-template twonodessuite*(name: string, debug1: string, debug2: bool, body) =
-  twonodessuite(name, debug1, $debug2, body)
-
-template twonodessuite*(name: string, debug1: bool, debug2: string, body) =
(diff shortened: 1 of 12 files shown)


tests/integration/twonodes.nim
```diff
@@ -11,43 +11,66 @@ export codexclient
 export nodes
 
 template twonodessuite*(name: string, debug1, debug2: bool, body) =
+  twonodessuite(name, $debug1, $debug2, body)
+
+template twonodessuite*(name: string, debug1: string, debug2: bool, body) =
+  twonodessuite(name, debug1, $debug2, body)
+
+template twonodessuite*(name: string, debug1: bool, debug2: string, body) =
+  twonodessuite(name, $debug1, debug2, body)
```
AuHau, line 20: Most probably there is better way how to do this, so please if you have some idea please shoot ;-)

markspanbroek (dismissed):
Nice! Thanks @AuHau!

tests/integration/twonodes.nim
AuHau: Oh, I was not sure if you can do such a thing in Nim, great thanks!

PR #420, opened 2023-05-23
Update links to codex-storage organization

veaceslavdoina:
Some badges may not show status correctly and as we moved repository it is better to update all the links to point to the new organization.


markspanbroek (approved):
Nice!

PR #421, opened 2023-06-01
Docker dht debug

benbierens:


codex/rest/api.nim
```diff
@@ -45,6 +48,47 @@ proc validate(
...
+    "address": formatAddress(node.address),
+    "seen": $node.seen
+  }
+  return jobj
+
+proc formatTable(routingTable: rt.RoutingTable): JsonNode =
+  let jarray = newJArray()
+  for bucket in routingTable.buckets:
+    for node in bucket.nodes:
+      jarray.add(formatNode(node))
```
emizzle, line 70: If I'm not mistaken, in a fully populated p2p network, this could potentially become a very expensive operation, as the last k-bucket could contain half the network (known to the client).

I understand this is being called from the `debug/info` endpoint, but given that this is a single-threaded client and these operations are not async, this could lock up the node and be a DDOS vector.

Perhaps a solution would be to hide this endpoint behind a compile flag so that it doesn't inadvertently make its way into production (similar was done in https://github.com/codex-storage/nim-codex/pull/393/commits/331884a85d946b7f99bdeac64f85eb25310f56df).

docker/startCodex.sh
```diff
@@ -1,7 +1,15 @@
...
+NAME=""
+if [ -n "$CODEX_NODENAME" ]; then
+  NAME=" '$CODEX_NODENAME'"
+fi
+echo "Starting Codex node$NAME"
 
 args=""
 
+## local ip as NAT?
+do_nat=$(hostname -i)
```
emizzle, line 10: I don't think this is cross-platform compatible. On my mac,
```shell
❯ which hostname
/bin/hostname

❯ hostname -i
hostname: illegal option -- i
usage: hostname [-f] [-s | -d] [name-of-host]

❯ hostname
Es-MacBook-Pro.local
```

emizzle (commented):
Nice work! It might be a good idea to split this up into two PRs: one for the docker stuff and the other for the dht stuff, that is based on the docker PR. This way, the docker stuff can get merged independently from the dht stuff, and you can still run your tests off of one branch.

codex/rest/api.nim
benbierens: Good idea. I added the compile flag. :D

docker/startCodex.sh
benbierens: I believe that immediately, but I don't think it's a problem:
This runs in a docker, it's always on ubuntu/lunar.
In the future, this nat-workaround should be replaced by proper nat-behavior inside Codex. (It might not be applicable inside docker or the dist-test environment though... 🤔)

codex/rest/api.nim
emizzle: What do you mean by screwing up execution in a testnet?

docker/startCodex.sh
emizzle: Good point. 👍

codex/rest/api.nim
benbierens: Like you said, it could be an expensive operation in a real p2p network. So, if we want to be able to do something like this for inspecting a future test-net, then maybe we can come up with a way to implement it that doesn't block threads (so much).

emizzle: Hmm, yes good point. One possible solution might be the result set could be limited and paged perhaps.

dryajov:
This PR also seems to do too many things at the same time - from what I can parse, it adds some helper/debug functions + a bunch of docker related changes that don't appear to be related? Can we split this out?

PR #426, opened 2023-06-07
Block exchange debug

jessiebroke:
This draft PR along w/ the other PR linked were created to address issues that started from the "fully-interconnected-download tests" file exchange test outlined here: https://github.com/codex-storage/cs-codex-dist-tests/pull/25. 

Failing because of a problem w/ the block exchange not fetching the manifest block consistently and also then started to fail w/ fetching blocks.

This draft PR resolves the above test and block exchange download issues.

### DIFF
--- a/codex/blockexchange/engine/discovery.nim
+++ b/codex/blockexchange/engine/discovery.nim
@@ -162,13 +162,9 @@ proc discoveryTaskLoop(b: DiscoveryEngine) {.async.} =
 
           for i, f in dialed:
             if f.failed:
-              trace "dial failed"
               await b.discovery.removeProvider(peers[i].data.peerId)
-            else:
-              trace "dial successful", peerId = peers[i].data.peerId
-
               without cb =? b.blockLocatedCallback:
-                trace "blockLocated callback not set :o"
+                trace "blockLocated callback not set"
 
               await cb(cid, peers[i].data.peerId)
 
--- a/codex/blockexchange/engine/discovery.nim
+++ b/codex/blockexchange/engine/discovery.nim
@@ -41,6 +41,7 @@ const
   DefaultAdvertiseLoopSleep = 30.minutes
 
 type
+  BlockLocatedCallback* = proc(cid: Cid, peer: PeerId): Future[void] {.gcsafe, upraises:[].}
   DiscoveryEngine* = ref object of RootObj
     localStore*: BlockStore                                      # Local block store for this instance
     peers*: PeerCtxStore                                         # Peer context store
@@ -62,6 +63,8 @@ type
     inFlightDiscReqs*: Table[Cid, Future[seq[SignedPeerRecord]]] # Inflight discovery requests
     inFlightAdvReqs*: Table[Cid, Future[void]]                   # Inflight advertise requests
     advertiseType*: BlockType                                    # Advertice blocks, manifests or both
+
+    blockLocatedCallback*: ?BlockLocatedCallback
 
 proc discoveryQueueLoop(b: DiscoveryEngine) {.async.} =
   while b.discEngineRunning:
@@ -152,14 +155,22 @@ proc discoveryTaskLoop(b: DiscoveryEngine) {.async.} =
           let
             peers = await request
 
-          trace "Discovered peers", peers = peers.len
+          trace "Discovered peers for block", peers = peers.len, cid
           let
             dialed = await allFinished(
               peers.mapIt( b.network.dialPeer(it.data) ))
 
           for i, f in dialed:
             if f.failed:
+              trace "dial failed"
               await b.discovery.removeProvider(peers[i].data.peerId)
+            else:
+              trace "dial successful", peerId = peers[i].data.peerId
+
+              without cb =? b.blockLocatedCallback:
+                trace "blockLocated callback not set :o"
+
+              await cb(cid, peers[i].data.peerId)
 
         finally:
           b.inFlightDiscReqs.del(cid)
--- a/codex/blockexchange/engine/discovery.nim
+++ b/codex/blockexchange/engine/discovery.nim
@@ -41,7 +41,6 @@ const
   DefaultAdvertiseLoopSleep = 30.minutes
 
 type
-  BlockLocatedCallback* = proc(cid: Cid, peer: PeerId): Future[void] {.gcsafe, upraises:[].}
   DiscoveryEngine* = ref object of RootObj
     localStore*: BlockStore                                      # Local block store for this instance
     peers*: PeerCtxStore                                         # Peer context store
@@ -63,8 +62,6 @@ type
     inFlightDiscReqs*: Table[Cid, Future[seq[SignedPeerRecord]]] # Inflight discovery requests
     inFlightAdvReqs*: Table[Cid, Future[void]]                   # Inflight advertise requests
     advertiseType*: BlockType                                    # Advertice blocks, manifests or both
-
-    blockLocatedCallback*: ?BlockLocatedCallback
 
 proc discoveryQueueLoop(b: DiscoveryEngine) {.async.} =
   while b.discEngineRunning:
@@ -163,11 +160,6 @@ proc discoveryTaskLoop(b: DiscoveryEngine) {.async.} =
           for i, f in dialed:
             if f.failed:
               await b.discovery.removeProvider(peers[i].data.peerId)
-            else:
-              without cb =? b.blockLocatedCallback:
-                trace "blockLocated callback not set"
-
-              await cb(cid, peers[i].data.peerId)
 
         finally:
           b.inFlightDiscReqs.del(cid)
--- a/codex/blockexchange/engine/discovery.nim
+++ b/codex/blockexchange/engine/discovery.nim
@@ -163,6 +163,7 @@ proc discoveryTaskLoop(b: DiscoveryEngine) {.async.} =
           for i, f in dialed:
             if f.failed:
               await b.discovery.removeProvider(peers[i].data.peerId)
+            else:
               without cb =? b.blockLocatedCallback:
                 trace "blockLocated callback not set"
 
--- a/codex/blockexchange/engine/engine.nim
+++ b/codex/blockexchange/engine/engine.nim
@@ -164,6 +164,10 @@ proc requestBlock*(
 
   trace "Selecting peers who have", cid
   var peers = b.peers.selectCheapest(cid)
+  if peers.len <= 0:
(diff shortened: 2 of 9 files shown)


codex/blockexchange/engine/engine.nim
```diff
@@ -116,39 +116,66 @@ proc stop*(b: BlockExcEngine) {.async.} =
...
+  var peers = b.peers.selectCheapest(cid)
   if peers.len <= 0:
-    trace "No cheapest peers, selecting first in list", cid
-    peers = toSeq(b.peers) # Get any peer
-    if peers.len <= 0:
-      trace "No peers to request blocks from", cid
+    trace "We don't know of anyone who has this. Start discovery immediately."
+    b.discovery.queueFindBlocksReq(@[cid])
+
+  without blockPeer =? b.findCheapestPeerForBlock(peers):
```
dryajov, line 171: Why spawn discovery twice? There is no reason to do it twice I believe.

codex/blockexchange/engine/discovery.nim
```diff
@@ -152,14 +155,22 @@ proc discoveryTaskLoop(b: DiscoveryEngine) {.async.} =
...
             if f.failed:
+              trace "dial failed"
               await b.discovery.removeProvider(peers[i].data.peerId)
+            else:
+              trace "dial successful", peerId = peers[i].data.peerId
+
+              without cb =? b.blockLocatedCallback:
+                trace "blockLocated callback not set :o"
+
+              await cb(cid, peers[i].data.peerId)
```
dryajov, line 173: I'm not sure you need this callback, there are several reasons why:

- Doing anything with the peer only makes sense after the peer has been fully connected, which only happens after the `PeerEventKind.Joined` has been triggered
- The logic in the callback is already triggered in `setupPeer`, which is registered for the `PeerEventKind.Joined` event. For example the local block list for the peer's "Haves" should be updated on peer connect, which happens in `setupPeer` again

In other words, simply dialing the peer should be enough - it either succeeds upgrading and is fully connected, or it fails in which case there isn't much to be done.

codex/stores/networkstore.nim
```diff
@@ -32,17 +32,28 @@ type
...
+      ## Does not occure
+      trace "Unable to get block from exchange engine", cid
+      return failure error
+
+    ## Does not occure
+    trace "Got block from exchange engine. Adding to local store..."
+    let ttl =  Duration.none
+    # TODO: What should the TTL be?!
+    # depends: are we serving a marketplace contract? duration of contract probably.
+    let res = await self.localStore.putBlock(newBlock, ttl)
```
benbierens, line 52: This is also already done in engine.nim proc blocksHandler.

benbierens:
Also, note to me: Remove some of the (outdated) comments and traces before merging.

codex/blockexchange/engine/engine.nim
benbierens: OK, I reran the test `FullyConnectedDownloadTest(3,1)` (cs-codex-dist-tests repository, branch feature/fully-connected-download-tests) with and without these lines:

```
if peers.len <= 0:
  trace "We don't know of anyone who has this. Start discovery immediately."
  b.discovery.queueFindBlocksReq(@[cid])
```

With lines:
`Performed 12 downloads of 1MB in 13.8697523 seconds, for an average of 00:00:01.1558127 seconds per MB.` (Test passed)

Without lines:
`Performed 12 downloads of 1MB in 427.2357426 seconds, for an average of 00:00:35.6029786 seconds per MB.` (Test failed)

I agree the code looks like it's just duplicate. But somehow, there's a case where this call makes the whole thing many times faster. (There's a performance-threshold assert in the tests that cause the tests to fail when download is this slow.)

I don't understand exactly who it causes this difference. But the fact that it's close to 30 seconds, and the discovery-loop sleeps for 30 seconds is very... suspicious... Also I propose I don't put too much time into this mystery, since the design for the future block exchange engine looks different anyway. What do you think?

codex/blockexchange/engine/engine.nim
```diff
@@ -165,42 +192,26 @@ proc requestBlock*(
...
-  let remaining = peers[1..min(peers.high, b.peersPerRequest)]
-  trace "Sending want list to remaining peers", count = remaining.len
-  for p in remaining:
-    if cid notin p.peerHave:
-      # just send wants
-      await b.network.request.sendWantList(
-        p.id,
-        @[cid],
-        wantType = WantType.WantHave) # we only want to know if the peer has the block
+  await b.sendWantHave(cid, blockPeer, peers)
```
dryajov, line 198: Changing the order here makes retrieval less efficient. `WantBlock` initiates block retrieval, while `WantHave` only signals the remotes that we want this block and doesn't trigger transmittion, we're waiting for the `WantHave` message to be sent to all peers before starting download.

codex/blockexchange/engine/engine.nim
```diff
@@ -116,39 +116,66 @@ proc stop*(b: BlockExcEngine) {.async.} =
...
   if b.pendingBlocks.isInFlight(cid):
     trace "Request handle already pending", cid
     return await b.pendingBlocks.getWantHandle(cid, timeout)
 
-  let
-    blk = b.pendingBlocks.getWantHandle(cid, timeout)
-
-  var
-    peers = b.peers.selectCheapest(cid)
+  let blk = b.pendingBlocks.getWantHandle(cid, timeout)
```
dryajov, line 163: Making this changes everywhere just makes the codebase inconsistent. The codebase is already written in a specific style which we better stick to, at least in the code units where this style is prevalent.

codex/blockexchange/engine/engine.nim
```diff
@@ -165,42 +192,26 @@ proc requestBlock*(
...
-  trace "Sending want list to remaining peers", count = remaining.len
-  for p in remaining:
-    if cid notin p.peerHave:
-      # just send wants
-      await b.network.request.sendWantList(
-        p.id,
-        @[cid],
-        wantType = WantType.WantHave) # we only want to know if the peer has the block
+  await b.sendWantHave(cid, blockPeer, peers)
+  await b.sendWantBlock(cid, blockPeer)
```
dryajov, line 200: Also, this was removed above, and moved before the `blockHandleMonitor`, which is probably why, if you remove the second `queueFindBlocksReq` on line 171, it slows down significantly.

codex/blockexchange/engine/engine.nim
```diff
@@ -269,10 +277,6 @@ proc scheduleTasks(b: BlockExcEngine, blocks: seq[bt.Block]) {.async.} =
           break # do next peer
 
 proc resolveBlocks*(b: BlockExcEngine, blocks: seq[bt.Block]) {.async.} =
```
dryajov, line 279: Like I said above, making this slight style changes, just makes the codebase less consistent. Everyone has their own style and opinions about comments, spacing, alignment, etc - but for the sake of consistency, I would avoid making slight styling changes.

For clarity, having a comment briefly explaning what the code bellow does is better than not, we also want to prefix this with a double `##` and reserve `#` for inline comments.

dryajov (changes requested):

codex/blockexchange/engine/engine.nim
benbierens: I agree. I wish we had some kind of linter.

codex/blockexchange/engine/engine.nim
benbierens: That's right. I fixed this one.

codex/blockexchange/engine/engine.nim
benbierens: I don't like comments that add nothing and basically repeat the function name. It's just clutter. I want comments to be rare, so that when they show up they are not likely to be ignored.

codex/blockexchange/engine/engine.nim
benbierens: I moved the `if ... queueFindBlocksReq` back down and reran the tests and indeed that still works fine.

codex/blockexchange/engine/engine.nim
dryajov: > I don't like comments that add nothing and basically repeat the function name.

I like using comments as visual queues that are there to guide you through the codebase. It is the prose version of what the code (usually, immediately bellow) does. I've found it useful when browsing an unfamiliar codebase, or reviewing code I've written a while back, specially when you lack good code exploration tooling - IDEs, etc...

I think the argument about only using comments when strictly necessary fails in practice for various reasons, but primarily because we tend to underestimate how much contextual information we're opperating with whe writting the code. What seems clear and logical when we've just written it becomes undecipherable as soon as this context is lost, usually a few days/weeks after or even worst when just starting with the codebase. For that reason, I tend to favor commenting rather than not.

The argument that "too many comments cause people to ignore them" is easily addressed by providing additional hints to the importance of the comment. Using `TODO`, `XXX`, `WARN`, etc... Usually addresses this quite cleanly.

I do agree that if the comment simply repeats the name of the function it should probably not be there, but that is a smaller annoyance overall. In the case of this particular comment at least, I think it provided some good initial context.

Overall, I'd like us to stick with prefacing the function with some context of what it's doing rather than not, but I'm open to suggestions, we should probably raise this to it's own issue/discussion and see what others think about it.

codex/blockexchange/engine/engine.nim
```diff
@@ -309,24 +301,22 @@ proc blocksHandler*(
     peerCtx = b.peers.get(peer)
 
   if peerCtx != nil:
-    # we don't care about this blocks anymore, lets cleanup the list
     await b.payForBlocks(peerCtx, blocks)
+    ## shouldn't we remove them from the want-list instead of this:
```
dryajov, line 305: This is handled in `resolveBlocks`, not really sure if we want `cleanPresence` is doing there, it doesn't seem like it's needed and it's probably breaking something.

codex/stores/networkstore.nim
```diff
@@ -32,17 +32,17 @@ type
...
 method getBlock*(self: NetworkStore, cid: Cid): Future[?!bt.Block] {.async.} =
-  ## Get a block from a remote peer
-  ##
-
   trace "Getting block from local store or network", cid
 
   without blk =? await self.localStore.getBlock(cid), error:
     if not (error of BlockNotFoundError): return failure error
     trace "Block not in local store", cid
-    # TODO: What if block isn't available in the engine too?
```
dryajov, line 43: Are we handling this case? The first TODO seems important, unless it's already handled elsewhere?

dryajov (approved):
LGTM sans the few remarks.

PR #444, opened 2023-06-16
feat: allow any case for logLevel

AuHau:


markspanbroek (approved):
I like it, but it does deviate from the standard way of specifying log levels with chronicles.

AuHau:
> I like it, but it does deviate from the standard way of specifying log levels with chronicles.

I disagree ;-) It deviates from how `parseEnum` works. If you you look on how [`parseTopicDirectives`](https://github.com/status-im/nim-chronicles/blob/c9c8e58ec3f89b655a046c485f622f9021c68b61/chronicles/helpers.nim#L25) is implemented, then it actually allows lowercase levels, even shortened one like `inf=info`... I have created PR there to allow this behavior on `setLogLevel` level as well: https://github.com/status-im/nim-chronicles/pull/136

dryajov (approved):
It does deviate from Nimbus and other Status Nim software as well, but I like it.

PR #445, opened 2023-06-16
chore: remove unused imports

AuHau:


tests/integration/testproofs.nim
```diff
@@ -3,7 +3,6 @@ import codex/contracts/marketplace
 import codex/contracts/deployment
 import codex/periods
 import ../contracts/time
-import ../contracts/deployment
```
markspanbroek, line 6: Looks like this is one too many, it makes the integration test fail :D

markspanbroek (approved):
:pray: Thanks for cleaning this up! Good to go once the tests pass again.

tests/integration/testproofs.nim
AuHau: Ups 😅

PR #452, opened 2023-06-19
feat: bigint uses decimal over hex encoding

### LINKED ISSUE #435: REST API: BigInt Conversion of Hex-encoded numbers to String Literals/Decimals

AuHau:
Closes #435

### DIFF
--- a/codex/utils/stintutils.nim
+++ b/codex/utils/stintutils.nim
@@ -1,4 +1,4 @@
 import pkg/stint
 
-func fromString*(T: typedesc[StUint|StInt], s: string): T {.inline.} =
+func fromDecimal*(T: typedesc[StUint|StInt], s: string): T {.inline.} =
   parse(s, type result, radix = 10)
--- a/codex/utils/stintutils.nim
+++ b/codex/utils/stintutils.nim
@@ -1,0 +1,4 @@
+import pkg/stint
+
+func fromString*(T: typedesc[StUint|StInt], s: string): T {.inline.} =
+  parse(s, type result, radix = 10)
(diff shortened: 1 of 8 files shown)


codex/utils/stintutils.nim
```diff
@@ -0,0 +1,4 @@
+import pkg/stint
+
+func fromString*(T: typedesc[StUint|StInt], s: string): T {.inline.} =
```
markspanbroek, line 3: I would suggest to call this `fromDecimals()`, to be more explicit. And it matches the `fromHex()` naming convention.

markspanbroek (approved):
Nice :+1:

codex/utils/stintutils.nim
AuHau: Good idea, thanks! 👍

mratsim:
@jangko is revamping stint. Currently on serialization https://github.com/status-im/nim-stint/pull/132, that makes sense to be upstream.

I would name if `fromDecimal` instead of `fromDecimals` though since it's decimal representation.

AuHau:
>@jangko is revamping stint. Currently on serialization https://github.com/status-im/nim-stint/pull/132, that makes sense to be upstream.

Sorry, not sure what you mean here 😅 Do you want to have the `fromDecimal` directly in `stint`?

>I would name if fromDecimal instead of fromDecimals though since it's decimal representation or a decimal number.
 
Yeah that is actually what I used, so all good 👍

jangko:
okay, will add `fromDecimal` to stint

PR #453, opened 2023-06-20
Nimsuggest tweaks

elcritch:
- Fix weird nimsuggest failures with static getCodexVersion and rng.Rng. Now nimsuggest doesn't routinely crash. 
- Some formatting tweaks. It's a style I've seen "in the wild" in a few Nim projects and find that for procs with many variables it helps readability. In particular the return type is much quicker to pick out which helps with `Result` types.

### DIFF
--- a/codex.nimble
+++ b/codex.nimble
@@ -7,29 +7,30 @@ license = "MIT"
 binDir = "build"
 srcDir = "."
 
-requires "nim >= 1.2.0",
-         "asynctest >= 0.3.2 & < 0.4.0",
-         "bearssl >= 0.1.4",
-         "chronicles >= 0.7.2",
-         "chronos >= 2.5.2",
-         "confutils",
-         "ethers >= 0.2.4 & < 0.3.0",
-         "libbacktrace",
-         "libp2p",
-         "metrics",
-         "nimcrypto >= 0.4.1",
-         "nitro >= 0.5.1 & < 0.6.0",
-         "presto",
-         "protobuf_serialization >= 0.2.0 & < 0.3.0",
-         "questionable >= 0.10.6 & < 0.11.0",
-         "secp256k1",
-         "stew",
-         "upraises >= 0.1.0 & < 0.2.0",
-         "https://github.com/status-im/lrucache.nim#1.2.2",
-         "leopard >= 0.1.0 & < 0.2.0",
-         "blscurve",
-         "libp2pdht",
-         "eth"
+requires "nim >= 1.2.0"
+requires "asynctest >= 0.3.2 & < 0.4.0"
+requires "bearssl >= 0.1.4"
+requires "chronicles >= 0.7.2"
+requires "chronos >= 2.5.2"
+requires "confutils"
+requires "ethers >= 0.2.4 & < 0.3.0"
+requires "libbacktrace"
+requires "libp2p"
+requires "metrics"
+requires "nimcrypto >= 0.4.1"
+requires "nitro >= 0.5.1 & < 0.6.0"
+requires "presto"
+requires "protobuf_serialization >= 0.2.0 & < 0.3.0"
+requires "questionable >= 0.10.6 & < 0.11.0"
+requires "secp256k1"
+requires "stew"
+requires "upraises >= 0.1.0 & < 0.2.0"
+requires "toml_serialization"
+requires "https://github.com/status-im/lrucache.nim#1.2.2"
+requires "leopard >= 0.1.0 & < 0.2.0"
+requires "blscurve"
+requires "libp2pdht"
+requires "eth"
 
 when declared(namedBin):
   namedBin = {
--- a/codex/codex.nim
+++ b/codex/codex.nim
@@ -141,14 +141,17 @@ proc stop*(s: CodexServer) {.async.} =
 
   s.runHandle.complete()
 
-proc new*(T: type CodexServer, config: CodexConf, privateKey: CodexPrivateKey): T =
-
+proc new*(
+    T: type CodexServer,
+    config: CodexConf,
+    privateKey: CodexPrivateKey): CodexServer =
+  ## create CodexServer including setting up datastore, repostore, etc
   let
     switch = SwitchBuilder
     .new()
     .withPrivateKey(privateKey)
     .withAddresses(config.listenAddrs)
-    .withRng(Rng.instance())
+    .withRng(rng.Rng.instance())
     .withNoise()
     .withMplex(5.minutes, 5.minutes)
     .withMaxConnections(config.maxPeers)
@@ -221,6 +224,7 @@ proc new*(T: type CodexServer, config: CodexConf, privateKey: CodexPrivateKey): 
       .expect("Should start rest server!")
 
   switch.mount(network)
+
   T(
     config: config,
     codexNode: codexNode,
--- a/codex/codex.nim
+++ b/codex/codex.nim
@@ -151,7 +151,7 @@ proc new*(
     .new()
     .withPrivateKey(privateKey)
     .withAddresses(config.listenAddrs)
-    .withRng(rng.Rng.instance())
+    .withRng(Rng.instance())
     .withNoise()
     .withMplex(5.minutes, 5.minutes)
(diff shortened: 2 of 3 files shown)


codex/codex.nim
```diff
@@ -141,14 +141,18 @@ proc stop*(s: CodexServer) {.async.} =
 
   s.runHandle.complete()
 
-proc new*(T: type CodexServer, config: CodexConf, privateKey: CodexPrivateKey): T =
+
+proc newCodexServer*(
```
emizzle, line 145: We use `new` (for ref objects) and `init` (for objects) in many places in the codebase besides just this one. As an alternative to decorating our codebase with these changes everywhere, is this an issue we could potentially contribute to nimsuggest upstream to prevent the crashes?

codex/erasure/erasure.nim
```diff
@@ -63,10 +63,11 @@ type
...
 proc encode*(
-  self: Erasure,
-  manifest: Manifest,
-  blocks: int,
-  parity: int): Future[?!Manifest] {.async.} =
+    self: Erasure,
+    manifest: Manifest,
+    blocks: int,
+    parity: int
+): Future[?!Manifest] {.async.} =
```
emizzle, line 70: I definitely agree this is a much nicer way to pick out the return type. If we are going to make these formatting changes here, it would be my personally preference to do it uniformly across the entire codebase.

codex/codex.nim
elcritch: Ah, actually this one isn't directly related to the nimsuggest fix. 

I used it for tracking down the issue as the `new` proc's are a generic overload and nimsuggest won't fully type check them until they're used somewhere. 

For that reason with large complicated it's helpful to split into a concrete proc with a `new` wrapper. 

I could revert it back to `new`, but I think making the `newCodexServer` private would make `new` the external api but provide the compile checking.

codex/erasure/erasure.nim
elcritch: Nice, yah I could remove them from this PR and do a bigger PR for that styling in the repo. I didn't want to undertake that as a first PR. :)

codex/codex.nim
```diff
@@ -141,14 +141,17 @@ proc stop*(s: CodexServer) {.async.} =
 
   s.runHandle.complete()
 
-proc new*(T: type CodexServer, config: CodexConf, privateKey: CodexPrivateKey): T =
+proc createCodexServer(
```
elcritch, line 144: @emizzle I renamed it and made it private. What do you think?

codex/codex.nim
```diff
@@ -221,9 +224,18 @@ proc new*(T: type CodexServer, config: CodexConf, privateKey: CodexPrivateKey):
...
     config: config,
     codexNode: codexNode,
     restServer: restServer,
     repoStore: repoStore,
-    maintenance: maintenance)
+    maintenance: maintenance
+  )
+
+proc new*(
+    _: typedesc[CodexServer],
```
emizzle, line 237: I thought `typedesc[CodexServer]` was outdated in favour of `type CodexServer`?

codex/erasure/erasure.nim
emizzle: Sounds like a good idea :)

codex/codex.nim
emizzle: > Ah, actually this one isn't directly related to the nimsuggest fix.
>
>I used it for tracking down the issue as the new proc's are a generic overload and nimsuggest won't fully type check them until they're used somewhere.
>
>For that reason with large complicated it's helpful to split into a concrete proc with a new wrapper.
>
>I could revert it back to new, but I think making the newCodexServer private would make new the external api but provide the compile checking.

`CodexServer.new` is used on https://github.com/codex-storage/nim-codex/blob/master/codex.nim#L92. Shouldn't that allow for nimsuggest to compile check the `new` proc without needing the non-exported `newCodexServer`?

Whatever ends up being settled on, we should make an effort to be consistent across the codebase (change all instances), and document it somewhere.

> @emizzle I renamed it and made it private. What do you think?

My personal opinion is to have neither the `newCodexServer` nor `createCodexServer` workarounds and keep it as it is :) My suspicion is that nimsuggest shouldn't have issues compile checking the way it is, and that possibly there's a nimsuggest fix that could be looked into. But I don't know enough about nimsuggest to comment beyond speculation.

elcritch: > CodexServer.new is used on https://github.com/codex-storage/nim-codex/blob/master/codex.nim#L92. Shouldn't that allow for nimsuggest to compile check the new proc without needing the non-exported newCodexServer?

Yes it will compile & check it from that file. However, with the normal nimsuggest setup if you edit and modify the codex/codex.nim file it won't catch errors there. In this case it helped hide some other issues. 

> My personal opinion is to have neither the newCodexServer nor createCodexServer workarounds and keep it as it is :) 

I can see that point. Skimming through I see that most (but not all) of the `new` constructors are simple constructors. 

My inclination would be to have a separate name for more complicated constructors like this. A proc of `createCodexServer` or similar would help imply it's doing something more than a normal `new` proc. Though that could be left over habit from older style C++ constructors.

> Whatever ends up being settled on, we should make an effort to be consistent across the codebase (change all instances), and document it somewhere.

That's a good point. 

> My suspicion is that nimsuggest shouldn't have issues compile checking the way it is, and that possibly there's a nimsuggest fix that could be looked into. But I don't know enough about nimsuggest to comment beyond speculation.

Yah, I think it'd be a useful feature for nimsuggest (really nim). 

I believe the reason it's not type checked at declaration is that, is that actually a generic proc of type `new*[T: CodexServer](...): T` under the hood. 

Though in this instance the compiler has enough type information to be resolved. Given the ubiquity of the `new` pattern I'd think it'd make sense to get it added to nimsuggest / nim.

codex/codex.nim
elcritch: The `typedesc[Foo]` form is more common from what I've seen in most Nim libraries outside Status ones. Plus I believe that `type Foo` actually requires running `type` as a static function which returns `typedesc[Foo]`. I'll ask around about it.

elcritch: From what I can tell the recommendation appears to be `typedesc[T]`. It looks like the `type Foo` form is only used in a few status libs. I'd guess just because it's a slight bit easier, but it incurs a slight overhead in the compiler and overloads `type` more. 

Also, I searched through a dozen or so Nim libraries and the dominant usage was `newFoo`. I suspect because it avoids the generic overload and works better with nimsuggest type completetion. 

For now I backed out those changes though.

emizzle: I believe we are following the object construction recommendations in [the style guide](https://status-im.github.io/nim-style-guide/language.objconstr.html). If there are some improvements to be made to the style guide, it might be best to raise a PR in the style guide and have a discussion there. If those changes get merged, then we can go through the process of updating nim-codex.

elcritch: Thanks, that's a good idea.

markspanbroek (approved):

dryajov (approved):

codex.nimble
```diff
@@ -7,29 +7,30 @@ license = "MIT"
...
-         "questionable >= 0.10.6 & < 0.11.0",
-         "secp256k1",
-         "stew",
-         "upraises >= 0.1.0 & < 0.2.0",
-         "https://github.com/status-im/lrucache.nim#1.2.2",
-         "leopard >= 0.1.0 & < 0.2.0",
-         "blscurve",
-         "libp2pdht",
-         "eth"
+requires "nim >= 1.2.0"
```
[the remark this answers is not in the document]
elcritch: Stylistic but it can also break Nimble at times so I figured I'd throw it in.

codex/codex.nim
```diff
@@ -221,6 +224,7 @@ proc new*(T: type CodexServer, config: CodexConf, privateKey: CodexPrivateKey):
       .expect("Should start rest server!")
 
   switch.mount(network)
+
```
emizzle, line 227: Needed?

elcritch: It seems to be standard to have an ending newline, so I followed it.

PR #454, opened 2023-06-21
Improving proc/func formatting consistency

elcritch:
AKA learning the code base by trying to make proc formatting a bit more consistent! I also tried adding comments when I could. 

Goal of formatting style is to make it easy to pick out the return type and any pragmas.

General rule of thumb used to reformat a proc/func:
- more than 4 arguments
- complex return types and pragmas

Format used:

```nim
proc corruptBlocks*(
    store: BlockStore,
    manifest: Manifest,
    blks, bytes: int
): Future[seq[int]] {.async.} =
   ## comment
   discard "..."
```


markspanbroek (approved):
My rule of thumb is: if the signature is >80 characters, then format over multiple lines. It probably boils down to the same thing :smile:

PR #463, opened 2023-06-28
Tests for AsyncStreamWrapper

tbekas:


tests/codex/testasyncstreamwrapper.nim
```diff
@@ -0,0 +1,90 @@
...
+    server.start()
+
+    var transp = await connect(address)
+    var rstream = newAsyncStreamReader(transp)
+    var wrapper = AsyncStreamWrapper.new(reader = rstream)
+    var buf = newSeq[byte](data.len div 2)
+
+    let readLen = (await wrapper.readOnce(addr buf[0], buf.len))
+
+    await wrapper.closeImpl()
```
dryajov, line 65: We shouldn't call `closeImpl()` explicitly, it's a private field (hence `Impl`), calling `await wrapper.close()` is the proper way, which calls `closeImpl` internally. 

It's a bit unfortunate that `closeImpl` is marked as public, but I believe this is done because you can't have private inheritable `methods`.

dryajov (dismissed):

tests/codex/testasyncstreamwrapper.nim
tbekas: Thanks for the context. I also found the `Impl` suffix odd, but I thought it was a peculiar naming choice.

elcritch:
I'm not too familiar with the Stream / AsyncStream packages, but the code looks clean.

PR #475, opened 2023-07-10
Speed up clock.waitUntil()

markspanbroek:
`OnChainClock.waitUntil()` should now finish immediately when a new block arrives that has a timestamp after the due date. Previously it could take up to one second to complete the future. This messed with the timings in the
integration tests and made them less predictable. As an added benefit, we can now tighten the sleeps in the integration test so they finish faster.


AuHau (approved):
Pretty awesome nice change! 👍

PR #477, opened 2023-07-12
Dispatcher loop debug

benbierens:
This is proving to be quite tricky. Detecting the callback durations is easy. But identifying the source of outliers is not. I'm creating this draft PR to have an overview of the changes. I am not sure if this will end up as anything we want to merge.


dryajov:
Closing as I believe this has been addressed.

Issue #308, opened 2022-11-11
minprotobuf limits message size to 4mb


dryajov:
TL;DR

- minprotobuf is limited to 4mb message sizes
- manifests can hold up to ~131072 sha256  hashes
- largest file size that we can represent with a 4mb manifest is roughly ~8.5GB
- we can fix this with nested manifests
- it would be nice to set a more reasonable limit on manifest sizes - say ~10MB
- this needs allowing to specify/increasing the max size of the message in minprotobuf
- this would be needed anyway when exporting a manifest - even with nested manifests, you'd need to store all the manifest files together when exporting/importing

[minprotobuf](https://github.com/status-im/nim-libp2p/blob/unstable/libp2p/protobuf/minprotobuf.nim#L23) imposes a `4mb` size limit on messages. 

There isn't a limit that the standard imposes, but it largely makes sense to keep messages around a couple of megabytes as per the protobuf authors advice - https://stackoverflow.com/questions/34128872/google-protobuf-maximum-size. However, there are some legit use cases that might require larger message sizes, in our case, the `manifest` size grows proportionally to the block size and total dataset size and it could reach sizes of several dozens of megabytes. We have ways to mitigate this with nested manifests, but this needs to be implemented on the other hand, it would be nice if minprotobuf allowed to specify the max message size.

PR #479, opened 2023-07-12
Fix for decoding large manifests

### LINKED ISSUE #459: [BUG] - Codex crash on 10GB file download
**Describe the bug**
Codex node running via Docker Compose crashed when 10GB file is downloaded

**To Reproduce**
 1. Follow the guide [Codex Two-Client Test](https://github.com/codex-storage/nim-codex/blob/master/docs/TWOCLIENTTEST.md)
 2. Use Demo environment as a remote peer
 3. Run local node using Docker Compose
 4. Upload [10GB.bin](https://speed.hetzner.de) file to the remote node
    ```
    time \
    curl -X POST https://api.demo.codex.storage/node-01/api/codex/v1/upload \
      -H "Content-Type: application/octet-stream" \
      -H "Expect: 100-continue" \
      -T "10GB.bin" \
      -u $auth
    ```
    ```
    zdj7WmP6iwG3z1nebqrLaBpihWDDkHJqrrVaZfh9iXHhLyLkT
    real    22m52.401s
    user    0m8.916s
    sys     0m15.758s
    ```
 5. Download file via local node
    ```bash
    time \
    curl -m 86400 \
      --connect-timeout 86400 \
      http://localhost:8080/api/codex/v1/download/zdj7WmP6iwG3z1nebqrLaBpihWDDkHJqrrVaZfh9iXHhLyLkT \
      -o d-10GB.bin

      % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                    Dload  Upload   Total   Spent    Left  Speed
      0     0    0     0    0     0      0      0 --:--:--  0:07:00 --:--:--     0
    curl: (52) Empty reply from server

    real    7m0.744s
    user    0m0.018s
    sys     0m0.072s
    ```
    And at the same time in the Docker compose logs we see the following
    ```
    codex-node-codex-node1-1  | TRC 2023-06-26 13:44:45.342+00:00 Retrieved providers from persistent store  topics="discv5 providers manager" tid=1 cid=64*8454da len=0
    codex-node-codex-node1-1  | TRC 2023-06-26 13:44:45.343+00:00 Cleaning up query iterator                 topics="discv5 providers manager" tid=1
    codex-node-codex-node1-1  | TRC 2023-06-26 13:44:45.344+00:00 local providers:                           topics="discv5" tid=1 prov=
    codex-node-codex-node1-1  | TRC 2023-06-26 13:44:45.344+00:00 discv5 pending queries                     topics="discv5" tid=1 total=0
    codex-node-codex-node1-1  | TRC 2023-06-26 13:44:45.344+00:00 Closest nodes                              topics="discv5" tid=1 nodes=0
    codex-node-codex-node1-1  | TRC 2023-06-26 13:44:45.344+00:00 nearby:                                    topics="discv5" tid=1 nodesNearby=[]
    codex-node-codex-node1-1  | TRC 2023-06-26 13:44:45.344+00:00 getProviders collected:                    topics="discv5" tid=1 res=@[]
    codex-node-codex-node1-1  | TRC 2023-06-26 13:44:45.344+00:00 Discovered peers                           topics="codex discoveryengine" tid=1 peers=0
    codex-node-codex-node1-1  | TRC 2023-06-26 13:44:48.095+00:00 Pending WANT failed or expired             topics="codex pendingblocks" tid=1 exc="Timeout exceeded!"
    codex-node-codex-node1-1  | TRC 2023-06-26 13:44:48.095+00:00 Block handle success                       topics="codex blockexcengine" tid=1 cid=zdj*hLyLkT
    codex-node-codex-node1-1  | Traceback (most recent call last, using override)
    codex-node-codex-node1-1  | /src/codex.nim(130) codex
    codex-node-codex-node1-1  | /src/vendor/nim-chronos/chronos/asyncloop.nim(295) poll
    codex-node-codex-node1-1  | /src/vendor/nim-chronos/chronos/asyncfutures2.nim(369) futureContinue
    codex-node-codex-node1-1  | /src/vendor/nim-chronicles/chronicles/log_output.nim(740) fetchManifest
    codex-node-codex-node1-1  | /src/vendor/nimbus-build-system/vendor/Nim/lib/system/excpt.nim(631) signalHandler
    codex-node-codex-node1-1  | SIGSEGV: Illegal storage access. (Attempt to read from nil?)
    codex-node-codex-node1-1 exited with code 0
    ```

**Expected behavior**
 1. We should be able to download 10GB and bigger files.

**Screenshots**
Screenshot 2023-06-26 at 16 47 32

**Environment:**
```
Codex revision: 6afa42
Nim Compiler Version 1.6.11 [Linux: arm64]
```
[codexstorage/nim-codex:sha-6afa427](https://hub.docker.com/layers/codexstorage/nim-codex/sha-6afa427/images/sha256-0bc2bd161bfcf74425a8a115c73ed532eb210445927042c931abcc1a5a44aac9?context=explore)

**Additional context**

 1. Files we used to run remote and local nodes
    <details>
      <summary>files</summary>

      **run.sh**
      ```bash
      #!/bin/bash

      # Dynamic
      # PUBLIC_IP=$(curl -s ifconfig.me)
      PUBLIC_IP="0.0.0.0"

      # Variables
      export CODEX_METRICS="true"
      export CODEX_LISTEN_ADDRS="/ip4/${PUBLIC_IP}/tcp/30010"
      export CODEX_NAT="${PUBLIC_IP}"
      export CODEX_DISC_IP="${PUBLIC_IP}"
      export CODEX_DISC_PORT=20010
      export CODEX_AGENT_STRING="Codex Demo"
      export CODEX_API_BINDADDR="0.0.0.0"
      export CODEX_API_PORT=8080
      # export CODEX_BLOCK_TTL=172800
      export CODEX_STORAGE_QUOTA=16106127360

      # Run
      docker-compose down && docker-compose up -d
      ```

      **docker-compose.yaml**
      ```yaml
      services:
        codex-node1:
          image: codexstorage/nim-codex:sha-6afa427
          restart: always
          environment:
            - CODEX_LOG_LEVEL=${CODEX_LOG_LEVEL:-TRACE}
            - CODEX_METRICS=${CODEX_METRICS:-false}
            - CODEX_METRICS_ADDRESS=${CODEX_METRICS_ADDRESS:-0.0.0.0}
            - CODEX_METRICS_PORT=${CODEX_METRICS_PORT:-8008}
            - CODEX_DATA_DIR=${CODEX_DATA_DIR:-/datadir}
            - CODEX_LISTEN_ADDRS=${CODEX_LISTEN_ADDRS:-/ip4/0.0.0.0/tcp/2345}
            # - CODEX_NAT=${CODEX_NAT:-10.0.0.10}
            # - CODEX_DISC_IP=${CODEX_DISC_IP:-0.0.0.0}
            - CODEX_DISC_PORT=${CODEX_DISC_PORT:-8090}
            - CODEX_NET_PRIVKEY=${CODEX_NET_PRIVKEY:-key}
            # - CODEX_BOOTSTRAP_NODE=${CODEX_BOOTSTRAP_NODE}
            - CODEX_MAX_PEERS=${CODEX_MAX_PEERS:-160}
            - CODEX_AGENT_STRING=${CODEX_AGENT_STRING:-Codex}
            - CODEX_API_BINDADDR=${CODEX_API_BINDADDR:-127.0.0.1}
            - CODEX_API_PORT=${CODEX_API_PORT:-8080}
            - CODEX_REPO_KIND=${CODEX_REPO_KIND:-fs}
            - CODEX_STORAGE_QUOTA=${CODEX_STORAGE_QUOTA:-8589934592}
            - CODEX_BLOCK_TTL=${CODEX_BLOCK_TTL:-0}
            # - CODEX_BLOCK_MI=${CODEX_BLOCK_MI}
            - CODEX_BLOCK_MN=${CODEX_BLOCK_MN:-1000}
            - CODEX_CACHE_SIZE=${CODEX_CACHE_SIZE:-0}
            - CODEX_PERSISTENCE=${CODEX_PERSISTENCE:-false}
            - CODEX_ETH_PROVIDER=${CODEX_ETH_PROVIDER:-ws://localhost:8545}
            # - CODEX_ETH_ACCOUNT=${CODEX_ETH_ACCOUNT}
            # - CODEX_MARKETPLACE_ADDRESS=${CODEX_MARKETPLACE_ADDRESS:-0x59b670e9fA9D0A427751Af201D676719a970857b}
            - CODEX_VALIDATOR=${CODEX_VALIDATOR:-false}
            - CODEX_VALIDATOR_MAX_SLOTS=${CODEX_VALIDATOR_MAX_SLOTS:-1000}
          ports:
            - 8080:8080/tcp   # REST API
            - 8008:8008/tcp   # Metrics
            - 30010:30010/tcp # libp2p
            - 20010:20010/udp # DHT discovery
          volumes:
            - /opt/codex/datadir:/datadir
          networks:
            - codex

      networks:
        codex:
          driver: bridge

      ```
    </details>

 2. I will try to reproduce the issue with the latest version from the master branch. Tested version is just 4 days old.

tbekas:
Resolves #459
Related to #308

### DIFF
--- a/codex/blockexchange/protobuf/message.nim
+++ b/codex/blockexchange/protobuf/message.nim
@@ -5,6 +5,11 @@
 
 import pkg/libp2p/protobuf/minprotobuf
 
+import ../../units
+
+const
+  MaxBlockSize* = 100.MiBs.uint
+  MaxMessageSize* = 100.MiBs.uint
 
 type
   WantType* = enum
@@ -72,7 +77,7 @@ proc write*(pb: var ProtoBuffer, field: int, value: Wantlist) =
   pb.write(field, ipb)
 
 proc write*(pb: var ProtoBuffer, field: int, value: Block) =
-  var ipb = initProtoBuffer()
+  var ipb = initProtoBuffer(maxSize = MaxBlockSize)
   ipb.write(1, value.prefix)
   ipb.write(2, value.data)
   ipb.finish()
@@ -99,7 +104,7 @@ proc write*(pb: var ProtoBuffer, field: int, value: StateChannelUpdate) =
   pb.write(field, ipb)
 
 proc protobufEncode*(value: Message): seq[byte] =
-  var ipb = initProtoBuffer()
+  var ipb = initProtoBuffer(maxSize = MaxMessageSize)
   ipb.write(1, value.wantlist)
   for v in value.payload:
     ipb.write(3, v)
@@ -175,14 +180,14 @@ proc decode*(_: type StateChannelUpdate, pb: ProtoBuffer): ProtoResult[StateChan
 proc protobufDecode*(_: type Message, msg: seq[byte]): ProtoResult[Message] =
   var
     value = Message()
-    pb = initProtoBuffer(msg)
+    pb = initProtoBuffer(msg, maxSize = MaxMessageSize)
     ipb: ProtoBuffer
     sublist: seq[seq[byte]]
   if ? pb.getField(1, ipb):
     value.wantlist = ? Wantlist.decode(ipb)
   if ? pb.getRepeatedField(3, sublist):
     for item in sublist:
-      value.payload.add(? Block.decode(initProtoBuffer(item)))
+      value.payload.add(? Block.decode(initProtoBuffer(item, maxSize = MaxBlockSize)))
   if ? pb.getRepeatedField(4, sublist):
     for item in sublist:
       value.blockPresences.add(? BlockPresence.decode(initProtoBuffer(item)))
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,4 @@ codex.nims
 nimbus-build-system.paths
 docker/hostdatadir
 docker/prometheus-data
+.DS_Store
--- a/codex/blockexchange/network/network.nim
+++ b/codex/blockexchange/network/network.nim
@@ -84,6 +84,8 @@ proc send*(b: BlockExcNetwork, id: PeerId, msg: pb.Message) {.async.} =
       await b.inflightSema.acquire()
       trace "Sending message to peer", peer = id
       await peer[].send(msg)
+    except CatchableError as err:
+      error "Error sending message", peer = id, msg = err.msg
     finally:
       b.inflightSema.release()
   do:
--- a/codex/blockexchange/network/networkpeer.nim
+++ b/codex/blockexchange/network/networkpeer.nim
@@ -15,13 +15,11 @@ import pkg/chronicles
 import pkg/libp2p
 
 import ../protobuf/blockexc
+import ../protobuf/message
 import ../../errors
 
 logScope:
   topics = "codex blockexcnetworkpeer"
-
-const
-  MaxMessageSize = 100 * 1 shl 20 # manifest files can be big
 
 type
   ConnProvider* = proc(): Future[Connection] {.gcsafe, closure.}
@@ -45,12 +43,12 @@ proc readLoop*(b: NetworkPeer, conn: Connection) {.async.} =
   try:
     while not conn.atEof or not conn.closed:
       let
-        data = await conn.readLp(MaxMessageSize)
+        data = await conn.readLp(MaxMessageSize.int)
         msg = Message.protobufDecode(data).mapFailure().tryGet()
       trace "Got message for peer", peer = b.id
       await b.handler(b, msg)
-  except CatchableError as exc:
-    trace "Exception in blockexc read loop", exc = exc.msg
+  except CatchableError as err:
+    warn "Exception in blockexc read loop", msg = err.msg
   finally:
     await conn.close()
 


codex/blockexchange/protobuf/message.nim
```diff
@@ -5,6 +5,15 @@
 
 import pkg/libp2p/protobuf/minprotobuf
 
+const
+  DefaultMaxSize = 100 * 1 shl 20
```
elcritch, line 9: Could we change this to something easier to read at a glance? 

You could do `10.MiBs.int` [units.nim](https://github.com/codex-storage/nim-codex/blob/a89938478a1a0c58b48ff8ec3bddfd5c96b51522/codex/units.nim#L49C6-L49C10).

elcritch:
Looks like there's a few other usages of `initProtobuffers`. Should these be updated too? 

https://github.com/codex-storage/nim-codex/blob/a89938478a1a0c58b48ff8ec3bddfd5c96b51522/codex/manifest/coders.nim#L35

dryajov:
> Looks like there's a few other usages of `initProtobuffers`. Should these be updated too?
> 
> https://github.com/codex-storage/nim-codex/blob/a89938478a1a0c58b48ff8ec3bddfd5c96b51522/codex/manifest/coders.nim#L35

I think we're fine with the default there, the manifest is a special case, but it should also get mostly resolved once we move over to merkle trees.

dryajov (approved):
LGTM.

tbekas:
> Looks like there's a few other usages of `initProtobuffers`. Should these be updated too?
> 
> https://github.com/codex-storage/nim-codex/blob/a89938478a1a0c58b48ff8ec3bddfd5c96b51522/codex/manifest/coders.nim#L35

I agree that those other places may become an issue in the future. So ideally I would like to see this parameter to be set in a single place as a default configuration for a Protobuf decoder. However doing it this way would require a larger refactor. So for now I'm just trying to fix a single issue.

veaceslavdoina:
@tbekas, thanks for the help to test the fix

```
              Upload   --> Codex Demo
            /                      |
HTTP Client                        |
            \                      |
              Download <-- Codex Local
```
```
10GB.bin - upload   - 13m24.079s
10GB.bin - download - 32m19.352s
```

Upload was so fast probably because file was already uploaded?

elcritch (dismissed):

codex/blockexchange/protobuf/message.nim
```diff
@@ -5,6 +5,17 @@
 
 import pkg/libp2p/protobuf/minprotobuf
 
+import ../../units
+
+const
+  DefaultMaxSize = 100.MiBs.uint
+  MaxEntrySize* = DefaultMaxSize
```
dryajov, line 12: I would not actually set this to 100mb, the default of 4mb(?) is more than enough for everything else but the manifest.

dryajov:
> > Looks like there's a few other usages of `initProtobuffers`. Should these be updated too?
> > https://github.com/codex-storage/nim-codex/blob/a89938478a1a0c58b48ff8ec3bddfd5c96b51522/codex/manifest/coders.nim#L35
> 
> I agree that those other places may become an issue in the future. So ideally I would like to see this parameter to be set in a single place as a default configuration for a Protobuf decoder. However doing it this way would require a larger refactor. So for now I'm just trying to fix a single issue.

The current default in the core protobuf (which is 4mb) should be more than enough. The manifest is a special case and should be mostly addressed by https://github.com/codex-storage/codex-research/pull/169. I would in fact only set the manifest limit past 4mb, if we're using protobuf for anything larger than that, then protobuf is the wrong tool for the job. One of the issues with protobuf is that it cannot be streamed and as such it can become a bottleneck or even an attack vector.

dryajov (changes requested):

codex/blockexchange/protobuf/message.nim
tbekas: Yes, the default is 4mb. Until we have https://github.com/codex-storage/codex-research/pull/169 I would leave 100mb for the manifest, it would mean that with the current block size we can handle downloading of files up to ~160GB.

tbekas:
> Upload was so fast probably because file was already uploaded?

@veaceslavdoina  No, upload is faster because in the current implementation it's a simple local operation, there's no p2p communication involved. Download on the other hand require DHT lookups and p2p comms.

dryajov (approved):
LGTM

PR #483, opened 2023-07-13
Taiko L2

markspanbroek:
Several fixes to allow a codex node to run against the Taiko L2 testnet.

- Adds CLI parameter to specify an ethereum private key
- Adds deployed address of Marketplace contract on Taiko
- Waits on 1 confirmation for the ERC20 approve() call before doing further calls
- Adds integration test that can be run with `make testTaiko`


tests/contracts/testContracts.nim
```diff
@@ -38,10 +38,10 @@ ethersuite "Marketplace contracts":
     request.client = await client.getAddress()
 
     switchAccount(client)
-    await token.approve(marketplace.address, request.price)
+    discard await token.approve(marketplace.address, request.price)
```
dryajov, line 41: probably want `asyncSpawn`, otherwise you're swallowing exceptions, which is fun to debug.

markspanbroek: I'm not discarding the future, I'm discarding the transaction receipt, after awaiting the future :smile:

codex/conf.nim
```diff
@@ -227,6 +227,12 @@ type
         name: "eth-account"
       .}: Option[EthAddress]
 
+      ethPrivateKey* {.
```
dryajov, line 230: Should we really allow passing a private hey from the cmd? It seems very insecure as it gets logged in multiple place, like your history, or `ps -ef`, etc... I would suggest we use a file instead similar to https://github.com/codex-storage/nim-codex/blob/master/codex/utils/keyutils.nim

codex/contracts/market.nim
```diff
@@ -36,7 +34,7 @@ proc approveFunds(market: OnChainMarket, amount: UInt256) {.async.} =
   let tokenAddress = await market.contract.token()
   let token = Erc20Token.new(tokenAddress, market.signer)
 
-  await token.approve(market.contract.address(), amount)
+  discard await token.approve(market.contract.address(), amount).confirm(1)
```
dryajov, line 37: `asyncSpawn` unless you really like suffering - `discard` will swallow exceptions.

tests/contracts/testContracts.nim
dryajov: ah, you're right :)

codex/conf.nim
markspanbroek: Yes, good point. It wasn't meant to be the final solution, just something to get some testing going on an L2. But it shouldn't be too much effort to use a file instead.

markspanbroek: Fixed in 7017a91e1436c4e80fd208116345bccf48e8ec96

tests/integration/nodes.nim
```diff
@@ -81,13 +53,48 @@ proc startNode*(args: openArray[string], debug: string | bool = false): NodeProc
   node.start()
   node
 
+proc dataDir(node: NodeProcess): ?string =
+  for argument in node.arguments:
+    if argument.startsWith("--data-dir="):
+      return some argument[len("--data-dir=")..<argument.len]
+    if argument.startsWith("-d "):
+      return some argument[len("-d ")..<argument.len]
+  none string
```
emizzle, line 62:
```suggestion
import pkg/chronicles
import pkg/confutils
import pkg/libp2p
import pkg/codex/conf

proc dataDir(node: NodeProcess): string =
  let config = CodexConf.load(cmdLine = node.arguments)
  return config.dataDir
```

emizzle (dismissed):
Well done, Mark, looks really good! A few minor suggestions, but nothing blocking.

dryajov (approved):
LGTM, I would apply @emizzle suggestions tho. They should work and are much cleaner.

codex/codex.nim
```diff
@@ -50,29 +52,40 @@ type
...
     quit QuitFailure
 
   let provider = JsonRpcProvider.new(config.ethProvider)
-  let signer = provider.getSigner(account)
+  var signer: Signer
+  if account =? config.ethAccount:
+    signer = provider.getSigner(account)
+  elif keyFile =? config.ethPrivateKey:
+    without isSecure =? checkSecureFile(keyFile) and isSecure:
+      error "Ethereum private key file does not have safe file permissions"
```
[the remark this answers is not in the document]
dryajov: `checkSecureFile` should check for file presence already?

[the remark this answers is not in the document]
emizzle: > `checkSecureFile` should check for file presence already?

It does, which would return `IoResult[bool, IoErrorCode]` with the `IoErrorCode` error set if the file was not found. The problem is that this statement checks both the Result error and value and if neither are correct, returns an error message that only mentions safe file permissions, but does not address file not found (or other `IoErrorCode`).

codex/contracts/market.nim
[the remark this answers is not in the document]
markspanbroek: I looked into this a bit more, but it doesn't seem that we can combine the `token.approve()` and a call into the marketplace into a single transaction. I was thinking of using [multicall](https://docs.openzeppelin.com/contracts/4.x/utilities#multicall), but that only works for batching calls to the same contract.

Other [sources](https://github.com/mds1/multicall#batch-contract-writes) seem to confirm that it is not possible to combine calls into different contracts, because it messes with `msg.sender`.

codex/codex.nim
```diff
@@ -50,29 +52,40 @@ type
...
+      quit QuitFailure
+    without key =? keyFile.readAllChars():
+      error "Unable to read Ethereum private key file"
+      quit QuitFailure
```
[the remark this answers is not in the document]
markspanbroek: Should be fixed now; I updated nim-ethers to not crash when the private key is wrong (https://github.com/codex-storage/nim-ethers/pull/52), and used that to give a better error message, as you suggested.

codex/codex.nim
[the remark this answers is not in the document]
markspanbroek: Should be fixed now

tests/integration/nodes.nim
markspanbroek: Should be fixed now

tests/integration/nodes.nim
```diff
@@ -81,13 +53,48 @@ proc startNode*(args: openArray[string], debug: string | bool = false): NodeProc
...
+  var address = "127.0.0.1"
+  var port = "8080"
+  for argument in node.arguments:
+    if argument.startsWith("--api-bindaddr="):
+      address = argument[len("--api-bindaddr=")..<argument.len]
+    elif argument.startsWith("--api-port="):
+      port = argument[len("--api-port=")..<argument.len]
+    elif argument.startsWith("-p "):
+      port = argument[len("-p ")..<argument.len]
+  "http://" & address & ":" & port & "/api/codex/v1"
```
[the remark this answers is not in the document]
markspanbroek: Should be fixed now

markspanbroek:
Seeing some failing tests in `testSales` after rebasing onto master, could be related to #494

PR #485, opened 2023-07-18
split windows CI tests into parts

elcritch:
Splite the unit, contract, and integration tests into separate parts. 

This incurs a slight overhead per Windows VM for duplicating `nimble setup` stages (~3min) but should cut each windows run down to ~20-25min. 

Another possible issue would be CI time. IIRC, Github Actions charge per minute. If so that'd mean the total minutes of CI runner time would roughly be about the same (give or take 3x * 3min or so).

If anyone else knows of other downsides let me know!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,23 +27,23 @@ jobs:
             cpu: amd64
             builder: windows-latest
             shell: msys2
-            tests: part1
+            tests: unittest
           - os: windows
             cpu: amd64
             builder: windows-latest
             shell: msys2
-            tests: part2
+            tests: contract
           - os: windows
             cpu: amd64
             builder: windows-latest
             shell: msys2
-            tests: part3
+            tests: integration
 
     defaults:
       run:
         shell: ${{ matrix.shell }} {0}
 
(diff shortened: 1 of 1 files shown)


markspanbroek (approved):
Nice, thanks @elcritch!

dryajov (approved):

dryajov:
Fixes #489

veaceslavdoina:
[About billing for GitHub Actions](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions)
> GitHub Actions usage is **free for standard GitHub-hosted runners in public repositories**, and for self-hosted runners. For private repositories, each GitHub account receives a certain amount of free minutes and storage for use with GitHub-hosted runners, depending on the product used with the account. Any usage beyond the included amounts is controlled by spending limits.

[BuildJet - macOS M1 runners & Windows Runners](https://buildjet.com/for-github-actions/docs/runners/hardware#mac-os-m1-runners-and-windows-runners)
> We can offer both macOS M1 and Windows runners, but due to their respective licensing terms, we can't offer it as part of regular per minute offering. Therefore, we **only offer runners per month, and we only take orders of 10 runners or more**.

PR #492, opened 2023-07-27
Bump deps take2

elcritch:
Update dependencies to mostly match `codexdht`. ~~Requires bumping Nim to 1.6.14.~~ Nimbus has been updated to 1.6.14. 

Needed to add:

* vendor/nim-protobuf-serialization
* vendor/nim-results
* vendor/nim-testutils
* vendor/npeg


elcritch:
Depends on https://github.com/codex-storage/nim-codex-dht/pull/80 and https://github.com/codex-storage/questionable/pull/36 

Though the `questionable` change could be changed to basic Result type.

codex/errors.nim
```diff
@@ -9,15 +9,22 @@
...
+
 type
   CodexError* = object of CatchableError # base codex error
   CodexResult*[T] = Result[T, ref CodexError]
 
-template mapFailure*(
-    exp: untyped,
-    exc: typed = type CodexError
-): untyped =
+proc mapFailure*[T, V, E](
```
dryajov, line 18: I really don't think this should be a proc, we want this inlined as much as possible.

codex/utils/keyutils.nim
```diff
@@ -26,18 +28,17 @@ proc setupKey*(path: string): ?!PrivateKey =
   if not path.fileAccessible({AccessFlags.Find}):
     info "Creating a private key and saving it"
     let
-      res = ? PrivateKey.random(Rng.instance()[]).mapFailure(CodexKeyError)
-      bytes = ? res.getBytes().mapFailure(CodexKeyError)
+      res = ? PrivateKey.random(Rng.instance()[]).mapFailure()
```
dryajov, line 31: This explicitly maps key related errors (which is an enum iirc), into a `CodexKeyError`, what the exception type it's being mapped to now?

tests/codex/blockexchange/discovery/testdiscovery.nim
```diff
@@ -5,7 +5,7 @@ import std/tables
 import pkg/asynctest
 import pkg/chronos
 
-import pkg/libp2p
+import pkg/libp2p except setup
```
dryajov, line 8: in order to avoid doing `except setup`, we might want to create a top level re-export (in the test directory) with

```nim
import pkg/libp2p

export libp2p except setup
```

elcritch: Yah it crossed my mind too, after updating 20 of them. :) It'd be nice just to have it. We do already have a `helper.nim` test thing, but maybe we'd want a new one? 

I'll let it stew a bit over the weekend and see on Monday.

codex/errors.nim
elcritch: Good point. It was giving compilation errors so I moved it to a proc to help figure them out. 

But I think a typed template would work as well. Especially now I know more about the source of the issues.

codex/utils/keyutils.nim
elcritch: I was having problems getting compilation to work with `mapFailure(CodexKeyError)` so I tried making the error types more generic. Part of it was my misunderstanding about the Result types and generic type conversions on  them, and part of it changes to compiler.

But I figured out the issue with `mapFailure` and should be able to change this back.

codex/errors.nim
elcritch: Hmmm, so here's been the annoying bit. When `mapFailure` is a proc type everything compiles. 

However if I change `mapFailure` back to a template I get a compile error: 

```nim
/Users/jaremycreechley/projs/status/nim-codex/codex/manifest/types.nim(14, 11) Warning: imported and not used: 'questionable' [UnusedImport]
/Users/jaremycreechley/projs/status/nim-codex/codex/manifest/coders.nim(143, 40) template/generic instantiation of `mapFailure` from here
/Users/jaremycreechley/projs/status/nim-codex/codex/errors.nim(27, 6) Error: type mismatch: got <Result[cid.Cid, cid.CidError], proc (e`gensym76: CidError): ref CatchableError{.noSideEffect, gcsafe, locks: 0.} | proc (e`gensym70: CidError): ref CatchableError{.noSideEffect, gcsafe, locks: 0.}>
but expected one of:
func mapErr[T; E0: not void; E1: not void](self: Result[T, E0];
    f: proc (x: E0): E1): Result[T, E1]
  first type mismatch at position: 2
  required type for f: proc (x: E0: not void): E1: not void{.closure.}
  but expression 'convertToErr' is of type: None
func mapErr[T; E0: not void](self: Result[T, E0]; f: proc (x: E0)): Result[T,
    void]
  first type mismatch at position: 2
  required type for f: proc (x: E0: not void){.closure.}
  but expression 'convertToErr' is of type: None
2 other mismatching symbols have been suppressed; compile with --showAllMismatches:on to see them

expression: mapErr(init(Cid, blockBuf), convertToErr)
stack trace: (most recent call last)
/Users/jaremycreechley/.asdf/installs/nim/1.6.14/lib/system/nimscript.nim(429, 18)
/Users/jaremycreechley/projs/status/nim-codex/build.nims(41, 16) testTask
/Users/jaremycreechley/projs/status/nim-codex/build.nims(28, 8) testCodexTask
/Users/jaremycreechley/projs/status/nim-codex/build.nims(21, 15) test
/Users/jaremycreechley/projs/status/nim-codex/build.nims(18, 7) buildBinary
/Users/jaremycreechley/.asdf/installs/nim/1.6.14/lib/system/nimscript.nim(273, 7) exec
/Users/jaremycreechley/.asdf/installs/nim/1.6.14/lib/system/nimscript.nim(273, 7) Error: unhandled exception: FAILED: nim c --out:build/testCodex -d:codex_enable_proof_failures=true tests/testCodex.nim [OSError]
```

If I make the inner `convertToErr` into a template too I get compile errors elsewhere. These are similar errors that the original `mapFailure` template gave: 

```nim
/Users/jaremycreechley/projs/status/nim-codex/codex/manifest/types.nim(14, 11) Warning: imported and not used: 'questionable' [UnusedImport]
/Users/jaremycreechley/projs/status/nim-codex/codex/manifest/manifest.nim(103, 1) template/generic instantiation from here
/Users/jaremycreechley/projs/status/nim-codex/codex/errors.nim(27, 14) Error: convertToErr can raise an unlisted exception: Exception
stack trace: (most recent call last)
/Users/jaremycreechley/.asdf/installs/nim/1.6.14/lib/system/nimscript.nim(429, 18)
/Users/jaremycreechley/projs/status/nim-codex/build.nims(41, 16) testTask
/Users/jaremycreechley/projs/status/nim-codex/build.nims(28, 8) testCodexTask
/Users/jaremycreechley/projs/status/nim-codex/build.nims(21, 15) test
/Users/jaremycreechley/projs/status/nim-codex/build.nims(18, 7) buildBinary
/Users/jaremycreechley/.asdf/installs/nim/1.6.14/lib/system/nimscript.nim(273, 7) exec
/Users/jaremycreechley/.asdf/installs/nim/1.6.14/lib/system/nimscript.nim(273, 7) Error: unhandled exception: FAILED: nim c --out:build/testCodex -d:codex_enable_proof_failures=true tests/testCodex.nim [OSError]
```

elcritch: Phew, okay this version works: 

```nim
template mapFailure*[T, V, E](
    exp: Result[T, V],
    exc: typedesc[E],
): Result[T, ref CatchableError] =
  ## Convert `Result[T, E]` to `Result[E, ref CatchableError]`
  ##

  exp.mapErr(proc (e: V): ref CatchableError = (ref exc)(msg: $e))
```

That reminds me of the issue with `libp2p except setup`. I'd setup the inner `convertToErr` proc to try and resolve some other compile issues with type resolution. However, using a a named inner `convertToErr` proc/template blows up. However, an anonymous (typed) proc works! Confused yet? I am! 😂

Ok, so `libp2p` export issue is with the `asynctest` macros redefine a local `template setup` which has a different type and captures the `unittest.setup` template. That works until you import another `setup` name which seems to cause the whole template name resolution order to get weird.

However, Nim 1.6.14 includes fixes for all the defect / raises handling from the Chronos refactor. So... a bit of whack a mole.

Last commit works better.

codex/blockexchange/network/networkpeer.nim
```diff
@@ -79,7 +79,7 @@ proc broadcast*(b: NetworkPeer, msg: Message) =
 
   asyncSpawn sendAwaiter()
 
-func new*(
+proc new*(
```
dryajov, line 82: Why change this? Usually we don't want side effects?

codex/conf.nim
```diff
@@ -292,9 +293,17 @@ proc defaultDataDir*(): string =
 
   getHomeDir() / dataDir
 
-proc parseCmdArg*(T: type MultiAddress, input: string): T
+proc parseCmdArg*(T: typedesc[MultiAddress],
+                  input: string): MultiAddress
                  {.upraises: [ValueError, LPError].} =
-  MultiAddress.init($input).tryGet()
+  var ma: MultiAddress
```
dryajov, line 299: should probably use `without ma =? ...:`

codex/conf.nim
```diff
@@ -337,6 +346,18 @@ proc readValue*(r: var TomlReader, val: var SignedPeerRecord) =
 
   val = SignedPeerRecord.parseCmdArg(uri)
 
+proc readValue*(r: var TomlReader, val: var MultiAddress) =
+  without input =? r.readValue(string).catch, err:
+    error "invalid MultiAddress configuration value", error = err.msg
+    quit QuitFailure
+
+  let res = MultiAddress.init(input)
```
dryajov, line 354: Same, `without` is probably more straight forward here.

codex/blockexchange/network/networkpeer.nim
elcritch: Not sure, I might've been testing if affected some of the compile issues. I'll revert it back as it's not.

codex/conf.nim
elcritch: Looks like without is useable here.

codex/conf.nim
elcritch: I'm not sure if that works since `val` is an argument? Though maybe it'd work with a `val = res` at the end.

tests/codex/blockexchange/discovery/testdiscovery.nim
dryajov: Yea, probably a dedicated `lphelper` or something like that.

elcritch: I just stuck it in `tests/helper.nim` since it's already used in all of the unit tests that used libp2p.

vendor/urls.rules
dryajov, line 1: why do we need this file?

codex/conf.nim
elcritch: @dryajov, changing this to use `without` causes a compilation failure on Windows. It's something I've run into a few times and only occurs on Windows (?). 

```nim
D:\a\nim-codex\nim-codex\codex\conf.nim(299, 59) Error: Invalid node kind nnkStmtList for macros.`$`
```

Here's the error log: https://github.com/codex-storage/nim-codex/actions/runs/5720835236/job/15501441283#step:8:386

vendor/urls.rules
elcritch: It's part of atlas setup. I used atlas to help me update the deps and figure out the new deps.

codex/errors.nim
elcritch: I think @emizzle 's changes to FuturesTracking actually helped fix up some of these issues. :)

codex/conf.nim
elcritch: I filed an issue for this in questionable since it's popped up randomly a couple of times: https://github.com/codex-storage/questionable/issues/39

tests/nimlldb.py
dryajov, line 1: did this get checked in by mistake?

elcritch: I'd added that for @tbekas and @gmega for debugging. It's only on Nim devel and figured it'd be handy to include for now to easy debugging.

PR #493, opened 2023-07-27
Cancel and wait for asyncstatemachine futures when stopping

### LINKED ISSUE #490: Cancel and wait for async state machine futures when stopping the state machine
Currently, the async state machine creates futures for the current state's run method, as well as the internal state machine's scheduler. During `stop`, these futures are cancelled, but the cancel operation should be waited for before completing `stop`.

Can use a `TrackedFutures` in #455 on the state machine type, track both futures, and `machine.trackedFutures.cancelTracked` in `stop`.

emizzle:
Fixes #490.

Previously, the async state machine created futures for the current state's run method, the internal state machine's scheduler, and the `AsyncQueue.pop` that waits for events to be pushed to the queue. During `stop`, all three of these futures are now cancelled and waited before completing `stop`.

The implementation uses the `TrackedFutures` to track both futures, and `machine.trackedFutures.cancelTracked` in `stop`.

Simplifications to the `then` (promise) api were made that removed multiple callbacks in `.then`, borrowed from the javascript promise api. Additionally, the `then` tests were revamped for simplicity.

The `TrackedFutures` module was updated to remove tracked futures when a future is cancelled, and all futures are removed from the table once `cancelTracked` completes cancelling (and waiting) all futures.

### DIFF
--- a/codex/utils/asyncstatemachine.nim
+++ b/codex/utils/asyncstatemachine.nim
@@ -1,7 +1,10 @@
+import std/sugar
 import pkg/questionable
 import pkg/chronos
 import pkg/chronicles
 import pkg/upraises
+import ./trackedfutures
+import ./then
 
 push: {.upraises:[].}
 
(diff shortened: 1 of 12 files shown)


codex/utils/asyncstatemachine.nim
```diff
@@ -84,18 +90,20 @@ proc start*(machine: Machine, initialState: State) =
 
   if machine.scheduled.isNil:
     machine.scheduled = newAsyncQueue[Event]()
-  machine.scheduling = machine.scheduler()
+
   machine.started = true
+  machine.scheduler()
+    .track(machine)
+    .catch((err: ref CatchableError) =>
+      error("Error in scheduler", error = err.msg)
```
markspanbroek, line 98: Good that you make the error handling explicit here. The scheduler should never fail. If it does, then it's a programmer error. So ideally I'd like the program to crash, not hide the error in the logs.

markspanbroek (approved):

codex/utils/asyncstatemachine.nim
emizzle: Oops, I forgot to address this. Will open a new PR.

Issue #494, opened 2023-07-28
Slot queue test fails occasionally


markspanbroek:
This test fails occasionally (once every 10 runs or so):

https://github.com/codex-storage/nim-codex/blob/1d161d383ec2738cdc3a0eac16e8b0788f218edb/tests/codex/sales/testsales.nim#L178

I couldn't see anything obviously wrong, could you take a look @emizzle?

emizzle:
Locally, in master, I was able to successfully run `testsales` 20 times with no errors, and `make test` 10 times with no errors.

There also doesn't seem to be any errors in CI relating to this.

Are you definitely still seeing this error?

markspanbroek:
It's a tough one to reproduce for some reason. On my older laptop I see the error quite frequently, but on my newer desktop see it much less frequently. I was able to reproduce it on CI as well, but it did not occur as often as I expected: https://github.com/codex-storage/nim-codex/actions/runs/6010028817/job/16300630141?pr=539#step:4:2933

```
testsales.nim(277, 28): Check failed: eventuallyCheck itemsProcessed.len == request.ask.slots.int
```

I had to run the test 10000 times :smile:  #539

PR #495, opened 2023-07-28
[sales] Fix error: "the length of the table changed while iterating"

markspanbroek:
Occasionally occurs when running the sales tests. Not related to the test failure in #494.

### DIFF
--- a/codex/sales/trackedfutures.nim
+++ b/codex/sales/trackedfutures.nim
@@ -1,5 +1,6 @@
 import std/sugar
 import std/tables
+import std/sequtils
 import pkg/chronicles
 import pkg/chronos
 import ../utils/then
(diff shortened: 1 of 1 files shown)


emizzle (commented):
Thanks for picking this up! I saw this when modifying the sales statemachine (#493) and fixed it in a slightly different way. Instead of copying to a seq first, `.track` was modified to prevent changes to the table when the cancelling flag is set: https://github.com/codex-storage/nim-codex/blob/d33ab8cf283b43ad5c2887d19e36b114024f955c/codex/utils/trackedfutures.nim#L22-L24

I *think* with the changes in #493, we shouldn't need the changes in this PR.

emizzle:
I've now merged #493 into master, so let's see if we still encounter this error.

PR #496, opened 2023-07-31
Initial support for nimble lock files

yyoncho:
- Add parallel ci-lock.yml
- In order to sync the lock file do the following:

``` bash
nimble lock --developFile=nimble-update-lock.develop
```

Note that --developFile support for lock is part of  this PR https://github.com/nim-lang/nimble/pull/1125 to be merged. If it is not merged, then you can rename nimble-update-lock.develop to nimble.develop and `nimble lock` will work just fine.

- I have used different `nim` version for the lock file because current `nim` does not have the proper nimble structure. I can bump the version in NBS if you want. I have also changed the versions of some of the packages in `vendor` folder to pass nimble version checks


elcritch:
Looks good. There will be some conflicts with my PR to update the deps, but I can handle merging the conflicts. That should resolve the build errors you're seeing.

elcritch:
@yyoncho I still have this in my work queue. It's just been bumped as I needed to work on some other things first.

elcritch:
@dryajov mind reviewing this? It'd be nice to get Nimble running aside NBS.

gmega:
I'm killing this as we've sort of embraced nimbus-build-system, so our lock files are the commit versions in the `/vendor` folder. Until package maintainers do proper versioning (so that the lockfile reflects one specific solution to dep constraints), I also don't see much of a win in using lockfiles as they are more complicated to deal with than submodules.

dryajov:
Keep in mind that the plan is still to migrate to nimble and lockfiles, once we migrate to nim 2.0

PR #497, opened 2023-08-02
Simplify empty cid handling

elcritch:


codex/blocktype.nim
```diff
@@ -37,11 +37,10 @@ type
...
     data*: seq[byte]
 
-template EmptyCid*: untyped =
-  var
-    EmptyCid {.global, threadvar.}:
-      array[CIDv0..CIDv1, Table[MultiCodec, Cid]]
+var
+  EmptyCid {.threadvar.}: array[CIDv0..CIDv1, Table[MultiCodec, Cid]]
 
+proc emptyCid*(version: CidVersion, codex: MultiCodec): ?!Cid =
```
elcritch, line 43: @dryajov not sure what but the previous templates combined with ORC were causing lots of defects trying to access a Result error that didn't exist.

My hunch is that all the error catches and gets needed for the tables were causing issues unrelated to ORC itself. The templates were giving me issues on the 1.6.14 update as well. So perhaps just trying to do all the error wrapping in a once block in the template was causing issues. 

Probably a compiler limitation but I figure no reason to push it there. Plus this makes it easier to call.

PR #504, opened 2023-08-02
Merkle tree construction

tbekas:


codex/merkletree/merkletree.nim
```diff
@@ -0,0 +1,111 @@
...
+import pkg/libp2p
+import pkg/stew/byteutils
+
+import pkg/questionable
+import pkg/questionable/results
+
+
+type
+  MerkleHash* = MultiHash
+  MerkleTree* = ref object of RootObj
```
elcritch, line 15: Does this need to be a type of `RootObj`? Not sure we'll need methods for this.

codex/merkletree/merkletree.nim
```diff
@@ -0,0 +1,111 @@
...
+    mcodec: ?MultiCodec
+    leaves: seq[MerkleHash]
+  MerkleProof* = ref object of RootObj
+    path: seq[MerkleHash]
+
+proc `$`*(t: MerkleTree): string =
+  result &= "height: " & $t.nodes.len
+  result &= "\nnodes: " & $t.nodes
+
+proc addLeaf*(self: MerkleTreeBuilder, hash: MerkleHash): ?!void =
```
elcritch, line 27: This could just be called `add`. It's pretty common in Nim to overload common operations that have similar semantics like `len`.

codex/merkletree/merkletree.nim
```diff
@@ -0,0 +1,111 @@
...
+
+  # copy leaves
+  nodes[0] = newSeq[MerkleHash](self.leaves.len)
+  for j in 0..<self.leaves.len:
+    nodes[0][j] = self.leaves[j]
+
+  # calculate internal nodes
+  for i in 1..<height:
+    let levelWidth = (nodes[i-1].len + 1) div 2
+    nodes[i] = newSeq[MerkleHash](levelWidth)
```
elcritch, line 68: Since this is a balanced merkle tree could we just use a flat `seq[MerkleHash]` and do indexing children via powers of 2? It's a bit more work here, but a flat `seq` is easy to serialize and copy. 

I guess it'd make add dynamic leaves a bit more expensive though?

tests/codex/merkletree/testmerkletree.nim
```diff
@@ -0,0 +1,195 @@
+import std/unittest
+import std/bitops
+import std/random
+import std/sequtils
+import pkg/libp2p
+import codex/merkletree/merkletree
+import ../helpers
+import pkg/questionable/results
+
+checksuite "merkletree":
```
elcritch, line 10: Great looking test suite!

codex/merkletree/merkletree.nim
```diff
@@ -0,0 +1,111 @@
...
+  self.leaves.add(hash)
+  return success()
+
+func calcTreeHeight(leavesCount: int): int =
+  if isPowerOfTwo(leavesCount): 
+    fastLog2(leavesCount) + 1
+  else:
+    fastLog2(leavesCount) + 2
+
+proc build*(self: MerkleTreeBuilder): ?!MerkleTree =
```
elcritch, line 43: Definitely bike shedding here, but the whole `mybuilder.build()` always seemed redundant to me :sweat_smile: Now `mybuilder.toTree()` :sunglasses:

```suggestion
proc toTree*(self: MerkleTreeBuilder): ?!MerkleTree =
```

Ok I'll stop bikes shedding for now. ;)

codex/merkletree/merkletree.nim
tbekas: I removed the builder altogether and replaced it with a single proc `newTree`

codex/merkletree/merkletree.nim
tbekas: It's not necessarily a balanced tree, since not all subtrees are balanced. We could still make it with a 1D array but it makes indexing a bit more complicated though. For now I would stay with 2D array and change it to 1D once it necessary (probably for serialization).

codex/merkletree/merkletree.nim
tbekas: Builder got replaced with a single `newTree` proc.

codex/merkletree/merkletree.nim
dryajov: Yes, I agree, as we discussed in the initial call, this should be a flat array for the reasons that @elcritch points out.

codex/merkletree/merkletree.nim
dryajov: Not unless we're planing to extend this, which we are not. We want this to just be an object.

```nim
  MerkleTree* = object
```

codex/merkletree/merkletree.nim
dryajov: I really recommend taking a look at the [status style guide](https://status-im.github.io/nim-style-guide/). 

The tl;dr is that you want to use `new` to construct `ref` types and `init` to construct objects, since we want to make `MerkleTree` an object instead of a reference, we should be using `init`.

codex/merkletree/merkletree.nim
```diff
@@ -0,0 +1,128 @@
...
+import pkg/questionable/results
+
+
+type
+  MerkleHash* = MultiHash
+  MerkleTree* = ref object of RootObj
+    nodes: seq[seq[MerkleHash]]
+  MerkleProof* = ref object of RootObj
+    index*: int
+    path*: seq[MerkleHash]
```
benbierens, line 19: Might it be wise to make the MerkleProof index and path read-only? You know: A constructor to set them, and then read-only procs to read them.

benbierens (approved):
We can rap-battle some more about the 1D array index stuff, if you'd like. (Or we can delay it till we get to serialization where we need to do it anyway.)

codex/merkletree/merkletree.nim
tbekas: Thanks for the info guys, changing it to `object`.

codex/merkletree/merkletree.nim
tbekas: Ok, switched to 1D array.

codex/merkletree/merkletree.nim
tbekas: Thanks for the info. I used `init` as suggested.

codex/merkletree/merkletree.nim
tbekas: Makes sense, adding accessor methods 👍

codex/merkletree/merkletree.nim
```diff
@@ -0,0 +1,164 @@
...
+    tree.setNode(0, j, leaves[j])
+
+  # calculate intermediate nodes
+  for i in 1..<tree.height:
+    for j in 0..<tree.getWidth(i):
+      let (left, right) = tree.getChildren(i, j)
+      
+      without mhash =? combine(left, right), error:
+        return failure(error)
+      tree.setNode(i, j, mhash)
```
benbierens, line 126: I like how there are many short functions. Very nice! But, there is quite a bit of `i` and `j` going on in these functions as well. Perhaps some of them can be renamed? index, level, or something else more descriptive?

benbierens (approved):

codex/merkletree/merkletree.nim
```diff
@@ -0,0 +1,164 @@
...
+    return failure("Index " & $index & " out of range [0.." & $self.leaves.high & "]" )
+
+  var path = newSeq[MerkleHash](self.height - 1)
+  for i in 0..<path.len:
+    let j = index div (1 shl i)
+    path[i] = self.getSibling(i, j)
+
+  success(MerkleProof(index: index, path: path))
+
+proc initTreeFromLeaves(leaves: seq[MerkleHash]): ?!MerkleTree =
```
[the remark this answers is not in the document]
dryajov: Ah, I see, this is just a helper

elcritch:
> We can rap-battle some more about the 1D array index stuff, if you'd like. (Or we can delay it till we get to serialization where we need to do it anyway.)

But @benbierens it's like `log n` less allocations!! ;) You can tell my embedded background shows -- I'm allergic to allocations.

codex/merkletree/merkletree.nim
tbekas: What do you think about this naming scheme: `level, i` instead of `i, j`? We would interpret that as an `i'th` node on a level `level`

codex/merkletree/merkletree.nim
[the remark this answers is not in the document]
tbekas: Yes it's a helper. I can inline it if you want. I just didn't want to make the `init*` proc too long.

benbierens:
Oh yeah, no doubt the 1D thing is much better in this way. It's just, we were just starting and needed a way to think about what we were trying to do. So we started with lots of allocs.

codex/merkletree/merkletree.nim
tbekas: We had an offline conversation, and `level, i` seem to be good enough naming scheme.

gmega (commented):
I have what are mostly petty comments. :slightly_smiling_face: Overall it looks good, particularly given that it passes the tests.

codex/merkletree/merkletree.nim
```diff
@@ -0,0 +1,184 @@
...
+
+proc height*(self: MerkleTree): int =
+  calcTreeHeight(self.leavesCount)
+
+proc `$`*(self: MerkleTree): string =
+  result &= "leavesCount: " & $self.leavesCount
+  result &= "\nnodes: " & $self.nodes
+
+proc getProof*(self: MerkleTree, index: int): ?!MerkleProof =
+  if index > self.leaves.high or index < 0:
```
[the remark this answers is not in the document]
tbekas: We can, thanks for catching it. It's kind of a leftover from previous 2D seq implementation.

codex/merkletree/merkletree.nim
```diff
@@ -0,0 +1,184 @@
...
+  for level in 0..<path.len:
+    let i = index div (1 shl level)
+    path[level] = self.getSibling(level, i)
+
+  success(MerkleProof(index: index, path: path))
+
+proc initTreeFromLeaves(leaves: openArray[MerkleHash]): ?!MerkleTree =
+  without mcodec =? leaves.?[0].?mcodec and
+          digestSize =? leaves.?[0].?size:
+    return failure("At least one leaf is required")
```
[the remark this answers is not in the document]
tbekas: Like this?
```nim
if leaves.len == 0:
  return failure("At least one leaf is required")
  
let mcodec = leaves[0].mcodec
let digestSize = leaves[0].size
```

We can, but the syntax I used is overall shorter and in my opinion less error prone to accidental bugs when refactoring since returning `failure(...)` is directly connected to failure in obtaining a value.

codex/merkletree/merkletree.nim
```diff
@@ -0,0 +1,184 @@
...
+
+  if not leaves.allIt(it.mcodec == mcodec):
```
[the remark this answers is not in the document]
tbekas: Yeah, initially I started implementing it this way, but the new code has to be compatible with already existing code and it's just easier to do it this way than converting to/from `MultiHash` in each public function. For the serialization though we will likely use a tree property instead of a node property.

codex/merkletree/merkletree.nim
```diff
@@ -0,0 +1,184 @@
...
+    return failure("All leaves must use the same codec")
+
+  let totalSize = getTotalSize(leaves.len)
```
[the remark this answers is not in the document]
tbekas: Makes sense, I even called it initially `calcTotalSize` but for whatever the reason I renamed it to `getTotalSize`. I like `computeTotalSize` more though, changing it 👍

codex/merkletree/merkletree.nim
```diff
@@ -0,0 +1,184 @@
...
+#
+# Proof for B is [H0, H4]
+
+func calcTreeHeight(leavesCount: int): int =
+  if isPowerOfTwo(leavesCount): 
+    fastLog2(leavesCount) + 1
+  else:
+    fastLog2(leavesCount) + 2
+
+func getLowHigh(leavesCount, level: int): (int, int) =
```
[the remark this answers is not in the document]
gmega: Ahah, oh, nim...

codex/merkletree/merkletree.nim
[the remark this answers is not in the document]
gmega: Got it

gmega (approved):
Looks good to me!

elcritch (approved):

elcritch (approved):

elcritch:
Ok, created an issue to follow up on the pre-image attack thing to get this unblocked: https://github.com/codex-storage/nim-codex/issues/511

codex/merkletree/merkletree.nim
```diff
@@ -0,0 +1,189 @@
...
+type
+  MerkleHash* = MultiHash
+  MerkleTree* = object
+    leavesCount: int
+    nodes: seq[MerkleHash]
+  MerkleProof* = object
+    index: int
+    path: seq[MerkleHash]
+
+# Tree constructed from leaves H0..H2 is
```
dryajov, line 29: Thanks for documenting this!

dryajov (approved):
OK, I think this is looking good for a first pass. 

We're missing proof validation (or am I blind?) and and tree validation when constructed with a merkle root. We can cover that in a separate PR tho.

We should also address the security concernes, but I'd leave it for a separate PR as well.

codex/merkletree/merkletree.nim
elcritch: Thanks! Note if you use `##` here then it'd show up if we ever run docgen. :)

benbierens (approved):

tbekas:
I'm going to address all of the comments in the upcoming PR.

tests/codex/merkletree/testmerkletree.nim
```diff
@@ -0,0 +1,108 @@
...
+    let tree = MerkleTree.init(leaves[0..1]).tryGet()
+
+    check:
+      tree.leaves == leaves[0..1]
+      tree.len == 3
+      tree.root == expectedRoot
+
+  test "tree with three leaves has expected root":
+    let
+      expectedRoot = combine(combine(leaves[0], leaves[1]), combine(leaves[2], leaves[2]))
```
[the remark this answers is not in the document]
tbekas: Fixed in the #516

codex/merkletree/merkletree.nim
```diff
@@ -0,0 +1,164 @@
...
+): MerkleTree =
+  let totalSize = getTotalSize(leavesCount)
+  var nodes = newSeq[MerkleHash](totalSize)
+  nodes[^1] = root
+  MerkleTree(nodes: nodes, leavesCount: leavesCount)
+
+proc init*(
+  T: type MerkleTree,
+  leaves: seq[MerkleHash]
+): ?!MerkleTree =
```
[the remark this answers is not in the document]
tbekas: Fixed in the #516

codex/merkletree/merkletree.nim
```diff
@@ -0,0 +1,164 @@
...
+  result &= "\npath: " & $self.path
+
+func `==`*(a, b: MerkleProof): bool =
+  (a.index == b.index) and (a.path == b.path)
+
+proc init*(
+  T: type MerkleProof,
+  index: int,
+  path: seq[MerkleHash]
+): MerkleProof =
```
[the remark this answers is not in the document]
tbekas: Fixed in the #516

codex/merkletree/merkletree.nim
```diff
@@ -0,0 +1,189 @@
+## Nim-Codex
+## Copyright (c) 2022 Status Research & Development GmbH
```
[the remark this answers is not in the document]
tbekas: Fixed in #516

codex/merkletree/merkletree.nim
tbekas: Fixed in #516

codex/merkletree/merkletree.nim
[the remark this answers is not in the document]
tbekas: I've address this by hardcoding a hashing algorithm for merkle trees tho SHA-256 in #516

PR #507, opened 2023-08-05
chore: move the docker-compose volume

vpavlin:
I was playing around with Codex and noticed the `datadir` is created in root (i.e. `/`) of my disk, which is not great.

This PR moves the `datadir` to `docker/` folder when you run something like

```
docker-compose -f docker/docker-compose.yaml up
```

I am also adding `:z` which is useful when you use SELinux to properly label the volume  (https://prefetch.net/blog/2017/09/30/using-docker-volumes-on-selinux-enabled-servers/)

### DIFF
--- a/docker/docker-compose.yaml
+++ b/docker/docker-compose.yaml
@@ -35,7 +35,7 @@ services:
       - 2345:2345/tcp # libp2p
       - 8090:8090/udp # DHT discovery
     volumes:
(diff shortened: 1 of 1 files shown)


dryajov:
@benbierens looks like a good change and I'd like to merge it, unless this will mess your setup?

benbierens (approved):

PR #509, opened 2023-08-08
Debug/extra logging

benbierens:


emizzle (approved):
Nice work.

Would it not be better to store the `BlockReq.startTime` as a `MonoTime` instead of `int64` (for operations such as `-` which is used later on)?

PR #510, opened 2023-08-08
fix: load slots on sales module start

### LINKED ISSUE #294: Add node restart

### LINKED ISSUE #505: Load Slots on Node Start
Recover the state from blockchain upon node start

AuHau:
Closes #294  
Closes #505

### DIFF
--- a/tests/codex/sales/testsales.nim
+++ b/tests/codex/sales/testsales.nim
@@ -73,21 +73,21 @@ asyncchecksuite "Sales - start":
     await sales.stop()
     await repo.stop()
 
+  proc fillSlot(slotIdx: UInt256 = 0.u256) {.async.} =
+    let address = await market.getSigner()
+    let slot = MockSlot(requestId: request.id,
+                        slotIndex: slotIdx,
+                        proof: proof,
+                        host: address)
+    market.filled.add slot
+    market.slotState[slotId(request.id, slotIdx)] = SlotState.Filled
+
   test "load slots when Sales module starts":
     let me = await market.getSigner()
 
     request.ask.slots = 2
     market.requested = @[request]
     market.requestState[request.id] = RequestState.New
-
-    proc fillSlot(slotIdx: UInt256 = 0.u256) {.async.} =
-      let address = await market.getSigner()
-      let slot = MockSlot(requestId: request.id,
-                          slotIndex: slotIdx,
-                          proof: proof,
-                          host: address)
-      market.filled.add slot
-      market.slotState[slotId(request.id, slotIdx)] = SlotState.Filled
 
     let slot0 = MockSlot(requestId: request.id,
                      slotIndex: 0.u256,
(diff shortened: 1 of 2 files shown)


tests/codex/sales/testsales.nim
```diff
@@ -22,6 +22,104 @@ import ../helpers/eventually
...
+    # randomly selected for the agent, and we also won't know the value of
+    # `failed`/`fulfilled`/`cancelled` futures, so we need to compare
+    # the properties we know
+    # TODO: when calling sales.load(), slot index should be restored and not
+    # randomly re-assigned, so this may no longer be needed
+    proc `==` (data0, data1: SalesData): bool =
+      return data0.requestId == data1.requestId and
+             data0.request == data1.request
+
+    check eventuallyCheck sales.agents.len == 2
```
emizzle, line 120: I think we can use `eventually` (instead of `eventuallyCheck`) from `asynctest`, no?

tests/codex/sales/testsales.nim
```diff
@@ -22,6 +22,104 @@ import ../helpers/eventually
...
+    let expected = SalesData(requestId: request.id, request: some request)
+    # because sales.load() calls agent.start, we won't know the slotIndex
+    # randomly selected for the agent, and we also won't know the value of
+    # `failed`/`fulfilled`/`cancelled` futures, so we need to compare
+    # the properties we know
+    # TODO: when calling sales.load(), slot index should be restored and not
+    # randomly re-assigned, so this may no longer be needed
+    proc `==` (data0, data1: SalesData): bool =
+      return data0.requestId == data1.requestId and
+             data0.request == data1.request
```
emizzle, line 118: Slot index and the state should now be restored, so we don't need this comment nor do I believe we need the `==`, as slot index should be compared.

emizzle (approved):
Nice work Adam! I'll approve now, but I think we should probably remove the `==`.

tests/codex/sales/testsales.nim
AuHau: I saw your message in Discord, make sense 👍

tests/codex/sales/testsales.nim
AuHau: Okey, I was not really analyzing this part, simply duplicated the test setup from bellow. I will modify it there as well then.

tests/codex/sales/testsales.nim
AuHau: Hmm, it seems there are some differences between `eventuallyCheck` and `eventually`. 😳 The tests in this file that still use `eventuallyCheck` started to fail when moved to `eventually`. 😳

I will merge it in its current form, and once we decide to make the full switch in the codebase, we should take a look at this.

tests/codex/sales/testsales.nim
emizzle: Thank you 🙏

PR #513, opened 2023-08-14
[repostore] Retrieve empty blocks

emizzle:
Empty blocks were not retrievable by the RepoStore, and this was causing integration tests to fail when K > 1, as empty blocks were used for padding protected manifests. The failures, however, were only being witnessed on ARM64 (Apple silicon M2's and likely M1's, though untested), and not on AMD64, for reasons that are still unknown. @dryajov was looking into checking that the erasure coding was indeed being done correctly.

### DIFF
--- a/codex/stores/repostore.nim
+++ b/codex/stores/repostore.nim
@@ -136,6 +136,10 @@ method putBlock*(
   ## Put a block to the blockstore
   ##
 
+  if blk.isEmpty:
+    trace "Empty block, ignoring"
+    return success()
+
   without key =? makePrefixKey(self.postFixLen, blk.cid), err:
     trace "Error getting key from provider", err = err.msg
     return failure(err)
@@ -209,6 +213,10 @@ method delBlock*(self: RepoStore, cid: Cid): Future[?!void] {.async.} =
 
   trace "Deleting block", cid
 
+  if cid.isEmpty:
+    trace "Empty block, ignoring"
+    return success()
+
   if blk =? (await self.getBlock(cid)):
     if key =? makePrefixKey(self.postFixLen, cid) and
       err =? (await self.repoDs.delete(key)).errorOption:
@@ -236,6 +244,10 @@ method delBlock*(self: RepoStore, cid: Cid): Future[?!void] {.async.} =
 method hasBlock*(self: RepoStore, cid: Cid): Future[?!bool] {.async.} =
   ## Check if the block exists in the blockstore
   ##
+
+  if cid.isEmpty:
+    trace "Empty block, ignoring"
+    return true.success
 
   without key =? makePrefixKey(self.postFixLen, cid), err:
     trace "Error getting key from provider", err = err.msg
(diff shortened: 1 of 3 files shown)


codex/stores/repostore.nim
```diff
@@ -85,6 +85,10 @@ method getBlock*(self: RepoStore, cid: Cid): Future[?!Block] {.async.} =
   ## Get a block from the blockstore
   ##
 
+  if cid.isEmpty:
```
dryajov, line 88: We want this checks for all the blockstore methods, get/put/del/etc...

emizzle: Updated.

emizzle:
Once I added the empty block tests, which calls `hasBlock`, there was an ambiguous call:
```
Error: ambiguous call; both repostore.hasBlock(self: RepoStore, cid: Cid) [proc declared in /Users/egonat/repos/status-im/nim-codex/codex/stores/repostore.nim(339, 6)] and repostore.hasBlock(self: RepoStore, cid: Cid) [method declared in /Users/egonat/repos/status-im/nim-codex/codex/stores/repostore.nim(244, 8)] match for: (RepoStore, Cid)
```
It appears that `hasBlock` was duplicated in `repostore` but was not being called anywhere else in the tests, or maybe even the codebase, so adding the `hasBlock` test for empty blocks was a good thing.

dryajov (approved):

veaceslavdoina:
@emizzle, in case you need in the future other ARM64/AMD64, we can continue our discussion here - https://github.com/codex-storage/infra-codex/issues/14.

dryajov:
For context, re erasure coding issue on M2 mac.

The issue is that, currently we don't support erasure coding a single block, there has to be at least two blocks to be able to erasure code them. 

Leopard will not encode a single buffer, it will simply copy it back. Blocks are treated as a list of symbols and each block occupies one buffer. So, for all intents and purposes, when using one block, we're sending a vector of L*K elements, but K=1 in this case, so leopard simply copies it back. 

The options are:

Split the block into two and erasure code the two pieces together
Erasure code with a  dummy block
???

In short, it's not a big deal, its a corner case that is probably not worth addressing right now, and when we need to address it, the fix should be relatively straightforward.

For your tests to work, just generate at least two random blocks, other than that it should works as expected.

dryajov:
btw, is there an issue to track this already, I couldn't find it.

emizzle:
> Erasure code with a dummy block

@dryajov, isn't the K=1 case already being padded with an empty block, hence the issue arising where empty blocks were not retrievable from the RepoStore? Or by "dummy" block, do you mean something different than an empty block?

The issue that seemed odd was that AMD CPUs seemed not to pad with an empty block (and therefore did not not run into issues retrieving empty blocks), whereas this issue was present on ARM silicon.

dryajov:
> Or by "dummy" block, do you mean something different than an empty block?

Other than empty block. Yeah, I'm guessing it was a combination of several factors, but I've been able to reproduce it with a single block, when there are more than one blocks, K=2, it works as expected, at least with my experiments.

emizzle:
Issue created: https://github.com/codex-storage/nim-codex/issues/522

PR #514, opened 2023-08-14
Improve integration testing client (CodexClient) and json serialization

emizzle:
The current client used for integration testing against the REST endpoints for Codex accepts and passes primitive types. This caused a hard to diagnose bug where a `uint` was not being deserialized correctly.

In addition, the json de/serializing done between the CodexClient and REST client was not easy to read and was not tested.

These changes bring non-primitive types to most of the CodexClient functions, allowing us to lean on the compiler to ensure we're providing correct typings when doing integration testing. More importantly, a json de/serialization util was created as a drop-in replacement for the std/json lib, with the main two differences being that field serialization is opt-in (instead of opt-out as in the case of json_serialization) and serialization errors are captured and logged, making debugging serialization issues much easier.

### Note
~~This PR is based on #513~~

### DIFF
--- a/codex/utils/json.nim
+++ b/codex/utils/json.nim
@@ -1,0 +1,339 @@
+
+import std/json except `%`, `%*`
+import std/macros
+import std/options
+import std/strutils
+import std/strformat
+import std/tables
+import std/typetraits
+import pkg/chronicles
+from pkg/libp2p import Cid, init
+import pkg/contractabi
+import pkg/stew/byteutils
+import pkg/stint
+import pkg/questionable/results
+import ../errors
+
+export json except `%`, `%*`
+
+logScope:
+  topics = "json serialization"
+
+type
+  SerializationError = object of CodexError
+  UnexpectedKindError = object of SerializationError
+
+template serialize* {.pragma.}
+
+proc newUnexpectedKindError(
+  expectedType: type,
+  expectedKinds: string,
+  json: JsonNode
+): ref UnexpectedKindError =
+  let kind = if json.isNil: "nil"
+             else: $json.kind
+  newException(UnexpectedKindError,
+    &"deserialization to {$expectedType} failed: expected {expectedKinds} " &
+    &"but got {kind}")
+
+proc newUnexpectedKindError(
+  expectedType: type,
+  expectedKinds: set[JsonNodeKind],
+  json: JsonNode
+): ref UnexpectedKindError =
+  newUnexpectedKindError(expectedType, $expectedKinds, json)
+
+proc newUnexpectedKindError(
+  expectedType: type,
+  expectedKind: JsonNodeKind,
+  json: JsonNode
+): ref UnexpectedKindError =
+  newUnexpectedKindError(expectedType, {expectedKind}, json)
+
+template expectJsonKind(
+  expectedType: type,
+  expectedKinds: set[JsonNodeKind],
+  json: JsonNode
+) =
+  if json.isNil or json.kind notin expectedKinds:
+    return failure(newUnexpectedKindError(expectedType, expectedKinds, json))
+
+template expectJsonKind(
+  expectedType: type,
+  expectedKind: JsonNodeKind,
+  json: JsonNode
+) =
+  expectJsonKind(expectedType, {expectedKind}, json)
+
+proc fromJson*(
+  _: type string,
+  json: JsonNode
+): ?!string =
+  if json.isNil:
+    let err = newException(ValueError, "'json' expected, but was nil")
+    return failure(err)
+  elif json.kind == JNull:
+    return success("null")
+  elif json.isNil or json.kind != JString:
+    return failure(newUnexpectedKindError(string, JString, json))
+  catch json.getStr
+
+proc fromJson*(
+  _: type bool,
+  json: JsonNode
+): ?!bool =
+  expectJsonKind(bool, JBool, json)
+  catch json.getBool
+
+proc fromJson*(
+  _: type int,
+  json: JsonNode
+): ?!int =
+  expectJsonKind(int, JInt, json)
+  catch json.getInt
+
+proc fromJson*[T: SomeInteger](
+  _: type T,
+  json: JsonNode
+): ?!T =
+  when T is uint|uint64 or (not defined(js) and int.sizeof == 4):
+    expectJsonKind(T, {JInt, JString}, json)
+    case json.kind
+    of JString:
+      let x = parseBiggestUInt(json.str)
+      return success cast[T](x)
+    else:
+      return success T(json.num)
+  else:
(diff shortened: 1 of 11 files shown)


codex/utils/json.nim
```diff
@@ -0,0 +1,339 @@
...
+    return failure(newUnexpectedKindError(expectedType, expectedKinds, json))
+
+template expectJsonKind(
+  expectedType: type,
+  expectedKind: JsonNodeKind,
+  json: JsonNode
+) =
+  expectJsonKind(expectedType, {expectedKind}, json)
+
+proc fromJson*(
```
elcritch, line 68: Why not use [nim-json-serialization](https://github.com/status-im/nim-json-serialization)?

emizzle: It was my first choice so that we could keep consistency across the code base, but opted out of it for a couple of reasons `nim-json-serialization` is so hard to debug that I consistently spend far too much time trying to implement anything. The errors almost always get swallowed and the application crashes in inconspicuous and hard-to-debug ways. It also doesn't help that most of its guts are written in macros, so even attempting to quickly output the serialization error just to see what it was, takes ages to figure out where that is even possible (and even then it's just not that easy). Having a serialization utility inside of codex (or could even be its own lib) that is not overtly magical and not written entirely in macros allows us the ability to debug, to not swallow errors, and to do things like having opt-in field serialization (versus opt-out like nim-json-serialization) was generally the reasoning behind not using `nim-json-serialization`. More discussion around these topics can be found here: https://discord.com/channels/613988663034118151/616299964242460682/789378950484852756.

Additionally, for the life of me, I could not override serialisation of `UInt256` (so it was not hex-encoded) using `nim-json-serialization` inside of `codex/utils/json` or `codex/rest/json`. I could override deserialization (`readValue`) but not serialization (`writeValue`), which I really think may boil down to a dependency issue. I'd like to maybe set some time up with you to see if we can work through that, as it was really not obvious and honestly quite frustrating.

elcritch: Thanks! That's really helpful to get context about.

Ah, yah I can see that. I ran into some  of the difficulty in debugging `nim-json-serialization` and the TOML one in the `conf.nim` module, and that was a very basic change. There seemed to be a lot of complicated magic and took me a while to debug what was even causing the issue, which was a pain.

> Additionally, for the life of me, I could not override serialisation of UInt256 (so it was not hex-encoded) using nim-json-serialization inside of codex/utils/json or codex/rest/json. I could override deserialization (readValue) but not serialization (writeValue), which I really think may boil down to a dependency issue.

That's a pain. 🤔 I thought I saw some UInt256 serialization stuff in `stint` or somewhere. That could explain some of those issues.

> I'd like to maybe set some time up with you to see if we can work through that, as it was really not obvious and honestly quite frustrating.

Sure! I'd be up for that sometime and like figuring out puzzles. It'd be nice to chat with you too, I don't think we've had the chance to. :)

I'll take a closer look at your setup as well to see if I can spot anything. It looked pretty straightfoward.

P.S. My favorite serde library in Nim is [msgpack4nim](https://github.com/jangko/msgpack4nim). It can serialize directly into or from an object making. This avoids allocations making it near the fastest one I've benchmarked. It rivaled `flatbuffers` which is famed for it's no-copy speed!

emizzle:
> I thought I saw some UInt256 serialization stuff in stint or somewhere. That could explain some of those issues

It could have been https://github.com/status-im/nim-eth/blob/master/eth/common/eth_types_json_serialization.nim#L26-L30. Importing these was definitely something I tried, but could not get working.

> Sure! I'd be up for that sometime and like figuring out puzzles. It'd be nice to chat with you too, I don't think we've had the chance to. :)

Let's do it 💪  DM me in discord and we can setup a date/time.

> P.S. My favorite serde library in Nim is [msgpack4nim](https://github.com/jangko/msgpack4nim). It can serialize directly into or from an object making. This avoids allocations making it near the fastest one I've benchmarked. It rivaled flatbuffers which is famed for it's no-copy speed!

msgpack4nim looks pretty cool! After looking at all the places that we use serialization in codex, i do think a universal  solution should be implemented at some point. The issue that I see right now is that [`nim-chronicles` uses `nim-json-serialization`](https://github.com/status-im/nim-chronicles#teaching-chronicles-about-your-types) due to [our use of the json sink](https://github.com/codex-storage/nim-codex/blob/master/config.nims#L114). I don't think it's possible to use a different serializer for a `json` sink, unfortunately. We could possibly remove the json sink (using `formatIt` instead), but I imagine we'll probably need to use the json log output for consumption in eg ElasticSearch or Loggly at some point.

codex/utils/json.nim
emizzle: > I thought I saw some UInt256 serialization stuff in stint or somewhere. That could explain some of those issues

It could have been https://github.com/status-im/nim-eth/blob/master/eth/common/eth_types_json_serialization.nim#L26-L30. Importing these was definitely something I tried, but could not get working.

> Sure! I'd be up for that sometime and like figuring out puzzles. It'd be nice to chat with you too, I don't think we've had the chance to. :)

Let's do it 💪  DM me in discord and we can setup a date/time.

> P.S. My favorite serde library in Nim is [msgpack4nim](https://github.com/jangko/msgpack4nim). It can serialize directly into or from an object making. This avoids allocations making it near the fastest one I've benchmarked. It rivaled flatbuffers which is famed for it's no-copy speed!

msgpack4nim looks pretty cool! After looking at all the places that we use serialization in codex, i do think a universal  solution should be implemented at some point. The issue that I see right now is that [`nim-chronicles` uses `nim-json-serialization`](https://github.com/status-im/nim-chronicles#teaching-chronicles-about-your-types) due to [our use of the json sink](https://github.com/codex-storage/nim-codex/blob/master/config.nims#L114). I don't think it's possible to use a different serializer for a `json` sink, unfortunately. We could possibly remove the json sink (using `formatIt` instead), but I imagine we'll probably need to use the json log output for consumption in eg ElasticSearch or Loggly at some point.

tests/integration/testIntegration.nim
```diff
@@ -9,11 +10,16 @@ import ../contracts/deployment
 import ../codex/helpers/eventually
 import ./twonodes
 
+
 # For debugging you can enable logging output with debugX = true
 # You can also pass a string in same format like for the `--log-level` parameter
 # to enable custom logging levels for specific topics like: debug2 = "INFO; TRACE: marketplace"
 
-twonodessuite "Integration tests", debug1 = false, debug2 = false:
+twonodessuite "Integration tests", debug1 = "INFO;TRACE:node,restapi,erasure", debug2 = false:
```
AuHau, line 18: I guess we can turn off the `debug1` logging before merging?

AuHau (dismissed):
LGTM! Some small nitpicks that are non-blocking 👍

PR #516, opened 2023-08-15
Refine merkle tree construction

### LINKED ISSUE #511: Upgrade Merkle Tree to avoid second-preimage attacks
Balancing the merkle tree when there's an odd number of leaves by duplicating the last node can lead to pre-image attacks.

See:

- [Merkle tree - Second preimage attack](https://en.wikipedia.org/wiki/Merkle_tree#Second_preimage_attack)
- [Bitcoin pre-image attack disclosure](https://bitcointalk.org/?topic=102395)
- [CVE-2012-2459](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2012-2459)
- [proposed fix for Bitcoin Merkle Tree CVE](https://bips.dev/98/)

TLDR: The original Bitcoin merkle trees used a duplicate node at the end of merkle tree to balance the tree when there were an even number of transactions. As I understand it, this led to the possibility of crafting a special block of transactions which would fail to validate, but would be cached. This effectively caused the bitcoin node to get stuck and DDOS the network.

It may all not apply to us, but I think prudence would say to use the fixes others did in case. Seems there's a few ways to solve it:

- record the length of the tree
- use a special empty node
- prefix hashes of leaf and internal nodes differently

tbekas:
Resolves #511

### DIFF
--- a/codex/merkletree/merkletree.nim
+++ b/codex/merkletree/merkletree.nim
@@ -85,7 +85,7 @@ proc build*(self: MerkleTreeBuilder): ?!MerkleTree =
   ## |       |       |
   ## A       B       C
   ##
-  ## where HZ=H(0x0b)
+  ## where HZ=0x0b
   ##
   ## Memory layout is [H0, H1, H2, H3, H4, H5]
   ##
@@ -159,7 +159,7 @@ proc getProof*(self: MerkleTree, index: Natural): ?!MerkleProof =
   ## - 1,[H0, H4] for data block B
   ## - 2,[HZ, H3] for data block C
   ## 
-  ## where HZ=H(0x0b)
+  ## where HZ=0x0b
   ## 
   if index >= self.leavesCount:
     return failure("Index " & $index & " out of range [0.." & $self.leaves.high & "]" )
--- a/codex/merkletree/merkletree.nim
+++ b/codex/merkletree/merkletree.nim
@@ -9,23 +9,30 @@
 
 import std/math
 import std/bitops
-import std/strutils
+import std/sequtils
+import std/sugar
 
 import pkg/questionable/results
 import pkg/nimcrypto/sha2
+import pkg/libp2p/[multicodec, multihash, vbuffer]
 
-const digestSize = sha256.sizeDigest
+import ../errors
 
 type
-  MerkleHash* = array[digestSize, byte]
   MerkleTree* = object
+    mcodec: MultiCodec
+    digestSize: Natural
     leavesCount: Natural
-    nodes: seq[MerkleHash]
+    nodesBuffer: seq[byte]
   MerkleProof* = object
+    mcodec: MultiCodec
+    digestSize: Natural
     index: Natural
-    path: seq[MerkleHash]
+    nodesBuffer: seq[byte]
   MerkleTreeBuilder* = object
-    buffer: seq[MerkleHash]
+    mcodec: MultiCodec
+    digestSize: Natural
+    buffer: seq[byte]
 
 ###########################################################
 # Helper functions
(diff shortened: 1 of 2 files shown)


codex/merkletree/merkletree.nim
```diff
@@ -1,96 +1,136 @@
...
+
+proc digestFn(data: openArray[byte], output: var MerkleHash): void =
+  var digest = sha256.digest(data)
+  copyMem(addr output, addr digest.data[0], digestSize)
+
+###########################################################
+# MerkleHash
+###########################################################
+
+var zeroHash: MerkleHash
```
tbekas, line 58: Addressing the [comment](https://github.com/codex-storage/nim-codex/pull/504#discussion_r1287729437) from previous PR from @elcritch 

This is the `nullHash` used whenever there's a missing operand to a combine function (like odd number of leaves).

codex/merkletree/merkletree.nim
```diff
@@ -1,96 +1,136 @@
...
+  ##      /            \
+  ##    H3=H(H0 & H1)   H4=H(H2 & HZ)
+  ##   /    \          /
+  ## H0=H(A) H1=H(B) H2=H(C)
+  ## |       |       |
+  ## A       B       C
+  ##
+  ## where HZ=H(0x0b)
+  ##
+  ## Memory layout is [H0, H1, H2, H3, H4, H5]
```
tbekas, line 90: Adressing the [comment](https://github.com/codex-storage/nim-codex/pull/504#discussion_r1287730466) from previous PR from @elcritch 
 
I added some docs to individual methods instead of a module.

codex/merkletree/merkletree.nim
```diff
@@ -99,76 +139,56 @@ proc leaves*(self: MerkleTree): seq[MerkleHash] =
...
+  ##    H3       H4
+  ##   /  \     /
+  ## H0    H1 H2
+  ## |     |  |
+  ## A     B  C
+  ##
+  ## Proofs of inclusion (index and path) are
+  ## - 0,[H1, H4] for data block A
+  ## - 1,[H0, H4] for data block B
+  ## - 2,[HZ, H3] for data block C
```
tbekas, line 160: Addressing the [comment](https://github.com/codex-storage/nim-codex/pull/504#discussion_r1287730949) by @elcritch  from previous PR.

This is the docs I've added. Let me know if anything is unclear.

elcritch:
Also I created an issue for the pre-image attack bit: https://github.com/codex-storage/nim-codex/issues/511. You could probably close it with this PR?

tbekas:
> Also I created an issue for the pre-image attack bit: #511. You could probably close it with this PR?

I just marked this PR as resolving the #511

codex/merkletree/merkletree.nim
```diff
@@ -1,96 +1,136 @@
...
-import pkg/stew/byteutils
-import pkg/questionable
 import pkg/questionable/results
+import pkg/nimcrypto/sha2
+
+const digestSize = sha256.sizeDigest
 
 type
-  MerkleHash* = MultiHash
+  MerkleHash* = array[digestSize, byte]
```
dryajov, line 20: We still want to use `Multihash` for the interior nodes, we only want to allow generic byte streams for leaf nodes.

codex/merkletree/merkletree.nim
```diff
@@ -1,96 +1,136 @@
...
 
-func getLowHigh(leavesCount, level: int): (int, int) =
-  var width = leavesCount
-  var low = 0
-  for _ in 0..<level:
-    low += width
-    width = (width + 1) div 2
+func computeLevels(leavesCount: int): seq[tuple[offset: int, width: int]] =
+  let height = computeTreeHeight(leavesCount)
+  result = newSeq[tuple[offset: int, width: int]](height)
```
dryajov, line 42: We usually avoid using `result` due to a slew of pitfalls. More info here https://status-im.github.io/nim-style-guide/language.result.html

codex/merkletree/merkletree.nim
dryajov: This is probably not going to work with `Multihash`, a potentially better approach is to prepend with `0` for leaf nodes and `1` for interior nodes.

tbekas: Ok, I've also heard about this approach as it also solves one of potential vulnerabilities. I'm going to change it soon.

codex/merkletree/merkletree.nim
tbekas: Thanks for the info

dryajov:
From our sync discussion: 

- We want to store the global multihash codec for the interior nodes as a type field and only store the digest bytes for each node
- We want to additionally store the CID codec for the merkle root so we can continue using the existing repo as a storage backend
- We should probably rework the tree to be a full tree, which means that it has to contain an even number of nodes at each level
  - The overhead from that would be one additional hash per level, but it will greatly simplify the logic for calculating the number of nodes, height, offsets, etc. For reference look at the https://en.wikipedia.org/wiki/Binary_tree#Properties_of_binary_trees 
  - For serialization purposes, we don't need to store the duplicate/dummy nodes, so the overhead is mostly just in memory

codex/merkletree/merkletree.nim
tbekas: Fixed in the latest batch of changes

codex/merkletree/merkletree.nim
tbekas: Fixed in the latest batch of changes

codex/merkletree/merkletree.nim
```diff
@@ -1,96 +1,136 @@
...
+
+  result[0].offset = 0
+  result[0].width = leavesCount
+  for i in 1..<height:
+    result[i].offset = result[i - 1].offset + result[i - 1].width
+    result[i].width = (result[i - 1].width + 1) div 2
+
+proc digestFn(data: openArray[byte], output: var MerkleHash): void =
+  var digest = sha256.digest(data)
```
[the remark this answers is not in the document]
tbekas: Fixed in the latest batch of changes. Codec is now a property of a tree.

tbekas:
@dryajov Let me know if the current data model as well as accessors are sufficient. Internally tree is stored just a `seq[byte]` while on the API `MultiHash` is used (bytes are converted to `MultiHash`)
```nim
MerkleTree* = object
    mcodec: MultiCodec
    digestSize: Natural
    leavesCount: Natural
    nodesBuffer: seq[byte]
    
proc root*(self: MerkleTree): MultiHash

proc leaves*(self: MerkleTree): seq[MultiHash]
```

tbekas:
Moved to a different PR

PR #535, opened 2023-08-28
[marketplace] Availability improvements

### LINKED ISSUE #534: [marketplace] improve Availabilities
Closes #

## Problem
When Availabilities are created, the amount of bytes in the Availability are reserved in the repo, so those bytes on disk cannot be written to otherwise. When a request for storage is received by a node, if a previously created Availability is matched, an attempt will be made to fill a slot in the request (more accurately, the request's slots are added to the SlotQueue, and eventually those slots will be processed). During download, bytes that were reserved for the Availability were released (as they were written to disk). To prevent more bytes from being released than were reserved in the Availability, the Availability was marked as used during the download, so that no other requests would match the Availability, and therefore no new downloads (and byte releases) would begin. The unfortunate downside to this, is that the number of Availabilities a node has determines the download concurrency capacity. If, for example, a node creates a single Availability that covers all available disk space the operator is willing to use, that single Availability would mean that only one download could occur at a time, meaning the node could potentially miss out on storage opportunities.

## Solution
To alleviate the concurrency issue, each time a slot is processed, a Reservation is created, which takes size (aka reserved bytes) away from the Availability and stores them in the Reservation object. This can be done as many times as needed as long as there are enough bytes remaining in the Availability. Therefore, concurrent downloads are no longer limited by the number of Availabilities. Instead, they would more likely be limited to the SlotQueue's `maxWorkers`.

emizzle:
Fixes #534.

## Problem
When Availabilities are created, the amount of bytes in the Availability are reserved in the repo, so those bytes on disk cannot be written to otherwise. When a request for storage is received by a node, if a previously created Availability is matched, an attempt will be made to fill a slot in the request (more accurately, the request's slots are added to the SlotQueue, and eventually those slots will be processed). During download, bytes that were reserved for the Availability were released (as they were written to disk). To prevent more bytes from being released than were reserved in the Availability, the Availability was marked as used during the download, so that no other requests would match the Availability, and therefore no new downloads (and byte releases) would begin. The unfortunate downside to this, is that the number of Availabilities a node has determines the download concurrency capacity. If, for example, a node creates a single Availability that covers all available disk space the operator is willing to use, that single Availability would mean that only one download could occur at a time, meaning the node could potentially miss out on storage opportunities.

## Solution
To alleviate the concurrency issue, each time a slot is processed, a Reservation is created, which takes size (aka reserved bytes) away from the Availability and stores them in the Reservation object. This can be done as many times as needed as long as there are enough bytes remaining in the Availability. Therefore, concurrent downloads are no longer limited by the number of Availabilities. Instead, they would more likely be limited to the SlotQueue's `maxWorkers`.

From a database design perspective, an Availability has zero or more Reservations, and looks like this:
```
                                                      +--------------------------------------+
                                                      |            RESERVATION               |
+--------------------------------------+              |--------------------------------------|
|            AVAILABILITY              |              | ReservationId  | id             | PK |
|--------------------------------------|              |--------------------------------------|
| AvailabilityId | id            | PK  |<-||-------o<-| AvailabilityId | availabilityId | FK |
|--------------------------------------|              |--------------------------------------|
| UInt256        | size          |     |              | UInt256        | size           |    |
|--------------------------------------|              |--------------------------------------|
| UInt256        | duration      |     |              | SlotId         | slotId         |    |
|--------------------------------------|              +--------------------------------------+
| UInt256        | minPrice      |     |
|--------------------------------------|
| UInt256        | maxCollateral |     |
+--------------------------------------+
```

Reservations are persisted in the RepoStore's metadata, along with Availabilities. The metadata store key path for Reservations is ` meta / sales / reservations / <availabilityId> / <reservationId>`, while Availabilities are stored one level up, eg `meta / sales / reservations / <availabilityId> `, allowing all Reservations for an Availability to be queried (this is not currently needed, but may be useful when work to restore Availability size is implemented, more on this later).

### Lifecycle
When a reservation is created, its size is deducted from the Availability, and when a reservation is deleted, any remaining size (bytes not written to disk) is returned to the Availability. If the request finishes, is cancelled (expired), or an error occurs, the Reservation is deleted (and any undownloaded bytes returned to the Availability). In addition, when the Sales module starts, any Reservations that are not actively being used in a filled slot, are deleted.

Having a Reservation persisted until after a storage request is completed, will allow for the originally set Availability size to be reclaimed once a request contract has been completed. This is a feature that is yet to be implemented, however the work in this PR is a step in the direction towards enabling this.

### Unknowns
Reservation size is determined by the `StorageAsk.slotSize`. If during download, more bytes than `slotSize` are attempted to be downloaded than this, then the Reservation update will fail, and the state machine will move to a `SaleErrored` state, deleting the Reservation. This will likely prevent the slot from being filled.

### Notes
Based on #514


codex/sales/reservations.nim
```diff
@@ -217,144 +300,184 @@ proc reserve*(
...
 
-  var iter = AvailabilityIter()
+  var iter = StorableIter()
   let query = Query.init(ReservationsKey)
-
-  without results =? await self.repo.metaDs.query(query), err:
-    return failure(err)
-
-  proc next(): Future[?Availability] {.async.} =
+  when T is Availability:
```
markspanbroek, line 416: Wouldn't it be nicer to have two procs here, one for Availability, and one for Reservation? Now the compiler won't help you if you were to introduce another SomeStorableObject type later.

codex/sales/reservations.nim
```diff
@@ -217,144 +300,184 @@ proc reserve*(
...
     await idleAsync()
     iter.finished = results.finished
     if not results.finished and
-      r =? (await results.next()) and
-      serialized =? r.data and
-      serialized.len > 0:
+       res =? (await results.next()) and
+       res.data.len > 0 and
+       key =? res.key and
+       key.namespaces.len == defaultKey.namespaces.len:
```
markspanbroek, line 435:
```suggestion
      res =? (await results.next()) and
      res.data.len > 0 and
      key =? res.key and
      key.namespaces.len == defaultKey.namespaces.len:
```

indentation

codex/sales/reservations.nim
```diff
@@ -217,144 +300,184 @@ proc reserve*(
...
 
-  without availabilities =? (await r.availabilities), err:
-    return failure(err)
+  without storables =? (await self.storables(T)), error:
+    return failure(error)
 
-  for a in availabilities:
-    if availability =? (await a) and not availability.used:
-      ret.add availability
+  # NOTICE: there is a swallowed deserialization error
```
markspanbroek, line 454: Should we log it?

codex/sales/reservations.nim
```diff
@@ -217,144 +300,184 @@ proc reserve*(
...
 
-      if used == availability.used and
-        size <= availability.size and
-        duration <= availability.duration and
-        collateral <= availability.maxCollateral and
-        minPrice >= availability.minPrice:
+      if size <= availability.size and
+         duration <= availability.duration and
+         collateral <= availability.maxCollateral and
+         minPrice >= availability.minPrice:
```
markspanbroek, line 478:
```suggestion
        duration <= availability.duration and
        collateral <= availability.maxCollateral and
        minPrice >= availability.minPrice:
```

codex/sales/states/preparing.nim
```diff
@@ -50,20 +51,33 @@ method run*(state: SalePreparing, machine: Machine): Future[?State] {.async.} =
...
       request.ask.pricePerSlot,
-      request.ask.collateral,
-      used = false):
-    info "no availability found for request, ignoring",
-      slotSize = request.ask.slotSize,
-      duration = request.ask.duration,
-      pricePerSlot = request.ask.pricePerSlot,
-      used = false
+      request.ask.collateral):
+    info "no availability found for request, ignoring"
```
markspanbroek, line 68: Maybe log as 'debug', instead of 'info'? Otherwise a user is going to see a lot of these...

markspanbroek (approved):
I like these changes a lot, the codes cleans up nicely by moving to the reservations model instead of using availability.used.

codex/sales/reservations.nim
emizzle: I partially agree, but I decided against it simply because it is not a small proc and it would be a lot of repetition. I opted instead to include both `Availability` and `Reservation` type checks in the `when`, and added an `else` condition that will fail when the type is neither of those two. That way there will be compiler support if a new type is added to `SomeStorableObject`.

codex/sales/reservations.nim
emizzle: Surprisingly, I can't seem to log it because there is an odd compilation error, which I suspect has something to do with questionable. If I change this to:

```nim
for storable in storables.items:
    if bytes =? (await storable):
      without obj =? T.fromJson(bytes), e:
        error "deserialization error", error = e.msg
        continue
      ret.add obj
```
I get the compilation error:
```
/Users/egonat/repos/status-im/nim-codex/tests/codex/sales/testsales.nim(442, 41) template/generic instantiation of `all` from here
/Users/egonat/repos/status-im/nim-codex/codex/sales/reservations.nim(458, 15) template/generic instantiation of `without` from here
/Users/egonat/repos/status-im/nim-codex/vendor/questionable/questionable/withoutresult.nim(36, 3) template/generic instantiation of `without` from here
/Users/egonat/repos/status-im/nim-codex/codex/sales/reservations.nim(459, 15) template/generic instantiation of `error` from here
/Users/egonat/repos/status-im/nim-codex/vendor/nim-chronicles/chronicles.nim(363, 10) template/generic instantiation of `log` from here
/Users/egonat/repos/status-im/nim-codex/vendor/nim-chronicles/chronicles.nim(332, 21) Error: undeclared identifier: 'activeChroniclesStream'
```

This is extrememly odd because there are other uses of `without` and chronicles error logging in the same module and there are no compilation errors there.

Any ideas?

emizzle: For future reference: needed to export chronicles.

Due to the generic type parameter, the context at which this proc is instantiated and called (which is different than at compile time) does not contain chronicles in its scope unless exported by the current module.

codex/sales/reservations.nim
emizzle: Updated

emizzle:
This should be g2g mark, ready for another look.

AuHau (commented):
Pretty great work! I like the testing! Although I have a few comments, please have a look ;-)

codex/node.nim
```diff
@@ -284,7 +284,13 @@ proc requestStorage*(
   let request = StorageRequest(
     ask: StorageAsk(
       slots: nodes + tolerance,
-      slotSize: (encoded.blockSize.int * encoded.steps).u256,
+      # TODO: Specify slot-specific size (as below) once dispersal is
+      # implemented. The current implementation downloads the entire dataset, so
+      # it is currently set to be the size of the entire dataset. This is
+      # because the slotSize is used to determine the amount of bytes to reserve
+      # in a Reservations
```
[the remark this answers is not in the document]
emizzle: The slotSize will be determined by the erasure coding mechanism (that has yet to be finalised), which determines the number of steps required. Because the erasure coding mechanism isn't finalised, we are simply downloading the entire dataset, and we need to ensure that the slotSize accurately represents the *max* size of the data that will need to be downloaded, because this is the amount of storage that will need to be carved off of the Availability.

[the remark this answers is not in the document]
dryajov: The network level erasure coding is there and usable, however, we're lacking the dispersal part which will distribute chunks to nodes in a way that keeps them uncorrelated, this part is yet to be finalized. For now, we can stub the dispersal mechanism and just simple split the dataset into even rows and each node gets a row.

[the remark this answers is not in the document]
emizzle: Fair point. I've created an issue for this: https://github.com/codex-storage/nim-codex/issues/557

Downloading the entire dataset seems to be very stable in all the testing (integration, distributed), so it would be good to have this handled in a separate PR so we can isolate the changes.

codex/sales.nim
```diff
@@ -139,21 +194,23 @@ proc mySlots*(sales: Sales): Future[seq[Slot]] {.async.} =
...
-  for slot in slots:
+  for slot in activeSlots:
     let agent = newSalesAgent(
       sales.context,
       slot.request.id,
       slot.slotIndex,
       some slot.request)
 
-    agent.context.onCleanUp = proc {.async.} = await sales.remove(agent)
+    agent.onCleanUp = proc {.async.} = await sales.remove(agent)
```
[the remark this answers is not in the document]
emizzle: Good catch. I must've missed this because of the rebased changes (on top of your changes that modified at which point we complete the processing future).

The loaded slots are not processed in the context of the slot queue, however, I've created a dummy future to pass in to `sales.cleanUp`.

codex/utils/trackedfutures.nim
```diff
@@ -42,6 +40,8 @@ proc track*[T, U](future: Future[T], self: U): Future[T] =
 proc cancelTracked*(self: TrackedFutures) {.async.} =
   self.cancelling = true
 
+  trace "cancelling tracked futures"
```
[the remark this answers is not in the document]
emizzle: I actually think one is a bit helpful without being too noisy. I removed the other one simply because it created too much noise and doesn't contain contextual information (contained in the layer above).

codex/sales/salesagent.nim
```diff
@@ -21,6 +21,10 @@ type
     context*: SalesContext
     data*: SalesData
     subscribed: bool
+    onCleanUp*: OnCleanUp
+
+  OnCleanUp* = proc: Future[void] {.gcsafe, upraises: [].}
```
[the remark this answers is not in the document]
emizzle: Took me a while to remember, thank god for well-documented git comments 😄 

You're right, this should not be duplicated, which has been updated (by removing `OnCleanUp` from `SalesContext`.

The reason this was moved was because every time `processSlot` was called, the context's `onCleanUp` callback was overwritten, taking the `done` future in its closure with it. Effectively, `done` was only called for the *last* processed slot, and the other workers would remain blocked. This will likely also happen with `onFilled`, so I've moved that to `SalesAgent`. 

We have slot/agent-level callbacks, that are specific to the slot and the callback closure will be overwritten for each slot that is process. Then we have sales-level callbacks, in which their closure will not be overwritten for each slot processed. I've added som comments regarded this.

`OnSale` is a sales-level callback, which is triggered once a slot is filled. It is only used in the tests, which I don't particularly care for, but I don't think we have better options for testing that `filled` is called at this point.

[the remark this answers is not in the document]
emizzle: See https://github.com/codex-storage/nim-codex/pull/535/commits/df7ac216a01f3d0bae2effb4c1cc53b80c0403f2 for all the above changes.

tests/codex/sales/testreservations.nim
```diff
@@ -1,186 +1,278 @@
...
 asyncchecksuite "Reservations module":
   var
     repo: RepoStore
     repoDs: Datastore
     metaDs: SQLiteDatastore
-    availability: Availability
     reservations: Reservations
 
   setup:
+    randomize()
```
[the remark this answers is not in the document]
emizzle: Good idea. Updated.

codex/sales.nim
```diff
@@ -117,15 +148,39 @@ proc processSlot(sales: Sales, item: SlotQueueItem, done: Future[void]) =
...
+    logScope:
+      reservationId = reservation.id
+      availabilityId = reservation.availabilityId
+
+    if err =? (await reservations.deleteReservation(
+      reservation.id, reservation.availabilityId
+    )).errorOption:
+      error "failed to delete unused reservation", error = err.msg
+    else:
+      trace "deleted unused reservation"
```
AuHau, line 192: This will log this message for every reservation deleted, right? Maybe only one would be sufficient? 😅

tests/codex/sales/testreservations.nim
```diff
@@ -21,7 +21,7 @@ asyncchecksuite "Reservations module":
     reservations: Reservations
 
   setup:
-    randomize()
+    randomize(1.int64) # create reproducible results
```
AuHau, line 24: This is also possible, but it will run with the same inputs every time. If it is not a problem then keep it.

AuHau (dismissed):
👍

PR #537, opened 2023-08-28
fix: sales concurrency bug

### LINKED ISSUE #532: Amount of active Slots that host can host is limited to SlotQueue's `maxWorkers` (default 3)
After merging Proving and Sales state machines in https://github.com/codex-storage/nim-codex/pull/469, the sale state machine's process finishes **only after the whole request's duration is passed and a request is paid out** (if not taking into consideration the error states and other "edge-cases").

This is a problem because currently, the SlotQueue works in a way that it delegates the Sales process for the chosen Slot to pool of workers (default is 3) and **it blocks the worker until the Sales process is finished**, which is now much longer then it used (before it was till Slot is filled and then for proving phase it was handed out to the other state machine). So essentially, this limits the number of concurrently hosted slots to the number of workers.

AuHau:
Closes #532.

Since the `onSale` callback can potentially be overridden (for example, like in [`testsales`](https://github.com/codex-storage/nim-codex/blob/9c487d9f028db4e59e3efa96bbf2806baf160a04/tests/codex/sales/testsales.nim#L408)), I have decided to create a new `onFilled` function instead. This function is not public and should only be used to clear the futures of the Queue.

Other options include making the `onSale` function "private" or supporting multiple functions in hooks.

### DIFF
--- a/codex/sales.nim
+++ b/codex/sales.nim
@@ -101,11 +101,8 @@ proc remove(sales: Sales, agent: SalesAgent) {.async.} =
   if sales.running:
     sales.agents.keepItIf(it != agent)
 
-proc cleanUp(sales: Sales,
-             agent: SalesAgent,
-             processing: Future[void]) {.async.} =
-  await sales.remove(agent)
-  # signal back to the slot queue to cycle a worker
+proc filled(sales: Sales,
+             processing: Future[void]) =
   if not processing.isNil and not processing.finished():
     processing.complete()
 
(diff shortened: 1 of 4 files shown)


emizzle (approved):
LGTM!
I think we can simply remove OnSale, as it's not really being used, and `OnFilled` is more accurate.

AuHau:
> I think we can simply remove OnSale, as it's not really being used, and OnFilled is more accurate.

The problem with replacing `OnSale` with `OnFilled` is that if `OnFilled` would be used by some external component (like tests for example) than it would potentially break the Sales module as currently the hooks allow only *one* hook to be used and if another is added than it replaces the original. And we use the `OnFilled` for unblocking the queue workers, so if tests would use it for some tests assertions then it would break the unblocking of the queue workers mechanism. I will add comment about it to the code and then merge it in the current form.

PR #540, opened 2023-08-30
Use latest tag for Tags and Default branch only

veaceslavdoina:
This is a quick fix for possible issue discovered by @AuHau when we can override `latest` tags with a builds from a custom branch. And now
- `tag_latest` will be set to true for `default branch`
- `tag_latest` will be set to true for `tagged commit`

I would like to review that behaviour in the future and I will reborn https://github.com/codex-storage/cs-codex-dist-tests/issues/42 slightly later.


AuHau (approved):
LGTM 👍 Thanks!

PR #541, opened 2023-08-30
Storing and retrieving data using merkle trees

tbekas:
Work in progress. Done so far:
* storing and retrieving data using merkle trees works locally
* merkle trees are stored as regular blocks

To do:
* fix all TODOs
* fix block exchange - it uses old mechanism that doesn't work anymore

### DIFF
--- a/codex/blocktype.nim
+++ b/codex/blocktype.nim
@@ -37,6 +37,14 @@ type
     cid*: Cid
     data*: seq[byte]
 
+  BlockAddress* = object
+    case leaf*: bool
+    of true:
+      treeCid*: Cid
+      index*: Natural
+    else:
+      cid*: Cid
+
 template EmptyCid*: untyped =
   var
     EmptyCid {.global, threadvar.}:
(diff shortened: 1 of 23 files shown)


codex/stores/repostore.nim
```diff
@@ -97,7 +102,53 @@ method getBlock*(self: RepoStore, cid: Cid): Future[?!Block] {.async.} =
     return failure(newException(BlockNotFoundError, err.msg))
 
   trace "Got block for cid", cid
-  return Block.new(cid, data)
+  return Block.new(cid, data, verify = true)
+
+proc getMerkleTree(self: RepoStore, cid: Cid): Future[?!MerkleTree] {.async.} =
+  try:
```
dryajov, line 108: I'd suggest avoiding exceptions (specially mixing wiht result), you can either use `.catch` from `questionable` (?) or the `withValue` apis (this are ugly but performant).

codex/blocktype.nim
```diff
@@ -147,24 +147,25 @@ func new*(
     cid: cid,
     data: @data).success
 
-func new*(
+proc new*(
```
dryajov, line 150: Any reason to change to a proc, func has no side effects and are overall "safer", I don't see anything that would trigger a side effect here?

codex/stores/repostore.nim
tbekas: Fixed in #566 , I'm using `.getOption` on `LruCache` instead

codex/blocktype.nim
tbekas: Fixed in #566 , it's `func` once again

PR #542, opened 2023-08-31
Fix slot queue push

### LINKED ISSUE #494: Slot queue test fails occasionally

### LINKED ISSUE #515: Migrate from `eventuallyCheck` to `eventually`
We have the utility's `eventuallyCheck`, but there is `asynctest`'s `eventually` function, which enhances the functionality. We should replace `eventuallyCheck` with `asynctest`'s version.

The only issue is that the `asynctest` version is not a direct replacement, as certain tests in [this link](https://github.com/codex-storage/nim-codex/blob/master/tests/codex/sales/testsales.nim) begin to fail. We need to investigate.

markspanbroek:
Removes the check for available storage before pushing to the slot queue.

The current code of the sales slot queue has a problem, because the proc to push something onto the slot queue is *asynchronous*. When a number of slots come in at the same time, then the push calls are implicitly queued up in the chronos dispatcher. Because these implicitly queued slots are not yet visible to the slot queue, duplicated processing of slots can happen. This is why the sales tests sometimes fail (#494), and why the sales tests started failing when the test timing is slightly altered (#515).

By removing the check for availability from the push call we can make it synchronous.

I think that removing the preliminary check for available storage is acceptable, because:
- the check is done as the first step when picking up a queue item anyway
- when there are too many requests coming in, you want to drop them immediately (which the slot queue already does)
- when there are many requests, you want to check availability for the most profitable ones first (which the slot queue already does)

fixes #494
fixes #515


emizzle (commented):
The main reason we checked availability on push was due to not having the queue add every single new slot being requested on the network. 

I do however, think this is an acceptable solution, for now at least, as the slot queue has a capped size (of 64 by default) and queue workers will only be blocked for the duration of download (see #532, fixed with #537). The most profitable slots coming in from the network will bubble up to the top of the queue, but that doesn't mean the node has availability to service them, which means that potentially less profitable, but serviceable, slots may have been dropped from the queue due to its capped size. Because of this, it may be a good idea to increase the size of the slot queue max size with the proposed change in this PR. 

How much to increase the slot queue max size by, or what the value should be is hard to say. I actually think it should be a function of the node's available memory as it will, at worst, be consuming constant memory if the queue stays full. But perhaps, this could be tackled as an optimisation in the future.

PR #543, opened 2023-09-01
Slows down block maintenance iteration

benbierens:
This combined with https://github.com/codex-storage/nim-codex/pull/536 have greatly improved node stability and network stability in the continuous testing environment.

### DIFF
--- a/codex/stores/maintenance.nim
+++ b/codex/stores/maintenance.nim
@@ -42,9 +42,9 @@ proc new*(
     clock: Clock = SystemClock.new()
 ): BlockMaintainer =
   ## Create new BlockMaintainer instance
-  ## 
+  ##
   ## Call `start` to begin looking for for expired blocks
-  ## 
+  ##
   BlockMaintainer(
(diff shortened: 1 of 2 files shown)


codex/stores/maintenance.nim
```diff
@@ -78,6 +78,7 @@ proc runBlockCheck(self: BlockMaintainer): Future[void] {.async.} =
     if be =? await maybeBeFuture:
       inc numberReceived
       await self.processBlockExpiration(be)
+      await sleepAsync(50.millis)
```
tbekas, line 81: Maybe we can do it as a configuration param?

tbekas (approved):

codex/stores/maintenance.nim
elcritch: Or something we can remove once we get non-blocking IO in. It looks like this might just give time for the network async items to catch up? But yah, it'd be nice to have something a bit more concrete long term.

PR #545, opened 2023-09-06
Erasure cleanup

dryajov:


dryajov:
@emizzle @markspanbroek just a heads up, this reworks one of the tests, because right now we don't support encoding a single block - https://github.com/codex-storage/nim-codex/pull/545/files#diff-77966ace2c23c80291b17b166fbe06bdb2c48e01c76335b15e81d49b4e9d4f8eR59-R100. I'll work on a solution, but for now, we should just use some random generated blocks instead.

markspanbroek:
> this reworks one of the tests, because right now we don't support encoding a single block

This might break tests at other levels as well, e.g. the distributed or continuous tests. If possible, I would suggest to fix single block encoding before merging this.

dryajov:
> > this reworks one of the tests, because right now we don't support encoding a single block
> 
> This might break tests at other levels as well, e.g. the distributed or continuous tests. If possible, I would suggest to fix single block encoding before merging this.

We should for now not test with one block, which are really a special case. 

Our encoding is block level, which means that a single block is one buffer to be encoded, this is to say that the encoded block is just one symbol lined up in a row, and there is just nothing to encoded it with. This is very convenient in the case of two or more blocks, where we stack the two blocks together and encode the two buffers in a row, which makes two symbols encoded into `<=2` symbols. 

In order to support 1 block encoding, we probably need to add a dummy block, otherwise the structure of the encoding becomes too complicated and it would be hard to make it match the dispersal and the nodes/slots.

Overall, it's not a bug, this is more of a corner case that we should eventually support but it is not critical for the overall system to work.

dryajov:
Just to clarify, I do plan to spend time on this in the coming weeks, but for now this is an improvement on the existing code and we now also understand why the previous test was failing. So, in terms of this PR, it's simply a cleanup and addressing the current limitation in the tests with the appropriate comments and an accompanying issue #551.

tests/integration/testIntegration.nim
```diff
@@ -65,6 +81,20 @@ twonodessuite "Integration tests", debug1 = false, debug2 = false:
...
+  #   let cid = client1.upload("some file contents").get
+  #   let id = client1.requestStorage(cid, duration=1.u256, reward=2.u256, proofProbability=3.u256, expiry=expiry, collateral=200.u256, nodes=2, tolerance=1).get
+  #   let request = client1.getPurchase(id).get.request.get
+  #   check request.ask.duration == 1.u256
+  #   check request.ask.reward == 2.u256
+  #   check request.ask.proofProbability == 3.u256
+  #   check request.expiry == expiry
+  #   check request.ask.collateral == 200.u256
+  #   check request.ask.slots == 3'u64
+  #   check request.ask.maxSlotLoss == 1'u64
```
emizzle, line 96: Instead of commenting out this test, could we increase the dataset size instead so that there's more than one Block being encoded?

codex/stores/repostore.nim
```diff
@@ -211,7 +218,14 @@ method delBlock*(self: RepoStore, cid: Cid): Future[?!void] {.async.} =
...
 
-  trace "Deleting block", cid
+  logScope:
+    cid = cid
+
+  trace "Deleting block"
+
+  if cid.isEmpty:
+    trace "Empty block, ignoring"
+    return success()
```
emizzle, line 228: Duplicate of lines below (rebase?)
```suggestion
```

emizzle (commented):
Looks good, but as per my comment in the code, could we increase the dataset size to avoid the one Block edge case instead of removing the integration test altogether?

tests/integration/testIntegration.nim
dryajov: Thats what I did, look above the commented out test, it's the same but with several blocks.

dryajov: In fact, I left the name of the original test as is and left this one as the edge case we'd want to enable after adding support for it.

emizzle (approved):

PR #548, opened 2023-09-07
chore(logging): decimal stint values

AuHau:
If the normal "to-string" operator ($) or default conversion is used on StInt values in logging, then hex encoded values are returned. This changes it to decimal values, but only for one entry 😳

Is there some way to say chronicles to use the `stint`'s `toString` function (which by default returns decimal values) on StInt types?

### DIFF
--- a/codex/node.nim
+++ b/codex/node.nim
@@ -254,7 +254,7 @@ proc requestStorage*(
   ## - Run the PoR setup on the erasure dataset
   ## - Call into the marketplace and purchasing contracts
   ##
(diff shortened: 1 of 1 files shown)


emizzle:
Yes there is. You need to create and/or import a nim-serialization `writeValue` that produces decimals (can use `$` to achieve this). There is [one in nim-eth](https://github.com/status-im/nim-eth/blob/0d7e7448c4a2a50136e726be7d4e4a38e094e8e3/eth/common/eth_types_json_serialization.nim#L28) that can be imported for convenience or we can roll our own.

PR #552, opened 2023-09-15
Integrate io threads

dryajov:
Integrating https://github.com/codex-storage/nim-datastore/pull/48


dryajov:
this is approach is scrapped

PR #554, opened 2023-09-22
add support to optional log-to-file

gmega:
This adds support to a log-file option which will write to a file in addition to stdout when specified. I started with the idea of using a chronicles file sink to implement this, but turns out you can't have an optional file sink in chronicles: if a file sink is declared in config.nims, then it will always output to a file. Even if you don't call `output.open()` on it, it will choose a default log file name (codex.X.log) and output.

I therefore implemented an in-codex tee which just writes to a file we manage ourselves. The file never gets closed, but chronicles also doesn't close its files as far as I can tell, so we're doing no worse than them. Dynamic sinks also don't help as I'd still have to manage the file myself, in addition to configuring the sink and making sure it matches the selected format (json, text, text with ANSI codes).

### DIFF
--- a/codex/conf.nim
+++ b/codex/conf.nim
@@ -78,6 +78,10 @@ type
       defaultValue: LogKind.Auto
       name: "log-format" }: LogKind
 
+    logFile* {.
+      desc: "If specified, writes logs to the specified file in addition to writing to stdout"
+      name: "log-file" }: Option[OutFile]
+
     metricsEnabled* {.
       desc: "Enable the metrics server"
       defaultValue: false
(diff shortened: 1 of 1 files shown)


AuHau:
As a workaround LGTM, but as Chronicles is Status owned, what about contributing to it with this functionality? If you don't want to code it yourself, then at least I would suggest adding at least an issue about it and maybe adding a comment to the relevant section of your PR to point future readers to the issue?

I will leave the approving to the selected reviewers ;-)

codex/conf.nim
```diff
@@ -444,47 +448,63 @@ proc updateLogLevel*(logLevel: string) {.upraises: [ValueError].} =
...
-    proc writeAndFlush(f: File, msg: LogOutputStr) =
-      try:
-        f.write(msg)
-        f.flushFile()
-      except IOError as err:
-        logLoggingFailure(cstring(msg), err)
 
-    proc stdoutFlush(logLevel: LogLevel, msg: LogOutputStr) =
-      writeAndFlush(stdout, msg)
+    # Managing the file ourselves may seem distasteful, but chronicles does not
```
dryajov, line 462: This doesn't sound correct as per chronicles own docs - https://github.com/status-im/nim-chronicles

@zah could you please clarify what would be the proper way of handling logging to either a file, stdout or both?

gmega: Ahah your opinion is just as valuable, @AuHau, just didn't think of dragging too many people into such a minor issue. At this point I'll wait for @zah's take, but what I could grok from docs + code + behavior is that these things are statically defined and that you can't mute a file sink - if it's defined, it will want to output.

gmega: Ahah your opinion is just as valuable, @AuHau, just didn't think of dragging too many people into such a minor issue. At this point I'll wait for @zah's take, but what I could grok from docs + code + behavior is that these things are statically defined and that you can't mute a file sink - if it's defined, it will want to output. The only thing you get to define dynamically is the name of the output file. You could in principle use that and send it to `/dev/null` to mute it, but that's not portable.

The other option would be having a dynamic sink which is also a file sink (we use a null writer trick to mute dynamic sinks), but I haven't found a way to do that - it's either one or the other. And while this could be built on top of a dynamic sink, the resulting code will be more complicated than what's currently there.

gmega:
Ahah your opinion is just as valuable, @AuHau, I just didn't think of dragging too many people into such a minor issue.

emizzle:
Created https://github.com/codex-storage/nim-codex/pull/558 using a dynamic sink as a possible alternative to this PR.

PR #561, opened 2023-10-03
feat: partial payouts for cancelled requests

### LINKED ISSUE #556: Sales statemachine should call `freeSlot()` for failed slots
`freeSlot()` is called for payout, but it is also used for removing the slot from the node's on-chain slots list, which is, for example, used during node's bootstrapping for loading the node's current slots ([here](https://github.com/codex-storage/nim-codex/blob/master/codex/sales.nim#L141). In this way, if the slot fails and its state machine reaches the Errored state, then the next time the node restarts, the slot is loaded again from the on-chain list. At the moment, it should not really cause any problems, as it will simply go again into an Errored state, but possibly with PR #535 there could be some issues with Reservations being missing. And generally we should keep a consistent state.

AuHau:
Adds support for partial payouts of canceled requests.

Also closes #556 and fixes off-by-one bug for calling `withdrawFunds()` contract function for canceled requests which leads to errors "Request is not expired yet".

The integration test is failing, though, because of the Hardhat's mining mode - automine. The time needs to be advanced (eq. mined block) before calling the `withdrawFunds()` function for the canceled request, but I have not figured out any way how to do it in the test's code. I have tried switching the Hardhat's mining mode into mining block every second and for that the test worked. Maybe we should consider changing the mining mode? It had brought us more problems recently 😳

The PR is blocked by the contract's PR: https://github.com/codex-storage/codex-contracts-eth/pull/69

Also closes https://github.com/codex-storage/codex-contracts-eth/issues/57

### DIFF
--- a/codex/purchasing/states/submitted.nim
+++ b/codex/purchasing/states/submitted.nim
@@ -33,7 +33,7 @@ method run*(state: PurchaseSubmitted, machine: Machine): Future[?State] {.async.
     await subscription.unsubscribe()
 
   proc withTimeout(future: Future[void]) {.async.} =
-    let expiry = request.expiry.truncate(int64)
+    let expiry = request.expiry.truncate(int64) + 1
     await future.withTimeout(clock, expiry)
 
   try:
--- a/codex/purchasing/states/submitted.nim
+++ b/codex/purchasing/states/submitted.nim
@@ -33,15 +33,6 @@ method run*(state: PurchaseSubmitted, machine: Machine): Future[?State] {.async.
     await subscription.unsubscribe()
 
   proc withTimeout(future: Future[void]) {.async.} =
-    # while true:
-    #   await clock.waitUntil(max(clock.now+1, request.expiry.truncate(int64)))
-    #
-    #   if (state =? await agent.retrieveRequestState()) and state == RequestState.Cancelled:
-    #     agent.schedule(cancelledEvent(request))
-    #     break
-    #
-    #   trace "Request is not yet cancelled, even it should be. Waiting some more."
-
     let expiry = request.expiry.truncate(int64) + 1
     await future.withTimeout(clock, expiry)
 
--- a/codex/purchasing/states/submitted.nim
+++ b/codex/purchasing/states/submitted.nim
@@ -33,6 +33,15 @@ method run*(state: PurchaseSubmitted, machine: Machine): Future[?State] {.async.
     await subscription.unsubscribe()
 
   proc withTimeout(future: Future[void]) {.async.} =
+    # while true:
+    #   await clock.waitUntil(max(clock.now+1, request.expiry.truncate(int64)))
+    #
+    #   if (state =? await agent.retrieveRequestState()) and state == RequestState.Cancelled:
+    #     agent.schedule(cancelledEvent(request))
+    #     break
+    #
+    #   trace "Request is not yet cancelled, even it should be. Waiting some more."
+
     let expiry = request.expiry.truncate(int64) + 1
     await future.withTimeout(clock, expiry)
 
--- a/codex/sales/states/unknown.nim
+++ b/codex/sales/states/unknown.nim
@@ -55,4 +55,4 @@ method run*(state: SaleUnknown, machine: Machine): Future[?State] {.async.} =
   of SlotState.Failed:
     return some State(SaleFailed())
   of SlotState.Cancelled:
-    return some State(SalePayout())
+    return some State(SaleCancelled())
--- a/codex/sales/states/unknown.nim
+++ b/codex/sales/states/unknown.nim
@@ -54,3 +54,5 @@ method run*(state: SaleUnknown, machine: Machine): Future[?State] {.async.} =
     return some State(SaleFinished())
   of SlotState.Failed:
     return some State(SaleFailed())
+  of SlotState.Cancelled:
+    return some State(SalePayout())
--- a/tests/codex/sales/testsales.nim
+++ b/tests/codex/sales/testsales.nim
@@ -453,10 +453,34 @@ asyncchecksuite "Sales":
       return success()
     createAvailability()
     await market.requestStorage(request)
+
+    # If we would not await, then the `clock.set` would run "too fast" as the `subscribeCancellation()`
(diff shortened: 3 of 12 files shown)


codex/sales/states/unknown.nim
```diff
@@ -51,3 +52,5 @@ method run*(state: SaleUnknown, machine: Machine): Future[?State] {.async.} =
     return some State(SaleFinished())
   of SlotState.Failed:
     return some State(SaleFailed())
+  of SlotState.Cancelled:
+    return some State(SalePayout())
```
markspanbroek, line 56: Why not go to the Cancelled state instead?

tests/integration/testIntegration.nim
```diff
@@ -134,3 +134,27 @@ twonodessuite "Integration tests", debug1 = false, debug2 = false:
     await provider.advanceTime(duration)
 
     check eventually (await token.balanceOf(account2)) - startBalance == duration*reward
+
+  test "expired request partially pays out for stored time":
```
markspanbroek, line 138: This test is failing in CI?

markspanbroek (commented):
Nice, thanks!

Looks like the newly added test is failing on CI.

tests/integration/testIntegration.nim
AuHau: Yes as per the comment in the PR description:

> The integration test is failing, though, because of the Hardhat's mining mode - automine. The time needs to be advanced (eq. mined block) before calling the withdrawFunds() function for the canceled request, but I have not figured out any way how to do it in the test's code. I have tried switching the Hardhat's mining mode into mining block every second and for that the test worked. Maybe we should consider changing the mining mode? It had brought us more problems recently 😳

Do you have any thoughts? If you don't have any ideas on how to accomplish this, nor you are against changing the Hardhat's mining mode, then I will drop (comment out) the test for it to pass...

codex/sales/states/unknown.nim
AuHau: Yup you are right, that is a left-over from an earlier version where I used the `SalePayout` state for freeing the slot, but then thought it will be cleaner to have the `freeSlot` calls directly in the states... Will fix that.

tests/integration/testIntegration.nim
markspanbroek: Of course, sorry :facepalm:

You can advance the time in the integration tests as well using:
```nim
    await provider.advanceTimeTo(expiry)
```

But it seems that the test is failing earlier; the requestStorage() call fails.

Changing the mining mode has a couple of problems that I can see:
- the tests take longer to run if you have to wait a second for a tx to be mined
- most tests would need to be rewritten because they assume manual mining
- tests might become brittle because of the timing differences when things are mined

AuHau: > You can advance the time in the integration tests as well using:

I am aware of this, but the problem is to call this at the right moment. It needs to be called once the state machines (both for seller and buyer) reach `Cancelled` state, which I don't know how to do that. I could either add some callback which would be called in `Cancelled` state (that would not actually much help here, as this is integration tests), I could try to time it (don't think that is good idea).

Or we could simply drop this test, because I have similar one ready in the Dist-testing framework: https://github.com/codex-storage/cs-codex-dist-tests/pull/55

markspanbroek: > It needs to be called once the state machines (both for seller and buyer) reach `Cancelled` state

Why do we need to advance the time after the request is cancelled on chain? (I'm probably missing something obvious here)

In thought it should work as follows:
1. client requests storage on 2 nodes
2. only 1 node fills a slot
3. the request expires (initiated by the advanceTime() call)
4. both client and the node that filled a slot withdraw their funds

AuHau:
I have added the verification of the truly canceled state and also the termination of canceled requests in cancelled states (both in purchasing and sales).

codex/purchasing/states/submitted.nim
```diff
@@ -33,7 +33,16 @@ method run*(state: PurchaseSubmitted, machine: Machine): Future[?State] {.async.
...
   proc withTimeout(future: Future[void]) {.async.} =
-    let expiry = request.expiry.truncate(int64)
+    # while true:
+    #   await clock.waitUntil(max(clock.now+1, request.expiry.truncate(int64)))
+    #
+    #   if (state =? await agent.retrieveRequestState()) and state == RequestState.Cancelled:
+    #     agent.schedule(cancelledEvent(request))
+    #     break
+    #
+    #   trace "Request is not yet cancelled, even it should be. Waiting some more."
```
markspanbroek, line 43: Can be removed?

tests/codex/sales/testsales.nim
```diff
@@ -450,10 +450,35 @@ asyncchecksuite "Sales":
       return success()
     createAvailability()
     await market.requestStorage(request)
-    clock.set(request.expiry.truncate(int64))
+
+    # If we would not await, then the `clock.set` would run "too fast" as the `subscribeCancellation()`
+    # would otherwise not set the timeout early enough as it uses `clock.now` in the deadline calculation.
+    await sleepAsync(chronos.milliseconds(100))
```
markspanbroek, line 456: I understand the need to wait a bit here, but can't we wait on something specific? Maybe we can wait here until onStore is called, for instance?

Waiting for a number of milliseconds has two unfortunate side-effects:
1. You're probably waiting too long most of the time, slowing down the test run
2. You're probably waiting not long enough some of the time, making the test fail every now and then in CI

:smile:

markspanbroek (approved):
Nicely done @AuHau!

tests/codex/sales/testsales.nim
AuHau: I agree with you, but I have not found a way to wait for something concrete. I have tried using the `onStore` callback, but even that is "too fast" and does not give enough time to properly set up the expiry timeouts. So the only possibility I see is adding a special callback called "onCancellationSet" or something similar, which could be used here. However, in my honest opinion, that is not worthwhile. Honestly, I have spent way too much time on this PR (again) and would like to move on to other tasks, so I would leave it as it is for now.

markspanbroek: Fair enough!

PR #566, opened 2023-10-12
Blockexchange uses merkle root and index to fetch blocks

### LINKED ISSUE #498: Merkelizing Block Data
Initial design writeup https://github.com/codex-storage/codex-research/blob/master/design/metadata-overhead.md
Work break down and review for Ben and Tomasz (epic coming up)
This is required to integrate the proving system

 Changes to the Repo Store
- Creating a merkle tree after upload (non-blocking, tree in memory)
- Storing merkle tree as an array on disk (where to put it on disk?)
- Inserting path into a partial merkle tree
- Apart from requesting block by <hash>, we can request block by 
<merkle root, index>
<merkle root, seq[index]> (optional)
<merkle root, from index, to index> (optional)
- Reconstruct the entire data set by traversing leaf indicies instead of traversing seq[cid] of blocks

Changes in the Block Exchange

- Additional message, to support querying by:
<merkle root, index>
<merkle root, seq[index]> (optional)
<merkle root, from index, to index> (optional)

Download response:
seq[<block, merkle proofs>]

Update the network store to support new access patterns

Cleanup

### LINKED ISSUE #511: Upgrade Merkle Tree to avoid second-preimage attacks

tbekas:
Resolves #511 
Resolves #498


codex/stores/repostore.nim
```diff
@@ -311,8 +333,8 @@ proc createBlockExpirationQuery(maxNumber: int, offset: int): ?!Query =
 method getBlockExpirations*(
   self: RepoStore,
   maxNumber: int,
-  offset: int): Future[?!BlockExpirationIter] {.async, base.} =
-  ## Get block expirations from the given RepoStore
+  offset: int): Future[?!AsyncIter[?BlockExpiration]] {.async, base.} =
+  ## Get block experiartions from the given RepoStore
```
dryajov, line 337:
```suggestion
  ## Get block expirations from the given RepoStore
```

codex/codex.nim
```diff
@@ -231,6 +231,8 @@ proc new*(
     wallet = WalletRef.new(EthPrivateKey.random())
     network = BlockExcNetwork.new(switch)
 
+    treeReader = TreeReader.new()
```
dryajov, line 234: This don't seem to be used anywhere?

dryajov (changes requested):
Just a quick preliminary review.

codex/stores/treereader.nim
```diff
@@ -0,0 +1,120 @@
...
+const DefaultTreeCacheCapacity* = 10    # Max number of trees stored in memory
+
+type
+  GetBlock = proc (cid: Cid): Future[?!Block] {.upraises: [], gcsafe, closure.}
+  DelBlock = proc (cid: Cid): Future[?!void] {.upraises: [], gcsafe, closure.}
+  TreeReader* = ref object of RootObj
+    getBlockFromStore*: GetBlock
+    treeCache*: LruCache[Cid, MerkleTree]
+
+method getTree*(self: TreeReader, cid: Cid): Future[?!MerkleTree] {.async.} =
```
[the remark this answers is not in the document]
tbekas: Right, I will make it non-virtual 👍

codex/codex.nim
tbekas: Yes, fixing it.

codex/node.nim
```diff
@@ -106,18 +109,26 @@ proc fetchBatched*(
...
+
+    let blocks = collect:
+      for i in 0..<batchSize:
+        if not iter.finished:
+          iter.next()
+
+
+    # let 
+    #   indexRange = (batchNum * batchSize)..<(min((batchNum + 1) * batchSize, manifest.blocksCount))
+    #   blocks = indexRange.mapIt(node.blockStore.getBlock(manifest.treeCid, it))
```
tbekas, line 130: I'll remove this comment block with the next batch of changes.

codex/stores/blockstore.nim
```diff
@@ -26,24 +28,32 @@ type
...
+  
+  raiseAssert("Not implemented!")
+
+method getBlockAndProof*(self: BlockStore, treeCid: Cid, index: Natural): Future[?!(Block, MerkleProof)] {.base.} =
+  ## Get a block and associated inclusion proof by Cid of a merkle tree and an index of a leaf in a tree
+  ## 
+  
+  raiseAssert("Not implemented!")
+
+method getBlocks*(self: BlockStore, treeCid: Cid, leavesCount: Natural, merkleRoot: MultiHash): Future[?!AsyncIter[?!Block]] {.base.} =
```
dryajov, line 51: Maybe it makes sense to change the interface to take a `BlockAddress` everywhere?

codex/blocktype.nim
```diff
@@ -154,17 +109,58 @@ func new*(
...
-      version = cid.cidver,
-      codec = cid.mcodec,
-      mcodec = mhash.mcodec)
+  const
+    Sha256 = multiCodec("sha2-256")
+    Raw = multiCodec("raw")
+    DagPB = multiCodec("dag-pb")
+    DagJson = multiCodec("dag-json")
+
+  var index {.global, threadvar.}: Table[(CIDv0, Sha256, DagPB), Result[Cid, CidError]]
```
dryajov, line 137: The correct type here is `MultiCodec`, the reason this kinda works with this declaration, is because `DagPB` is an alias to `MultiCodec`, but this declaration is weird, because you then use `DagJson` and `Raw`, which are clearly not `DagPB`. I would declare this as `Table[(CidVersion, MultiCodec, MultiCodec), Cid]` (without the encapsulated result, so that we don't have to unpack it every time).

dryajov (changes requested):

dryajov (changes requested):
Adding some more comments as I move along.

codex/blockexchange/engine/engine.nim
```diff
@@ -279,50 +347,54 @@ proc scheduleTasks(b: BlockExcEngine, blocks: seq[bt.Block]) {.async.} =
...
-  trace "Got blocks from peer", peer, len = blocks.len
-  for blk in blocks:
-    if isErr (await b.localStore.putBlock(blk)):
-      trace "Unable to store block", cid = blk.cid
+  blocksDelivery: seq[BlockDelivery]) {.async.} =
+  trace "Got blocks from peer", peer, len = blocksDelivery.len
+
+  for bd in blocksDelivery:
+    if isErr (await b.localStore.putBlock(bd.blk)):
+      trace "Unable to store block", cid = bd.blk.cid
```
dryajov, line 381: Shouldn't we filter out failed blocks from the `blocksDelivery`?

dryajov:
One thing I'm seeing right now, is that we're have a `treeCid` and a `merkleRoot`. We're storing the tree as a block and derive a hash and cid from it. This create indirection and duplication. I think there is an easy way of overcoming this by simply naming the on disc merkle tree block with the merkle root hash, this doesn't violate content addressability at all, since we can easily verify that the contents of the file are indeed correct by rebuilding the root hash from the stored hashes.

codex/blocktype.nim
```diff
@@ -37,91 +39,44 @@ type
     cid*: Cid
```
dryajov, line 39: I think this can simplify things a lot, if we make the `cid` and `BlockAddress`?

codex/streams/storestream.nim
```diff
@@ -32,18 +33,20 @@ const
...
 
 type
-  # Make SeekableStream from a sequence of blocks stored in Manifest
-  # (only original file data - see StoreStream.size)
-  StoreStream* = ref object of SeekableStream
+  StoreStream* = ref object of LPStream
     store*: BlockStore          # Store where to lookup block contents
     manifest*: Manifest         # List of block CIDs
     pad*: bool                  # Pad last block to manifest.blockSize?
+    iter*: AsyncIter[?!Block]
```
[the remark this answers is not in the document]
tbekas: Right, this property shouldn't be exposed. I'll remove the public modifier `*`.

codex/utils/asynciter.nim
```diff
@@ -0,0 +1,141 @@
+import std/sugar
```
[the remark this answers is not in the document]
tbekas: The `items` iterator is already implemented in line 22 of this file. I'll add the `pairs` though, I didn't thought about it initially, but I agree that it makes sense to have it.

codex/stores/treereader.nim
```diff
@@ -0,0 +1,101 @@
...
+        return failure(err)
+
+      without blk =? await self.getBlockFromStore(leafCid), err:
+        return failure(err)
+
+      return success(blk)
+    )
+  return success(iter)
+
+proc new*(T: type TreeReader, treeCacheCap = DefaultTreeCacheCapacity): TreeReader =
```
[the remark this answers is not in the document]
tbekas: I agree, fixing it.

codex/blockexchange/engine/pendingblocks.nim
```diff
@@ -36,12 +43,121 @@ type
...
+
+proc checkIfAllDelivered(p: PendingBlocksManager, treeReq: TreeReq): void =
+  if treeReq.deliveredCount.some == treeReq.leavesCount:
+    without tree =? buildTree(treeReq), err:
+      error "Error building a tree", msg = err.msg
+      p.trees.del(treeReq.treeCid)
+      return
+    p.trees.del(treeReq.treeCid)
+    try:
+      asyncSpawn p.onTree(tree)
```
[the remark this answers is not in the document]
tbekas: Righ, I will change the log message, but I have to leave the `try/exept` because otherwise I get a compilation error:
> nim-codex/codex/blockexchange/engine/pendingblocks.nim(111, 24) Error: Exception can raise an unlisted exception: Exception

codex/blockexchange/engine/pendingblocks.nim
```diff
@@ -73,36 +189,100 @@ proc getWantHandle*(
...
+            let
+              startTime = pending[].startTime
+              stopTime = getMonoTime().ticks
+              retrievalDurationUs = (stopTime - startTime) div 1000
+            codexBlockExchangeRetrievalTimeUs.set(retrievalDurationUs)
+            trace "Block retrieval time", retrievalDurationUs
+      else:
+        warn "Delivery cid doesn't match block cid", deliveryCid = bd.address.cid, blockCid = bd.blk.cid
+
     # resolve any pending blocks
```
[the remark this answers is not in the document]
tbekas: Not sure if I understand this comment. You're suggesting to replace this:

```nim
if not bd.address.leaf:
  # ...
 
if bd.address.leaf:
  # ....
```

with this:
```nim
if bd.address.leaf:
  # ...
else:
 # ...
```
?

codex/blockexchange/engine/pendingblocks.nim
```diff
@@ -73,36 +189,100 @@ proc getWantHandle*(
...
+
+                inc treeReq.deliveredCount
+
+                p.checkIfAllDelivered(treeReq[])
+              else:
+                warn "Invalid proof provided for a block", address = bd.address
+            else:
+              warn "Missing proof for a block", address = bd.address
+          else:
+            trace "Ignore veryfing proof for already delivered block", address = bd.address
```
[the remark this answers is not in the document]
tbekas: Thanks for catching this. My IDE isn't configured yet to catch typos (I'll try to fix it).

codex/blockexchange/engine/engine.nim
```diff
@@ -124,22 +120,27 @@ proc stop*(b: BlockExcEngine) {.async.} =
 
   trace "NetworkStore stopped"
 
-proc sendWantHave(b: BlockExcEngine, cid: Cid, selectedPeer: BlockExcPeerCtx, peers: seq[BlockExcPeerCtx]): Future[void] {.async.} =
-  trace "Sending wantHave request to peers", cid
+
+proc sendWantHave(b: BlockExcEngine, address: BlockAddress, selectedPeer: BlockExcPeerCtx, peers: seq[BlockExcPeerCtx]): Future[void] {.async.} =
```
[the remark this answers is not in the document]
tbekas: I agree, this line is way too wide, I will break it. If you don't mind I would like to leave the explicit return `Future[void]` type as it's more clear in my opinion.

codex/blocktype.nim
tbekas: Right, sorry I was rushing and my brain for some reason put wrong types here, the correct types are `(CidVersion, MultiCodec, MultiCodec)` as you pointed.

I'll also fix the value type to just `Cid` instead of `Result[Cid, CidError]`.

codex/blockexchange/engine/engine.nim
tbekas: Hmm, that's good question. I think we can filter it out for now. It's a bit of an edge case - you can imagine a scenario where we can't put a block to the local store - maybe because it's full - but we can still process the block (i.e. stream it to a tcp socket of the user requesting the dataset via the rest api).

tbekas:
@dryajov  I agree with your comment. `treeCid: Cid` and `treeRoot: MultiHash` are redundant. Fixing it is not that simple though. One thing that I see is that whenever we read or transmit a block, we verify its content against the `Cid` by computing a hash from the data and checking if it matches the hash obtained from `Cid`. We would have to handle this special case somehow (we would have to recognize that we're reading a tree and use different verification method). 

In order to not introduce additional flow, I would do the following:
We no longer store trees as blocks - we would continue to store them in the `key/value` datastore however, but under the `trees` subdirectory (until we decide to store proofs only). There's no longer `treeCid: Cid` in `BlockAddress`, there's just `treeRoot: MultiHash` instead. What do you think?

codex/blockexchange/engine/pendingblocks.nim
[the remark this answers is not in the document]
dryajov: Interesting, this should be handled elsewhere then, at the point where the exception is being initiated. The problem is that this isn't really doing anything at all and the correct way of handling the base `Exception` (which should never be raised explicitely), is to convert it to a `Defect` and re-raise. So we need to do some diggeing to determine what exactly is trying to raise the `Exception`.

codex/blockexchange/engine/pendingblocks.nim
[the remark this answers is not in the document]
dryajov: Yes, precisely.

codex/blockexchange/engine/pendingblocks.nim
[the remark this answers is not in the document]
markspanbroek: You can try to mark the `TreeHandler` type with an explicit `{.raises:[].}`

I have seen that sometimes that isn't enough, and the compiler mistakenly thinks that a closure can raise an `Exception`.  Then the only way to work around it is to add a `try:except` block, and raise a `Defect`.

dryajov:
I merged https://github.com/codex-storage/nim-codex/pull/593, which simply removes the storageproofs directory and tests for now. If you can't get it to merge cleanly, just delete the directory and tests...

dryajov:
I rebased locally, and it does so cleanly, so it should work without issues.

codex/blockexchange/engine/engine.nim
```diff
@@ -205,12 +206,79 @@ proc requestBlock*(
...
 
   return await blk
 
+proc requestBlock(
+  b: BlockExcEngine,
+  treeReq: TreeReq,
+  index: Natural,
+  timeout = DefaultBlockTimeout
+): Future[Block] {.async.} =
+  let address = BlockAddress(leaf: true, treeCid: treeReq.treeCid, index: index)
```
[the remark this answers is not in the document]
tbekas: Just pushed changes with unified flow.

codex/stores/blockstore.nim
tbekas: I added a generic method below, but I also left the individual methods for requesting either by `cid` or `(treeCid, index)` as well.
```
getBlock*(self: BlockStore, address: BlockAddress)
```

codex/blocktype.nim
tbekas: Not really, block can belong to more than a single tree.

codex/blockexchange/engine/engine.nim
```diff
@@ -338,14 +398,14 @@ proc wantListHandler*(
 
     logScope:
       peer      = peerCtx.id
-      cid       = e.cid
+      # cid       = e.cid
       wantType  = $e.wantType
 
     if idx < 0: # updating entry
-      trace "Processing new want list entry", cid = e.cid
+      trace "Processing new want list entry", address = e.address
```
dryajov, line 405: you should be able to move this to the `logScope`?

dryajov (commented):
A few comments as I go through.

codex/stores/blockstore.nim
```diff
@@ -27,23 +29,31 @@ type
...
+  ## Get a block from the blockstore
+  ##
 
-  BlockStore* = ref object of RootObj
+  raiseAssert("getBlock by cid not implemented!")
 
-iterator items*(self: BlocksIter): Future[?Cid] =
-  while not self.finished:
-    yield self.next()
+method getBlock*(self: BlockStore, treeCid: Cid, index: Natural): Future[?!Block] {.base.} =
```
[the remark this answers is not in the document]
tbekas: Ad1. But backwards index is a different access pattern and hence a different parameter type. This is what we have in `AsyncHeapQueue`
```nim
proc `[]`*[T](heap: AsyncHeapQueue[T], i: Natural) : T {.inline.} =
  ## Access the i-th element of ``heap`` by order from first to last.
  ## ``heap[0]`` is the first element, ``heap[^1]`` is the last element.
  heap.queue[i]

proc `[]`*[T](heap: AsyncHeapQueue[T], i: BackwardsIndex) : T {.inline.} =
  ## Access the i-th element of ``heap`` by order from first to last.
  ## ``heap[0]`` is the first element, ``heap[^1]`` is the last element.
  heap.queue[len(heap.queue) - int(i)]
```

Ad2. There's a method to get block by address (line 46), but in some cases we don't have an address object in the scope, instead we have separately `treeCid` and `index` variables. We could allocate the memory and create the `BlockAddress` object, but: 1. that seem a bit wasteful and 2. it's an additional inconvenience to the client.

codex/stores/repostore.nim
```diff
@@ -82,6 +77,62 @@ func available*(self: RepoStore): uint =
...
 
+proc encode(cidAndProof: (Cid, MerkleProof)): seq[byte] =
+  let 
+    (cid, proof) = cidAndProof
+    cidBytes = cid.data.buffer
+    proofBytes = proof.encode
+
+  var buf = newSeq[byte](1 + cidBytes.len + proofBytes.len)
+
+  buf[0] = cid.data.buffer.len.byte # cid shouldnt be more than 255 bytes?
```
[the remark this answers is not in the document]
tbekas: Going to fix it.

codex/stores/repostore.nim
```diff
@@ -502,7 +608,7 @@ proc stop*(self: RepoStore): Future[void] {.async.} =
 
   self.started = false
 
-func new*(
+proc new*(
```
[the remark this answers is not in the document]
tbekas: Going to fix it

codex/blockexchange/engine/engine.nim
tbekas: Yep, you're right. Fixing it.

codex/blockexchange/engine/engine.nim
```diff
@@ -279,50 +282,107 @@ proc scheduleTasks(b: BlockExcEngine, blocks: seq[bt.Block]) {.async.} =
...
     trace "Sending payment for blocks", price
     await sendPayment(peer.id, payment)
 
-proc blocksHandler*(
+proc validateBlockDelivery(
+  b: BlockExcEngine,
+  bd: BlockDelivery
+): ?!void =
+  if bd.address notin b.pendingBlocks:
+    return failure("Received block is not currently a pending block")
```
tbekas, line 318: @dryajov this is a semantic change. Previously we stored any blocks delivered to us. Now I added this validation step to check if we were awaiting for a delivered block in the first place. If not we're discarding it. 

I added it to prevent someone filling up our disk with some random blocks, that we didn't ask for.

dryajov: Yeah, good check

codex/blockexchange/engine/pendingblocks.nim
```diff
@@ -37,88 +40,121 @@ type
...
-    # resolve any pending blocks
-    p.blocks.withValue(blk.cid, pending):
-      if not pending[].handle.completed:
-        trace "Resolving block", cid = blk.cid
-        pending[].handle.complete(blk)
+  for bd in blocksDelivery:
+    p.blocks.withValue(bd.address, blockReq):
+      trace "Resolving block", address = bd.address
+
+      if not blockReq.handle.finished:
```
dryajov, line 98: you want to use the `[]` for ref/pointer types, nim will try to infere the type, but it doesn't work always...

codex/erasure/erasure.nim
```diff
@@ -64,236 +70,199 @@ type
...
+  proc isFinished(): bool = pendingBlocks.len == 0
 
-    trace "Retrieved block"
-    some (blk, blockIdx[idx])
-
-  iter.next = next
-  success iter
+  proc genNext(): Future[(?!bt.Block, int)] {.async.} = 
+    let completedFut = await one(pendingBlocks)
+    pendingBlocks.del(pendingBlocks.find(completedFut))
```
dryajov, line 107: find can return `-1`, which will crash `del`

tests/codex/blockexchange/protobuf/testpresence.nim
```diff
@@ -26,12 +25,7 @@ checksuite "block presence protobuf messages":
     check message.price == @(price.toBytesBE)
 
   test "decodes CID":
-    check Presence.init(message).?cid == cid.some
-
-  test "fails to decode when CID is invalid":
```
dryajov, line 31: Why is this removed?

tests/codex/testerasure.nim
```diff
@@ -22,25 +22,13 @@ asyncchecksuite "Erasure encode/decode":
...
-  var metaDs: SQLiteDatastore
 
   setup:
     rng = Rng.instance()
     chunker = RandomChunker.new(rng, size = dataSetSize, chunkSize = BlockSize)
-    manifest = !Manifest.new(blockSize = BlockSize)
-    repoDs = SQLiteDatastore.new(Memory).tryGet()
-    metaDs = SQLiteDatastore.new(Memory).tryGet()
-    store = RepoStore.new(repoDs, metaDs)
+    store = CacheStore.new(cacheSize = (dataSetSize * 8), chunkSize = BlockSize)
```
dryajov, line 29: I would not use the a different store from the repostore here (or elsewhere). I understand the desire to isolate the tests as much as possible, but the more we rely on concrete components instead of mocks, the more robust our tests are going to be. In this case, the repostore is perfectly usable with an in-memory sql and doesn't require any mocking.

tests/codex/testnode.nim
```diff
@@ -114,7 +103,7 @@ asyncchecksuite "Test Node":
 
     check:
       fetched.cid == manifest.cid
```
dryajov, line 105: Probably comparing the manifests directly is better?

```suggestion
      fetched == manifest
```

codex/stores/blockstore.nim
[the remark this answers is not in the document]
dryajov: > Ad1. But backwards index is a different access pattern and hence a different parameter type. This is what we have in AsyncHeapQueue

OK, not a big deal - we can't use `[]` operator with async and we'll unlikely to use a `BackwardsIndex` either way.

> Ad2. There's a method to get block by address (line 46), but in some cases we don't have an address object in the scope, instead we have separately treeCid and index variables. We could allocate the memory and create the BlockAddress object, but: 1. that seem a bit wasteful and 2. it's an additional inconvenience to the client.

Hmm, that seems like a problem of not using `BlockAddress` consistently rather than anything else?

dryajov (commented):
Looking much better. I'll try to go through it once more, but so far I think this is meargeable.

dryajov:
@CodiumAI-Agent /review

dryajov:
@CodiumAI-Agent /improve --extended

codex/stores/repostore.nim
```diff
@@ -82,6 +77,62 @@ func available*(self: RepoStore): uint =
...
+  self: RepoStore,
+  treeCid: Cid,
+  index: Natural,
+  blockCid: Cid,
+  proof: MerkleProof
+): Future[?!void] {.async.} =
+  ## Put a block to the blockstore
+  ##
+
+  without key =? createBlockCidAndProofMetadataKey(treeCid, index), err:
```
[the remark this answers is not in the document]
tbekas: I don't like such long names either, but I like consistency in a codebase. Procs above this one are of similar length:
```nim
proc createBlockExpirationMetadataKey
proc createBlockExpirationMetadataQueryKey
proc createBlockCidAndProofMetadataKey
```

I could shorten it up to `forCidAndProof` and it would be used like this `Key.forCidAndProof(...)` but it would be different than other procs, so I think I should refactor other procs as well, to keep it consistent. Wdyt?

codex/erasure/erasure.nim
tbekas: This issue was already prior to my changes. But I will fix it though - I think the best handling would be to return a failed future from this method - this will stop the iterator and then it will propagate the error via `Result.failure` back to the caller of decode/encode.

tests/codex/testnode.nim
tbekas: You're right, fixing it.

tests/codex/blockexchange/protobuf/testpresence.nim
tbekas: Because we're no longer decoding from raw bytes to `Cid` inside of `Presence.init` proc. We're decoding it in [message.nim](https://github.com/codex-storage/nim-codex/pull/566/commits/8c1d97d75c97314c626b8544e17fe8b47d2d2613#diff-77bd53597755a1938432f3d6205bbfb4f9754f2ae7350f7faa8fc32f682c3b8cR161) (line 161)

codex/blockexchange/engine/pendingblocks.nim
tbekas: But it works in this case, should I add `[]` anyway?

tests/codex/testerasure.nim
tbekas: Ok, I agree. I'm going to use the same type of stores that were used before.

codex/stores/blockstore.nim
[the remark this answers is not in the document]
tbekas: >Hmm, that seems like a problem of not using BlockAddress consistently rather than anything else?

So in couple of places we have a following pattern:
```nim
for i in 0..<manifest.blocksCount:
   await getBlock(manifest.treeCid, i)
```

Would you like it to be replaced by following code?
```nim
for i in 0..<manifest.blocksCount:
   let address = BlockAddress.init(manifest.treeCid, i)
   await getBlock(address)
```

If so I could get rid of the method `getBlock(treeCid: Cid, index: Natural)`, but in my opinion the 2nd code is simply worse because it's:
- less efficient (new object created on each iteration)
- more verbose than necessary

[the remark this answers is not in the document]
dryajov: Yeah, I understand the reasoning, but IMO having two slightly distinct interfaces will just make the codebase less consistent - IMO, consistency trumps everything else.

We could add an initializer from a tuple for example 

```nim
init(_: type (Cid, Natural)): BlockAddress = ...
```

The point is that we stick with one convention and follow that, otherwise we'll end up with (slightly) distinct looking flows, which will make the codebase harder to read and maintain.

codex/stores/repostore.nim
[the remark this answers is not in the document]
dryajov: Heh, yeah - sorry for bikeshedding, there is already a lot going on here and we can certainly leave it for another PR. That said, I like the above proposal.

codex/blockexchange/engine/pendingblocks.nim
dryajov: Yeah, it doesn't work _always_ and it's also easy to miss that we're really handling a pointer when reading the code.

codex/stores/blockstore.nim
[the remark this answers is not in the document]
tbekas: Alright. I understand the reasoning and I agree that unification has certain benefits (if at some point we will remove fetching individual blocks by just `cid`, we will have less to refactor).

Also there's more to unify if we go this path:

```nim
method ensureExpiry*(self: BlockStore, cid: Cid, expiry: SecondsSince1970)

method delBlock*(self: BlockStore, cid: Cid)
method delBlock*(self: BlockStore, treeCid: Cid, index: Natural)

method hasBlock*(self: BlockStore, cid: Cid)
method hasBlock*(self: BlockStore, tree: Cid, index: Natural)

proc contains*(self: BlockStore, blk: Cid)
proc contains*(self: BlockStore, address: BlockAddress)
````

I could do this as a part of this PR or in a separate one. Wdyt?

**Comment by tbekas** [comment]
I see that integrationTests are not successful, I will try to fix it on Monday.

**Comment by dryajov** [comment]
> I see that integration tests are not successful, I will try to fix it on Monday.

It seems to be failing on marketplace related stuff, which has been very flaky lately.

File: codex/node.nim [hunk 28]
```diff
@@ -108,18 +111,21 @@ proc fetchBatched*(
...
   trace "Fetching blocks in batches of", size = batchSize
-  for blks in manifest.blocks.distribute(max(1, batches), true):
-    try:
-      let
-        blocks = blks.mapIt(node.blockStore.getBlock( it ))
 
+  let iter = Iter.fromSlice(0..<manifest.blocksCount)
+    .map((i: int) => node.blockStore.getBlock(BlockAddress.init(manifest.treeCid, i)))
+
+  for batchNum in 0..<batchCount:
```
dryajov, line 122: This can potentialy throw, we probably want this to in the try.

dryajov (approved):
Lets get this merged, we can address outstanding issues in subsequent PRs

PR #596, opened 2023-10-24
test: integration tests listen on localhost

AuHau:
When I run integration tests on my Mac with some security enabled I get:

image

This fixes it by listening on localhost instead of `0.0.0.0` which is a no-no for macOS 😅

### DIFF
--- a/tests/integration/multinodes.nim
+++ b/tests/integration/multinodes.nim
@@ -86,6 +86,7 @@ template multinodesuite*(name: string,
         "--api-port=" & $(8080 + index),
         "--data-dir=" & datadir,
         "--nat=127.0.0.1",
+        "--listen-addrs=/ip4/127.0.0.1/tcp/0",
         "--disc-ip=127.0.0.1",
         "--disc-port=" & $(8090 + index),
         "--eth-account=" & $accounts[index]]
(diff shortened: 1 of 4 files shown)


markspanbroek (approved):
Looks good to me!

dryajov:
Hold on, I never seen this issue, also, why would it be a problem on macos?

AuHau:
> Hold on, I never seen this issue, also, why would it be a problem on macos?

@dryajov do you have turned on your Firewall? That is most probably the reason. macOS Firewall doesn't like "random binaries" (eq. not signed by Apple certificate) to open internet-wide port listening (e.g. "accept incoming network connections"). If you say "Allow" then it remembers your choice only if the binary does not change, and as Codex binary changes all the time, it is not really a solution ;-) Only listening for localhost connections is...

dryajov:
> > Hold on, I never seen this issue, also, why would it be a problem on macos?
> 
> @dryajov do you have turned on your Firewall? That is most probably the reason. macOS Firewall doesn't like "random binaries" (eq. not signed by Apple certificate) to open internet-wide port listening (e.g. "accept incoming network connections"). If you say "Allow" then it remembers your choice only if the binary does not change, and as Codex binary changes all the time, it is not really a solution ;-) Only listening for localhost connections is...

OK, I see this changes are made to the tests, which makes sense then. I though this was being made to the command line defaults, in which case it's something we definitely don't want 👍

AuHau:
Oh yeah, that would not make sense otherwise 😅 👍

PR #597, opened 2023-10-24
feat: ensure block expiry

AuHau:
This is the first part of the work on https://github.com/codex-storage/nim-codex/issues/436

### DIFF
--- a/codex/stores/networkstore.nim
+++ b/codex/stores/networkstore.nim
@@ -17,6 +17,7 @@ import pkg/libp2p
 
 import ../blocktype as bt
 import ../utils/asyncheapqueue
+import ../clock
 
 import ./blockstore
 import ../blockexchange
@@ -63,6 +64,21 @@ method putBlock*(
   await self.engine.resolveBlocks(@[blk])
   return success()
 
+method updateExpiry*(
+    self: NetworkStore,
+    cid: Cid,
+    expiry: SecondsSince1970
+): Future[?!void] {.async.} =
+  ## Updates block's assosicated expiry TTL in store
+  ##
+
+  without blk =? await self.localStore.getBlock(cid), error:
+    if not (error of BlockNotFoundError): return failure error
+    trace "Updating expiry - block not in local store", cid
+    return
+
+  await self.localStore.updateExpiry(cid, expiry)
+
 method delBlock*(self: NetworkStore, cid: Cid): Future[?!void] =
   ## Delete a block from the blockstore
   ##
@@ -91,8 +107,8 @@ proc new*(
   engine: BlockExcEngine,
   localStore: BlockStore
 ): NetworkStore =
-  ## Create new instance of a NetworkStore 
-  ## 
+  ## Create new instance of a NetworkStore
+  ##
   NetworkStore(
       localStore: localStore,
       engine: engine)
--- a/codex/stores/networkstore.nim
+++ b/codex/stores/networkstore.nim
@@ -64,20 +64,21 @@ method putBlock*(
   await self.engine.resolveBlocks(@[blk])
   return success()
 
-method updateExpiry*(
+method ensureExpiry*(
     self: NetworkStore,
     cid: Cid,
     expiry: SecondsSince1970
 ): Future[?!void] {.async.} =
-  ## Updates block's assosicated expiry TTL in store
+  ## Ensure that block's assosicated expiry is at least given timestamp
+  ## If the current expiry is lower then it is updated to the given one, otherwise it is left intact
   ##
 
-  without blk =? await self.localStore.getBlock(cid), error:
-    if not (error of BlockNotFoundError): return failure error
+  if (await self.localStore.hasBlock(cid)).tryGet:
+    return await self.localStore.ensureExpiry(cid, expiry)
+  else:
     trace "Updating expiry - block not in local store", cid
-    return
 
-  await self.localStore.updateExpiry(cid, expiry)
+  return success()
 
 method delBlock*(self: NetworkStore, cid: Cid): Future[?!void] =
   ## Delete a block from the blockstore
--- a/codex/stores/repostore.nim
+++ b/codex/stores/repostore.nim
@@ -129,21 +129,33 @@ proc getBlockExpirationEntry(
   let duration = ttl |? self.blockTtl
   self.getBlockExpirationEntry(cid, self.clock.now() + duration.seconds)
 
-method updateExpiry*(
+method ensureExpiry*(
     self: RepoStore,
     cid: Cid,
     expiry: SecondsSince1970
 ): Future[?!void] {.async.} =
-  ## Updates block's assosicated expiry TTL in store
+  ## Ensure that block's assosicated expiry is at least given timestamp
+  ## If the current expiry is lower then it is updated to the given one, otherwise it is left intact
   ##
+
   logScope:
     cid = cid
 
-  without blockExpEntry =? self.getBlockExpirationEntry(cid, expiry), err:
+  if expiry <= 0:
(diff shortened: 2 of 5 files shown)


tests/codex/stores/testrepostore.nim
```diff
@@ -231,6 +231,34 @@ asyncchecksuite "RepoStore":
...
       response[0].data == expectedExpiration.toBytes
 
+  test "Should update block expiration timestamp":
+    let
+      duration = 10.seconds
+      blk = createTestBlock(100)
+
+    let
+      expectedExpiration: SecondsSince1970 = 123 + 10
+      updatedExpectedExpiration: SecondsSince1970 = expectedExpiration + 10
```
emizzle, line 241:
```suggestion
    let
      duration = 10
      blk = createTestBlock(100)
      expectedExpiration: SecondsSince1970 = now + duration
      updatedExpectedExpiration: SecondsSince1970 = expectedExpiration + 1
      
      # ...
      
      (await repo.putBlock(blk, some duration.seconds)).tryGet
```

emizzle (commented):
Looks good. A few smaller suggestions.

What happens if a user supplies a negative number to `updateExpiry`?

codex/stores/networkstore.nim
```diff
@@ -63,6 +64,22 @@ method putBlock*(
...
+method ensureExpiry*(
+    self: NetworkStore,
+    cid: Cid,
+    expiry: SecondsSince1970
+): Future[?!void] {.async.} =
+  ## Ensure that block's assosicated expiry is at least given timestamp
+  ## If the current expiry is lower then it is updated to the given one, otherwise it is left intact
+  ##
+
+  if (await self.localStore.hasBlock(cid)).tryGet:
```
benbierens, line 76: the code in the repostore can handle the case where the block is not found, right?
If yes, this method can directly pass to localstore.
If no, it probably should.

tests/codex/stores/testrepostore.nim
```diff
@@ -231,6 +231,80 @@ asyncchecksuite "RepoStore":
...
+
+    (await repo.ensureExpiry(blk.cid, updatedExpectedExpiration)).tryGet
+
+    response = await queryMetaDs(expectedKey)
+
+    check:
+      response.len == 1
+      !response[0].key == expectedKey
+      response[0].data == expectedExpiration.toBytes
+
```
benbierens, line 307: Nice!

codex/stores/repostore.nim
```diff
@@ -107,23 +107,59 @@ method getBlock*(self: RepoStore, cid: Cid): Future[?!Block] {.async.} =
...
+  logScope:
+    cid = cid
+
+  if expiry <= 0:
+    return failure(newException(ValueError, "Expiry timestamp must be larger then zero"))
+
+  let expiryKey = createBlockExpirationMetadataKey(cid)
+  let currentExpiry = await self.metaDs.get(expiryKey.tryGet)
+
+  if currentExpiry.isOk and expiry < (!currentExpiry).toSecondsSince1970:
```
tbekas, line 150: I think that just treating any error as a missing value is wrong. You should check if error is an instance of `DatastoreKeyNotFound`, if so treat it as missing value, otherwise you should propagate the error to the caller, so `return failure(err)`.

codex/stores/networkstore.nim
```diff
@@ -63,6 +64,22 @@ method putBlock*(
...
+    return await self.localStore.ensureExpiry(cid, expiry)
+  else:
+    trace "Updating expiry - block not in local store", cid
```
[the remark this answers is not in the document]
AuHau: I disagree. If this is called with CID of not locally stored block, then it will create an expiry entry for a non-existing block, which might lead to potential unexpected behavior. Will that happen? Most probably not, but still, this is IMHO safer approach.

codex/stores/repostore.nim
AuHau: Good point, thanks! 👍

codex/stores/networkstore.nim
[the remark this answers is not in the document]
AuHau: My previous comment is bit wrong in the point of creating expiry for non-existing blocks - that should not happen. But I still think that this part of the Network Store implementation is correct.

@benbierens, what do you mean by "handle non-existing blocks"? Fail or ignore for blocks not found?

IMHO in NetworkStore we should ignore non-local (e.g. non-existing) blocks, but in the context of LocalStore IMHO we should fail for non-existing blocks... WDYT?

codex/stores/repostore.nim
```diff
@@ -144,18 +144,22 @@ method ensureExpiry*(
...
-  let currentExpiry = await self.metaDs.get(expiryKey.tryGet)
+  without expiryKey =? createBlockExpirationMetadataKey(cid), err:
+    return failure(err)
 
-  if currentExpiry.isOk and expiry < (!currentExpiry).toSecondsSince1970:
-    trace "Current expiry is larger then the specified one, no action needed"
-    return success()
+  without currentExpiry =? await self.metaDs.get(expiryKey), err:
+    if err of DatastoreKeyNotFound:
+      error "No current expiry exists for the block"
```
tbekas, line 152: In this branch I think you should return:
`failure(newException(BlockNotFoundError, err.msg))`

codex/stores/repostore.nim
```diff
@@ -144,18 +144,22 @@ method ensureExpiry*(
...
+    else:
+      error "Could not read datastore key", err = err.msg
 
-  without (key, data) =? self.getBlockExpirationEntry(cid, expiry), err:
-    trace "Unable to create updated expiration metadata key", err = err.msg
     return failure(err)
 
-  if err =? (await self.metaDs.put(key, data)).errorOption:
+  if expiry < currentExpiry.toSecondsSince1970:
+    trace "Current expiry is larger then the specified one, no action needed"
```
tbekas, line 159: A minior change:
```nim
  if expiry <= currentExpiry.toSecondsSince1970:
    trace "Current expiry not shorter then the specified one, no action needed"
```

codex/stores/networkstore.nim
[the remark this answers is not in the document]
tbekas: As stated previously, I don't think there's a nice implementation of this method in the network store, as we cannot modify what's stored in the network, only locally.

tbekas (changes requested):
Small comments, otherwise looks good.

codex/stores/networkstore.nim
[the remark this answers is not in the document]
AuHau: Sooo I keep it? 😅

codex/stores/repostore.nim
AuHau: It is maybe bit long stretch, but agree

codex/stores/repostore.nim
AuHau: Ok, btw. do you know the "Suggestion" feature for PRs? ;-) This would be great adept for that, for example...

https://haacked.com/archive/2019/06/03/suggested-changes/

tbekas (approved):
LGTM

codex/stores/networkstore.nim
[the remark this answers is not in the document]
tbekas: You can keep it I think :) Maybe we will remove it in the future when we decide to decouple networkstore from the blockstore interface.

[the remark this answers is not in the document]
benbierens: We have noticed in several places that the blockstore interface is causing us trouble and not giving us any or enough benefits. So it should go. I agree there's no nice way to implement this.
The repostore implementation though should definitely not allow the creation of metadata for non-existing blocks.

benbierens (approved):
Yeah!

PR #609, opened 2023-11-09
Storage/purchases endpoint

benbierens:


openapi.yaml
```diff
@@ -412,6 +412,50 @@ paths:
         "503":
           description: Sales are unavailable
 
+  "/storage/request/{cid}":
```
dryajov, line 415: I would generally change all the api to conform to `/noun/{specific-resource}/noun`. 

For example a `POST` to `/storage/{cid}` creates a request for storage, `GET` to `/storage/purchases` to list the purchase `id`s, `GET/POST/DELETE` to ``GET` to `/storage/purchases/{id}` to manipulate a specific purchase id.

dryajov: for reference take a look at #612

benbierens: I'm confused, simply because of the terms.
What we're creating when we POST, is a 'storage request'. This yields a 'purchase ID', which lets you see the state of your purchase.

If we change it to POST `/storage/{cid}`, don't we make it a lot less clear that this is the creation of a request-for-storage?
If we consider `/storage` to be the root path of the storage API, then we have `/request/{cid}`, `/purchases`, and `/purchases/{cid}`, which seems to fit pretty nicely with your template, doesn't it?

dryajov: Usually, you want aspects (nouns) of the specific resource to come after a specific entity name/identifier and the HTTP methods themself identify some CRUD operation.

But maybe you're right that it works just fine in this case and we have `/storage/requests` to list all the requests and `/storage/requests/{cid}` to create a new request, etc...

AuHau (approved):
LGTM 👍

PR #612, opened 2023-11-09
Fix REST endpoints semantics

dryajov:
This is an attempt to fix the semantics of the `/data` endpoints. 

- `/data` allows operating on all the data resources under it as well as creating _new_ resources (i.e. uploading a file/data)
  - A `GET` request to the `/data` endpoint should list all the top level `cid`s (i.e. manifest, etc...)
  - A `POST` request to the `/data` endpoint should accept a file upload and create a new top level `cid`
- `/data/{cid}` endpoint identifies a specific resource under the `/data` endpoint
  - A `GET` request to `/data/{cid}` will attempt to retrieve the cid from the **local** node, if it doesn't exist in the local node, it should respond with a `404`
  - A `GET` request to `/data/{cid}/network` will attempt to retrieve `cid` from the **network** if it doesn't exist locally

The `/data/{cid}` and `/data/{cid}/network` seems to be a better default than `/data/{cid}/local` or similar, which would default issuing a network request, which should be an explicit operation.

All our endpoints should eventually follow this convention - which are standard REST conventions.


tbekas:
@dryajov 
How about instead of doing `data/{cid}/network` we would do `/data/{cid}?network=true` (or alternatively `?local=false`)?

The common pattern in URL design is: `resource/identifier/resource`. Endpoint `data/{cid}/network` would indicate that `network` is a subresource of `data`.

dryajov:
> @dryajov How about instead of doing `data/{cid}/network` we would do `/data/{cid}?network=true` (or alternatively `?local=false`)?
> 
> The common pattern in URL design is: `resource/identifier/resource`. Endpoint `data/{cid}/network` would indicate that `network` is a subresource of `data`.

I'd argue that `network` is a resource. Query params are betters suited as request modifiers, like pagination or optional information in the response they don't usually change where or how the request is generated, which is the case here. Resource identifiers don't always have to imply a hierarchy, it can also be used to indicate a distinct aspect of the subresource, such as network or local.

tests/integration/testIntegration.nim
```diff
@@ -19,6 +19,7 @@ import ./twonodes
 # You can also pass a string in same format like for the `--log-level` parameter
 # to enable custom logging levels for specific topics like: debug2 = "INFO; TRACE: marketplace"
 
+# twonodessuite "Integration tests", debug1 = "TRACE", debug2 = "TRACE":
```
benbierens, line 22: just leftover

tests/integration/testblockexpiration.nim
```diff
@@ -44,19 +44,21 @@ ethersuite "Node block expiration tests":
...
+      contentId & (if local: "" else: "/network")
+
     let content = client.get(downloadUrl)
     client.close()
     content
 
   proc hasFile(contentId: string): bool =
     let client = newHttpClient(timeout=3000)
-    let dataLocalUrl = baseurl & "/local"
+    let dataLocalUrl = baseurl & "/data/" & contentId
```
benbierens, line 58: I think this proc was intended to check the local-content-listing API. Downloading the file sure checks that the file is there. But I suppose we should have another check then to see that the content-listing API is OK.

dryajov: Yeah, that makes sense.

We might also further refine the api by making `/data/{cid}` return a `200/404` response codes to check for presence, and then have dedicated `/data/{cid}/network` and `/data/{cid}/local` to fetch the file. I'm not that concerned with the namespasing as long as it's descriptive.

dryajov: oh, btw - in this particular case, there should have been an explicit test to check the `listing` api endpoint, I can add it.

AuHau:
I am against this PR. IMHO, this makes the API much less intuitive and understandable in my eyes. Can you please first elaborate on why you think that the current `/data` endpoint and its semantics are broken/wrong?

When I think about APIs, I think about it from the user perspective, how they will understand and use the API. Quite sometimes, I see a more "involved" approach from the people who are very deep into the project and its technical side, designing APIs around the tech without keeping the users in mind.

The biggest problem I see here is the division between "local" and "network" data fetch. In my expectation, I think 95% use cases will be about fetching the data without really caring if it is local or from the network. Most probably, most of the time, they will be fetched from the network. At the same time, upload/download data will be the most common and basic use case we provide, and hence, I think the API should be as simple as possible. I am not against supporting "local-only" data fetch, but I would go about it with the same proposal as @tbekas suggested - having query parameter `?local=true` which would fetch it only from the local store if available.

Another thing I am not in favor of is the `/data` listing. I have raised the same point with the original PR. "Listing data" here really means "listing local manifests", but having that on `/data` endpoint would suggest that it could list all the content in the network. I am not sure what benefit would have listing local manifests on the `/data` endpoint? Over `/local`? Maybe we could improve on the naming with `/local-manifests`? Although not sure if that is really needed.

In general, I would suggest to keep the API as flat and simple as possible. Instead of using multi-level path nestings, have more endpoints. Support special and edge-cases features using query parameters. 

Whatever path we take, please don't forget to update the OpenAPI definition when modifying API ;-)

dryajov:
> I am against this PR. IMHO, this makes the API much less intuitive and understandable in my eyes. Can you please first elaborate on why you think that the current `/data` endpoint and its semantics are broken/wrong?
> 
> When I think about APIs, I think about it from the user perspective, how they will understand and use the API. Quite sometimes, I see a more "involved" approach from the people who are very deep into the project and its technical side, designing APIs around the tech without keeping the users in mind.
> 
> The biggest problem I see here is the division between "local" and "network" data fetch. In my expectation, I think 95% use cases will be about fetching the data without really caring if it is local or from the network. Most probably, most of the time, they will be fetched from the network. At the same time, upload/download data will be the most common and basic use case we provide, and hence, I think the API should be as simple as possible. I am not against supporting "local-only" data fetch, but I would go about it with the same proposal as @tbekas suggested - having query parameter `?local=true` which would fetch it only from the local store if available.
> 
> Another thing I am not in favor of is the `/data` listing. I have raised the same point with the original PR. "Listing data" here really means "listing local manifests", but having that on `/data` endpoint would suggest that it could list all the content in the network. I am not sure what benefit would have listing local manifests on the `/data` endpoint? Over `/local`? Maybe we could improve on the naming with `/local-manifests`? Although not sure if that is really needed.
> 
> In general, I would suggest to keep the API as flat and simple as possible. Instead of using multi-level path nestings, have more endpoints. Support special and edge-cases features using query parameters.
> 
> Whatever path we take, please don't forget to update the OpenAPI definition when modifying API ;-)

There is a good reason why presto, the rest library we use, doesn't support the same prefix url for multiple endpoints - it's semantically wrong and against several principles of REST api design, it isn't a limitation of the framework, it is an intentional design decision, which I support as well.

> The biggest problem I see here is the division between "local" and "network" data fetch. In my expectation, I think 95% use cases will be about fetching the data without really caring if it is local or from the network.

This is a question of sane defaults, which should be, above all, _safe_. 

There should be a clear separation between local and network requests. Id argue that defaulting all operations to local first is safer in principle and in practice. We might argue that it isn't consistent right now, but this is exactly why the PR exists, to fix that.

> Another thing I am not in favor of is the `/data` listing. I have raised the same point with the original PR. "Listing data" here really means "listing local manifests", but having that on `/data` endpoint would suggest that it could list all the content in the network. I am not sure what benefit would have listing local manifests on the `/data` endpoint? Over `/local`? Maybe we could improve on the naming with `/local-manifests`? Although not sure if that is really needed.

See above - if everything defaults to local, retrieving a list of _local_ CIDs from the `/data` endpoint on a `GET`, would be consistent with the rest of the system/model.

Lastly, the endpoints are intended to be used by applications, not end users and from a developer standpoint, as long as they are consistent, it should be fine. Eventually, we should provide a client library or one can be generated from the swagger file.

AuHau:
> There is a good reason why presto, the rest library we use, doesn't support the same prefix url for multiple endpoints - it's semantically wrong and against several principles of REST api design, it isn't a limitation of the framework, it is an intentional design decision, which I support as well.

I'm not sure how this comment is related here. I know that at some point we talked about the `/data/{cid}` and `/data/local`, which is not possible with Presto. However, I am not bringing up this point here at all.

> This is a question of sane defaults, which should be, above all, safe.
There should be a clear separation between local and network requests. Id argue that defaulting all operations to local first is safer in principle and in practice. We might argue that it isn't consistent right now, but this is exactly why the PR exists, to fix that.

Why do you think that doing network fetch is "not safe"? Why do you think we should separate these concepts? From the decentralized storage client's APIs I worked with, I have never seen this separation. The only thing I can think of is that in the future, network fetch will result in bandwidth costs, but that is not yet implemented. Even if it were implemented, I would still argument against this separation. And yes, network fetch is more costly on network bandwidth and resources, but if that would be a concern to the user then he maybe should not run the full node himself and use something like "light-client" (in the future).

From API use-cases the only place where I could see "local fetch" and "local listing manifests" is some node management software, where you want to potentially see and manage data stored locally. But I would argue that it is minor use-case (which we still should support for sure!). But the main use case of Codex (and its API) is that you share data **with the network** and you download data **from the network** and that is why I think we should optimize for this use case even on the API level.

If I were a third-party developer working with this API, I would just default to use `/data/{cid}/network` route, because it will _just-work™️_ and behave as I would want (fetch locally if possible, and if not then from network) and so I would prefer to just simplify this for everybody...

> Lastly, the endpoints are intended to be used by applications, not end users and from a developer standpoint, as long as they are consistent, it should be fine. Eventually, we should provide a client library or one can be generated from the swagger file.

Of course, when I talk about "users" I mean developers, but they are users as well. And having a simple API leads to having good developer experience. You are right though, that most probably there won't be many people interacting directly with the API, but rather some client library.

tbekas:
>From API use-cases the only place where I could see "local fetch" and "local listing manifests" is some node management software (...) the main use case of Codex (and its API) is that you share data with the network and you download data from the network and that is why I think we should optimize for this use case even on the API level.

You've raised a very good point I think.

Also adding another perspective: fetching data from local disk instead of the network is just an optimisation (a detail), it shouldn't change the address of the resource.

dryajov:
> Why do you think that doing network fetch is "not safe"?

> is that in the future, network fetch will result in bandwidth costs, but that is not yet implemented.

So, you're saying that inadvertently spending the user's funds, is no big deal? I'm quite surprised by this reasoning. 

Of course, this isn't the only issue, albeit perhaps the most severe. The point is that there are two completely different flows with different execution and cost profiles. This needs to be somehow reflected in the interaction, one way is through a more fine grained endpoint, another is to use a query parameter.

> Also adding another perspective: fetching data from local disk instead of the network is just an optimisation (a detail), it shouldn't change the address of the resource.

Is it just an implementation detail? In this context without the separation, you will inadvertently trigger separate flows that have very different consequences for the user, both in performance and cost. This sin't a simple implementation detail and should be reflected by the resource name.

I'm willing to concede that either a query parameter or an endpoint are both valid in this context, but a dedicated endpoint is more appropriate because it has a better semantic fit in this context.

AuHau:
> So, you're saying that inadvertently spending the user's funds, is no big deal? I'm quite surprised by this reasoning.

No, because it is not "inadvertently". If the user calls the API, then they are making a conscious request to fetch the data they want. Yes, there is some chance that it will already be locally stored (or parts of the dataset), which is great as it will save them money. However, this probability is very low, so most likely Codex will have to perform a network fetch.

Again, I have a hard time imagining a case where the user would upload their data to their node and then would want to exclusively download their local data. I mean, for those cases, there are other systems with many proven years in production for local network storage. I have a hard time believing Codex would be used like this, but in any way, I am not saying that it should not be possible. Only that this is such a marginal use-case that we should design for the main use-case, which is network downloading/uploading.

I agree that there should be a way to make the user aware of the costs related to fetching datasets, but in my mind, this is a task of UI, which of course needs to be supported by API, but we are not there yet...

dryajov:
> > So, you're saying that inadvertently spending the user's funds, is no big deal? I'm quite surprised by this reasoning.
> 
> No, because it is not "inadvertently". If the user calls the API, then they are making a conscious request to fetch the data they want. Yes, there is some chance that it will already be locally stored (or parts of the dataset), which is great as it will save them money. However, this probability is very low, so most likely Codex will have to perform a network fetch.

This is where you're wrong - REST is about semantics, if the semantics are wrong, then you're not doing REST. The semantics should maximally reflect the meaning of the interaction. 

> Again, I have a hard time imagining a case where the user would upload their data to their node and then would want to exclusively download their local data. I mean, for those cases, there are other systems with many proven years in production for local network storage. I have a hard time believing Codex would be used like this, but in any way, I am not saying that it should not be possible. Only that this is such a marginal use-case that we should design for the main use-case, which is network downloading/uploading.

This isn't just about usability, this point is being missed.

> I agree that there should be a way to make the user aware of the costs related to fetching datasets, but in my mind, this is a task of UI, which of course needs to be supported by API, but we are not there yet...

It isn't, the whole point of RESTful APIs is about correctly communicating the meaning of the interaction, otherwise there is no point to sticking to REST at all, just use POST and write RPC style endpoints.

emizzle:
> the whole point of RESTful APIs is about correctly communicating the meaning of the interaction

What's the reasoning behind adhering to a strict RESTful API?

dryajov:
> > the whole point of RESTful APIs is about correctly communicating the meaning of the interaction
> 
> What's the reasoning behind adhering to a strict RESTful API?

Otherwise you end up fighting the underlying framework/tooling for no good reason. 

However, this discussion has mostly coalesced around defaults and semantics more than anything else. 

The main question as I see it is - do we want everything to default to network or local?

I'm of the opinion that defaulting to local is safer, both for future use cases, where network requests will cost users funds, and so, this change will have to happen at some point in the future either way. And from a design perspective - where network vs local isn't simple an implementation detail, it is an important aspect of the call that implies fundamentally different cost vs performance characteristics and thus, its better represented by a path element rather than a query parameter.

emizzle:
>  I know that at some point we talked about the /data/{cid} and /data/local, which is not possible with Presto

Are there particular RESTful design guidelines that Presto follows? There are quite a [lot](https://restful-api-design.readthedocs.io/en/latest/urls.html) [of](https://github.com/byrondover/api-guidelines/blob/master/Guidelines.md) [variances](https://github.com/microsoft/api-guidelines) in guidelines on the internet from what i can find. I can't seem to find any guidelines that indicate `/data/{cid}` and `/data/{cid}/local` shouldn't be used 🤷 .

> this change will have to happen at some point in the future either way

Perhaps we should implement a best guess for now (for testnet launch) and revisit this later? The reason I say this is because we do not have a complete bandwidth incentives picture yet, and the approach we land on now could change with the final bandwidth incentives design.

## My 2c
FWIW, my opinion here is that Codex is a network-centric protocol, interacting with the Codex network should be the default, and the interaction should be as frictionless as possible. From an application's/user's perspective (outside-in API design approach), they are going to care about a few things:
1. That they get the data they want to download
2. The performance of the query (local only)
3. The cost of the query

### Getting the data they want to download
If we default to local, the most important point may not be adhered to: the application/user may not even get the data they are looking for. To ensure they get what they are looking for when downloading, my opinion is the default behaviour (for `/data/{cid}`) should be to check local first (fast) and then check the network, which would incur a performance penalty and potentially a cost (see below).

### Query performance
Documenting here all the options 
In the case where they want to optimise performance, eg only fetch local data, we could use (in order of preference, some have already been mentioned, but adding here for completeness and API design reasoning):
1. [variant format](https://restful-api-design.readthedocs.io/en/latest/urls.html#variants) (eg `/data/{cid};local`):
> The use of the semicolon to provide options that are specific to a path segment is explicitly alowed in [RFC3986](http://tools.ietf.org/html/rfc3986#section-3.3) The advantage over using a ”?variant” query argument is that this format is specific to a path segment only, and would allow you to e.g. work on the saved variant of a sub-resource in a sub-collection of the current variant of another resource
2. [a controller](https://github.com/byrondover/api-guidelines/blob/master/Guidelines.md#controller) or [action](https://restful-api-design.readthedocs.io/en/latest/methods.html#actions) (eg `/data/{cid}/local`) as it is a separate action to be performed internally (although this kind of disgresses from the outside-in approach of API design)
2. query parameter (eg `?local=true` or `?fast=true`) as this could be considered a filter, with `fast` possibly being more descriptive to the concerns of an application/user than `local`
3. a separate endpoint (eg `/local/{cid}` or `/fast/{cid}`) as this could be considered its own collection in a way, but it's an internal separation, so again digresses from the outside-in approach.

### Query cost
Regarding spending of funds, requesting various cids could end up costing a lot of money depending on how big the dataset is, so when it comes to bandwidth incentives design, there might need to be a way to estimate cost before the download occurs, which could be its own [action](https://restful-api-design.readthedocs.io/en/latest/methods.html#actions) endpoint (eg `/data/{cid}/estimate`, and `/data/{cid};local/estimate` would always return `0`). An application/user would get an estimate first, before executing the download. If we want to avoid inadvertent spending of funds via download, then perhaps this could be an opt-in feature somehow (via config, its own REST endpoint, cli flag, etc), however we need to consider that it would affect the node's ability to perform provider (host) services. This is why I think we should hold off final design discussions until the bandwidth incentive design is finalised.

dryajov:
> Perhaps we should implement a best guess for now (for testnet launch) and revisit this later?

This is what we've been doing so far, but we've accumulated a baseline of network endpoints that require a bit more structure. I see endpoints being added left and right and we're running into "limitations" of the framework. This to me is an indication that we've reached some critical point where we need to do a bit of cleanup. 

But to put things into perspective, let me summarize the type arguments I've seen brought up so far:

1. Convenience/usability - these revolve around defaults from a user/developer stand point
2. Correctness - these revolve around the semantics that the structure of the endpoint communicates
3. Sane defaults - these revolve around default behaviors. I see this connected to the above two argument and more or less split into two contrasting standpoints:
   1. Safety centric - safety over everything, including usability 
   2. Usability centric - usability over everything, including safety

They aren't mutually exclusive, but they are intertwined and they do affect each other. In order to reach some sort of a common ground we need to organize this discussion around each individual topic/argument.

I'll summarize my reasoning arround this aspects bellow.

1. **Sane defaults:** This is probably the most important aspect, I come at this from a safety centric standpoint, even if it sacrifices usability, tho it doesn't have to. My reasons are:
   1. We're building software that deals with user funds and as such it should avoid unexpected or runaway costs. Thus defaulting to local makes sense because it makes network level interactions, that potentially involve spending user funds, less likely.
   2. Performance. In our case, each network level invocation is costlier than a local invocation, thus defaulting to local makes sense, because it reduces the potential cost to the system in general.
   3. Finally, there are security considerations. Even in a network centric application such as this, an unwanted network request could have potential undesirable consequences for the user. Making network requests explicit is the best course of action as it reduces both our and the users liability.
   4. With that said, none of the above affect ussability, because it doesn't preclude network requests it simply make them explicit.
2. **Correctness:** Whats correct in the context of RESTful apis is contested and non-standard, but there are a few commonalities across the spectrum. I think this [resource](https://restful-api-design.readthedocs.io/en/latest/urls.html) (thanks for citing it) describes what I want to see quite well. In addition the guidelines outline there, a requirement I have is that the endpoint should maximally describe the specific resource being interacted with, in this conext I consider a path element more appropriate because:
   1. It communicates clearly that we're dealing with a resource from the network, rather than local
   2. In contrast to a query parameter, which is more appropriate as a request modifier, e.g. pagination, optional fields, etc...
   3. Variants could be another option, but they are definitely less common and will undoubtly confuse users even more. Still an interesting alternative to the path variant.
3. **Convenience/Usability:** Finally, none of the design decision being made here affect convenience or usability. More on this bellow.

-----------------------

> FWIW, my opinion here is that Codex is a network-centric protocol, interacting with the Codex network should be the default,

I see this claim being made several times, but I don't see any clear articulation of the reason why, network-centric by itself, doesn't imply that all interactions should default to the network on its own.

> and the interaction should be as frictionless as possible

I don't see how this proposal adds friction in any way, this simply requires an **explicit** "marker" in the url string that will trigger the desired flow. I believe I have articulated the reasoning behind the "marker" it at the very least, consistently.

> From an application's/user's perspective (outside-in API design approach)

I think the outside in design argument is a bit out of place. It might apply to commercial api offerings, such as google maps, where you're competing with other similar producs on a pure api offerings grounds. However, this isn't the case for us, we have other more important considerations at play here than simply a nice API. One such consideration, as I already mentioned is performance, after all, you're running this on some user's computer and have 0 control over their setup and as such you can't hide all the inherent complexity that such an approach brings, behind layers and layers of middleware and infrastructure.

> they are going to care about a few things:
>
> 1. That they get the data they want to download
> 2. The performance of the query (local only)
> 3. The cost of the query

> If we default to local, the most important point may not be adhered to: the application/user may not even get the data they are looking for. To ensure they get what they are looking for when downloading, my opinion is the default behaviour (for /data/{cid}) should be to check local first (fast) and then check the network, which would incur a performance penalty and potentially a cost (see below).

Again, I don't see any coherent reasoning as to why it should automaticaly fetch from the network, this is so far, as you yourself note, an opinion but without any coherent argumentation.

> there might need to be a way to estimate cost before the download occurs

Good point, but orthogonal to this discussion I think?

To conclude, I haven't seen any good reasons to default to network, besides feelings or opinions (no sarcasm here), I'm still unconvinced that this isn't just that, feelings and oppinions.

P.S.

Thanks for the links, they are quite useful!

tests/integration/testblockexpiration.nim
dryajov: I've added a dedicated test for the `listing` endpoint.

AuHau:
I won't be arguing here anymore because it takes me too much time to formulate my thoughts in this format. I agree with Eric that now is not a great time for this discussion (both because we don't have time and there are unknowns around bandwidth incentive design), but as Dmitry seems to be set on this approach, I am fine with going forward and revisit it later preferably in-person, or at least in call-format.

PR #614, opened 2023-11-13
Fixes json formatting for NodeId

benbierens:


AuHau (approved):
LGTM

dryajov (approved):

benbierens:
OK yeah so,
NodeId is a UInt256, and it's not distinct. So defining the JSON method for it defined it for all UInt256 and that's why the tests are failing. So I guess these two lines are gonne need some more time.

codex/rest/json.nim
```diff
@@ -39,7 +39,7 @@ type
     manifest* {.serialize.}: Manifest
 
   RestNode* = object
-    nodeId* {.serialize.}: NodeId
+    nodeId* {.serialize.}: RestNodeId
```
dryajov, line 42: Why would you need a separate object for `NodeId` since it's already a `UInt256`, we simply need to add a serialized /deserializer for it.

dryajov: > OK yeah so, NodeId is a UInt256, and it's not distinct. So defining the JSON method for it defined it for all UInt256 and that's why the tests are failing. So I guess these two lines are gonne need some more time.

I think having a serializer for `UInt256` is totally fine, usually you want to serialize it as `hex`, but we could also add modifiers to `serialize` that specify how we'd like to serialize it.

benbierens: `NodeId` is a type-alias for `UInt256`. So the serialize method I originally added for NodeId (which outputs as hex) was being applied to all UInt256, messing up other API calls where UInt256 is used and we don't want to use hex.
If you know how I can apply the .serialize. attribute and set a custom function, please let me know.

dryajov: How is `UInt256` serialized in other places, it should default to `hex` as well, this is the customary way of serializing such types. This is also the point where we might have to add attributes to the `serialize` annotation, we could have something like `serialize: hex|xhex` for non-refixed and prefixed `0x0000...`, `serialize: dec`, etc...

PR #619, opened 2023-11-14
feat: update expiry when data downloaded and slot filled

### LINKED ISSUE #436: Reservations module: set Block TTL to request expiry, then update to request end once started

AuHau:
Closes #436


codex/clock.nim
```diff
@@ -13,6 +14,12 @@ method now*(clock: Clock): SecondsSince1970 {.base, upraises: [].} =
 method waitUntil*(clock: Clock, time: SecondsSince1970) {.base, async.} =
   raiseAssert "not implemented"
 
+method start*(clock: Clock) {.base, async.} =
+  discard
+
+method stop*(clock: Clock) {.base, async.} =
+  discard
```
AuHau, line 21: Added these, so we could use the base `Clock` type everywhere (and hence the `MockClock`), also in the contract's components. It was needed because of the `OnChainClock` implementation.

codex/stores/blockstore.nim
```diff
@@ -88,6 +88,18 @@ method ensureExpiry*(
 
   raiseAssert("Not implemented!")
 
+method ensureExpiry*(
```
AuHau, line 91: Had to extend the `ensureExpiry` with another implementation that adapts to the Merkelization PR that was merged recently

tbekas (approved):
Looks good to me, but please wait for a review from marketplace team member before merging, as I'm not competent enough have a say about marketplace code.

AuHau:
What changes? I thought we agreed on me adding this during the Maintenance call...
I would prefer to merge it and then when the changes are made, the work can be tweaked by the person who will be making the changes...

AuHau:
Or you mean this PR? https://github.com/codex-storage/nim-codex/pull/623

tbekas:
@dryajov do you mean changes for the expiration to be per `treeCid`? I think it shouldn't block this change, specifically that we want to have working software for testnet release, independent of what's planned in the future.

codex/clock.nim
emizzle: Is there a reason they're `async`? The `OnChainClock` implementation doesn't seem to need the async.

codex/stores/repostore.nim
```diff
@@ -234,7 +234,7 @@ method ensureExpiry*(
       return failure(err)
 
   if expiry <= currentExpiry.toSecondsSince1970:
-    trace "Current expiry is larger then the specified one, no action needed"
+    trace "Current expiry is larger then the specified one, no action needed", current = currentExpiry.toSecondsSince1970, ensuring = expiry
```
emizzle, line 237:
```suggestion
    trace "Current expiry is larger than or equal to the specified one, no action needed", current = currentExpiry.toSecondsSince1970, ensuring = expiry
```

codex/node.nim
```diff
@@ -104,11 +104,28 @@ proc fetchManifest*(
...
+
+  return success()
+
 proc fetchBatched*(
   node: CodexNodeRef,
   manifest: Manifest,
   batchSize = FetchBatch,
-  onBatch: BatchProc = nil): Future[?!void] {.async, gcsafe.} =
+  onBatch: BatchProc = nil,
+  expiry: ?SecondsSince1970 = SecondsSince1970.none): Future[?!void] {.async, gcsafe.} =
```
emizzle, line 128:
```suggestion
  expiry = SecondsSince1970.none): Future[?!void] {.async, gcsafe.} =
```

codex/node.nim
```diff
@@ -415,13 +436,21 @@ proc start*(node: CodexNodeRef) {.async.} =
       # since fetching of blocks will have to be selective according
       # to a combination of parameters, such as node slot position
       # and dataset geometry
-      if fetchErr =? (await node.fetchBatched(manifest, onBatch = onBatch)).errorOption:
+      if fetchErr =? (await node.fetchBatched(manifest, onBatch = onBatch, expiry = some request.expiry.toSecondsSince1970)).errorOption:
         let error = newException(CodexError, "Unable to retrieve blocks")
         error.parent = fetchErr
         return failure(error)
```
emizzle, line 442: I find it a bit odd to be updating a block's expiry inside of `fetchBatched`. I would imagine a caller could call `fetchBatched` without understanding that it would iterate all blocks and update their expiry. Perhaps it might be better to have `fetchBatched` return `seq[Blocks]` and then `ensureExpiry` called on that sequence after `fetchBatched` completes?
```nim
without blocks =? (await node.fetchBatched(manifest, onBatch = onBatch)), fetchErr:
  let error = newException(CodexError, "Unable to retrieve blocks")
  error.parent = fetchErr
  return failure(error)

await allFuturesThrowing(blocks.mapIt(node.blockStore.ensureExpiry(it.read.get.cid, request.expiry.toSecondsSince1970)))
```

Additionally, using `.get.cid` may crash the node with a defect if `it.read` doesn't contain a block, so possibly could flesh this out a bit more, which may also help with readability:
```nim
await allFuturesThrowing(
  blocks.map(
    proc(blok: Future[?!Block]): Future[?!void] = (
      without blk =? await blok, err:
        return failure(err)
      return node.blockStore.ensureExpiry(blk.cid, request.expiry.toSecondsSince1970)
    )
  )
)
```

emizzle (commented):
Thanks Adam. Most looks good, some minor comments in the code. Thanks for adding the comments inline in the code, they helped a lot.

codex/clock.nim
AuHau: I see now why it is async because of the `await clock.provider.(un)subscribe(...)` calls so, it needs to stay `async`...

codex/stores/blockstore.nim
```diff
@@ -88,6 +88,18 @@ method ensureExpiry*(
...
+    self: BlockStore,
+    treeCid: Cid,
+    index: Natural,
+    expiry: SecondsSince1970
+): Future[?!void] {.base.} =
+  ## Ensure that block's assosicated expiry is at least given timestamp
```
markspanbroek, line 97:
```suggestion
  ## Ensure that block's associated expiry is at least given timestamp
```

codex/stores/cachestore.nim
```diff
@@ -230,6 +230,17 @@ method ensureExpiry*(
 
   discard # CacheStore does not have notion of TTL
 
+method ensureExpiry*(
+    self: CacheStore,
+    treeCid: Cid,
+    index: Natural,
+    expiry: SecondsSince1970
+): Future[?!void] {.async.} =
+  ## Updates block's assosicated TTL in store - not applicable for CacheStore
```
markspanbroek, line 239:
```suggestion
  ## Updates block's associated TTL in store - not applicable for CacheStore
```

codex/stores/networkstore.nim
```diff
@@ -108,6 +108,23 @@ method ensureExpiry*(
 
   return success()
 
+method ensureExpiry*(
+    self: NetworkStore,
+    treeCid: Cid,
+    index: Natural,
+    expiry: SecondsSince1970
+): Future[?!void] {.async.} =
+  ## Ensure that block's assosicated expiry is at least given timestamp
```
markspanbroek, line 117:
```suggestion
  ## Ensure that block's associated expiry is at least given timestamp
```

codex/stores/networkstore.nim
```diff
@@ -108,6 +108,23 @@ method ensureExpiry*(
...
+  ## If the current expiry is lower then it is updated to the given one, otherwise it is left intact
+  ##
+
+  if (await self.localStore.hasBlock(treeCid, index)).tryGet:
```
markspanbroek, line 121: `tryGet` is dangerous here, it will raise an exception, instead of returning a failed result.

As a general rule, I'd avoid `get`, `tryGet`, `!`, etc. except in tests.

codex/stores/repostore.nim
```diff
@@ -243,6 +243,20 @@ method ensureExpiry*(
 
   return success()
 
+method ensureExpiry*(
+    self: RepoStore,
+    treeCid: Cid,
+    index: Natural,
+    expiry: SecondsSince1970
+): Future[?!void] {.async.} =
+  ## Ensure that block's assosicated expiry is at least given timestamp
```
markspanbroek, line 252:
```suggestion
  ## Ensure that block's associated expiry is at least given timestamp
```

tests/codex/testnode.nim
```diff
@@ -163,3 +195,112 @@ asyncchecksuite "Test Node":
...
+    store: NetworkStore
+    sales: Sales
+    node: CodexNodeRef
+    blockDiscovery: Discovery
+    peerStore: PeerCtxStore
+    pendingBlocks: PendingBlocksManager
+    discovery: DiscoveryEngine
+    manifest: Manifest
+    manifestCid: string
+
```
markspanbroek, line 224: This is a lot of duplication, can we do this in one place?

tests/codex/testnode.nim
```diff
@@ -163,3 +195,112 @@ asyncchecksuite "Test Node":
...
+    close(file)
+    await node.stop()
+
+  test "onExpiryUpdate callback":
+    let
+      # The blocks have set default TTL, so in order to update it we have to have larger TTL
+      expectedExpiry: SecondsSince1970 = clock.now + DefaultBlockTtl.seconds + 11123
+
+    without expiryUpdateCallback =? sales.onExpiryUpdate:
+      raiseAssert "onExpiryUpdate not set"
```
markspanbroek, line 280: This can be a separate test

tests/codex/testnode.nim
```diff
@@ -163,3 +195,112 @@ asyncchecksuite "Test Node":
...
+    for index in 0..<manifest.blocksCount:
+      let blk = (await localStore.getBlock(manifest.treeCid, index)).tryGet
+      let expiryKey = (createBlockExpirationMetadataKey(blk.cid)).tryGet
+      let expiry = await localStoreMetaDs.get(expiryKey)
+
+      check (expiry.tryGet).toSecondsSince1970 == expectedExpiry
+
+  test "onStore callback":
+    without onStore =? sales.onStore:
+      raiseAssert "onStore not set"
```
markspanbroek, line 293: Same here, I'd make it a separate test to check that the onStore callback is set

markspanbroek (commented):
Looks good @AuHau, thanks! I left a few comments.

tests/codex/testnode.nim
AuHau: Honestly, I don't really see a way, unless we want to run the above tests (the one in "Test Node" suite) with Host contracts enabled, it is an option, but I thought better to separate these.

Node object (and hence its tests) require a lot of dependencies and that is why there is lot of duplication. Also as tests depends on closure a lot, having some "setup utility function" is not really option here.

codex/node.nim
dryajov: > This design comes from the limitation of the BlockStore.getBlock() abstraction. The "correct" approach is calling putBlock() with the expected TTL, but because of how NetworkStore works, it actually calls putBlock during getBlock (eq. when it fetches data from network) and because of the interface we can't specify TTL for the getBlock.

But sticking it into `fetchBatched` isn't addressing this particular limitation at all it simply adds an unexpected side effect to `fetchBatched`. What @emizzle is suggesting is the correct approach here.

As for the perceived limitation of the `network` store interface, an explicit specialization with an additional `method` or `proc`, for example `getOrFetch(cid, ttl)`, that `get`s from the local store or `fetches` from the network, would address this quite nicely.

dryajov (approved):
LGTM, but lets clean it up with the suggestions, specially removing the expiry update from `fetchBatched`

codex/node.nim
AuHau: I don't perceive it as a problem. In my mind, `fetchBatched` is for "downloading data from the network" and the expiry extends this premise to "download data from network and store it for X time" which is IMHO tightly coupled. Btw. if you don't set `expiry` nothing is set... The default expiry is `none` so by no behavior is changed from the original one.

That said, I don't mind changing it as you suggested, but my question then is, how should I address this problem?
>There is also a potential problem that if you are downloading some very big dataset that might exceed the DefaultTTL (currently 24 hours), the maintenance could start deleting the data from the beginning of the dataset before you finish the download. Hence that is why the ensureExpiry is called in the onBatch callback.

If we bring up the expected dataset size (or slot size? not sure) of 1 TB, then if the download is slower than 11MB/s  (`1 000 000MB / (24*60*60)`) then we will run into this problem, which is IMHO quite plausible. So should I move the expiry update call into the Marketplace's `onBatch` implementation? Or should I ignore this, call it edge-case and update only once `fetchBatched` is finished?

Also, keep in mind that the `DefaultTTL` is configurable by the user through CLI parameter...

AuHau:
I will merge this as is so we can move forward as the deadline is approaching. If you feel like something should be changed please create an issue (feel free to assign it to me) and I will address it later on when we have some time to do the cleanup.

AuHau:
@dryajov can you please clarify what fixes? Are you referring to refactoring the mapping to remove the `.get` or removing the expiry from `fetchBatched`?

I have asked for your input earlier and didn't get it so I decided to move forward...

codex/node.nim
dryajov: > I don't perceive it as a problem. In my mind, fetchBatched is for "downloading data from the network" and the expiry extends this premise to "download data from network and store it for X time" which is IMHO tightly coupled. Btw. if you don't set expiry nothing is set... The default expiry is none so by no behavior is changed from the original one.

The point is that there is no way to tell that there is some ttl related logic going on during download, this isn't explicit at all and it's unexpected behavior, overall this is a bad design choice. 

> There is also a potential problem that if you are downloading some very big dataset that might exceed the DefaultTTL (currently 24 hours), the maintenance could start deleting the data from the beginning of the dataset before you finish the download. Hence that is why the ensureExpiry is called in the onBatch callback.

> So should I move the expiry update call into the Marketplace's onBatch implementation? 

No need for that, just create another `onBatch` in the `onStore` callback update the expiry and the call the marketplace `onBatch` after that. 

Like so:

```nim
hostContracts.sales.onStore = proc(request: StorageRequest,
                                    slot: UInt256,
                                    onBatch: BatchProc): Future[?!void] {.async.} =

  proc myOnBatch(blocks: seq[bt.Block]) = 
    # update expiry here
    blocks.mapIt( store.ensureExpiry(...) )
   onBatch()
  

  if fetchErr =? (await node.fetchBatched(manifest, onBatch = myOnBatch)).errorOption:
    let error = newException(CodexError, "Unable to retrieve blocks")
    error.parent = fetchErr
    return failure(error)

```

> Or should I ignore this, call it edge-case and update only once fetchBatched is finished?

Definitely no, I think at the very least the above works fine.

Alternatively, you could also make the network store final and add a `getOrFetch(cid, ttl)` function - this is probably the cleanest solution and it addresses the limitations you're facing. The top level consumers of the interface probably only care about this use either way.

-----------


Also, this:

```nim
  let iter = Iter.fromSlice(0..<manifest.blocksCount)
    .map((i: int) => node.blockStore.getBlock(BlockAddress.init(manifest.treeCid, i)))

  for batchNum in 0..<batchCount:
    let blocks = collect:
      for i in 0..<batchSize:
        if not iter.finished:
          iter.next()
```

Looks awfully weird, I understand whats going on, but we probably want to add a `collect` call or something similar to the async iterators for this types of use cases.

dryajov:
> @dryajov can you please clarify what fixes? Are you referring to refactoring the mapping to remove the `.get` or removing the expiry from `fetchBatched`?
> 
> I have asked for your input earlier and didn't get it so I decided to move forward...

Answered in https://github.com/codex-storage/nim-codex/pull/619#discussion_r1402689414

codex/node.nim
AuHau: Great! Thanks a lot for the input. I will put together fix tomorrow 🙏

AuHau: > Alternatively, you could also make the network store final and add a getOrFetch(cid, ttl) function - this is probably the cleanest solution and it addresses the limitations you're facing. The top level consumers of the interface probably only care about this use either way.

Agree that this would be best, but IMHO lets go with the simplest now. We can refactor this later.

PR #621, opened 2023-11-15
Adds option to automatically announce codex by its public IP

benbierens:


veaceslavdoina:
That service use a [Rate Limits](https://ipinfo.io/developers#rate-limits).

Can we also implement a mechanism to use a call with token when a variable like `NAT_PUBLIC_IP_AUTO_TOKEN` is defined?

docker/docker-entrypoint.sh
```diff
@@ -3,7 +3,12 @@
 # Parameters
 if [[ "${NAT_IP_AUTO}" == "true" ]]; then
   export CODEX_NAT=$(hostname --ip-address)
-  echo "Set CODEX_NAT: ${CODEX_NAT}"
+  echo "Internal: Set CODEX_NAT: ${CODEX_NAT}"
+fi
+
+if [[ "${NAT_PUBLIC_IP_AUTO}" == "true" ]]; then
+  export CODEX_NAT=$(curl https://ipinfo.io/ip)
```
dryajov, line 10: I don't know if we can rely on this, it's probably better to setup our own endpoint for that.

dryajov:
> That service use a [Rate Limits](https://ipinfo.io/developers#rate-limits).
> 
> Can we also implement a mechanism to use a call with token when a variable like `NAT_PUBLIC_IP_AUTO_TOKEN` is defined?

Yes, I don't think we can rely on it. I would just set up our own service for this, it should be reasonable simple with any existing web framework.

veaceslavdoina:
For example - AWS [CloudFront](https://aws.amazon.com/cloudfront/) with [Lambda@Edge](https://aws.amazon.com/lambda/edge/)/[Functions](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/cloudfront-functions.html) in 2 different Regions with a failover for resilience.

Should we use self-made service just for our nodes or it would be a publicly available?

AuHau:
libp2p has a protocol for exactly this use-case called AutoNAT: https://docs-libp2p-io.ipns.dweb.link/concepts/nat/autonat/

I am fine with using this approach for now as a workaround, but later on, we should switch to this protocol as it is not dependant on any third-party (nor our own running service)

veaceslavdoina:
We can use a very generic approach like
```shell
NAT_PUBLIC_IP_AUTO="https://ipinfo.io/ip"

if [[ -n "${NAT_PUBLIC_IP_AUTO}" ]]; then
  export CODEX_NAT=$(curl -m 5 "${NAT_PUBLIC_IP_AUTO}")
  echo "Public: Set CODEX_NAT: ${CODEX_NAT}"
fi
```

In that way, we just need to set `NAT_PUBLIC_IP_AUTO` with the value of the service which will return a Public IP. It can be a self-managed or a different one.

And maybe with some resilience, otherwise node may fail to start - `Error while processing the nat= parameter: IP Address string is empty`
<details>
<summary>code</summary>

```shell
NAT_PUBLIC_IP_AUTO="https://ipinfo.io/ip"

# Get Public IP
if [[ -n "${NAT_PUBLIC_IP_AUTO}" ]]; then
  # Waith 60 seconds
  WAIT=60
  SECONDS=0
  while (( SECONDS < WAIT )); do
    export CODEX_NAT=$(curl -m 5 "${NAT_PUBLIC_IP_AUTO}")
    # Check the exit code is 0 and returned value is not empty
    [[ $? -eq 0 && -n "${CODEX_NAT}" ]] && { echo "Public: Set CODEX_NAT: ${CODEX_NAT}"; break; }
    # Sleep and check again
    sleep 5
  done
fi
```
</details>

veaceslavdoina:
Ben, looks like i forgot some important curl switches
- `-s, --silent`
    >              Silent or quiet mode. Do not show progress meter or error messages. Makes Curl mute. It will still output the data you ask for, potentially even to the terminal/stdout unless you redirect it.
- `-f, --fail`
   > (HTTP) Fail fast with no output at all on server errors. This is useful to enable scripts and users to better deal with failed attempts. In normal cases when an HTTP server fails to deliver a document, it returns an HTML document stating so (which often also describes why and more). This flag will prevent curl from outputting that and return error 22.

Also, we should use `NAT_IP_AUTO` only when `NAT_PUBLIC_IP_AUTO` is not defined. And we stop execution if we can't get Public IP in defined wait time.

Last check looks like that
Screenshot 2023-11-16 at 15 22 29

veaceslavdoina (approved):

docker/docker-compose.yaml
```diff
@@ -29,6 +29,8 @@ services:
       # - CODEX_MARKETPLACE_ADDRESS=${CODEX_MARKETPLACE_ADDRESS:-0x59b670e9fA9D0A427751Af201D676719a970857b}
       - CODEX_VALIDATOR=${CODEX_VALIDATOR:-false}
       - CODEX_VALIDATOR_MAX_SLOTS=${CODEX_VALIDATOR_MAX_SLOTS:-1000}
+      - NAT_IP_AUTO=false
+      - NAT_PUBLIC_IP_AUTO=https://ipinfo.io/ip
```
dryajov, line 33: Are we sure we're going to be able to rely on this service? It seems like it would be kinda flaky. I strongly suggest we host our own or as has been suggested rely on libp2p infrastructure for this.

veaceslavdoina: We do not rely on it?

https://github.com/codex-storage/nim-codex/blob/750fe2392feb57e6043b824e8e939a99d866e30a/docker/docker-entrypoint.sh#L9-L15

We are using a generic approach and a service defined in the `NAT_PUBLIC_IP_AUTO` variable will be used.

Looks like we specified `https://ipinfo.io/ip` in [docker-compose.yaml](https://github.com/codex-storage/nim-codex/blob/master/docker/docker-compose.yaml). But it is fine for now and just as an example?

And we have a task to build our own - https://github.com/codex-storage/infra-codex/issues/91

dryajov: Ah, great - yeah, it's a bit confusing to have it default to that, not sure whats better - maybe no default. I'm sure our IPs are gonna get banned pretty quickly if we run a large number of nodes or too many frequent tests against it.

dryajov: Got you, thanks for clarifying.

Issue #620, opened 2023-11-15
Hardcoded empty block hash?


cskiraly:
I see hardcoded empty block hashes in
https://github.com/codex-storage/nim-codex/blob/4d50ecf5048b014f6f5657cd8eb36015088aab4c/codex/blocktype.nim#L150C1-L151C1

However, after decode, I'm getting different hashes. Not 100% this is the problem I'm seeing, but I'm wondering if this hardcoding is the right approach.

hardcoded: 
`(CIDv1, Sha256, Raw): ? Cid.init("zb2rhmy65F3REf8SZp7De11gxtECBGgUKaLdiDj7MCGCHxbDW").mapFailure`

What I get:
`zb2rhd4qdxnB7ce5gfrcbY1DMBdxDUkud1rhrGr6MLW73XHiN`

cskiraly:
Above was with a block size of 1024. Hash is also size dependent, so I think we need some other approach here

PR #622, opened 2023-11-15
Multidimensional encoding

cskiraly:
This adds multidimensional encoding capabilities to the manifest.

In this patch set
- the manifest is changed to recursive, containing another manifest instead of just a few encoding parameters
- encoding is done recursively, each step adding a layer of RS coding
- the "steps" parameter, which previously defined both the interleaving used, and the number of times the encoding had to be applied, is now split into two parameters: interleaving, and steps
- the default interleaving is to keep the old behaviour, i.e. interleaving=steps
- a workaround is included to handle empty block decoding  (https://github.com/codex-storage/nim-codex/issues/620)

The concept of recursive manifest is questionable. It might be better to support multiple K,M parameters in a single manifest, describing the whole multi-d RS encoding structure without recursion. This would allow to handle originalBlockCount and padding blocks better.
It should be relatively easy to convert this implementation to a non-recursive version. ID mapping changes, but operations are the same, and encoding is done dimension-by-dimension anyway.


dryajov:
can you rebase this with master?

cskiraly:
rebase on master now

PR #623, opened 2023-11-15
Fix block delete by treeCid and index

tbekas:


codex/stores/repostore.nim
```diff
@@ -100,6 +100,12 @@ proc decode(_: type (Cid, MerkleProof), data: seq[byte]): ?!(Cid, MerkleProof) =
     proof = ? MerkleProof.decode(data[sizeof(uint64) + n..^1])
   success((cid, proof))
 
+proc decodeCid(_: type (Cid, MerkleProof), data: seq[byte]): ?!Cid =
+  let
+    n = uint64.fromBytesBE(data[0..<sizeof(uint64)]).int
+    cid = ? Cid.init(data[sizeof(uint64)..<sizeof(uint64) + n]).mapFailure
```
benbierens, line 106: I suppose if you wanted to, you could find a way to remove the duplication with `proc decode`.
This is pretty intense logic. It would be great not to have it duplicated.

benbierens (commented):
I think this PR is an excellent improvement.

codex/stores/repostore.nim
tbekas: I agree and I will fix it 👍

tbekas: I will address this comment in the upcoming PR with an encoder/decoder for tuples. Doing it in current PR would just add some code that will be deleted very soon anyway.

PR #629, opened 2023-11-20
fix: require expiry from api

### LINKED ISSUE #611: Require `expiry` to be provided in API
- Remove the "default 0" from `expiry`
- Check that it is in future (also in smart-contract)

AuHau:
This PR makes the `expiry` parameter of the `requestStorage` API call mandatory, as it is not trivial to estimate it so we are "offloading" this to the user.

As I wanted to check that the "expiry is in future", I had to refactor `node.nim` to have the `clock` instance directly there and available in the API handler.

The branch is based on the `feat/set-ttl-for-requests-expiry` because of the changes to the `Clock` interface there.

Closes #611

### DIFF
--- a/codex/codex.nim
+++ b/codex/codex.nim
@@ -65,7 +65,6 @@ proc bootstrapInteractions(
     config = s.config
     repo = s.repoStore
 
-  s.codexNode.clock = SystemClock()
 
   if not config.persistence and not config.validator:
     if config.ethAccount.isSome or config.ethPrivateKey.isSome:
@@ -113,6 +112,8 @@ proc bootstrapInteractions(
 
   if config.validator or config.persistence:
     s.codexNode.clock = clock
+  else:
+    s.codexNode.clock = SystemClock()
 
   if config.persistence:
     # This is used for simulation purposes. Normal nodes won't be compiled with this flag
--- a/codex/codex.nim
+++ b/codex/codex.nim
@@ -34,6 +34,7 @@ import ./utils/fileutils
 import ./erasure
 import ./discovery
 import ./contracts
+import ./systemclock
 import ./contracts/clock
 import ./contracts/deployment
 import ./utils/addrutils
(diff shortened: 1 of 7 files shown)


tests/integration/codexclient.nim
```diff
@@ -1,6 +1,7 @@
 import std/httpclient
 import std/strutils
 from pkg/libp2p import Cid, `$`, init
+# from std/json import `[]=`
```
markspanbroek, line 4:
```suggestion
```

tests/integration/codexclient.nim
```diff
@@ -31,6 +32,36 @@ proc upload*(client: CodexClient, contents: string): ?!Cid =
...
+  ## Call request storage REST endpoint
+  ##
+  let url = client.baseurl & "/storage/request/" & $cid
+  let json = %*{
+      "duration": duration,
+      "reward": reward,
+      "proofProbability": proofProbability,
+      "collateral": collateral,
+      "nodes": nodes,
+      # "expiry": expiry,
```
markspanbroek, line 56:
```suggestion
```

codex/codex.nim
```diff
@@ -55,12 +56,16 @@ type
...
+    s: CodexServer
+): Future[void] {.async.} =
   ## bootstrap interactions and return contracts
   ## using clients, hosts, validators pairings
   ##
+  let
+    config = s.config
+    repo = s.repoStore
+
+  s.codexNode.clock = SystemClock()
```
markspanbroek, line 68: Can we set this later as the `else` branch of the `if config.validator or config.persistence:`?

Then the assignments to node.clock are in one place, and a bit easier to reason about.

markspanbroek (approved):
Thanks @AuHau! A few small comments below.

tests/integration/codexclient.nim
AuHau: Thanks, forgot these from debugging.

PR #631, opened 2023-11-21
Safe block deletion (with ref count)

tbekas:


tests/codex/stores/testrepostore.nim
```diff
@@ -262,56 +243,45 @@ asyncchecksuite "RepoStore":
...
 
     check:
-      response.len == 1
-      !response[0].key == expectedKey
-      response[0].data == updatedExpectedExpiration.toBytes
+      expectedExpiration notin updatedExpirations
+      updatedExpectedExpiration in updatedExpirations
 
-  test "Should not update block expiration timestamp when current expiration is farther then new one":
+  test "Should update block expiration timestamp when new expiration is farther":
```
benbierens, line 265: This one does the opposite, which is correct. The title is a copy-paste error.

codex/rest/json.nim
```diff
@@ -61,10 +62,10 @@ type
...
 
   RestRepoStore* = object
-    totalBlocks* {.serialize.}: uint
-    quotaMaxBytes* {.serialize.}: uint
-    quotaUsedBytes* {.serialize.}: uint
-    quotaReservedBytes* {.serialize.}: uint
+    totalBlocks* {.serialize.}: Natural
+    quotaMaxBytes* {.serialize.}: NBytes
+    quotaUsedBytes* {.serialize.}: NBytes
+    quotaReservedBytes* {.serialize.}: NBytes
```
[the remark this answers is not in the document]
tbekas: If I do `curl -s 127.0.0.1:8080/api/codex/v1/space | jq` I get

```json
{
  "totalBlocks": 0,
  "quotaMaxBytes": 8589934592,
  "quotaUsedBytes": 0,
  "quotaReservedBytes": 0
}
``` 

I think that the `bytes` suffix should stay on properties.

codex/stores/typeddatastore.nim
```diff
@@ -0,0 +1,109 @@
...
+type
+  ModifyT*[T] = proc(v: ?T): Future[?T] {.upraises: [CatchableError], gcsafe, closure.}
+  ModifyTGetU*[T, U] = proc(v: ?T): Future[(?T, U)] {.upraises: [CatchableError], gcsafe, closure.}
+
+  BytesTuple = (?seq[byte], seq[byte])
+
+  KeyVal[T] = (?Key, ?!T)
+  ResIter[T] =  Future[KeyVal[T]]
+
+
```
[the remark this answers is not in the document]
tbekas: It's tested transitively via the `testrepostore.nim`. I think we can live without a dedicated test suit for this file.

codex/utils/asynciter.nim
```diff
@@ -58,8 +58,18 @@ proc new*[T](_: type Iter, genNext: GenNext[T], isFinished: IsFinished, finishOn
...
+proc emptyIter*[T](): Iter[T] =
+  ## Creates an empty iterator
+  ##
+
+  proc genNext(): T {.upraises: [CatchableError].} =
+    raise newException(CatchableError, "Next item requested from an empty iterator")
+  proc isFinished(): bool = true
+
+  Iter.new(genNext, isFinished)
+
```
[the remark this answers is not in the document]
tbekas: I agree, I will add tests for this file.

codex/stores/repostore.nim
```diff
@@ -41,119 +45,277 @@ declareGauge(codex_repostore_bytes_reserved, "codex repostore bytes reserved")
...
-    quotaMaxBytes*: uint          # maximum available bytes
-    quotaUsedBytes*: uint         # bytes used by the repo
-    quotaReservedBytes*: uint     # bytes reserved by the repo
+    quotaMaxBytes*: NBytes
+    quotaUsage*: QuotaUsage
+    totalBlocks*: Natural
     blockTtl*: Duration
     started*: bool
 
+  QuotaUsage* = object
```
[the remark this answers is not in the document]
tbekas: `QuotaUsage` is db entity, while `quotaMaxBytes` is a configuration property that's stays immutable through the lifetime of an app. Therefore I think they belong in different spaces. If we want to present them together, they could be combined in a presentation layer (Json) as it's already done in the `RestRepoStore` object.

Also I think that either all db entities should be in separate files, or none of them, to keep the code base consistent. Moving `QuotaUsage` to separate file while keeping other entities is inconsistent.

codex/utils/genericcoders.nim
```diff
@@ -0,0 +1,64 @@
...
+      let fBytes = encode(f)
+      pb.write(index, fBytes)
+      index.inc
+    else:
+      {.error: "provide `proc encode(a: " & $typeof(f) & "): seq[byte]` to use autoencode".}
+
+  pb.finish
+  pb.buffer
+
+proc autodecode*(T: typedesc[tuple | object], bytes: seq[byte]): ?!T =
```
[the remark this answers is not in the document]
tbekas: I agree, I will add few simple test cases.

codex/stores/repostore.nim
```diff
@@ -41,119 +45,277 @@ declareGauge(codex_repostore_bytes_reserved, "codex repostore bytes reserved")
...
+  )
+
+proc getLeafMetadata(self: RepoStore, treeCid: Cid, index: Natural): Future[?!LeafMetadata] {.async.} =
   without key =? createBlockCidAndProofMetadataKey(treeCid, index), err:
     return failure(err)
 
-  without value =? await self.metaDs.get(key), err:
+  without leafMd =? await self.metaDs.get(key), err:
     if err of DatastoreKeyNotFound:
       return failure(newException(BlockNotFoundError, err.msg))
```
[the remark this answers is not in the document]
tbekas: I thought about it, thing is it wouldn't make anything simpler. 

Ideally would be to use `Option[T]` but then the return type would be `Future[Result[Option[T], E]]` or using questionable it would be `Future[?![?T]]`. In my opinion that's just too long and too complicated. 
Currently we always return `failure(BlockNotFoundError)` for both blocks and blocks metadata and since it always translates to the same behaviour - handling missing block routine - we don't need to introduce another type for `failure(BlockMetadataNotFoundError)`, making the handling uniform in the client code (as it would be with handling `T.none` if we would use `Option[T]`).

codex/stores/repostore.nim
```diff
@@ -41,119 +45,277 @@ declareGauge(codex_repostore_bytes_reserved, "codex repostore bytes reserved")
...
     else:
       return failure(err)
 
-  return (Cid, MerkleProof).decode(value)
+  LeafMetadata.decode(leafMd)
+
+proc updateTotalBlocksCount(self: RepoStore, plusCount: Natural = 0, minusCount: Natural = 0): Future[?!void] {.async.} =
```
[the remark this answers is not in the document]
tbekas: >Give given how much of a pain all the type conversions are becoming, I'm not sure we should change this.

That's exactly the reason. My initial implementation was to use single `deltaCount: int` parameter. However in the end it's actually waay easier to avoid conversion errors and arithmetic overflow/underflow errors if we have two parameters. We could also go with `amount: Nautral, op: Add | Subtract` but that's still 2 params.

codex/stores/repostore.nim
```diff
@@ -41,119 +45,277 @@ declareGauge(codex_repostore_bytes_reserved, "codex repostore bytes reserved")
...
+        if currMd.size == blk.data.len.NBytes:
+          md = BlockMetadata(size: currMd.size, expiry: max(currMd.expiry, minExpiry), refCount: currMd.refCount)
+          res = StoreResult(kind: AlreadyInStore)
+
+          # making sure that the block acutally is stored in the repoDs
+          without hasBlock =? await self.repoDs.has(blkKey), err:
+            raise err
+
+          if not hasBlock:
+            warn "Block metadata is present, but block is absent. Restoring block.", cid = blk.cid
```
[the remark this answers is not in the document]
tbekas: I agree, I will add a test case for it :)

codex/stores/repostore.nim
```diff
@@ -41,119 +45,277 @@ declareGauge(codex_repostore_bytes_reserved, "codex repostore bytes reserved")
...
+      var
+        maybeMeta: ?BlockMetadata
+        res: DeleteResult
+
+      if currMd =? maybeCurrMd:
+        if currMd.refCount == 0 or currMd.expiry > expiryLimit:
+          maybeMeta = BlockMetadata.none
+          res = DeleteResult(kind: Deleted, released: currMd.size)
+
+          if err =? (await self.repoDs.delete(blkKey)).errorOption:
```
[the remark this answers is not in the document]
tbekas: Yes, it's done implicitly, by returning `none` from the `modify` function. In line 292 we assign `none` to the result, so the `modify` will delete the entry. Relevant docs: https://github.com/codex-storage/nim-datastore/blob/3c4daf41983747800f15d24f9cc076cfb4db6116/datastore/datastore.nim#L72

codex/stores/repostore.nim
```diff
@@ -41,119 +45,277 @@ declareGauge(codex_repostore_bytes_reserved, "codex repostore bytes reserved")
...
+    return failure(err)
+
+  await self.metaDs.modifyTGetU(metaKey,
+    proc (maybeCurrMd: ?BlockMetadata): Future[(?BlockMetadata, DeleteResult)] {.async.} =
+      var
+        maybeMeta: ?BlockMetadata
+        res: DeleteResult
+
+      if currMd =? maybeCurrMd:
+        if currMd.refCount == 0 or currMd.expiry > expiryLimit:
```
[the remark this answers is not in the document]
tbekas: You're 100% right, there's a logic error here. It should be `currMd.expiry > expiryLimit` and the default value of `expiryLimit` should be `SecondsSince1970.low` instead of `SecondsSince1970.high`.

Thank you for catching this! I will add a test case for this.

codex/stores/repostore.nim
```diff
@@ -274,17 +389,37 @@ method ensureExpiry*(
...
-    trace "Error total blocks key!", err = err.msg
+  without res =? await self.putLeafMetadata(treeCid, index, blkCid, proof), err:
     return failure(err)
+
+  if res == Stored:
+    if err =? (await self.updateBlockMetadata(blkCid, plusRefCount = 1)).errorOption:
+      return failure(err)
+    trace "Leaf metadata stored, block refCount incremented"
+  else:
+    trace "Leaf metadata already exists"
```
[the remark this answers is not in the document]
tbekas: No, the refCount refers to how many trees are referring this block.

If you store just a block (without any references to the tree yet), what should be the initial refCount for this block? In my opinion it's `0` - since there are no references to it.

Then when adding new references to this block (new leaf metadata) we increase the refCount by 1.

tests/codex/stores/testrepostore.nim
```diff
@@ -85,167 +85,148 @@ asyncchecksuite "RepoStore":
...
-      repo.quotaReservedBytes == 0
-
-    expect DatastoreKeyNotFound:
-      discard (await metaDs.get(QuotaReservedKey)).tryGet
+      repo.totalUsed == 100'nb
+      repo.quotaUsedBytes == 100'nb
+      repo.quotaReservedBytes == 0'nb
 
   test "Should release bytes":
     discard createTestBlock(100)
```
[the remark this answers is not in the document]
tbekas: I haven't added this code, but as far as I can tell the goal is to verify that the quota usage changes after creating a block.

tests/codex/stores/testrepostore.nim
tbekas: Ah, you're right, thanks for catching it!

codex/stores/repostore.nim
```diff
@@ -297,131 +432,64 @@ method putBlock*(
...
-  if err =? (await self.metaDs.put(batch)).errorOption:
-    trace "Error updating quota bytes", err = err.msg
-
-    if err =? (await self.repoDs.delete(key)).errorOption:
-      trace "Error deleting block after failed quota update", err = err.msg
+  if res.kind == Stored:
+    trace "Block Stored"
+    if err =? (await self.updateQuotaUsage(plusUsed = res.used)).errorOption:
+      # rollback changes
+      without delRes =? await self.tryDeleteBlock(blk.cid), err:
```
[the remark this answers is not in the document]
tbekas: I will think about a way to address it, but really it's not that trivial. First of all we don't know if we should attempt to increase the quota unless we try to store a block - because it may turn out that the block is already stored and there was no reason to increase the quota in the first place.

It seems to be easy at surface. But there might occur concurrency errors, that eventually may lead to permanent inconsistencies in db, which should be avoided at all cost. Keep in mind that we need to keep consistent 4 different records in 2 different stores:
- block in repoDs
- block metadata in metaDs
- quota usage in metaDs
- total blocks in metaDs

codex/stores/repostore.nim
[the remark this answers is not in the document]
dryajov: Hos is this edge case even possible?

codex/stores/repostore.nim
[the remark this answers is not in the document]
dryajov: All of this should be written as a batch in one go with the `method put*(self: Datastore, batch: seq[BatchEntry])` method, this will avoid inconsistencies across this stores, the only exception being the `repoDs` which is right now a separate datastore. Another approach would be to moving the `repoDs` to the same table as the rest of the keys, but that might have additional performance consequences - but still worth a try.

We can also think about adding a concurrent store on top that will update/rollback a batch of `put` oprations in one transation.

codex/stores/repostore.nim
```diff
@@ -41,126 +45,277 @@ declareGauge(codex_repostore_bytes_reserved, "codex repostore bytes reserved")
...
-    cid = ? Cid.init(data[sizeof(uint64)..<sizeof(uint64) + n]).mapFailure
-    proof = ? CodexProof.decode(data[sizeof(uint64) + n..^1])
-  success((cid, proof))
-
-proc decodeCid(_: type (Cid, CodexProof), data: seq[byte]): ?!Cid =
-  let
-    n = uint64.fromBytesBE(data[0..<sizeof(uint64)]).int
-    cid = ? Cid.init(data[sizeof(uint64)..<sizeof(uint64) + n]).mapFailure
-  success(cid)
+proc encode(t: Cid): seq[byte] = t.data.buffer
```
dryajov, line 96: this file is already quite busy, can we perhaps split this helpres out to their own unit?

dryajov (commented):
This is quite a lot of work, I think we're at a point now that we can start merging this into master, can we rebase this with master so I can give it a proper review?

codex/stores/repostore.nim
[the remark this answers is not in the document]
tbekas: The only case I can think of is manual block deletion or a bug. Anyway it makes sense to handle this case though. If someone calls `putBlock` the expectation after successful execution is that the block is stored _regardless of the existing state in stores_. If we would remove this check and the existing state would be: metadata exists, but block doesn't exist, then after successful call to `putBlock` we wouldn't have block stored.

codex/stores/repostore.nim
tbekas: Except for `Cid` all `encode/decode` procs are defined here because all corresponding types are also defined here.

dryajov: Then it's probably good to extract the types to their own `types` module and share across both, this is the commonly accepted approach.

codex/stores/typeddatastore.nim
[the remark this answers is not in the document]
dryajov: This looks like something that needs to live in datastore proper? And, I'd actually add some tests for it as suggested.

codex/stores/repostore.nim
dryajov: This file needs to be split into several functional units that group related functionality. 

I suggest the following structure: 

- `respostore/types.nim` - common types shared by all other functional unites
- `repostore/coders.nim` - encoding/decoding primitives
- `repostore/operations.nim` - the update/get/modify operation
- `repostore/repostore.nim/store.nim` - the actual repostore functionality
- `repostore.nim` - top level module that exports the required public functinal units/modules

This structure would make the changes easier to track and compare with previous versions. It's quite hard to follow whats going on right now and the file is already quite large and busy with unrelated functionality.

codex/utils/asynciter.nim
```diff
@@ -4,8 +4,8 @@ import pkg/questionable
 import pkg/chronos
 
 type
-  Function*[T, U] = proc(fut: T): U {.raises: [CatchableError], gcsafe, noSideEffect.}
-  IsFinished* = proc(): bool {.raises: [], gcsafe, noSideEffect.}
+  Function*[T, U] = proc(fut: T): U {.raises: [CatchableError], gcsafe, closure.}
```
dryajov, line 7: Closure is unnecessary here, when declaring a proc type, it's implicitely a closure.

codex/utils/executor.nim
```diff
@@ -0,0 +1,5 @@
+
+
+type
```
dryajov, line 3: leftover from other work?

codex/stores/repostore.nim
tbekas: Ok

codex/utils/executor.nim
tbekas: It is 🤦 Removing it

codex/stores/repostore.nim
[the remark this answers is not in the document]
tbekas: Following up on a offline discussion regarding this issue. We agreed that all the items we store in `metaDs`, which are:
* block metadata
* quota byte usage
* total number of blocks

should be saved with a single batch and a single transaction. However currently we don't have `modify` API that would accept batch input, and adding such API would require some additional development effort in `nim-datastore`, so we're fine with lower consistency guarantee, which is just consistency between block metadata
and the block data (a file).

benbierens (approved):
Overall good improvements.
Commit sha-5e952f0 passes dist-tests. 👍

codex/stores/repostore/store.nim
```diff
@@ -0,0 +1,398 @@
...
+  ## Stop repo
+  ##
+  if not self.started:
+    trace "Repo is not started"
+    return
+
+  trace "Stopping repo"
+  await self.close()
+
+  self.started = false
```
dryajov, line 398: nitpick, we usually have new lines at EOF, maybe we should add a github hook to enforce it...

codex/stores/repostore/coders.nim
```diff
@@ -0,0 +1,53 @@
...
+proc encode*(i: uint64): seq[byte] =
+  @(i.toBytesBE)
+
+proc decode*(T: type uint64, bytes: seq[byte]): ?!T =
+  if bytes.len >= sizeof(uint64):
+    success(uint64.fromBytesBE(bytes))
+  else:
+    failure("Not enough bytes to decode `uint64`")
+
+proc encode*(i: int): seq[byte] = cast[uint64](i).encode
```
dryajov, line 46: Avoid using `cast` as much as possible. Nim has safe type conversion and unsafe type conversion, `cast` is explicitly unsafe.

Just doing:

```suggestion

proc encode*(i: int): seq[byte] = uint64(i).encode

```

should suffice.

codex/stores/repostore/coders.nim
```diff
@@ -0,0 +1,53 @@
...
+proc decode*(T: type int, bytes: seq[byte]): ?!T = uint64.decode(bytes).map((ui: uint64) => cast[int](ui))
```
dryajov, line 47: same as above

codex/stores/repostore/coders.nim
```diff
@@ -0,0 +1,53 @@
...
+
+proc encode*[T: enum](e: T): seq[byte] = e.ord().encode
+proc decode*(T: typedesc[enum], bytes: seq[byte]): ?!T = int.decode(bytes).map((ui: int) => T(ui))
+
+proc encode*(i: Natural): seq[byte] = cast[int](i).encode
+proc decode*(T: type Natural, bytes: seq[byte]): ?!T = int.decode(bytes).map((ui: int) => cast[Natural](ui))
```
dryajov, line 53: ditto

codex/stores/repostore/operations.nim
```diff
@@ -0,0 +1,213 @@
...
+
+declareGauge(codex_repostore_blocks, "codex repostore blocks")
+declareGauge(codex_repostore_bytes_used, "codex repostore bytes used")
+declareGauge(codex_repostore_bytes_reserved, "codex repostore bytes reserved")
+
+proc putLeafMetadata*(self: RepoStore, treeCid: Cid, index: Natural, blkCid: Cid, proof: CodexProof): Future[?!StoreResultKind] {.async.} =
+  without key =? createBlockCidAndProofMetadataKey(treeCid, index), err:
+    return failure(err)
+
+  await self.metaDs.modifyGet(key,
```
dryajov, line 39: If we're simply storing the value, without having to update it, why not just use a simple `put`?

codex/stores/repostore/operations.nim
```diff
@@ -0,0 +1,213 @@
...
+
+  without leafMd =? await get[LeafMetadata](self.metaDs, key), err:
+    if err of DatastoreKeyNotFound:
+      return failure(newException(BlockNotFoundError, err.msg))
+    else:
+      return failure(err)
+
+  success(leafMd)
+
+proc updateTotalBlocksCount*(self: RepoStore, plusCount: Natural = 0, minusCount: Natural = 0): Future[?!void] {.async.} =
```
dryajov, line 67: This should probably return new count for completenness?

dryajov: Can be left for a separate PR tho...

codex/stores/repostore/operations.nim
```diff
@@ -0,0 +1,213 @@
...
+          currCount + plusCount - minusCount
+        else:
+          plusCount - minusCount
+
+      self.totalBlocks = count
+      codex_repostore_blocks.set(count.int64)
+      count.some
+  )
+
+proc updateQuotaUsage*(
```
dryajov, line 81: Same as above, probably makes sense to return the new counts?

codex/stores/repostore/operations.nim
```diff
@@ -0,0 +1,213 @@
...
+        md: BlockMetadata
+        res: StoreResult
+
+      if currMd =? maybeCurrMd:
+        if currMd.size == blk.data.len.NBytes:
+          md = BlockMetadata(size: currMd.size, expiry: max(currMd.expiry, minExpiry), refCount: currMd.refCount)
+          res = StoreResult(kind: AlreadyInStore)
+
+          # making sure that the block acutally is stored in the repoDs
+          without hasBlock =? await self.repoDs.has(blkKey), err:
```
dryajov, line 155: is this necessary, seems like `modifyGet` would be broken if that wasn't the case?

dryajov: Ah, nevermind, this are different DS.

codex/stores/repostore/operations.nim
```diff
@@ -0,0 +1,213 @@
...
+    return failure(err)
+
+  await self.metaDs.modifyGet(metaKey,
+    proc (maybeCurrMd: ?BlockMetadata): Future[(?BlockMetadata, DeleteResult)] {.async.} =
+      var
+        maybeMeta: ?BlockMetadata
+        res: DeleteResult
+
+      if currMd =? maybeCurrMd:
+        if currMd.refCount == 0 or currMd.expiry < expiryLimit:
```
dryajov, line 190: is there really a case where we want to delete things even if they have a refcount > 0?

codex/stores/maintenance.nim
```diff
@@ -75,11 +75,11 @@ proc runBlockCheck(self: BlockMaintainer): Future[void] {.async.} =
...
   var numberReceived = 0
-  for maybeBeFuture in iter:
-    if be =? await maybeBeFuture:
-      inc numberReceived
-      await self.processBlockExpiration(be)
-      await sleepAsync(50.millis)
+  for beFut in iter:
+    let be = await beFut
+    inc numberReceived
+    await self.processBlockExpiration(be)
```
dryajov, line 81: instead of deleting one by one, we should accomulate a batch of blocks and delete. Something for a future PR.

dryajov (approved):
OK, besides the `cast`, which should be a very straightforward change, most of the comments can be addressed in separate PRs.

codex/stores/repostore/operations.nim
tbekas: Because we need an information if such entry was already in the store or is it a new entry. If it's a new entry we're going to increase block ref count by 1.

codex/stores/repostore/operations.nim
tbekas: I'm a fan of having minimal code that does the job. If nothing is going to depend on the returned value, there's no need to return anything.

codex/stores/repostore/operations.nim
tbekas: Yes, that's correct.

codex/stores/repostore/operations.nim
tbekas: Block expiry is such case. This change still assumes that expiry is property of a block, not of a dataset, PR that changes that is here https://github.com/codex-storage/nim-codex/pull/678 In this PR this condition is changed to

```nim
if currMd.refCount == 0:
```

codex/stores/maintenance.nim
tbekas: Block expiration is going to be a property of a dataset if this PR goes through https://github.com/codex-storage/nim-codex/pull/678

Regardless of that deleting in a batch may be incorrect (depending on the implementation). In above PR we're still deleting blocks one by one because each block needs a careful inspection of ref-count. Delete happens only if block has ref-count = 0.

codex/stores/repostore/operations.nim
tbekas: I would have the same argument as above, so let's just keep the discussion in the thread above.

codex/stores/repostore/coders.nim
tbekas: I already had that discussion with @emizzle link: https://github.com/codex-storage/nim-codex/pull/678#discussion_r1527647116

I insist on keeping the cast here. But if you know anything better (simpler) that keeps this property always true, thank I'm open to change.

```
x == decode(encode(x))
```

tbekas: Unfortunately it won't work. I already had this discussion with @emizzle here https://github.com/codex-storage/nim-codex/pull/678#discussion_r1527647116

This is the property that has to be satisfied for a successful pair of `encode` and `decode` procs:
```
let x
x == decode(encode(x))
```

This code will fail when we'll use type conversion in place of casting. It will fail for all negative values of int.

dryajov: Firstly, we need to add the appropriate missing tests then.

tbekas: I agree. I'm going to add tests for these coders.

dryajov: I would still narrow this down to only specific numeric types - `Natural | enum` and leave `int` out of the cast, because that opens this up for a lot of potential mistakes. 

This works just fine, right now:

```
proc encode*(i: Natural | enum): seq[byte] = cast[uint64](i).encode
proc decode*(T: typedesc[Natural | enum], bytes: seq[byte]): ?!T = uint64.decode(bytes).map((ui: uint64) => cast[T](ui))
```

tbekas: I've added the `testcoders.nim` that checks this property for random values of different types.

dryajov: Lets also constrain the encoding/decoding away from `int` as in my example above. Otherwise this is just way too unsafe.

PR #634, opened 2023-11-23
refactor: move expiry update from fetchBatched

AuHau:
This is follow up PR after #619 which refactors the updating expiry part, moving it from `fetchBatched` into separate `onBatch` wrapper.

### DIFF
--- a/codex/node.nim
+++ b/codex/node.nim
@@ -38,6 +38,7 @@ import ./discovery
 import ./contracts
 import ./node/batch
 import ./utils
+import ./errors
 
 export batch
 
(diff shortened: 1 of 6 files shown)


codex/node.nim
```diff
@@ -441,11 +437,13 @@ proc start*(node: CodexNodeRef) {.async.} =
...
       trace "Fetching block for manifest", cid
-      # TODO: This will probably require a call to `getBlock` either way,
-      # since fetching of blocks will have to be selective according
-      # to a combination of parameters, such as node slot position
-      # and dataset geometry
-      if fetchErr =? (await node.fetchBatched(manifest, onBatch = onBatch, expiry = some request.expiry.toSecondsSince1970)).errorOption:
+      let expiry = request.expiry.toSecondsSince1970
+      proc expiryUpdateOnBatch(blocks: seq[bt.Block]): Future[void] {.async.} =
+        let ensureExpiryFutures = blocks.mapIt(node.blockStore.ensureExpiry(it.cid, expiry))
+        await allFuturesThrowing(ensureExpiryFutures)
```
dryajov, line 443: This will throw on a failed future and since this is a callback it will most likely crash the process, we should surround this with a `try/except` or use `catch` from results to convert it into a result.

AuHau: Good point! Thanks for the catch! 👍

dryajov (approved):
LGTM

PR #639, opened 2023-11-27
Proof/data sampler

benbierens:


benbierens:
Heads-up: There's a small bug in the assembly in Constantine that needs to get fixed before this thing will work. The PR for that is open here: https://github.com/mratsim/constantine/pull/306

benbierens:
The constantine update was pulled in and the sampler output has been brough up-to-date with the circuit reference code.
Discussed with Dmitriy and we're holding off merging this one until the merkle tree update has been finished and interested.

benbierens:
Data sampler should now be aligned with slot builder.

codex/proof/datasampler.nim
```diff
@@ -0,0 +1,196 @@
...
+# "<ContainerType><ElementType>Index" => The index of an ElementType within a ContainerType.
+# Some examples:
+# SlotBlockIndex => The index of a Block within a Slot.
+# DatasetBlockIndex => The index of a Block within a Dataset.
+
+logScope:
+  topics = "codex datasampler"
+
+type
+  DataSampler* = ref object of RootObj
```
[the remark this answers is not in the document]
benbierens: The expected lifecycle for DataSampler objects is to stay in-memory for as long as a slot is being hosted. Combined with the constructor/start proc split, this thing should be a ref type.

codex/slots/converters.nim
```diff
@@ -0,0 +1,70 @@
...
+func toVerifiableProof*(proof: CodexProof): ?!Poseidon2Proof =
+  let
+    verifiableProof = Poseidon2Proof(
+      index: proof.index,
+      nleaves: proof.nleaves,
+      path: proof.path.mapIt(
+        ? Poseidon2Hash.fromBytes(it.toArray32).toResult
+      ))
+
+  success verifiableProof
```
dryajov, line 70: new line at end of file.

codex/proof/slotblocks.nim
```diff
@@ -0,0 +1,110 @@
+import std/sequtils
```
dryajov, line 1: This isn't needed, just use `slotsBuilder.buildSlot(idx)` and everything will be taken care of.

codex/proof/datasampler.nim
```diff
@@ -0,0 +1,214 @@
...
+  if err =? (await slotBlocks.start()).errorOption:
+    error "Failed to create SlotBlocks object for slot", error = err.msg
+    return failure(err)
+
+  let
+    manifest = slotBlocks.manifest
+    numberOfCellsInSlot = getNumberOfCellsInSlot(self.slot)
+    blockSize = manifest.blockSize.uint64
+
+  without starter =? (await startDataSampler(self.blockStore, manifest, self.slot)), e:
```
dryajov, line 80: you can initialize the builder with a verifiable manifest and it will unpack and set the slots and root hashes.

dryajov (changes requested):
Good progress, but lets wait for #668 to land, as there are additional changes that should simplify this considerably.

benbierens:
There's a lot of overlap between what's happening in the slot builder and parts of the data sampler code. But they mismatch in subtle but important ways. For example: slot builder digests block-cells to get their block-roots. Data sampler does the same thing but needs the tree to get proof instead. Slot builder builds and stores the tree for a slot. Data sampler does the same, but only if the tree can't be loaded from disk. Slot builder retrieves blocks if needed. Data sampler does the same, but batched and with an onBatch callback to support so marketplace shenanigans. We can unify these things make it work, but it'll take some changes.

PR #642, opened 2023-11-29
Verifiable manifests

benbierens:


codex/manifest/types.nim
```diff
@@ -22,8 +24,22 @@ const
 type
   ManifestCoderType*[codec: static MultiCodec] = object
   DagPBCoder* = ManifestCoderType[multiCodec("dag-pb")]
+  VerificationHash* = F
```
dryajov, line 27: We cannot hardcode this, this needs to be parametrizable with the correct hash type.

dryajov:
In general, the manifest should never be hardcoded with any particular set of hashes or other types. We should use a combination of templates/generics and multicodecs to keep the manifest generic and hash type agnostic.

codex/manifest/types.nim
benbierens: That makes perfect sense. I would like a nice generalized way we can wrap up the Poseidon2 field elements and plug them into the manifest, and drop the `VerificationHash` type completely.

tbekas (approved):
Overall looks good. I agree that it would be nice to keep manifest agnostic to the hash type. But I don't know how to do it. With multicodec we're not able to specify poseidon2 as a codec.

dryajov:
> Overall looks good. I agree that it would be nice to keep manifest agnostic to the hash type. But I don't know how to do it. With multicodec we're not able to specify poseidon2 as a codec.

This is just a matter of adding the missing multicodecs and I already did it in https://github.com/status-im/nim-libp2p/pull/974

PR #607, opened 2023-11-09
Multi-node integration test refactor

emizzle:
Refactoring of the multi node integration tests, into a `marketplacetestsuite`. Allows us to run many node integration tests on our local machine, which saves quite a lot of time as compared to the distributed tests.

Notable features:
- Allows arbitrary number of nodes for `clients`, `providers`, and `validators`.
- Allows log files for each started node to be captured (including hardhat), saved in `tests/integration/logs/<start_datetime> <suite_name>/<test_name>/<node_role>_<node_idx>.log`.
- Allows log topics to be specified
- Allows simulation of proof failures to be specified
- Allows log output to stdout

Two notable changes:
1. `OnChainClock.now` was changed so that it always get the time from the last block time. Without this change, the clock was drifting from the actual last block time, which was causing validators to mark proofs as missing when not required, or not mark them missing when required, and providers were sending proofs when the period had not yet ended. Once the change was in, these errors disappeared.
2. Cancellation of the `SaleProving` prove loop was removed as it was hanging during integration tests. This is not needed as the state machine will cancel `SaleProving.run`, of which the `proveLoop` is a child and should also be cancelled.

Leaving this in draft for now as I'd like to try moving the node config setup into each test, as opposed to being defined at the suite-level.

PR #643, opened 2023-12-04
remove waitFor cancelAndWait in proving state

emizzle:
In proving.onCancelled, the `waitFor state.loop.cancelAndWait()` was never completing. Turns out this was not needed, because when changing states, the current state's run is cancelled, which automatically cancels the state prove loop, because it is a child of proving.run. Therefore, the logic to cancelAndWait the prove loop was removed as it was not needed.

This is a PR that is a smaller chunk of https://github.com/codex-storage/nim-codex/pull/607 to attempt to make it smaller and more digestible.

### DIFF
--- a/codex/sales.nim
+++ b/codex/sales.nim
@@ -373,7 +373,7 @@ proc subscribeCancellation(sales: Sales) {.async.} =
   let queue = context.slotQueue
 
   proc onCancelled(requestId: RequestId) =
-    trace "request cancelled, removing all request slots from queue"
+    trace "request cancelled (via contract RequestCancelled event), removing all request slots from queue"
     queue.delete(requestId)
 
   try:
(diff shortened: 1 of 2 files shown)


markspanbroek (approved):
Nice!

gmega:
Yes, @emizzle literally deleted my problem. :-D

elcritch:
> Yes, @emizzle literally deleted my problem. :-D When can we merge this?

@gmega Looks like Mark already approved it, so it's good to go? I hit the update-branch button so it'd re-build.

emizzle:
> @gmega Looks like Mark already approved it, so it's good to go? I hit the update-branch button so it'd re-build.
> 
> 

I think this merged master into this branch (check out commits) which kinda muddles the history a bit. If it's ok I'll delete that commit. And if needed, will rebase on master instead.

PR #644, opened 2023-12-04
Workaround for Hardhat last block timestamp bug

emizzle:
Likely due to a Hardhat bug in which the callbacks for subscription events are called and awaited before updating its local understanding of the last block time, Hardhat will report a block time in the `newHeads` event that is generally 1 second before the time reported from `getLatestBlock.timestamp`. This was causing issues with the OnChainClock's offset and therefore the `now()` used by the `OnChainClock` would sometimes be off by a second (or more), causing tests to fail.

This commit introduce a `codex_use_hardhat` compilation flag, that when set, will always get the latest block timestamp from Hardhat via the `getLatestBlock.timestamp` RPC call for `OnChainClock.now` calls. Otherwise, the last block timestamp reported in the `newHeads` event will be used.

Update the docker dist tests compilation flag for simulated proof failures (it was not correct), and explicitly add the `codex_use_hardhat=false` for clarity.

This is a PR that is a smaller chunk of https://github.com/codex-storage/nim-codex/pull/607 to attempt to make it smaller and more digestible.

### Note
Bumps codex-contracts-eth to https://github.com/codex-storage/codex-contracts-eth/pull/75, which likely should be merged before merging this PR.

### DIFF
--- a/.github/workflows/docker-dist-tests.yml
+++ b/.github/workflows/docker-dist-tests.yml
@@ -24,7 +24,7 @@ jobs:
     name: Build and Push
     uses: ./.github/workflows/docker-reusable.yml
     with:
-      nimflags: '-d:disableMarchNative -d:codex_enable_api_debug_peers=true -d:codex_enable_simulated_proof_failures -d:codex_enable_log_counter=true'
+      nimflags: '-d:disableMarchNative -d:codex_enable_api_debug_peers=true -d:codex_enable_proof_failures=true -d:codex_use_hardhat=false -d:codex_enable_log_counter=true'
       nat_ip_auto: true
       tag_latest: ${{ github.ref_name == github.event.repository.default_branch || startsWith(github.ref, 'refs/tags/') }}
       tag_suffix: dist-tests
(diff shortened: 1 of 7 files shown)


markspanbroek:
Would it be possible to keep the local clock offset implementation when not testing with hardhat? Or does that lead to problems as well?

AuHau (approved):
LGTM, did you report the Hardhat bug upstream? I would suggest to do so ;-) 

Btw. just so you know, the `clock` is now used in other places in Codex as well (for example in the validation of Storage Request when calling POST API call for creating one). If the clock would get synced out a lot (eq. for some reason no block arriving) it could lead to potentially unexpected behaviour. But IMHO it should not really happen.

emizzle:
I've re-added the clock offset, and the only way to know if it works will be in the distributed tests unfortunately. I will build a new image from this branch and see if the only Marketplace test we have in the dist tests works.

emizzle:
I've run the MarketplaceExample distributed test and it expectedly fails, but that is not a problem introduced with the changes in this PR. I think that we are likely safe with these changes, though it is difficult to tell without more robust marketplace tests in the dist test suite. As we add more tests, if this becomes an issue, we will at least be aware of it.

AuHau (approved):

markspanbroek (approved):

PR #648, opened 2023-12-05
fix: prevent partial payout underflow

emizzle:
- Bump codex-contracts-eth to prevent an underflow from occurring when calculating partial payout amounts during fillSlot.

- catch 404 in CodexClient.getPurchase and return a failure instead of letting the exception bubble

These changes are also used in a larger PR (https://github.com/codex-storage/nim-codex/pull/607) and are being introduced to break up the PR.


emizzle:
Closed in favour of https://github.com/codex-storage/nim-codex/pull/641

Issue #626, opened 2023-11-16
Maintenance for Marketplace cleanup 


AuHau:
As discussed on the call, the outstanding issue for having properly functional cleanup as part of the Marketplace requires these three parts to be included:

 - [ ] Support for manually triggering maintenance (client team?)
 - [ ] Callback which signals when and how much data was freed (client team?)
 - [ ] Registering the callback and its logic for returning the freed capacity back to Availability (Adam)

Issue #467, opened 2023-06-30
Persistent availabilities


AuHau:
Currently, when a new availability is created, is filled up with Slots and then later on these Slots finishes, the availability is not "returned". While this might be useful for some use cases, I would argue that for simple "setup and forget" use cases it is not really convenient as it requires for the host's operators keep in mind when the Slots finishes in order to come back and "refill" the Availabilities.

I would therefore propose adding a "persistent" flag to the availabilities, which would behave in a way that when the slot finishes, the slot's capacity is returned to its original availability which could be automatically offered again.

This would require few things:
 - add REST API flag to the Availability JSON structure and keep track of it internally
 - add REST API capability to "inactivate" the Availability, in order not to offer it anymore (even with the returned capacity), for the cases when host wants to gracefully terminate its offerings
 - wire "refresh signal" for the Sales components (mainly the currently developed Slot Queue) to be noticed that there was a capacity returned to availability and that the Sales process should potentially kick in again

AuHau:
I have been scoping out the implementation for this, and I believe it would make the most sense to add the property `until` to the `Availability` concept. It would function as follows when a sale is completed:

-  If no `until` is specified, the capacity is returned to the availability indefinitely or until `until` is set.
-  If `until` is specified and is in the future, then the capacity is returned.
-  If `until` is specified and is in the past, then the capacity is **not** returned and will eventually be depleted.

One concern I have is that users might be a bit confused, as they may expect that "all sales will be finished when `until` expires." What do you think? We could even expand this modification to ensure that there are no ongoing sales linked to the Availability when `until` expires.

markspanbroek:
I really like the `until` property. Regarding the semantics; I would indeed ensure that all sales are finished before the `until` timestamp. That should make it easy for node owners to reason about.

AuHau:
Sounds good to me!

emizzle:
Sounds great, and I agree with @markspanbroek that on, what I assume is a REST api call to update an Availability, updating the `until` of an Availability would not succeed unless all active sales are completed before that time.

emizzle:
Thinking about this again, I feel like the `inactive` flag and the `until` flag should be two flags that control the same behaviour (slot queue + capacity recycling), which should be differentiated from deleting an Availability (something we haven't really discussed before). A node operator should be able to deactivate an Availability with active sales, but should not be able to delete it.

### `inactive`/`until` behaviour
[...]

PR #650, opened 2023-12-11
feat: cleanup returning availability

### LINKED ISSUE #411: Cleanup for `SaleErrored`
When the Sale state-machine reaches the `SaleErrored`, there should be a cleanup process that will release the potentially allocated resources such as:
- downloaded bytes
- restore the original availability size
- ... and potentially more

AuHau:
This pull request brings the final piece to close issue #411 with a naive implementation. It introduces the concept of "returning bytes to availability," which is done for failed sales. This is achieved by attempting to re-reserve the required bytes from the Repo store. If successful, the bytes are returned; otherwise, an error is logged and the availability is lost.

I am unsure about the API of the `returnBytes` function, so please let me know in the review if you prefer it as it is.

There will be follow-up work later on that will intelligently handle the returning of bytes using a currently non-existing mechanism in RepoStore. This mechanism will clean up space for us when it is full, in order to avoid losing availability. This is partially addressed in the following link: [https://github.com/codex-storage/nim-codex/issues/626](https://github.com/codex-storage/nim-codex/issues/626)

Closes #411.

### DIFF
--- a/codex/sales.nim
+++ b/codex/sales.nim
@@ -109,6 +109,7 @@ proc remove(sales: Sales, agent: SalesAgent) {.async.} =
 
 proc cleanUp(sales: Sales,
              agent: SalesAgent,
+             returnBytes: bool,
              processing: Future[void]) {.async.} =
 
   let data = agent.data
@@ -119,9 +120,19 @@ proc cleanUp(sales: Sales,
     reservationId = data.reservation.?id |? ReservationId.default,
     availabilityId = data.reservation.?availabilityId |? AvailabilityId.default
 
-  # TODO: return bytes that were used in the request back to the availability
-  # as well, which will require removing the bytes from disk (perhaps via
-  # setting blockTTL to -1 and then running block maintainer?)
+  # if reservation for the SalesAgent was not created, then it means
+  # that the cleanUp was called before the sales process really started, so
+  # there are not really any bytes to be returned
+  if returnBytes and request =? data.request and reservation =? data.reservation:
+    if returnErr =? (await sales.context.reservations.returnBytes(
+                        reservation.availabilityId,
+                        reservation.id,
+                        request.ask.slotSize
+                      )).errorOption:
+          error "failure returning bytes",
+            error = returnErr.msg,
+            availabilityId = reservation.availabilityId,
+            bytes = request.ask.slotSize
 
   # delete reservation and return reservation bytes back to the availability
   if reservation =? data.reservation and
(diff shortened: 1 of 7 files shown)


codex/sales/reservations.nim
```diff
@@ -359,6 +359,56 @@ proc createReservation*(
 
   return success(reservation)
 
+proc returnBytes*(
```
markspanbroek, line 362: Can we think of a more descriptive name for this function? It's a bit hard to know what it does outside of the context of this PR. Maybe `returnBytesToRepo()` or `cancelReservation()`?

codex/sales/reservations.nim
```diff
@@ -359,6 +359,56 @@ proc createReservation*(
...
+  # Update availability with returned size
+  if updateErr =? (await self.update(availability)).errorOption:
+
+    trace "Rolling back returning bytes"
+    if rollbackErr =? (await self.repo.release(bytesToBeReserved)).errorOption:
+      rollbackErr.parent = updateErr
+      return failure(rollbackErr)
+
+    return failure(updateErr)
+
```
markspanbroek, line 409: This pattern (trying to update something, and then rolling back if it doesn't work) is used in multiple places in this file. Perhaps we can extract it to a separate function?

codex/sales/states/cancelled.nim
```diff
@@ -29,6 +29,6 @@ method run*(state: SaleCancelled, machine: Machine): Future[?State] {.async.} =
     onClear(request, data.slotIndex)
 
   if onCleanUp =? agent.onCleanUp:
-    await onCleanUp()
+    await onCleanUp(true)
```
markspanbroek, line 32:
```suggestion
    await onCleanUp(returnBytes = true)
```

codex/sales/states/errored.nim
```diff
@@ -29,5 +29,5 @@ method run*(state: SaleErrored, machine: Machine): Future[?State] {.async.} =
     onClear(request, data.slotIndex)
 
   if onCleanUp =? agent.onCleanUp:
-    await onCleanUp()
+    await onCleanUp(true)
```
markspanbroek, line 32:
```suggestion
    await onCleanUp(returnBytes = true)
```

codex/sales.nim
```diff
@@ -108,6 +108,7 @@ proc remove(sales: Sales, agent: SalesAgent) {.async.} =
 
 proc cleanUp(sales: Sales,
              agent: SalesAgent,
+             returnBytes: bool,
```
markspanbroek, line 111: In what cases do we want this to be false? Shouldn't we always try to return bytes to repo?

codex/sales/reservations.nim
AuHau: I don't see simple way how to achieve that. Each "roll-back" action is different based on what had been done... If you see how to implement that, please let me know.

codex/sales.nim
AuHau: In the case where we successfully finish the state machine, in the current state, we don't have "persistent availabilities", so we don't return bytes once the sale is successfully finished. I created an issue about it here #467 some time ago. We can implement that afterwards, but this PR is not about that.

Issue #646, opened 2023-12-04
[BUG] Block verification fails


benbierens:
The following warning message show up during dist-testing:
```
2023-12-04 11:26:12 WRN 2023-12-04 10:26:12.079+00:00 Block validation failed                    topics="codex blockexcengine" tid=1 msg="Received block is not currently a pending block"
```
Message shows up in groups of 30 to 40 lines, multiple times during test runs.
Observed in Codex SHA-0e4387d

tbekas: It may happen due to a couple of reasons:
- some peer sends us a block twice
- we ask more than 1 peer for the same block

(Second item is unlikely: occurs in tests with 1 uploader and 1 downloader.)

benbierens:
Not seen in `TwoClient` test. Yes seen in `MarketplaceExample` test.
Validation failures follow interesting pattern. Blocks are identified by tree CID + index:
First section:
index: 5
index: 6
Some time later:
index: 5
index: 6
index: 7
index: 8
Some time later:
index: 7
index: 8
index: 9
index: 10
Some time later:
index: 7
index: 8
index: 9
index: 10
index: 11
index: 12

etc, etc.

benbierens:
Consider this code, from engine.nim
```nim
      trace "Sending blocks to peer", peer = task.id, blocks = blocksDelivery.len
      await b.network.request.sendBlocksDelivery(
        task.id,
        blocksDelivery
      )

      codex_block_exchange_blocks_sent.inc(blocksDelivery.len.int64)

      trace "About to remove entries from peerWants", blocks = blocksDelivery.len, items = task.peerWants.len
      # Remove successfully sent blocks
      task.peerWants.keepIf(
        proc(e: WantListEntry): bool =
          not blocksDelivery.anyIt( it.address == e.address )
      )
      trace "Removed entries from peerWants", items = task.peerWants.len
```

sendBlocksDelivery is called and works correctly. But, in some cases, the trace "About to remove..." is never reached. This suggests that the async function isn't returning, it's being interrupted by other async tasks.

To check this, I set this to 1.
```nim
DefaultConcurrentTasks = 10
```
Changing this to 1 solves the issue. Something in the concurrent tasks is preventing the task.peerWants list from being cleaned up after sending. It doesn't make sense to me yet because the peer tasks are kept in an async-heapqueue which I checked: In the test in question there are 2 nodes in the network and as expected the task queue is either empty, or has 1 peer in it. (Duplicate peers are not allowed in this queue and that seems to work OK)
[...]

PR #651, opened 2023-12-12
Fix block retransmit

benbierens:


codex/blockexchange/engine/engine.nim
```diff
@@ -581,7 +582,9 @@ proc blockexcTaskRunner(b: BlockExcEngine) {.async.} =
       peerCtx = await b.taskQueue.pop()
 
     trace "Got new task from queue", peerId = peerCtx.id
+    await peerCtx.lock.acquire()
     await b.taskHandler(peerCtx)
```
tbekas, line 586: Lock is never released if this line throws

tbekas (changes requested):

codex/blockexchange/engine/engine.nim
```diff
@@ -581,7 +582,9 @@ proc blockexcTaskRunner(b: BlockExcEngine) {.async.} =
       peerCtx = await b.taskQueue.pop()
 
     trace "Got new task from queue", peerId = peerCtx.id
+    await peerCtx.lock.acquire()
```
dryajov, line 585: Can you explain why this is needed at all?

dryajov (changes requested):
Can you please provide a more detailed reasoning for this change?

codex/blockexchange/engine/engine.nim
tbekas: Linked issue has some context https://github.com/codex-storage/nim-codex/issues/646

benbierens:
OK, so following our discussion during the weekly node-team meeting yesterday, I realized that we might be dispatching workers at the wrong level.

Currently, `engine.nim` has a default number of workers of 10 and an async-queue that feeds the workers with "PeerContext" objects. The task of the worker is basically "Do whatever needs to happen for this peer." -> whatever needs to happen in this case is depending on information that's stored in a different scope, and is subject to change asynchronously. That's fundamentally where this bug comes from.

The lock I introduced here prevents multiple workers from acting on peer information that another worker might already be working on, but hasn't updated yet. This is a solution to the bug. But... we can do better.

Instead of having our workers-dispatcher-queue mechanism in `engine.nim`, at the level of peers, we could change this to `network.nim`, at the level of messages: "Unit-of-work". Network.nim has the rpc-handler function that processes each incoming message. If we plug in the async queue here, and dispatch to 10 message-handling workers, then I think all the information needed for the worker is contained within the queued object. This would mean that `engine.nim` would be down-stream of an unknown number of separate workers. That's something that would need to be kept in mind while writing code there.

cskiraly:
Adding my analysis, hope it help.

So, what I see happening, all triggered by receiving a new block:
  1. when we receive a block in `networkstore.putBlock`, we always call `engine.resolveBlocks`. BTW, we always call it with a single block, except in tests.
  2. `resolveBlock` (why async?) calls `scheduleTasks`
  3. `scheduleTasks` (why async?) checks who needs it. In this check it does not consider in-flight info, since we don't have that on `PeerCtx`
  4. it updates the queue for each peer that needs this new block. (why heap, how is it sorted?)
  5. workers, in an infinite loop, pop from that heap in `blockexcTaskRunner`, and do their `await b.taskHandler(peerCtx)`
  6. taskHandler is sorting the want list, in descending order, based on priority  (are we setting it? If not, how is that sort behaving?),
  7. then it looks up all blocks, and waits for all of these (why are we waiting for all here, and not after sending?)
  8. then it sends all blocks with a single call to `sendBlocksDelivery`, queuing all blocks as one message (do I get it right? why a single message?) in the underlying `BlockExcNetwork`
  9. `BlockExcNetwork` seems to have a singe `inflightSema`, initialized with `maxInflight`, for all the peers. Inside that semaphore, both `await b.connect()` and `await conn.writeLp` are called, which is I suppose the reason we need more workers, but I'm wondering what's the relation between `maxInflight` (now 100) and the number of workers (now 10).

cskiraly:
It also feels strange to me that we are sending all blocks as a single message, if I got it right. I also agree that finer granularity scheduling would make sense. Moving it to the block level.

cskiraly:
As for the duplicates, I think the problem is as follows:
`taskHandler` is taking the info from the shared `task:BlockExcPeerCtx`, and it is copying it into its own `wantsBlocks`, and then starts it's async life. It only updates `task.peerWants` at the very end, when send returns. This means that:
- if a next block is coming in, the BlockExcPeerCtx on heap is updated, and another worker might try to handle the same peer, with the same, not yet updated want list
- also, if the same peer sends an updated IWANT list, IWANT state might get messed up. I think it is OK, but still, worth a check

dryajov:
> As for the duplicates, I think the problem is as follows: `taskHandler` is taking the info from the shared `task:BlockExcPeerCtx`, and it is copying it into its own `wantsBlocks`, and then starts it's async life. It only updates `task.peerWants` at the very end, when send returns. This means that:
> 
> * if a next block is coming in, the BlockExcPeerCtx on heap is updated, and another worker might try to handle the same peer, with the same, not yet updated want list
> * also, if the same peer sends an updated IWANT list, IWANT state might get messed up. I think it is OK, but still, worth a check
> 
> Adding inflight info would I think solve this, even without a lock, or at least with less locking.
> 
> Adding inFlight list to BlockExcPeerCtx:
> 
> * add to it when sent
> * remove from it when receiving have list
> * handle anomaly, when it grows? This is clearly to be discussed, as it depends on how HAVE lists and WANT lists are being sent. Anyway, having a reliable transport underneath, I don't think we should time out on it, IF we handle broken peer connections by clearing state correctly.

This seems to be the real issue - we're removing the blocks too late, this is later exacerbated by the fact that we're currently sending blocks one by one.

There are at least two solutions - remove the blocks as soon as they are selected for processing or add an in-flight flag that keeps track of them being, well, in-flight and skips them on the next request.

An in-flight flag would be more robust than simply removing the blocks from the iwant list, it will ensure that malicious or inconsistent peers, don't send spurious duplicate iwants requests. If a peer keeps sending iwant for blocks it already requested/sent a previous iwant, then ideally that peer is downgraded and eventually disconnected, but since we don't have any peer-scoring in place the in-flight flag at least handles the most basic cases.

On the topic of sending blocks on by one, to address this, we can add another queue that holds some number of blocks (say 10) and processes them when either 1) the queue is full or b) a (short <100ms!) timeout has elapsed at which point the blocks in the queue get processed/sent. (leaving this here as an Idea, but doesn't need to be addressed right away.)

------------------

All in all, locking per peer isn't a good idea, the design allows for concurrent processing of multiple tasks/requests for the same peer, we're simply not updating the current state of the peer correctly.

dryajov:
> Instead of having our workers-dispatcher-queue mechanism in engine.nim, at the level of peers, we could change this to network.nim, at the level of messages: "Unit-of-work". 

This is a bad idea for a couple of reasons:

- First, the networking layer should not do any decision making whatsoever, it is structured the way precisely for this reason, to prevent leaking application level logic into the networking layer, which should only be concerned with network level stuff. All application level logic should live in the engine. 
- Second, the only thing the engine does, is send blocks and (in the future) do bookkeeping per peer, thats why the task processing was simplified to queueing the peer - it's unclear what level of granularity we'd want per task, so I would not go that route right now, perhaps in the future __if__ more complex tasks/flows are required this might be needed, but for now this will simply add unneeded complexity.

benbierens:
Queueing incoming messages in the network layer wouldn't introduce any new decision making logic there.
The engine with its peer-context object is already doing a bunch of admin per peer, and it's causing the issue.
I'm going to give this one some more time, and see if I can solve this in another way.

codex/blockexchange/engine/engine.nim
```diff
@@ -529,7 +529,8 @@ proc taskHandler*(b: BlockExcEngine, task: BlockExcPeerCtx) {.gcsafe, async.} =
 
   trace "wantsBlocks", peer = task.id, n = wantsBlocks.len
   if wantsBlocks.len > 0:
-    trace "Got peer want blocks list", items = wantsBlocks.len
+    task.peerWants = task.peerWants.filterIt(it notin wantsBlocks)
```
dryajov, line 532: This works, but maybe `keepItIf()` is better, which doesn't make a copy.

dryajov (changes requested):
This works, but, I think the flag is a better overall solution as I already mentioned. Its a tiny bit more work, but it's worth it.

I'm ok with this if you have good arguments against it the flag tho...

benbierens:
This is the smallest, simplest change I can make that solves the issue. That's the only argument I have.
By flag, you mean keeping a list of in-flight blocks per peer context, right?

dryajov:
> This is the smallest, simplest change I can make that solves the issue. That's the only argument I have. By flag, you mean keeping a list of in-flight blocks per peer context?

Yes, you can add the flag directly to the `WantListEntry` object, and don't serialized it.

codex/blockexchange/engine/engine.nim
```diff
@@ -529,8 +529,13 @@ proc taskHandler*(b: BlockExcEngine, task: BlockExcPeerCtx) {.gcsafe, async.} =
 
   trace "wantsBlocks", peer = task.id, n = wantsBlocks.len
   if wantsBlocks.len > 0:
-    trace "Got peer want blocks list", items = wantsBlocks.len
-
+    # Consider only the wants not currently in-flight,
+    # Then mark them as in-flight.
+    wantsBlocks.keepItIf(not it.inFlight)
```
dryajov, line 534: You probably just want to only get the non-inflight in the `filterIt` above, instead of the `keepItIf`, like so:

```nim
  var
    wantsBlocks = task.peerWants.filterIt(
      it.wantType == WantType.WantBlock and not it.inFlight
    )
```

codex/blockexchange/engine/engine.nim
```diff
@@ -529,8 +529,13 @@ proc taskHandler*(b: BlockExcEngine, task: BlockExcPeerCtx) {.gcsafe, async.} =
...
+    let wantAddresses = wantsBlocks.mapIt(it.address)
+    for peerWant in task.peerWants.mitems:
+      if peerWant.address in wantAddresses:
+        peerWant.inFlight = true
```
tbekas, line 538: I think this impl is bugged. 

1. All blocks that were not in flight are marked as in flight.
2. A block that was marked as `inFlight=true` was not found in repostore.
3. Such block will never be delivered (even if we'll have it in repostore future).

I liked previous impl, since it was very simple and achieved the same result.

tbekas (changes requested):

codex/blockexchange/engine/engine.nim
benbierens: Good catch! I don't see how to detect if transmit failed. The send method doesn't return anything. I can set the flag back to false when the local read fails.

benbierens: It seems I can't figure out if the block transmit failed without changing the signatures all over the codebase. Also I can't write any unit tests for this because everything's so tightly coupled. I can check it's OK using the integration/dist-tests, but that it can't check that transmit-failures or local-read failures are handled correctly. At least, not without running the tests for a very long time and hoping one of those things happen.

benbierens:
Update:
Original bug (block retransmit) is fixed within the constraint that an in-flight flag must be used. Bug is captured by dist-tests and in-flight flag behavior is capture by unit-tests.
An open 'issue' that was not introduced here but was always present is: the send callback can fail silently. Network internals capture the exception and no result-type is returned. My attempt to change the return-type were not successful, would cause a lot of changes, and technically are not in scope for this bug. This means that when transmission fails, the want-blocks are cleared all the same. This means we don't have to bother with the in-flight flag in this case. This behavior is unchanged by this PR, and something we will need to fix when we pick up bandwidth incentives.

benbierens:
Branch image passes basic dist-tests.

codex/blockexchange/engine/engine.nim
```diff
@@ -555,13 +561,16 @@ proc taskHandler*(b: BlockExcEngine, task: BlockExcPeerCtx) {.gcsafe, async.} =
...
-    # Extract successfully received blocks
-    let
       blocksDelivery = blocksDeliveryFut
         .filterIt(it.completed and it.read.isOk)
         .mapIt(it.read.get)
 
+    # All the wants that failed local lookup must be set to not-in-flight again.
+    let
+      successAddresses = blocksDelivery.mapIt(it.address)
+      failedAddresses = wantsBlocks.mapIt(it.address).filterIt(it notin successAddresses)
```
tbekas, line 571: `wantsBlocks.mapIt(it.address)` seem to be already represented as `wantAddresses`

codex/blockexchange/engine/engine.nim
```diff
@@ -571,13 +580,9 @@ proc taskHandler*(b: BlockExcEngine, task: BlockExcPeerCtx) {.gcsafe, async.} =
...
       codex_block_exchange_blocks_sent.inc(blocksDelivery.len.int64)
 
-      trace "About to remove entries from peerWants", blocks = blocksDelivery.len, items = task.peerWants.len
-      # Remove successfully sent blocks
-      task.peerWants.keepIf(
-        proc(e: WantListEntry): bool =
-          not blocksDelivery.anyIt( it.address == e.address )
-      )
-      trace "Removed entries from peerWants", items = task.peerWants.len
+      let deliveryAddresses = blocksDelivery.mapIt(it.address)
```
tbekas, line 583: `blocksDelivery.mapIt(it.address)` is already represented as `successAddresses`

tbekas (commented):
Still I liked the previous approach with AsyncLock. Right now we have to mark some want entries temporarily as `inFlight = true` even though we never attempt to send them  if we don't have a corresponding block. It makes reading & understanding of the code more difficult.

codex/blockexchange/engine/engine.nim
benbierens: Good catch!

codex/blockexchange/engine/engine.nim
benbierens: Glad you caught these. Good improvement.

tbekas (approved):

dryajov:
> Still I liked the previous approach with AsyncLock. Right now we have to mark some want entries temporarily as `inFlight = true` even though we never attempt to send them if we don't have a corresponding block. It makes reading & understanding of the code more difficult.

This doesn't work, because you would prevent processing all block for a peer for the duration of the block. This isn't the intended behaviour.

codex/blockexchange/protobuf/message.nim
```diff
@@ -29,6 +29,7 @@ type
     cancel*: bool           # Whether this revokes an entry
     wantType*: WantType     # Note: defaults to enum 0, ie Block
     sendDontHave*: bool     # Note: defaults to false
+    inFlight*: bool         # Whether block sending is in progress. Not serialized.
```
dryajov, line 32: We really need an automatic way of handling serde.

dryajov (approved):
Lgtm.

The one thing to note is that this will not make the pesky message disappear, because locally resolves blocks also go thru the same flow, and locally resolves blocks aren't necessarily always requested. So the message stating that we're resolving blocks that weren't previously requested is a bit of a red hearing, this is still a valid fix tho.

We do want to ignore locally resolves blocks from being reported as not requested, since that is a totally valid condition.

codex/blockexchange/protobuf/message.nim
benbierens: I hear Eric's made a new library for this. We should take a look.

benbierens:
I haven't actually seen the retransmit warning message in any other cases than the marketplace-test. It didn't even show up in the two-client test, where I would expect it.

tbekas:
>This doesn't work, because you would prevent processing all block for a peer for the duration of the lock. This isn't the intended behaviour.

As far as I remember that fixed worked fine when @benbierens was testing it.

PR #654, opened 2023-12-15
Rework merkle tree

dryajov:


tests/codex/merkletree/testmerkledigest.nim
```diff
@@ -0,0 +1,56 @@
...
+      let
+        chunk = bytes[(i*2*KB)..<((i+1)*2*KB)]
+        digest = Sponge.digest(chunk, rate = 2)
+      leaves.add(digest)
+
+    let
+      digestTree = Poseidon2MerkleTree.digest(bytes, chunkSize = 2*KB).tryGet
+      tree = Poseidon2MerkleTree.init(leaves).tryGet
+      root = tree.root.tryGet
+      expected = digestTree.root.tryGet
```
markspanbroek, line 33: You could test against the implementation in nim-poseidon2:
```suggestion
      expected = SpongeMerkle.digest(bytes, chunkSize = 2*KB)
```
This makes the test simpler, and stronger because you're testing against a different implementation.

codex/merkletree/merkletree.nim
```diff
@@ -7,412 +7,157 @@
...
+  for layer in self.layers:
+    yield layer
+
+iterator nodes*[H, K](self: MerkleTree[H, K]): H =
+  for layer in self.layers:
+    for node in layer:
+      yield node
+
+func root*[H, K](self: MerkleTree[H, K]): ?!H =
+  let last = self.layers[self.layers.len - 1]
```
markspanbroek, line 62:
```suggestion
  let last = self.layers[^1]
```

codex/merkletree/merkletree.nim
```diff
@@ -7,412 +7,157 @@
...
+  var
+    m = proof.nleaves
+    j = proof.index
+    h = leaf
+    bottomFlag = K.KeyBottomLayer
+
+  for p in proof.path:
+    let oddIndex : bool = (bitand(j,1) != 0)
+    if oddIndex:
+      # the index of the child is odd, so the node tiself can't be odd (a bit counterintuitive, yeah :)
```
markspanbroek, line 112:
```suggestion
      # the index of the child is odd, so the node itself can't be odd (a bit counterintuitive, yeah :)
```

markspanbroek (approved):
Thanks @dryajov, looks good! I left a few comments.

codex/merkletree/poseidon2.nim
```diff
@@ -0,0 +1,116 @@
...
+  Poseidon2MerkleProof* = MerkleProof[Poseidon2Hash, PoseidonKeysEnum]
+
+converter toKey*(key: PoseidonKeysEnum): Poseidon2Hash =
+  case key:
+  of KeyNone: KeyNoneF
+  of KeyBottomLayer: KeyBottomLayerF
+  of KeyOdd: KeyOddF
+  of KeyOddAndBottomLayer: KeyOddAndBottomLayerF
+
+converter toBool*(x: CtBool): bool =
```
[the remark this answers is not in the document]
dryajov: > CtBool is a protection mechanism from constantine that ensures that you don't inadvertently use normal non-constant functions in your logic, and thereby introduce side channel attacks.

Yes, this is meant for tests and it should be moved there.

codex/merkletree/poseidon2.nim
```diff
@@ -0,0 +1,116 @@
...
+  while pos < nodes.len:
+    self.layers.add( nodes[pos..<(pos + layer)] )
+    pos += layer
+    layer = divUp(layer, 2)
+
+  let
+    index = Rng.instance.rand(nleaves - 1)
+    proof = ? self.getProof(index)
+
+  ? proof.verify(self.leaves[index], ? self.root) # sanity check
```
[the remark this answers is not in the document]
dryajov: > I'm not sure about this. I'd expect this kind of checking to be done in a unit or fuzzing test?

This is a sanity check to verify that a serialized tree was deserialized and loaded correctly. The `fromNodes` function does just that, takes a list of nodes and fills the tree up, without rebuilding it, to avoid incurring the cost multiple times. This checks that the tree was loaded/filled correctly by sampling a random proof. We might actually want to sample several random proofs for more confidence. 

We can eventually remove it, but I'm keeping it around in case we end up sending the entire tree over the network. I'm not settled on that yet, but it doesn't hurt anything for now.

Fuzzing should be used, but in a different context.

tests/codex/merkletree/testmerkledigest.nim
dryajov: We should probably have both checks

PR #657, opened 2023-12-18
fix: rename `provider` to `ethProvider` in tests

emizzle:
To support more than two levels of integration test suite template nesting, `provider` had to be renamed to `ethProvider`, otherwise the compiler would complain that the `provider` symbol was unidentified in the top-most level of the nested test suites.

Eg
- `ethersuite`: declares `provider`
  - `multinodesuite`
    - `marketplacesuite` (`provider` symbol can't be found)
 
Note:
1. `marketplacesuite` does not exist yet; it is introduced in https://github.com/codex-storage/nim-codex/pull/607.
2. This PR is part of reducing the size of a much larger PR https://github.com/codex-storage/nim-codex/pull/607.

### DIFF
--- a/tests/contracts/testClock.nim
+++ b/tests/contracts/testClock.nim
@@ -8,21 +8,21 @@ ethersuite "On-Chain Clock":
   var clock: OnChainClock
 
   setup:
-    clock = OnChainClock.new(provider)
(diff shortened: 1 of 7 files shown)


AuHau (approved):
LGTM

PR #661, opened 2023-12-20
fix: comment out, instead of skipping, intermittently failing test

emizzle:
`skip` seems to working only sometimes. It intermittently fails on Windows in CI. The reason for skipping the test in the first place was to prevent intermittent failures of the test itself, so adding `skip` did not improve the situation, unfortunately.


dryajov (approved):

tbekas:
Sorry for the incorrect suggestion. Also thanks for the explanation why you decided to get rid of it.

PR #664, opened 2023-12-21
chore: bump ethers, remove upraises from `onBlock` callback

emizzle:
Bumps `nim-ethers` to the latest version, `0.7.1`.

Removes `{.upraises: [].}` from `onBlock` callback


codex/contracts/clock.nim
```diff
@@ -26,7 +26,7 @@ method start*(clock: OnChainClock) {.async.} =
   if clock.started:
     return
 
-  proc onBlock(blck: Block) {.upraises:[].} =
+  proc onBlock(blck: Block) =
```
markspanbroek, line 29:
```suggestion
  proc onBlock(blck: Block) {.raises: [].} =
```

markspanbroek: We can remove the upraises library in Codex, because we don't support Nim 1.2 anymore.
But we should retain the property that no exceptions should be raised this callback, so I'd keep a `{.raises: [].}` pragma around.

markspanbroek: O wait, never mind. The compiler can deduce that it doesn't raise any exceptions, so we can indeed remove the pragma here.

markspanbroek (approved):

PR #666, opened 2023-12-22
Integrate slot builder

dryajov:


codex/slots/slotbuilder.nim
```diff
@@ -0,0 +1,267 @@
...
+          return failure(err)
+
+        # TODO: Remove this sleep. It's here to prevent us from locking up the thread.
+        # await sleepAsync(10.millis)
+
+        digest
+
+  success hashes
+
+proc buildSlotTree*(
```
benbierens, line 164: Please move the saving of the tree from `buildSlot` to `buildSlotTree`. The data sampler needs to build the tree and save it, *and* have the PoseidonTree object returned to grab proofs.

dryajov: > needs to build the tree and save it,

The sampler doesn't need to save anything, it should build the bottom and top trees of the fly, and use the repostore to access the proofs of the middle tree (which is the large tree and the one that gets persisted).

--------------------------------

There are 3 trees, two are built on the fly, the block trees and the top level slot tree, the middle tree is persisted. 

For the bottom layer, just select the block and use the `digestTree` function from `digest.nim`, and then the `tree.getProof(index)` for the specific block cell index. 

For the middle tree, use the `repostore.getCidAndProof` methods to fetch the required proof. 

The sampler should take a reference to the `RepoStore` and the `SlotBuilder`. The flow should be:

- Get the block where cell is located from the repostore, build a tree on top of it using `diget.digestTree` (or a corresponding `builder.getCellTree` which should take care of the padding), which returns a tree instance
- Invoke `cellTree.getProof(cellIndex)` for the cell proof
- Invoke `repostore.getCidAndProof(cellBlockCid)` to get the proof for the middle persisted tree, and finally, 
- Either invoke a yet to be added `buildSlotsTree` method, or just invoke the `Poseidon2Tree.init(slotRoots & self.rootsPadLeafs)` on it, to get the root of the slots and the corresponding proof on it.

I'll add the `buildCellTree` and `buildSlotsTree` methods as soon as I'm back, which are essentially invocations of `PoseidonTree.init(data & padding)`.

codex/indexingstrategy.nim
```diff
@@ -0,0 +1,61 @@
...
+
+# I'm choosing to use an assert here because:
+# 1. These are a programmer errors and *should not* happen during application runtime.
+# 2. Users don't have to deal with Result types.
+
+type
+  # Representing a strategy for grouping indices (of blocks usually)
+  # Given an interation-count as input, will produce a seq of
+  # selected indices.
+  IndexingStrategy* = ref object of RootObj
```
dryajov, line 12: This should really just be an iterator instead of a class hierarchy...

codex/slots/slotbuilder.nim
benbierens: This lines up with what I'm trying to do but it's missing one step:
When a slot is first filled, the middle tree doesn't exist in the local blockstore yet. (It was built on the client node, but is not present on the host node.) So, I set the data sampler to fetch the proofs for the middle tree using `repoStore.getCidAndProof`, but when it can't find it, the middle tree has to be created and saved. I figured I would call `slotBuilder.buildSlotTree` to do this, and if it would save the middle tree the problem would be solved. If you'd rather we move the initial middle-tree-building on the host node out of the data sampler, that's fine with me. It just needs to happen somewhere.

benbierens: I would rather not burden the users for `DataSampler` with the need to check the existence of the slot tree and recreate it if necessary. DataSampler is already doing work in this area, so I think the responsibility fits there. I'm working on this, I hope you'll approve: https://github.com/codex-storage/nim-codex/blob/b3da3eb024d2b2a1e0b915daa2a792685e228f49/codex/proof/datasamplerstarter.nim#L72

dryajov: > This lines up with what I'm trying to do but it's missing one step: When a slot is first filled, the middle tree doesn't exist in the local blockstore yet. (It was built on the client node, but is not present on the host node.) So, I set the data sampler to fetch the proofs for the middle tree using `repoStore.getCidAndProof`, but when it can't find it, the middle tree has to be created and saved. I figured I would call `slotBuilder.buildSlotTree` to do this, and if it would save the middle tree the problem would be solved. If you'd rather we move the initial middle-tree-building on the host node out of the data sampler, that's fine with me. It just needs to happen somewhere.

This should really be handled as a descrente step, outside of the sampler. It should happen at the time the node gets assigned the slot. We should simply call `buildSlot` at that moment and by the time the sampler gets called the slot should already be downloaded and stored by the node.

PR #667, opened 2023-12-23
Plumbing for Circom Rust runner and zk circuit modules

elcritch:
Currently just adding in the necessary submodules. 

Plan is to then:

- integrate cargo build for the rust circom runner circo
- setup the nim-rust ffi interface (manually or automated)
- link the resulting circom lib into the Codex build
- get basic msgpack test running from unit tests using example circuits
- setup the build so the new circuits can be added when ready


dryajov:
Closing in favor of #698

PR #668, opened 2024-01-09
Build slots

dryajov:


codex/erasure/erasure.nim
```diff
@@ -338,16 +343,16 @@ proc encodeData(
 proc encode*(
   self: Erasure,
   manifest: Manifest,
-  blocks: int,
-  parity: int): Future[?!Manifest] {.async.} =
+  blocks: uint,
+  parity: uint): Future[?!Manifest] {.async.} =
```
elcritch, line 347: Blocks is converted to an int here. Perhaps it'd make sense to use a `Natural` or even make a new `NoBlocks` int type?

codex/manifest/manifest.nim
```diff
@@ -308,5 +315,5 @@ proc new*(
     originalTreeCid: manifest.treeCid,
     originalDatasetSize: manifest.originalDatasetSize,
     verifiable: true,
-    verificationRoot: verificationRoot,
-    slotRoots: slotRoots))
+    slotsRoot: slotsRoot,
+    slotRoots: @slotRoots)
```
elcritch, line 319: Sure we want to copy here? It could be a `move` annotation in the future instead.

codex/node.nim
```diff
@@ -56,17 +59,47 @@ type
...
+    engine: BlockExcEngine
+    erasure: Erasure
+    discovery: Discovery
     contracts*: Contracts
     clock*: Clock
 
   OnManifest* = proc(cid: Cid, manifest: Manifest): void {.gcsafe, closure.}
 
+func switch*(self: CodexNodeRef): Switch =
+  return self.switch
```
elcritch, line 74: Why add accessors? Do you think we'll need accessors in the future or is it just to prevent writes?

codex/rest/api.nim
```diff
@@ -300,8 +300,8 @@ proc initPurchasingApi(node: CodexNodeRef, router: var RestRouter) =
       ## proofProbability - how often storage proofs are required
       ## reward           - the maximum amount of tokens paid per second per slot to hosts the client is willing to pay
       ## expiry           - timestamp, in seconds, when the request expires if the Request does not find requested amount of nodes to host the data
-      ## nodes            - minimal number of nodes the content should be stored on
-      ## tolerance        - allowed number of nodes that can be lost before pronouncing the content lost
+      ## nodes            - number of nodes the content should be stored on
+      ## tolerance        - allowed number of nodes that can be lost content is lost
```
elcritch, line 304: Not sure I understand either of those revised sentences 😇

codex/slots/builder.nim
```diff
@@ -252,24 +313,53 @@ proc new*(
...
+        ? Poseidon2Hash.fromBytes(
+          ( ? it.mhash.mapFailure ).digestBytes.toArray32
+        ).toFailure
+      )
+
+    without root =? self.buildRootsTree(slotRoots).?root(), err:
+      error "Failed to build slot roots tree", err = err.msg
+      return failure(err)
+
+    if not bool(slotRoot == root): # TODO: `!=` doesn't work for SecretBool
```
benbierens, line 359: We have a fix for this in poseidon2. See https://github.com/codex-storage/nim-poseidon2/commit/eef2603c11eeb5069eaed6c0296f3a1c5d23aad1
Used: https://github.com/codex-storage/nim-codex/blob/1836c2476c4ac2822f8274f4b2530e07e53c9cc4/codex/proof/datasamplerstarter.nim#L80

codex/erasure/erasure.nim
dryajov: Blocks are really the K param for the erasure coding, so uint should be fine.

codex/node.nim
dryajov: Prevent writes

codex/manifest/coders.nim
```diff
@@ -109,7 +109,7 @@ proc decode*(_: type Manifest, data: openArray[byte]): ?!Manifest =
     blockSize: uint32
     originalDatasetSize: uint32
     ecK, ecM: uint32
-    verificationRoot: seq[byte]
+    slotsRoot: seq[byte]
     slotRoots: seq[seq[byte]]
```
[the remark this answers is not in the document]
dryajov: Yeah, I agree, I like verify root or proofRoot.

codex/manifest/manifest.nim
dryajov: Move would destroy the source tho? We should probably use sync/lent, but I want to leave that for a future pass.

codex/erasure/erasure.nim
dryajov: Actually, `Natural` is probably better.

codex/manifest/manifest.nim
dryajov: So, I've always assumed that `@` from `openArray` would _not_ make a copy, so I checked `system.nim` and...

```nim
  proc `@`*[T](a: openArray[T]): seq[T] {.magic: "OpenArrayToSeq".} =
    ## Turns an *openArray* into a sequence.
    ##
    ## This is not as efficient as turning a fixed length array into a sequence
    ## as it always copies every element of `a`.
    newSeq(result, a.len)
    for i in 0..a.len-1: result[i] = a[i]
```

Both, from the comments and the implementation, it does seem to make a copy... thats weird, why would it? But, it has a magic attached `OpenArrayToSeq`, so I checked the generated c source and...

```C
asgnRef((void**) (&(*T14_)._protected_0_1._verifiable_1.slotRoots), at___OOZOOZOOZcodexZslotsZbuilder_9449(slotRoots, slotRootsLen_0));
```

where `asgnRef` is 

```C
static N_INLINE(void, asgnRef)(void** dest, void* src) {
        ....
	nimln_(288, "/nim-codex/vendor/nimbus-build-system/vendor/Nim/lib/system/gc.nim");
	(*dest) = src;
	popFrame();
}
```

`(*dest) = src;` does a pointer assignment

So, it doesn't look like it's doing a copy.

Btw, a move wouldn't be appropriate here, since again, I'm simply converting `openArray` to a `seq`

PR #676, opened 2024-01-16
Wire sampler

dryajov:


codex/node.nim
```diff
@@ -543,6 +544,47 @@ proc onStore(
...
+  without sampler =? DataSampler.new(slotIdx, self.blockStore, builder), err:
+    error "Unable to create data sampler", err = err.msg
+    return failure(err)
+
+  without proofInput =? await sampler.getProofInput(challenge, nSamples = 3), err:
+    error "Unable to get proof input for slot", err = err.msg
+    return failure(err)
+
+  # Todo: send proofInput to circuit. Get proof. (Profit, repeat.)
+  success(@[42'u8])
```
elcritch, line 586: Yes!! ;)

elcritch: So here's where we'd want to call into the Rust Arkcom?

dryajov: yep

PR #681, opened 2024-01-19
Debug/sampling

benbierens:
No unit-test nor implementation handling existed for creating proving samples from padded data cells. The unit tests used a data setup where each slot would always contain a power-of-two number of cells, causing padding to never happen and sampling in the range of padding to never happen.
This PR adjusts the test data so padding happens, and adds tests to check that sampling in this range behaves as expected.


benbierens:
I've built an image of this branch and used it to verify this indeed fixes the node crash.

dryajov (approved):
LGTM

dryajov:
@benbierens  Can we resolve conflicts and get this merged?

dryajov (approved):
LGTM

PR #688, opened 2024-01-29
Update actions to lates major versions

veaceslavdoina:
We have a lot of `node 16` deprecation warnings in the GitHub Actions.

PR updates all actions to the latest major versions and almost all of them in the release notes mentions node version upgrade.

We probably should perform same update in other repositories as well.

Screenshot 2024-01-29 at 14 44 14


markspanbroek (approved):
Looks good, thanks @veaceslavdoina!

PR #691, opened 2024-02-01
remove call to `waitFor` from clock.now and make it async

gmega:
`clock.now` is the last method of `Clock` that isn't yet async. The requirement for an async `clock.now` [already exists in test code](https://github.com/codex-storage/nim-codex/blob/master/codex/contracts/clock.nim#L54), but we've been avoiding making that change and have so far instead resorted to using `waitFor` and maintaining a non-async signature. This is however not recommended, and it breaks the profiler.

This PR makes the required changes so that clock.now becomes async. The alternative is figuring out a different way to implement the test so that the `waitFor` call is no longer required, and then we can kick the can down the road.

### DIFF
--- a/codex/clock.nim
+++ b/codex/clock.nim
@@ -1,6 +1,5 @@
 import pkg/chronos
 import pkg/stew/endians2
-import pkg/upraises
 import pkg/stint
 
 type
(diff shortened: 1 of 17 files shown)


markspanbroek (approved):
Looks good to me!

PR #692, opened 2024-02-01
feat(api): availabilities patch endpoint

AuHau:
I split the work on this PR from the work on #467. It is a prerequisite for implementing persistent availabilities as there will need to be a way how to update and terminate them. 

This PR brings several changes and breaking changes to the API endpoints:
- (breaking) Changing the `POST /sales/availability` successful response type to `201`
- (breaking) Changing the `Availability` structure where the `size` property was replaced by `totalSize` property which indicates the total capacity of the availability and `freeSize` which indicate the capacity of the availability that is available for selling
- Adding `PATCH /sales/availability/{id}` that allows updating the availability properties except the `freeSize`. The updated properties are then used only for asserting the new incoming requests. The currently running requests are not effected by the update availability
- Adding `GET /sales/availability/{id}/reservations` which returns the list of Reservations that are linked to the Availability.

Other notable addition is the `Optionalize()` macro which allows you to convert `object` type into `object` type with its properties converted into `Option[T]`.

Lastly, I have spent quite some time originally debugging an issue, where originally the `PATCH` endpoint was returning `204 No Content` successful reply, but there seems to be some bug in either of `nim-presto`, `chronos` server implementation or `HTTPClient` which I have summarized in reproduction repository [here](https://github.com/AuHau/nim-http204-error) and opted instead to use simple `200` response instead.


dryajov:
We should report the bug in presto

AuHau:
> We should report the bug in presto

For sure, I already reported it: https://github.com/status-im/nim-presto/issues/77

tests/integration/testIntegration.nim
```diff
@@ -243,8 +256,53 @@ twonodessuite "Integration tests", debug1 = false, debug2 = false:
...
+    client1.patchAvailability(availability.id, duration=100.u256.some, minPrice=200.u256.some, maxCollateral=200.u256.some)
+
+    var updatedAvailability = (client1.getAvailabilities().get).findItem(availability).get
+    check updatedAvailability.duration == 100
+    check updatedAvailability.minPrice == 200
+    check updatedAvailability.maxCollateral == 200
+    check updatedAvailability.totalSize == 140000
+    check updatedAvailability.freeSize == 140000
+
+    # Non existing availability
```
markspanbroek, line 277: Instead of testing all these things in one `test`, I'd rather see them in their own `test`s. Now each of these is dependent on the state that is left behind by the others.

(would also be great if we could test this outside of the integration tests, but perhaps that's something for another day)

markspanbroek (approved):
Very nice! I like the way you addressed the PATCH functionality, very elegant.

tests/integration/testIntegration.nim
AuHau: I know, I was also thinking about it and I agree with you. The thing is the way how the `testIntegration` is structured - it is a "bundle of everything" 😅 So I didn't really want to pollute it with "many" test cases. Also, I was worried about the performance hit of having multiple small tests like this.

But okey, let me refactor it ;-)

codex/sales/reservations.nim
```diff
@@ -118,17 +123,9 @@ func toArray(id: SomeStorableId): array[32, byte] =
...
-  x.requestId == y.requestId and
-  x.slotIndex == y.slotIndex
+  x.id == y.id
 proc `==`*(x, y: Availability): bool =
-  x.id == y.id and
-  x.size == y.size and
-  x.duration == y.duration and
-  x.maxCollateral == y.maxCollateral and
-  x.minPrice == y.minPrice
+  x.id == y.id
```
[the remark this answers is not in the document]
emizzle: I think that should be ok. I'm not entirely sure if I remember correctly, but I think maybe ids didn't exist when `Availability` was first created, hence the need for comparing the properties.

codex/rest/api.nim
```diff
@@ -271,24 +273,127 @@ proc initSalesApi(node: CodexNodeRef, router: var RestRouter) =
...
+        if isSome restAv.freeSize:
+          return RestApiResponse.error(Http400, "Updating freeSize is not allowed")
+
+        if size =? restAv.totalSize:
+          # we don't allow lowering the totalSize bellow currently utilized size
+          if size < (availability.totalSize - availability.freeSize):
+            return RestApiResponse.error(Http400, "New totalSize must be larger then current totalSize - freeSize, which is currently: " & $(availability.totalSize - availability.freeSize))
+
+          if err =? (await reservations.changeAvailabilitySize(availability, size)).errorOption:
+            return RestApiResponse.error(Http500, err.msg)
```
emizzle, line 343: My first thought would be that this should be handled in the `update` instead of separating it out.

codex/sales/reservations.nim
```diff
@@ -355,6 +354,24 @@ proc createReservation*(
...
+  if availability.totalSize < newSize: # storage added
+    if reserveErr =? (await self.repo.reserve((newSize - availability.totalSize).truncate(uint))).errorOption:
+      return failure(reserveErr.toErr(ReserveFailedError))
+
+  elif availability.totalSize > newSize: # storage removed
+    if reserveErr =? (await self.repo.release((availability.totalSize - newSize).truncate(uint))).errorOption:
+      return failure(reserveErr.toErr(ReleaseFailedError))
+
+  return success()
+
```
emizzle, line 374: Could this be called directly from `update*` and then not exported? It feels more like an "internal operation" of the reservations module.

codex/rest/api.nim
```diff
@@ -271,24 +273,127 @@ proc initSalesApi(node: CodexNodeRef, router: var RestRouter) =
...
+        without keyId =? id.key.tryGet.catch, error:
+          return RestApiResponse.error(Http400, error.msg)
+
+        let reservations = contracts.sales.context.reservations
+
+        if error =? (await reservations.get(keyId, Availability)).errorOption:
+          if error of NotExistsError:
+            return RestApiResponse.error(Http404, "Availability not found")
+          else:
+            return RestApiResponse.error(Http500, error.msg)
```
emizzle, line 385: Since this is used in multiple places, could we extract this into a template for better readability? eg
```suggestion
       template availabilityNotFound(error) =
          if error of NotExistsError:
            RestApiResponse.error(Http404, "Availability not found")
          else:
            RestApiResponse.error(Http500, error.msg)

        return availabilityNotFound(error)
```

codex/rest/api.nim
```diff
@@ -271,24 +273,127 @@ proc initSalesApi(node: CodexNodeRef, router: var RestRouter) =
...
+
+        without availabilitysReservations =? (await reservations.all(Reservation, id.some)), err:
+          return RestApiResponse.error(Http500, err.msg)
+
+        # TODO: Expand this structure with information about the linked StorageRequest not only RequestID
+        let json = %availabilitysReservations
+        return RestApiResponse.response($json, contentType="application/json")
```
emizzle, line 392:
```suggestion
        return RestApiResponse.response(availabilitysReservations.toJson, contentType="application/json")
```

codex/rest/api.nim
```diff
@@ -271,24 +273,127 @@ proc initSalesApi(node: CodexNodeRef, router: var RestRouter) =
...
+      except CatchableError as exc:
+        trace "Excepting processing request", exc = exc.msg
+        return RestApiResponse.error(Http500)
+
```
emizzle, line 396: Could also do a templated block instead of wrapping it all in try/catch. eg

```nim
template onError(body) =
  try:
    body
  except CatchableError as exc:
    trace "Excepting processing request", exc = exc.msg

onError:
  without contracts =? node.contracts.host:
    return RestApiResponse.error(Http503, "Sales unavailable")
  ....
```

codex/sales/reservations.nim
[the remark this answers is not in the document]
emizzle: There's a very vague recollection of just an `id` comparison not being enough, but for the life of me, I can't remember why.

codex/sales/reservations.nim
```diff
@@ -467,7 +485,7 @@ proc storables(
       return failure(error)
   elif T is Reservation:
     # should indicate key length of 5, but let the .key logic determine it
-    without defaultKey =? key(ReservationId.default, AvailabilityId.default), error:
+    without defaultKey =? key(ReservationId.default, availabilityId |? AvailabilityId.default), error:
```
emizzle, line 488: The addition of the `availabilityId` is probably not good for this routine. Below, we're simply comparing the number of namespaces for the key, which is different for Availabilities and Reservations (and admittedly a weak design because the namespaces can change). By specifying the availabiltyId, we're not changing the query logic to find reservations for a particular availability. 

Instead, I think better options might include either:
1. pass a filter predicate into this routine to add to the `next` query filtering, OR
2. query all Reservations then filter by AvailabilityId (in memory)

codex/sales/reservations.nim
```diff
@@ -497,12 +513,13 @@ proc storables(
 
 proc all*(
   self: Reservations,
-  T: type SomeStorableObject
+  T: type SomeStorableObject,
+  availabilityId: ?AvailabilityId = AvailabilityId.none
```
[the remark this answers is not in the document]
emizzle: Perhaps instead we could add an `all` overload that accepts an availabilityId instead of making it optional. That overload would create a filter predicate that can be passed to `storables`, which would insert that filtering predicate (matching an availabilityId) into the `next` query logic.

In this overload (without the availabilityId parameter), we would pass a `none` predicate to storables.

tests/codex/sales/testreservations.nim
```diff
@@ -68,7 +68,7 @@ asyncchecksuite "Reservations module":
   test "creating availability reserves bytes in repo":
     let orig = repo.available
     let availability = createAvailability()
-    check repo.available == (orig.u256 - availability.size).truncate(uint)
+    check repo.available == (orig.u256 - availability.freeSize).truncate(uint)
 
   test "can get all availabilities":
```
emizzle, line 73: Would probably need to include some tests for getting reservations for a particular Availability. And maybe another test showing that reservations included in a different Availability are not returned.

emizzle (commented):
Awesome job Adam, lots of great work here!

tests/integration/codexclient.nim
```diff
@@ -138,27 +138,68 @@ proc getSlots*(client: CodexClient): ?!seq[Slot] =
...
+  if duration =? duration:
+    json["duration"] = %duration
+
+  if minPrice =? minPrice:
+    json["minPrice"] = %minPrice
+
+  if maxCollateral =? maxCollateral:
+    json["maxCollateral"] = %maxCollateral
+
+  client.http.patch(url, $json)
```
emizzle, line 181: Serializing an optionalized type will output `null` for `none` fields, which are handled correctly in deserialization on the api side.
```suggestion
type OptRestAvailability = Optionalize(RestAvailability)
let availability = OptRestAvailability(totalSize: totalSize, freeSize: freeSize, duration: duration, minPrice: minPrice, maxCollateral: maxCollateral)

client.http.patch(url, availability.toJson)
```

Another thought: we could also consider exporting this type from `rest/json` instead of creating it each time.

tests/integration/testIntegration.nim
```diff
@@ -263,8 +276,78 @@ twonodessuite "Integration tests", debug1 = false, debug2 = false:
...
+
+    let updatedAvailability = (client1.getAvailabilities().get).findItem(availability).get
+    check updatedAvailability.duration == 100
+    check updatedAvailability.minPrice == 200
+    check updatedAvailability.maxCollateral == 200
+    check updatedAvailability.totalSize == 140000
+    check updatedAvailability.freeSize == 140000
+
+    let nonExistingResponse = client1.patchAvailabilityRaw(AvailabilityId.example, duration=100.u256.some, minPrice=200.u256.some, maxCollateral=200.u256.some)
+    check nonExistingResponse.status == "404 Not Found"
```
emizzle, line 298: Imo, this probably should be handled by REST unit tests (which don't exist yet), instead of an integration test. Definitely out of scope of this PR obviously, but something we should look to implement in the future

codex/sales/reservations.nim
AuHau: I wanted to do it originally like that, but then I thought I won't have the diff between the old/new values of size, but now I have realized that I can load it from the data store before updating to compare the values like that, so - yes, I will do that!
Thanks ;-)

tests/codex/sales/testreservations.nim
AuHau: Yeah, should have done that 🙈 Thanks for the catch!

tests/integration/codexclient.nim
AuHau: I was banging my head on why integration tests were not passing and it seemed like there were no changes updated on the node's side, while when I tested it locally it worked. It turned out that yours suggestion is actually serializing into empty JSON object `{}` 😳

It is because the `Optionalize` does not preserve the `serialize` pragmas. Honestly, I don't want to tinker with the macro anymore as that would be another time blackhole, so I will reject your proposal, even though I like it more than the original. 

If/when we find the need to extend the `Optionalize` to honor `serialize` pragmas then we can revisit this part.

codex/sales/reservations.nim
```diff
@@ -215,6 +214,39 @@ proc update(
...
+        if reserveErr =? (await self.repo.reserve((obj.totalSize - oldAvailability.totalSize).truncate(uint))).errorOption:
+          return failure(reserveErr.toErr(ReserveFailedError))
+
+      elif oldAvailability.totalSize > obj.totalSize: # storage removed
+        if reserveErr =? (await self.repo.release((oldAvailability.totalSize - obj.totalSize).truncate(uint))).errorOption:
+          return failure(reserveErr.toErr(ReleaseFailedError))
+  else:
+    let err = getResult.error()
+    if not (err of NotExistsError):
+      return failure(err)
```
AuHau, line 246: I know this is not the nicest construct, but I have not figured a way how to achieve this using some nice construct from Questionable, because I need to continue the execution of the proc in case when the error is `NotExistsError` and in the same time not to evaluate the sizing changes...

If you know of better way please let me know.

emizzle:
```nim
without availability =? await self.get, error:
  # do something with error 

# do something with availability 
```

tests/integration/codexclient.nim
emizzle: No problem, I understand. For future reference, you can annotate the type with `{.serialize.}` and then there's no need to annotate each property, ie all properties will be serialized by default.

AuHau:
Oh shi*, I have no idea how did I add this to the merge queue 😳 Please review it anyhow and if there are some outstanding issues, I will tackle them in follow up PR.

codex/sales/reservations.nim
AuHau: Hmmm, I know about this construct, but did not think of a way how I could wrap it on my use case. But you (again 😅) gave me idea :-D Thanks! Will fix it in follow-up PR.

codex/rest/api.nim
```diff
@@ -271,24 +273,122 @@ proc initSalesApi(node: CodexNodeRef, router: var RestRouter) =
...
+        without keyId =? id.key.tryGet.catch, error:
+          return RestApiResponse.error(Http400, error.msg)
+
+        let reservations = contracts.sales.context.reservations
+
+        if error =? (await reservations.get(keyId, Availability)).errorOption:
+          if error of NotExistsError:
+            return RestApiResponse.error(Http404, "Availability not found")
+          else:
+            return RestApiResponse.error(Http500, error.msg)
```
emizzle, line 381: I'm not sure, but this block (that gets an `Availability` by id) might be unnecessary, as we're fetching results from `all` afterwards which . I *think* that `all` would return `@[]` when no `Reservations` are found, then a `404` can be returned. The constraint that an `Availability` must exist for `Reservations` to exist should already be handled by the reservations module, so there really shouldn't be a case where `Reservations` exist for an `AvailabilityId`, but the `Availability` doesn't exist.

codex/utils/options.nim
```diff
@@ -0,0 +1,61 @@
...
+    fields.add(newFieldNode)
+
+  # Creates new object type T with the fields lists from steps above.
+  let tSym = genSym(nskType, "T")
+  nnkStmtList.newTree(
+      nnkTypeSection.newTree(
+        nnkTypeDef.newTree(tSym, newEmptyNode(), nnkObjectTy.newTree(newEmptyNode(), newEmptyNode(), fields))
+      ),
+      tSym
+  )
```
emizzle, line 55:
```diff
+proc addSerializePragma(typeNode: NimNode): NimNode =
+  nnkPragmaExpr.newTree(
+    typeNode,
+    nnkPragma.newTree(
+      newIdentNode("serialize")
+    )
+  )

+ let tSymAnnotated = tSym.addSerializePragma()
  let tree = nnkStmtList.newTree(
      nnkTypeSection.newTree(
-        nnkTypeDef.newTree(tSym, newEmptyNode(), nnkObjectTy.newTree(newEmptyNode(), newEmptyNode(), fields))
+        nnkTypeDef.newTree(tSymAnnotated, newEmptyNode(), nnkObjectTy.newTree(newEmptyNode(), newEmptyNode(), fields))
      ),
      tSym
  )
```

AuHau: Nice! Did you test it? Does it work? Would you be willing to create PR for it?
IMHO I would just not be adding the pragma automatically to all the fields. Either, it should be smart enough to detect that the pragma was present on the original field and only then add it to the new type, or a simpler approach could be adding some flag to the `Optionalize()` which would enable this behavior.

codex/rest/api.nim
AuHau: You are right that the fetched availability object is not really used for anything. But I don't think using the approach of "no reservation found == 404" is the right way, because what if Availability exists but simply does not have any Reservation yet?...

But it reminded me that there is `exist(key)` function in `reservations.nim` which would be better suited for this check compared to fetching the whole object 👍

PR #693, opened 2024-02-04
Add Nim-matrix workflow to run on merge queue

veaceslavdoina:
### Context

In PR [Build improvements #680](680), we raised the question about building Codex using different Nim versions, but **only before the merge** to not slow-down existing PR flow and to not waste resources on every push.

We can try to use [GitHub merge queue](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue) to implement that.

### PR

 - Add `build-status` job to `CI` workflow to check matrix execution status - will use it in branch rule
 - Add `Nim-matrix` workflow to **run only on merge queue** and only on Linux
 - [Using concurrency](https://docs.github.com/en/actions/using-jobs/using-concurrency) for `CI` workflow to cancel previous run at PR next commits

### Next steps

 - Are we ready for experiments ? 😄
 - Merge
 - Update `branch rules` with **just** `build / status` job
 - Enable `merge queue` in `branch rules`

-- Not ready? Just not enable `merge queue` and then run `Nim-matrix` workflow manually.

### Known issues

 - [There is no way to set different checks for `PR` and `merge queue`](https://github.com/orgs/community/discussions/46757?sort=old#discussioncomment-4909336) and this is why we use single `build-status` job name for both workflows
 - When merge queue checks fail we should press merge button again they will re-run them from scratch
 - [Jumping to the top of a merge queue](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue#jumping-to-the-top-of-the-queue) will cause a full rebuild of all in-progress pull requests
 - By running multiple additional jobs in `Nim-matrix` workflow we consume GitHub Org [`Total concurrent jobs`](https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#usage-limits) limit
 - `Nim-matrix` workflow fail very often on BuildJet runners at tests - we use GitHub ones for now
 - By adding an additional step before the merge, we will **delay merge up to 20 minutes**

<details>
<summary>More details</summary>

### Implementation

In the [current merge queue implementation](https://github.com/orgs/community/discussions/46757) we can't set branch rules separately for `PR` and `merge queue` checks and this is very confusing. In that way, both workflows should be run on `PR` and `merge queue` and we would need to run steps conditionally and that adds some ambiguity.

This is why we workaround that by using a single `build-status` job name for both workflows and run them separately. In that way, `PR` and `merge queue` checks will pass at their own stages.

Basically, flow will be the following

**PR**
 1. PR is created
 2. `CI` workflow will be triggered
 3. At next push, p.2 will be repeated
 4. When `CI` workflow will be finished successfully and auto-merge is enabled, PR will be moved into the merge queue

**Merge queue**
 1. `Nim-matrix` workflow will be triggered
 2. `Nim-matrix` workflow will mark branch as protected (push is not permitted) and will
     - create a temporary branch and merge a current one with the `master`
     - run build over merged code
     - report build status
 4. If `Nim-matrix` workflow will fail we should press auto-merge button again or start from PR-p.3
 5. When `Nim-matrix` checks will pass, PR will be merged automatically

More information can be found in [How merge queues work](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue#how-merge-queues-work).

### Speed

Merge queue will add an additional matrix of jobs to build over different Nim versions. Because they will run in parallel, max merge delay will be equal to the job  max execution time.

| Runner                        | Duration, max | Costs, job    | Costs, matrix |
| ----------------------------- | ------------- | ------------- | ------------- |
| `ubuntu-latest`               | `19m 23s`     | `Free`        | `Free`        |
| `buildjet-4vcpu-ubuntu-2204`  | `15m 33s`     | `0.128 $/job` | `0.64 $/run`  |
| `buildjet-8vcpu-ubuntu-2204`  | `13m 14s`     | `0.224 $/job` | `1.12 $/run`  |
| `buildjet-16vcpu-ubuntu-2204` | `13m 00s`     | `0.416 $/job` | `2.08 $/run`  |

</details>

### DIFF
--- a/.github/workflows/nim-matrix.yml
+++ b/.github/workflows/nim-matrix.yml
@@ -1,76 +1,30 @@
 name: Nim matrix
+
 on:
   merge_group:
   workflow_dispatch:
+
 env:
   cache_nonce: 0 # Allows for easily busting actions/cache caches
+  nim_version: v1.6.14, v1.6.16, v1.6.18, v2.0.0, v2.0.2
+
 jobs:
-  build:
-    strategy:
-      matrix:
-        nim_version: [v1.6.14, v1.6.16, v1.6.18, v2.0.0, v2.0.2]
-        include:
-          - os: linux
-            cpu: amd64
-            builder: ubuntu-latest
-            # builder: buildjet-4vcpu-ubuntu-2204
-            shell: bash --noprofile --norc -e -o pipefail
-            tests: all
-
-    defaults:
-      run:
-        shell: ${{ matrix.shell }} {0}
-
-    name: '${{ matrix.nim_version }}-${{ matrix.os }}-${{ matrix.cpu }}-tests-${{ matrix.tests }}'
-    runs-on: ${{ matrix.builder }}
-    timeout-minutes: 80
-    steps:
-      - name: Checkout sources
-        uses: actions/checkout@v4
-        with:
-          submodules: recursive
-
-      - name: Setup Nimbus Build System
-        uses: ./.github/actions/nimbus-build-system
-        with:
-          os: ${{ matrix.os }}
-          shell: ${{ matrix.shell }}
-          nim_version: ${{ env.nim_version }}
-
-      ## Part 1 Tests ##
-      - name: Unit tests
-        if: matrix.tests == 'unittest' || matrix.tests == 'all'
-        run: make -j${ncpu} test
-
-      # workaround for https://github.com/NomicFoundation/hardhat/issues/3877
-      - name: Setup Node.js
-        uses: actions/setup-node@v4
-        with:
-          node-version: 18.15
-
-      - name: Start Ethereum node with Codex contracts
-        if: matrix.tests == 'contract' || matrix.tests == 'integration' || matrix.tests == 'all'
-        working-directory: vendor/codex-contracts-eth
-        env:
-          MSYS2_PATH_TYPE: inherit
-        run: |
-          npm install
-          npm start &
-
-      ## Part 2 Tests ##
-      - name: Contract tests
-        if: matrix.tests == 'contract' || matrix.tests == 'all'
-        run: make -j${ncpu} testContracts
-
-      ## Part 3 Tests ##
-      - name: Integration tests
-        if: matrix.tests == 'integration' || matrix.tests == 'all'
-        run: make -j${ncpu} testIntegration
-
-  build-status:
-    if: always()
-    needs: [build]
+  matrix:
     runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.matrix.outputs.matrix }}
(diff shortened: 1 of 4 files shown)


markspanbroek (commented):
Thanks @veaceslavdoina for figuring this out, looks like it wasn't trivial :smile:

Is there something we can do to avoid the duplication between `ci.yml` and `nim-matrix.yml`? They both contain very similar steps, and I'm afraid that we'll forget to update one of them in the future.

veaceslavdoina:
@markspanbroek, thanks you for the idea. Now we use [Reusable workflows](https://docs.github.com/en/actions/using-workflows/reusing-workflows) for `CI` and `Nim-matrix`.

We probably should enable `merge queue` only after confirmation that builds with Nim v2 passes. We can trigger `Nim-matrix` workflow manually until that.

markspanbroek (approved):
Thanks for removing the duplication! It all looks good to me now.

veaceslavdoina:
Branch rule was changed and PR automatically merged

Screenshot 2024-02-06 at 12 56 20

.github/workflows/nim-matrix.yml
```diff
@@ -0,0 +1,30 @@
+name: Nim matrix
+
+on:
+  merge_group:
+  workflow_dispatch:
+
+env:
+  cache_nonce: 0 # Allows for easily busting actions/cache caches
+  nim_version: v1.6.14, v1.6.16, v1.6.18, v2.0.0, v2.0.2
```
gmega, line 9: Guys, why do we have 2.x in the compiler matrix? Is Codex supposed to work with 2.x?

dryajov: yes, we want to start moving to 2.x

dryajov: oh... but this will fail for now, since we haven't completed the move @veaceslavdoina

veaceslavdoina: That was [the idea](#issuecomment-1928398091)?
> We probably should enable merge queue only after confirmation that builds with Nim v2 passes. We can trigger Nim-matrix workflow manually until that.

Merge queue is not enabled and `Nim-matrix` workflow is not used yet, may be just to test how it works. If we want to use it right now, we then should remove v2 from the list.

Should we?

dryajov: Ah I see, well, if it isn't being used for now, then no harm in keeping it, but as soon as we do, it will break with 2.x and so it would need to be disabled until we port Codex over.

PR #694, opened 2024-02-07
rework indexing strategy to use an enum

dryajov:
Reworks the indexer as an enum, to allow saving the strategy in the manifest.

Changes split out of #686 for easier tracking.

### DIFF
--- a/codex/indexingstrategy.nim
+++ b/codex/indexingstrategy.nim
@@ -1,5 +1,3 @@
-import std/sequtils
-
 import ./errors
 import ./utils
 import ./utils/asynciter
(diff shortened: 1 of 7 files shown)


codex/indexingstrategy.nim
```diff
@@ -1,12 +1,23 @@
...
 
 {.push raises: [].}
 
 type
+  StrategyType* = enum
+    # Simplest approach:
+    # 0 => 0, 1, 2
+    # 1 => 3, 4, 5
+    # 2 => 6, 7, 8
+    LinearStrategy,
```
tbekas, line 13: Maybe it would be better to name it `ConsequtiveStrategy` instead of `LinearStrategy`? Examples in `SteppedStrategy` are also linear.

dryajov: Not against it, but honestly I think both names work, it's just a preference thing ;)

tbekas (approved):

PR #696, opened 2024-02-07
cleanup manifest and wire in strategy

dryajov:
Add indexing strategy to manifest and propagate to erasure


gmega (approved):
I can't assess this in depth as I don't really understand most of it, but on the surface looks good. :-)

PR #697, opened 2024-02-07
Rework builder & sampler

dryajov:


codex/merkletree/merkletree.nim
```diff
@@ -117,11 +117,8 @@ func reconstructRoot*[H, K](proof: MerkleProof[H, K], leaf: H): ?!H =
 
   return success h
 
-func verify*[H, K](proof: MerkleProof[H, K], leaf: H, root: H): ?!void =
-  return if bool(root == ? proof.reconstructRoot(leaf)):
-      success()
-    else:
-      failure("invalid proof")
+func verify*[H, K](proof: MerkleProof[H, K], leaf: H, root: H): ?!bool =
+  success bool(root == ? proof.reconstructRoot(leaf))
```
gmega, line 121: Do we still need a result type here? Won't just `bool(root == ? proof.reconstructRoot(leaf))` do it at this point?

gmega (approved):
Looks fine on the surface... :-)

codex/merkletree/merkletree.nim
dryajov: hmm, good catch, not really

dryajov: actually, it kinda does, tho we might want to rework this later on... `reconstructRoot` uses `CompressFn*[H, K] = proc (x, y: H, key: K): ?!H {.noSideEffect, raises: [].}`, who's signature return a result, this isn't strictly necessary, but the signature is such in case we use hashing interface that return a result, such as the `multihash.digest`...

PR #698, opened 2024-02-08
wire in circom backend

dryajov:
Wires in circom compat ffi backend and adds circom prover support.

Split out from #686


.github/actions/nimbus-build-system/action.yml
```diff
@@ -45,25 +45,29 @@ runs:
...
       with:
+        path-type: inherit
         msystem: MINGW32
         install: >
           base-devel
           git
           mingw-w64-i686-toolchain
           mingw-w64-i686-cmake
           mingw-w64-i686-ntldd-git
+          mingw-w64-i686-rust
```
gmega, line 70: Well looking at the CI changes it doesn't look that bad :joy:

codex/slots/proofs/backends/circomcompat.nim
```diff
@@ -0,0 +1,237 @@
...
+import pkg/circomcompat
+import pkg/poseidon2/io
+
+import ../../types
+import ../../../stores
+import ../../../merkletree
+
+import pkg/constantine/math/arithmetic
+import pkg/constantine/math/arithmetic
+import pkg/constantine/math/io/io_bigints
```
gmega, line 25: Do you follow any conventions for grouping/ordering imports? Would be nice to know which so I can follow the same.

codex/slots/proofs/backends/circomcompat.nim
```diff
@@ -0,0 +1,237 @@
...
+
+export circomcompat
+
+const
+  # TODO: this defaults need to be adjusted and/or replased with cli config params
```
gmega, line 30: replased -> replaced (grammar police)

codex/slots/proofs/backends/circomcompat.nim
```diff
@@ -0,0 +1,237 @@
...
+  # to the correct length
+  doAssert input.samples.len == self.numSamples,
+    "Number of samples does not match"
+
+  doAssert input.slotProof.len <= self.datasetDepth,
+    "Number of slot proofs does not match"
+
+  doAssert input.samples.allIt(
+    block:
+      (it.merklePaths.len <= self.slotDepth + self.blkDepth and
```
gmega, line 95: I'm guessing blkDepth is the subpath that goes into the 64k block, correct?

codex/slots/proofs/backends/circomcompat.nim
```diff
@@ -0,0 +1,237 @@
...
+        return failure("Failed to push cell data")
+
+  var
+    proofPtr: ptr Proof = nil
+
+  let proof =
+    try:
+      if self.backendCfg.proveCircuit(backend, proofPtr.addr) != ERR_OK or
+        proofPtr == nil:
+        return failure("Failed to prove")
```
gmega, line 174: Yuck, we get zero error info back from the backend on failure? :-/

codex/slots/proofs/prover.nim
```diff
@@ -0,0 +1,98 @@
...
+import ../builder
+import ../sampler
+
+import ./backends
+import ../types
+
+export backends
+
+type
+  AnyProof* = CircomProof
```
gmega, line 34: Why prefixing with AnyXXX and not just naming them XXX?

tests/codex/slots/backends/testcircomcompat.nim
```diff
@@ -0,0 +1,137 @@
...
+
+    builder = Poseidon2Builder.new(store, verifiable).tryGet
+    sampler = Poseidon2Sampler.new(slotId, store, builder).tryGet
+
+    # circom = CircomCompat.init(r1cs, wasm, zkey)
+    circom = CircomCompat.init(r1cs, wasm)
+    verifyingKeyPtr = circom.getVerifyingKey().tryGet
+    challenge = 1234567.toF.toBytes.toArray32
+
+    proofInput = (await sampler.getProofInput(challenge, samples)).tryGet
```
gmega, line 117: I'm not sure why we have this suite in addition to the above, but I'm assuming this is testing generation of proof inputs. If that's the case, I'd probably would have preferred to have those tests on their own (i.e. generating proof inputs and assertions on generated proof inputs) as smaller granularity tests make tracking down bugs simpler.

gmega (approved):
As far as I can understand it, looks good to me!

codex/slots/proofs/backends/circomcompat.nim
dryajov: ah, good catch I just saw it as well... 

no strict conventions and constantine imports are quite skewed

codex/slots/proofs/backends/circomcompat.nim
dryajov: exactly

codex/slots/proofs/backends/circomcompat.nim
dryajov: we get an error code, perhaps we should log it

codex/slots/proofs/prover.nim
dryajov: because these are generic types that will represent all the proofs that we can handle, something like:

```nim
AnyProof* = CircomProof or KZGProof or XYZProof
```

tests/codex/slots/backends/testcircomcompat.nim
dryajov: This is testing again control inputs that we test the circuit itself, this ensures that the circuit is instantiated properly in case we start seeing issues with the sampled chunks. This can probably go away at some point, but it makes me feel a little better having this test in there for now.

dryajov: btw, we do have more granular tests in `sampler/testsampler.nim`, which test the sampling itself rather than the circuit invocation.

codex/slots/proofs/backends/circomcompat.nim
```diff
@@ -0,0 +1,228 @@
...
+    self.backendCfg,
+    addr backend) != ERR_OK or backend == nil:
+    raiseAssert("failed to initialize CircomCompat backend")
+
+  var
+    entropy = input.entropy.toBytes
+    dataSetRoot = input.datasetRoot.toBytes
+    slotRoot = input.slotRoot.toBytes
+
+  if backend.pushInputU256Array(
```
elcritch, line 101: That looks nice, just binding the Circom builder more directly makes sense.

codex/slots/proofs/backends/circomcompat.nim
```diff
@@ -0,0 +1,228 @@
...
+      data = s.cellData
+
+    merklePaths.setLen(self.slotDepth) # zero pad inputs to correct size
+    if backend.pushInputU256Array(
+      "merklePaths".cstring,
+      merklePaths[0].addr,
+      uint (merklePaths[0].len * merklePaths.len)) != ERR_OK:
+        return failure("Failed to push merkle paths")
+
+    data.setLen(self.cellElms * 32) # zero pad inputs to correct size
```
elcritch, line 149: Wouldn't this be done elsewhere prior to running the prover? Maybe it could be an assert.

codex/slots/proofs/backends/circomcompat.nim
```diff
@@ -0,0 +1,228 @@
...
+        proofPtr == nil:
+        return failure("Failed to prove - err code: " & $res)
+
+      proofPtr[]
+    finally:
+      if proofPtr != nil:
+        proofPtr.addr.releaseProof()
+
+      if backend != nil:
+        backend.addr.releaseCircomCompat()
```
elcritch, line 173: Backend is allocated initialized up above but there's a few early returns on failures as well. Should this be done in a `defer` block by the init section?

elcritch: Otherwise seems like this could leak on proof failures.

dryajov: Yep, you're right. This still needs some more hands on debugging, but this is an obvious one.

codex/slots/proofs/backends/circomcompat.nim
dryajov: It's an assert up top, but this is set in the loop, to avoid (more) unnecessary reallocs.

codex/slots/proofs/backends/circomcompat.nim
dryajov: fix here https://github.com/codex-storage/nim-codex/pull/701

PR #700, opened 2024-02-10
rework cli to accept circuit params

dryajov:
Rework cli to use commands and accept the required circuit params

Work split out from #686


dryajov:
closing in favor of #702

PR #702, opened 2024-02-10
Wire prover into node

dryajov:
This wires the prover into the codex node - depends on #700 for more comprehensive cli rework

Work split out from #686


codex/slots/proofs/backends/circomcompat.nim
```diff
@@ -175,26 +158,27 @@ proc prove*[H](
...
+  try:
+    let res = verifyCircuit(proofPtr, inputs.addr, self.vkp)
+    if res == ERR_OK:
+      success true
+    elif res == ERR_FAILED_TO_VERIFY_PROOF:
+      success false
+    else:
+      failure("Failed to verify proof - err code: " & $res)
+  finally:
+    inputs.releaseCircomInputs()
```
elcritch, line 181: Does the proofPtr need releasing?

dryajov: no, thats a GC object

tests/integration/testproofs.nim
```diff
@@ -62,12 +62,17 @@ twonodessuite "Proving integration test", debug1=false, debug2=false:
...
+        "persistence",
+        "validator",
         "--data-dir=" & validatorDir,
         "--api-port=8089",
         "--disc-port=8099",
         "--listen-addrs=/ip4/127.0.0.1/tcp/0",
-        "--validator",
-        "--eth-account=" & $accounts[2]
+        "--eth-account=" & $accounts[2],
+        # TODO: this is wrong, validator shouldn't need circom stuff
```
dryajov, line 72: we can move validator to be it's own command it this isn't needed for it

dryajov: can you clarify what parameters are needed for the validator?

markspanbroek: The validator only needs to interact with the smart contract, so these parameters would apply:
--eth-provider
--eth-account
--eth-private-key
--marketplace-address

dryajov: OK, these are shared parameters between the prover and verifier, I can add a flag for `circuits`, which would then make it mandatory to either select either a `persistence & prover` or `persistence & validator`.

markspanbroek: > I can add a flag for `circuits`, which would then make it mandatory to either select either a `persistence & prover` or `persistence & validator`

I don't understand this? Keep in mind that the validator does not check zk proofs, that is the job of the smart contract. The validator only looks for missed proofs, and points them out to the smart contract. Only a host (prover) needs the circuit parameters, because it needs to create proofs.

dryajov:
```sh
Available sub-commands:

codex persistence [OPTIONS]... command

The following options are available:

     --eth-provider         The URL of the JSON-RPC API of the Ethereum node [=ws://localhost:8545].
     --eth-account          The Ethereum account that is used for storage contracts.
     --eth-private-key      File containing Ethereum private key for storage contracts.
     --marketplace-address  Address of deployed Marketplace contract.

Available sub-commands:

codex persistence prover [OPTIONS]...

The following options are available:

     --circom-r1cs          The r1cs file for the storage circuit [=/Users/dryajov/Library/Application
                            Support/Codex/circuits/proof_main.r1cs].
     --circom-wasm          The wasm file for the storage circuit [=/Users/dryajov/Library/Application
                            Support/Codex/circuits/proof_main.wasm].
     --circom-zkey          The zkey file for the storage circuit [=/Users/dryajov/Library/Application
                            Support/Codex/circuits/proof_main.zkey].
     --circom-no-zkey       Ignore the zkey file - use only for testing! [=false].
     --proof-samples        Number of samples to prove [=5].
     --max-slot-depth       The maximum depth of the slot tree [=32].
     --max-dataset-depth    The maximum depth of the dataset tree [=8].
     --max-block-depth      The maximum depth of the network block merkle tree [=5].
     --max-cell-elements    The maximum number of elements in a cell [=67].

codex persistence validator [OPTIONS]...

The following options are available:

     --validator-max-slots  Maximum number of slots that the validator monitors [=1000].
```

markspanbroek: Right, that makes sense.
The we'd also have to check that the `persistence prover` options are set when starting the sales module.

dryajov: However, this will only allow you to be a prover or a validator, not both. Is there a case where you can be both? In which case, we can just leave `--validator` as cmd argument rather than a command.

emizzle: > However, this will only allow you to be a prover or a validator, not both. Is there a case where you can be both? In which case, we can just leave `--validator` as cmd argument rather than a command.

I don't think there is anything limiting a node from being both a validator and a host (prover). A host should be able to mark proofs as missing (including its own).

dryajov: yep, thats what we agreed on, we should have `--validator` as a cmd param.

tests/integration/multinodes.nim
```diff
@@ -103,9 +103,9 @@ template multinodesuite*(name: string,
         "prover",
         "--eth-account=" & $accounts[index],
         # TODO: this is wrong, client shouldn't need circom stuff
-        "--circom-r1cs=tests/circuits/fixtures/proof_main.r1cs",
-        "--circom-wasm=tests/circuits/fixtures/proof_main.wasm",
-        "--circom-no-zkey"]
+        "--circom-r1cs=vendor/codex-contracts-eth/verifier/networks/hardhat/proof_main.r1cs",
```
dryajov, line 106: I think it's probably better to store codex specific assents in codex itself, preferably in `tests/circuits/fixtures`? It's probably fine to duplicate them for the contracts, but for codex tests, I'd rather we keep them in the same codebase.

markspanbroek: The problem that I tried to solve is that the circuit assets need to match between the codex-contracts-eth and nim-codex. And since nim-codex already has the contracts in its vendor folder, it seemed easiest to put the asset with the contracts.

You could also copy them into the codex-contracts-eth from nim-codex during testing, but that seemed less elegant?

dryajov: Well, the point is that in the context of codex, and specifically the integration tests, we're trying to make sure that all the subsystems - storage/networking, block selection and sampling as well as the contracts and marketplace, function correctly. 

There are a set of assets that are used in codex proper, regardless of what tests the individual components rely on and this should be under codex, because as I already said, they are shared by several subsystems. In this case, the integration tests aren't testing _just_ the contracts, they rely on all this subsystems functioning for the contract to execute correctly and we better test all of this against the same set of assets which, IMO, should be in the codex repo. Then each project, external to codex, can contain it's own set of assets as well, but this are unrelated to those under codex.

markspanbroek: That's fine too :+1:  I have no strong opinion on this.

benbierens:
I see the CLI changes are being split into another PR. That's great. This one still has a lot of unrelated code-style changes. I wouldn't ask to omit those. But maybe in the future, we should have dedicated code-style PRs so they don't increase the size of others.

elcritch:
> I see the CLI changes are being split into another PR. That's great. This one still has a lot of unrelated code-style changes. I wouldn't ask to omit those. But maybe in the future, we should have dedicated code-style PRs so they don't increase the size of others.

+1 to this idea, it makes it harder to review the PR and catch important changes. Also, it'd be nice to start using `nph` to format as well.

tests/integration/multinodes.nim
```diff
@@ -204,9 +204,12 @@ template multinodesuite*(name: string, body: untyped) =
       var config = conf
       config.addCliOption("--bootstrap-node", bootstrap)
       config.addCliOption(StartUpCmd.persistence, "--eth-account", $accounts[running.len])
-      config.addCliOption(PersistenceCmd.prover, "--circom-r1cs", "tests/circuits/fixtures/proof_main.r1cs")
-      config.addCliOption(PersistenceCmd.prover, "--circom-wasm", "tests/circuits/fixtures/proof_main.wasm")
-      config.addCliOption(PersistenceCmd.prover, "--circom-zkey", "tests/circuits/fixtures/proof_main.zkey")
+      config.addCliOption(PersistenceCmd.prover, "--circom-r1cs",
```
dryajov, line 207: lets move this into `tests/circuit/fixtures` please.

codex/stores/networkstore.nim
```diff
@@ -37,13 +36,16 @@ type
...
+    if not (err of BlockNotFoundError):
+      trace "Error getting block from local store", address, err = err.msg
+      return failure err
 
-    without newBlock =? (await self.engine.requestBlock(address)).catch, error:
-      trace "Unable to get block from exchange engine", address
-      return failure error
+    trace "Block not in local store", address, err = err.msg
+
+    without newBlock =? (await self.engine.requestBlock(address)), err:
```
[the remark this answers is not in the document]
dryajov: This can be extracted into its own pr, as many other things.

markspanbroek:
I split this PR into several smaller pieces:
- #731 
- #732 
- #733 
- #734 
- #735 
- #736 
- #737

markspanbroek:
Closing, because the PRs that were split out from this are now merged.

PR #703, opened 2024-02-10
Install Rust during Docker builds

veaceslavdoina:
PR install Rust for Docker builds.

Also, `NIMBLE_DIR` was removed from `PATH` via `"${HOME}/.bash_env"` - not sure how this is intended to work.

Was [tested manually](https://github.com/codex-storage/nim-codex/actions/runs/7854380843).

```shell
docker run codexstorage/nim-codex:sha-a0761bb-dist-tests codex --version

Codex revision: a0761bb
Nim Compiler Version 1.6.14 [Linux: arm64]
```

### DIFF
--- a/.github/workflows/ci-reusable.yml
+++ b/.github/workflows/ci-reusable.yml
@@ -27,7 +27,6 @@ jobs:
     name: '${{ matrix.os }}-${{ matrix.cpu }}-${{ matrix.nim_version }}-${{ matrix.tests }}'
     runs-on: ${{ matrix.builder }}
     timeout-minutes: 80
-    continue-on-error: true
     steps:
       - name: Checkout sources
         uses: actions/checkout@v4
--- a/docker/codex.Dockerfile
+++ b/docker/codex.Dockerfile
@@ -14,8 +14,8 @@ ARG MAKE_PARALLEL
 ARG NIMFLAGS
 
 RUN apt-get update && apt-get install -y git cmake curl make bash lcov build-essential nim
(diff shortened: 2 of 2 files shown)


veaceslavdoina:
One of the possible reference to the removed `.bash_env` - [Access $HOME in Dockerfile ENV instruction](https://stackoverflow.com/questions/70787967/access-home-in-dockerfile-env-instruction).

Anyway it didn't work as it probably was intended. In case of need we can implement it in a right way.

dryajov:
Why is nimble dir being removed? Is that necessary?

veaceslavdoina:
> Why is nimble dir being removed? Is that necessary?

Most probably it just make no sense, or I didn't found it
```dockerfile
RUN echo 'export NIMBLE_DIR="${HOME}/.nimble"' >> "${HOME}/.bash_env"
RUN echo 'export PATH="${NIMBLE_DIR}/bin:${PATH}"' >> "${HOME}/.bash_env"
```

I didn't find any part of the code which may use `.bash_env` and also checked https://github.com/status-im/nimbus-build-system/tree/fe9bc3f3759ae1add6bf8c899db2e75327f03782. No one is using it.

However, `$BASH_ENV` is described in the [6.2 Bash Startup Files](https://www.gnu.org/software/bash/manual/html_node/Bash-Startup-Files.html), and can be used to pass variables between layers, but
- `$BASH_ENV != .bash_env`
- By default `$BASH_ENV` is undefined
- Docker builds works well without these 2 lines

dryajov (approved):
Lgtm

PR #706, opened 2024-02-15
Fix Docker entrypoint NAT helper variables

veaceslavdoina:
This is a quick fix for existing Docker entrypoint

Since we've introduced [`NAT_PUBLIC_IP_AUTO`](https://github.com/codex-storage/nim-codex/commit/750fe2392feb57e6043b824e8e939a99d866e30a) helper variable in Docker entrypoint, we broke image run with default values
```shell
docker run codexstorage/nim-codex:latest

# Can't get Public IP in  seconds - Stop Codex run

# And we need to workaround that
docker run -e NAT_IP_AUTO=true codexstorage/nim-codex:latest
docker run -e NAT_PUBLIC_IP_AUTO=ip.codex.storage codexstorage/nim-codex:latest
```
In this PR, we use slightly different logic
1. When `CODEX_NAT` is passed - use it
2. `NAT_PUBLIC_IP_AUTO` (Public IP) have priority over `NAT_IP_AUTO`(Private IP)
3. Show separate error for `NAT_PUBLIC_IP_AUTO` and `NAT_IP_AUTO` when can't get IP address

### DIFF
--- a/docker/docker-entrypoint.sh
+++ b/docker/docker-entrypoint.sh
@@ -1,27 +1,38 @@
 #!/bin/bash
 
 # Parameters
-if [[ "${NAT_IP_AUTO}" == "true" && -z "${NAT_PUBLIC_IP_AUTO}" ]]; then
-  export CODEX_NAT=$(hostname --ip-address)
-  echo "Internal: Set CODEX_NAT: ${CODEX_NAT}"
+if [[ -z "${CODEX_NAT}" ]]; then
+  if [[ "${NAT_IP_AUTO}" == "true" && -z "${NAT_PUBLIC_IP_AUTO}" ]]; then
+    export CODEX_NAT=$(hostname --ip-address)
+    echo "Private: CODEX_NAT=${CODEX_NAT}"
+  elif [[ -n "${NAT_PUBLIC_IP_AUTO}" ]]; then
+    # Run for 60 seconds if fail
+    WAIT=120
+    SECONDS=0
+    SLEEP=5
+    while (( SECONDS < WAIT )); do
+      export CODEX_NAT=$(curl -s -f -m 5 "${NAT_PUBLIC_IP_AUTO}")
+      # Check if exit code is 0 and returned value is not empty
+      if [[ $? -eq 0 && -n "${CODEX_NAT}" ]]; then
+        echo "Public: CODEX_NAT=${CODEX_NAT}"
+        break
+      else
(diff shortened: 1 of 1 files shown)


docker/docker-entrypoint.sh
```diff
@@ -1,28 +1,39 @@
 #!/bin/bash
 
 # Parameters
-if [[ "${NAT_IP_AUTO}" == "true" && -z "${NAT_PUBLIC_IP_AUTO}" ]]; then
-  export CODEX_NAT=$(hostname --ip-address)
-  echo "Internal: Set CODEX_NAT: ${CODEX_NAT}"
+if [[ -z "${CODEX_NAT}" ]]; then
+  if [[ "${NAT_IP_AUTO}" == "true" && -z "${NAT_PUBLIC_IP_AUTO}" ]]; then
```
gmega, line 5: OK so the semantics here is that if you set NAT_PUBLIC_IP_AUTO that will override NAT_IP_AUTO, correct? I like it.

gmega (approved):
Nice! LGTM.

docker/docker-entrypoint.sh
veaceslavdoina: Yes, and this is done for dist-tests images, where we set [`nat_ip_auto: true`](https://github.com/codex-storage/nim-codex/blob/master/.github/workflows/docker-dist-tests.yml#L28) by default and we need a way to set Public IP if needed.

docker/docker-entrypoint.sh
```diff
@@ -1,28 +1,39 @@
...
+    export CODEX_NAT=$(hostname --ip-address)
+    echo "Private: CODEX_NAT=${CODEX_NAT}"
+  elif [[ -n "${NAT_PUBLIC_IP_AUTO}" ]]; then
+    # Run for 60 seconds if fail
+    WAIT=120
+    SECONDS=0
+    SLEEP=5
```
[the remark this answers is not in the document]
veaceslavdoina: Good question, we may changes that and just to consider some points
- for curl we have `-m 5` - 5 seconds timeout for IP Lookup service
- `sleep 5` - 5 seconds before a new lookup try

If IP Lookup service works fine and we get Public IP - start time will not be affected (~ 1 sec), but in case of the issue, it make sense to wait
- temporary network issue - short try make sense
- service overload - short try make no much sense
- relate limit - short try make no much sense

PR #708, opened 2024-02-16
Block and manifest as objects

dryajov:
split out from #702

### DIFF
--- a/codex/blocktype.nim
+++ b/codex/blocktype.nim
@@ -31,7 +31,7 @@ import ./codextypes
 export errors, logutils, units, codextypes
 
 type
-  Block* = ref object of RootObj
+  Block* = object of RootObj
     cid*: Cid
     data*: seq[byte]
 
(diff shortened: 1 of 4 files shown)


tests/codex/slots/testslotbuilder.nim
```diff
@@ -331,8 +331,8 @@ suite "Slot builder":
         protectedManifest,
         cellSize = cellSize).tryGet()
 
+    var
```
emizzle, line 334: needed?

codex/streams/storestream.nim
```diff
@@ -110,7 +110,7 @@ method readOnce*(
     if blk.isEmpty:
       zeroMem(pbytes.offset(read), readBytes)
     else:
-      copyMem(pbytes.offset(read), blk.data[blockOffset].addr, readBytes)
+      copyMem(pbytes.offset(read), blk.data[blockOffset].unsafeAddr, readBytes)
```
emizzle, line 113: Was this intentional? The [documentation](https://nim-lang.org/docs/system.html#unsafeAddr%2CT) says:
> Warning: unsafeAddr is a deprecated alias for addr, use addr instead.

emizzle (approved):
LGTM, but might possibly be a couple of small unneeded changes

codex/streams/storestream.nim
dryajov: That's for versions 2.0+

codex/blocktype.nim
```diff
@@ -31,7 +31,7 @@ import ./codextypes
 export errors, logutils, units, codextypes
 
 type
-  Block* = ref object of RootObj
+  Block* = object of RootObj
```
elcritch, line 34: I am concerned about this change. Changing Block to an object will incur a lot more copies and memory presusre since `seq[byte]` will be copied anytime the Block is copied.

dryajov: Yeah, as discussed in DMs, this is for safety and can be reverted once we move on to optimize things - but before that, we should actually measure how big the penalty is. 

The specific change was made to avoid some weirdness with ref/options/questionable.

PR #710, opened 2024-02-17
Node use correct stores

dryajov:
split out from #702

### DIFF
--- a/codex/node.nim
+++ b/codex/node.nim
@@ -231,7 +231,7 @@ proc retrieve*(
               self.networkStore,
               leoEncoderProvider,
               leoDecoderProvider)
-          without res =? (await erasure.decode(manifest)), error:
+          without _ =? (await erasure.decode(manifest)), error:
             trace "Unable to erasure decode manifest", cid, exc = error.msg
         except CatchableError as exc:
           trace "Exception decoding manifest", cid, exc = exc.msg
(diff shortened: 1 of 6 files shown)


codex/node.nim
```diff
@@ -226,7 +226,12 @@ proc retrieve*(
       proc erasureJob(): Future[void] {.async.} =
         try:
           # Spawn an erasure decoding job
-          without res =? (await self.erasure.decode(manifest)), error:
+          let
+            erasure = Erasure.new(
+              self.networkStore,
+              leoEncoderProvider,
+              leoDecoderProvider)
+          without res =? (await erasure.decode(manifest)), error:
```
emizzle, line 234: It seems like `res` is bound but never used, so what happens to the decoded manifest?

codex/node.nim
```diff
@@ -571,25 +584,28 @@ proc onProve(
 
   trace "Received proof challenge"
 
-  without cid =? Cid.init(cidStr).mapFailure, err:
-    error "Unable to parse Cid", cid, err = err.msg
-    return failure(err)
+  if prover =? self.prover:
+    trace "Prover enabled"
```
emizzle, line 588: Maybe easier to read by returning a failure if `self.prover.isNone`:
```suggestion
  without prover =? self.prover:
    return failure "no prover"
```

emizzle (commented):
LGTM, don't want to approve as it will auto-merge. Will give others a chance to comment.

codex/node.nim
```diff
@@ -61,9 +62,9 @@ type
   CodexNode* = object
     switch: Switch
     networkId: PeerId
-    blockStore: BlockStore
+    networkStore: NetworkStore
```
AuHau, line 65: So this means we are abandoning the `BlockStore` abstraction? Or moving `NetworkStore` from it?

dryajov: Network store inherits `BlockStore`, so no.

codex/node.nim
dryajov: We already have a decoded manifest, this is reconstruction happening on download, but `res` should be `_`...

codex/node.nim
AuHau: Well yeah - technically, but here you are moving away from supporting the `BlockStore` abstraction inside Codex in favor of `NetworkStore` so any other store won't be compatible here, which in my view is "abandoning `BlockStore` abstraction"...

I mean we have already bumped into this limitation many times, so it make sense to separate `NetworkStore` out of this, but I know you were not so in favor of that, so that is why I am wondering about it.

dryajov: Thats not how interfaces and inheritance work :), we're still using the same abstraction, only a more specialized version of it. So no, we aren't moving aways from the `BlockStore` abstraction as you suggested in your original comment.

codex/node.nim
dryajov: Yes, but I wont change it for now, because it will break the integration tests

PR #711, opened 2024-02-17
Prover works with zkey

dryajov:
split out from #702


codex/slots/proofs/backends/circomcompat.nim
```diff
@@ -32,40 +35,24 @@ type
...
-  CircomG2* = G2
-
-  CircomProof*  = Proof
-  CircomKey*    = VerifyingKey
-  CircomInputs* = Inputs
+    vkp*          : ptr CircomKey
 
 proc release*(self: CircomCompat) =
-  ## Release the backend
+  ## Release the ctx
```
gmega, line 43: Risking painting the bike shed: I understand a lot better what "backend" is than "ctx". Ctx reads like "context", but not sure why you'd want to call this a context. Plus there's still uses of the word "backend" here and there and it's not clear if they were just forgotten or if we're now making a distinction between ctx and backend.

codex/slots/proofs/prover.nim
```diff
@@ -30,25 +30,27 @@ import ../types
...
+  topics = "codex prover"
+
 type
-  AnyProof* = CircomProof
-  AnyInputs* = CircomInputs
-  AnyKeys* = CircomKey
-  AnyHash* = Poseidon2Hash
   AnyBackend* = CircomCompat
-  AnyBuilder* = Poseidon2Builder
+  AnyProof* = CircomProof
```
gmega, line 38: Still unclear to me why we call those AnyXXX but I guess I can ask you in person when we chat. :-)

tests/codex/slots/backends/testcircomcompat.nim
```diff
@@ -16,77 +17,65 @@ import pkg/codex/codextypes
...
-import pkg/constantine/math/io/io_bigints
-
 import ./helpers
 import ../helpers
 
 suite "Test Circom Compat Backend - control inputs":
   let
     r1cs = "tests/circuits/fixtures/proof_main.r1cs"
     wasm = "tests/circuits/fixtures/proof_main.wasm"
+    zkey = "tests/circuits/fixtures/proof_main.zkey"
```
gmega, line 27: One thing that bothers me a bit is that this test looks a lot like how it was before, except we now pass a zkey as an input. Meaning you could've built a test that looks exactly like this one (and passes) by just adding a dummy zkey argument in the previous test. Which to me smells of weak assertions - who's to say this test would fail if the incorrect zkey were supplied?

gmega (approved):
I have mostly minor comments/questions, it otherwise looks good to merge to me as far as I can understand it.

tests/codex/slots/backends/testcircomcompat.nim
dryajov: As the PR states, this is testing with an actual zkey file, which it wasn't doing previously. Not sure what is bothering you here?

codex/slots/proofs/backends/circomcompat.nim
dryajov: The way I think about it is - `this is the context for the backend` and from that standpoint it makes sense.

tests/codex/slots/backends/testcircomcompat.nim
gmega: I think the point is - if you supply and invalid zkey file (or no zkey file) and this passes cause, say, the implementation has a bug, you won't be catch it from this test. The change you make on the invalid input is too coarse, and it may well continue to fail simply because r1cs or wasm caught it.

dryajov: There are two ways to run the prover, with and without a zkey file. Both are valid.

If the zkey file is invalid, then the proof should not verify, running without a zkey file should only be possible in tests, but otherwise it's a valid scenario. This tests aren't intended to detect wether the file is being passed under the correct conditions, only that the backend is actually working with the provided inputs.

We might need additional tests to make sure that the file is being passed at the right times, but this would be the wrong place to make those checks/tests.

codex/slots/proofs/backends/circomcompat.nim
gmega: OK fine, it's sort of a context object which you init, put stuff into, hand over to the backend to get the proofs, then release. In that case I think the weird part is that you're interacting with an object that's called `cfg` to get things done (`backendCfg`), when I guess the expectation for a config object is that it contains read-only data.

dryajov: but, nothing is being written to the `cfg`, it does in fact, only contain data.

codex/slots/proofs/prover.nim
elcritch: Similar pattern is used in Nim with ints & floats with `SomeInteger = int | uint | int8 | ...` and `SomeFloat`.

elcritch: Guess it's like an "inverted" way of doing polymorphism, but at the type level instead of using inheritance.

PR #713, opened 2024-02-19
rework cli to use commands for `persistence` mode

dryajov:
Note that multinode tests are commented out because of lack of proper command options support, which @emizzle is addressing.

Split out from #702

### DIFF
--- a/codex.nim
+++ b/codex.nim
@@ -23,6 +23,7 @@ import ./codex/codex
 import ./codex/logutils
 import ./codex/units
 import ./codex/utils/keyutils
+import ./codex/codextypes
 
 export codex, conf, libp2p, chronos, logutils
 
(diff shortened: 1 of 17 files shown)


codex.nim
```diff
@@ -54,91 +55,93 @@ when isMainModule:
...
+    quit QuitFailure
 
-    if not(checkAndCreateDataDir((config.dataDir).string)):
-      # We are unable to access/create data folder or data folder's
-      # permissions are insecure.
-      quit QuitFailure
+  trace "Data dir initialized", dir = $config.dataDir
+
+  if not(checkAndCreateDataDir((config.dataDir / "repo"))):
+    # We are unable to access/create data folder or data folder's
```
benbierens, line 73: Should these comments and those on lines 66, 67 be warn or error logs?

tests/integration/testcli.nim
```diff
@@ -9,32 +9,14 @@ suite "Command line interface":
   let key = "4242424242424242424242424242424242424242424242424242424242424242"
 
   test "complains when persistence is enabled without ethereum account":
-    let node = startNode(@["--persistence"])
+    let node = startNode(@["persistence"])
     node.waitUntilOutput("Persistence enabled, but no Ethereum account was set")
     node.stop()
 
-  test "complains when validator is enabled without ethereum account":
```
benbierens, line 16: Are we sure we don't want to give these warnings to the user? They seem like config mistakes that are easily made.

codex.nim
dryajov: Ah yeah, definitely want to print an error there and in the other place you pointed out.

PR #714, opened 2024-02-20
Fixes active cancellation for pending want requests

gmega:
This PR fixes the active cancellation of pending want requests by adding the actual want list sends with the `cancel` flag set to true as part of block delivery, with accompanying tests.

### DIFF
--- a/codex/blockexchange/engine/engine.nim
+++ b/codex/blockexchange/engine/engine.nim
@@ -274,9 +274,8 @@ proc issueCancellations(b: BlockExcEngine, addrs: seq[BlockAddress]) {.async.} =
   trace "Sending block request cancellations to peers", addrs = addrs.len
 
   let sends = b.peers.mapIt(
-    (peer: it, request: b.network.request.sendWantList(id = it.id,
-      addresses = addrs, cancel = true))
-  )
+    (peer: it, request: b.network.request.sendWantCancellations(peer = it.id,
+      addresses = addrs)))
 
   discard await allFinished(sends.mapIt(it.request))
 
--- a/codex/blockexchange/engine/engine.nim
+++ b/codex/blockexchange/engine/engine.nim
@@ -120,7 +120,6 @@ proc stop*(b: BlockExcEngine) {.async.} =
 
   trace "NetworkStore stopped"
 
-
 proc sendWantHave(
   b: BlockExcEngine,
   address: BlockAddress,
@@ -145,16 +144,6 @@ proc sendWantBlock(
     blockPeer.id,
     @[address],
     wantType = WantType.WantBlock) # we want this remote to send us a block
-
-proc findCheapestPeerForBlock(b: BlockExcEngine, cheapestPeers: seq[BlockExcPeerCtx]): ?BlockExcPeerCtx =
-  if cheapestPeers.len <= 0:
-    trace "No cheapest peers, selecting first in list"
-    let
-      peers = toSeq(b.peers) # Get any peer
-    if peers.len <= 0:
-      return none(BlockExcPeerCtx)
-    return some(peers[0])
-  return some(cheapestPeers[0]) # get cheapest
 
 proc monitorBlockHandle(b: BlockExcEngine, handle: Future[Block], address: BlockAddress, peerId: PeerId) {.async.} =
   try:
@@ -231,7 +220,7 @@ proc blockPresenceHandler*(
         have      = presence.have
         price     = presence.price
 
-      trace "Updating precense"
+      trace "Updating presence"
       peerCtx.setPresence(presence)
 
   let
@@ -280,6 +269,17 @@ proc scheduleTasks(b: BlockExcEngine, blocksDelivery: seq[BlockDelivery]) {.asyn
 
           break # do next peer
 
+proc issueCancellations(b: BlockExcEngine, addrs: seq[BlockAddress]) {.async.} =
+  ## Tells neighboring peers that we're no longer interested in a block.
+  trace "Sending block request cancellations to peers", addrs = addrs.len
+
+  for neighbor in b.peers:
+    await b.network.request.sendWantList(
+      id = neighbor.id,
+      addresses = addrs,
+      cancel = true
+    )
+
 proc resolveBlocks*(b: BlockExcEngine, blocksDelivery: seq[BlockDelivery]) {.async.} =
   trace "Resolving blocks", blocks = blocksDelivery.len
 
@@ -290,6 +290,8 @@ proc resolveBlocks*(b: BlockExcEngine, blocksDelivery: seq[BlockDelivery]) {.asy
     cids.incl(bd.blk.cid)
     if bd.address.leaf:
       cids.incl(bd.address.treeCid)
+
+  asyncSpawn b.issueCancellations(blocksDelivery.mapIt(it.address))
   b.discovery.queueProvideBlocksReq(cids.toSeq)
 
 proc resolveBlocks*(b: BlockExcEngine, blocks: seq[Block]) {.async.} =
@@ -398,7 +400,7 @@ proc wantListHandler*(
 
   for e in wantList.entries:
     let
-      idx = peerCtx.peerWants.find(e)
+      idx = peerCtx.peerWants.findIt(it.address == e.address)
 
     logScope:
       peer      = peerCtx.id
--- a/codex/blockexchange/engine/engine.nim
+++ b/codex/blockexchange/engine/engine.nim
@@ -273,15 +273,15 @@ proc cancelBlocks(b: BlockExcEngine, addrs: seq[BlockAddress]) {.async.} =
   ## Tells neighboring peers that we're no longer interested in a block.
   trace "Sending block request cancellations to peers", addrs = addrs.len
 
-  let sends = b.peers.mapIt(
-    (peer: it, request: b.network.request.sendWantCancellations(peer = it.id,
-      addresses = addrs)))
+  let failed = (await allFinished(
+    b.peers.mapIt(
+      b.network.request.sendWantCancellations(
+        peer = it.id,
+        addresses = addrs))))
+    .filterIt(it.failed)
 
-  await allFutures(sends.mapIt(it.request))
-
-  let failed = sends.filterIt(it.request.failed).mapIt(it.peer.id)
   if failed.len > 0:
-    trace "Failed to send block request cancellations to peers", failed = $failed
+    trace "Failed to send block request cancellations to peers", peers = failed.len
 
 proc resolveBlocks*(b: BlockExcEngine, blocksDelivery: seq[BlockDelivery]) {.async.} =
   trace "Resolving blocks", blocks = blocksDelivery.len
--- a/codex/blockexchange/engine/engine.nim
+++ b/codex/blockexchange/engine/engine.nim
@@ -269,7 +269,7 @@ proc scheduleTasks(b: BlockExcEngine, blocksDelivery: seq[BlockDelivery]) {.asyn
 
           break # do next peer
 
-proc issueCancellations(b: BlockExcEngine, addrs: seq[BlockAddress]) {.async.} =
+proc cancelBlocks(b: BlockExcEngine, addrs: seq[BlockAddress]) {.async.} =
   ## Tells neighboring peers that we're no longer interested in a block.
   trace "Sending block request cancellations to peers", addrs = addrs.len
 
@@ -277,7 +277,7 @@ proc issueCancellations(b: BlockExcEngine, addrs: seq[BlockAddress]) {.async.} =
     (peer: it, request: b.network.request.sendWantCancellations(peer = it.id,
       addresses = addrs)))
 
-  discard await allFinished(sends.mapIt(it.request))
+  await allFutures(sends.mapIt(it.request))
 
   let failed = sends.filterIt(it.request.failed).mapIt(it.peer.id)
   if failed.len > 0:
@@ -294,7 +294,7 @@ proc resolveBlocks*(b: BlockExcEngine, blocksDelivery: seq[BlockDelivery]) {.asy
     if bd.address.leaf:
       cids.incl(bd.address.treeCid)
 
-  await b.issueCancellations(blocksDelivery.mapIt(it.address))
+  await b.cancelBlocks(blocksDelivery.mapIt(it.address))
   b.discovery.queueProvideBlocksReq(cids.toSeq)
 
 proc resolveBlocks*(b: BlockExcEngine, blocks: seq[Block]) {.async.} =
--- a/codex/blockexchange/engine/engine.nim
+++ b/codex/blockexchange/engine/engine.nim
@@ -273,12 +273,16 @@ proc issueCancellations(b: BlockExcEngine, addrs: seq[BlockAddress]) {.async.} =
   ## Tells neighboring peers that we're no longer interested in a block.
   trace "Sending block request cancellations to peers", addrs = addrs.len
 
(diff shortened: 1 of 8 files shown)


codex/blockexchange/engine/engine.nim
```diff
@@ -290,6 +290,8 @@ proc resolveBlocks*(b: BlockExcEngine, blocksDelivery: seq[BlockDelivery]) {.asy
     cids.incl(bd.blk.cid)
     if bd.address.leaf:
       cids.incl(bd.address.treeCid)
+
+  asyncSpawn b.issueCancellations(blocksDelivery.mapIt(it.address))
```
dryajov, line 294: `asyncSpawn` isn't needed here at all... what I would do is just a regular `await b.issueCancellations(blocksDelivery.mapIt(it.address))`, but issue the network requests in "parallel", collect the futures into a list and then await them all.

Something like this:

```nim
let
  addrs = blocksDelivery.mapIt(it.address)
  failed = (await allFinished(b.peers.mapIt(
  b.network.request.sendWantList(id = it.id,
      addresses = addrs,
      cancel = true) )))
      .filterIt( it.failed )

if failed.len > 0:
  ...
```

`asynSpawn` is used for "unattended" tasks, such as running something periodically, for the duration of the process, etc... It also has a few gotchas, mainly that, if all exceptions aren't handled in the async task runner, then it will crash chronos, because there is no `await` point to propagate the exception to. 

I would change this to something:

tests/codex/blockexchange/engine/testengine.nim
```diff
@@ -142,6 +142,15 @@ asyncchecksuite "NetworkStore engine handlers":
     localStore: BlockStore
     blocks: seq[Block]
 
+  const NOP_SEND_WANT_LIST = proc(
```
dryajov, line 145: You probably want to use a more nimish style of constant naming for this `NopSendWantList`

codex/utils.nim
```diff
@@ -35,6 +36,24 @@ proc orElse*[A](a, b: Option[A]): Option[A] =
   else:
     b
 
+template findIt*(s, pred: untyped): untyped =
```
dryajov, line 39: Not sure if this is needed, there is a `binarySearch` helper in `algorithm.nim` that does the same I believe.

```nim
    assert binarySearch(["a", "b", "c", "d"], "d", system.cmp[string]) == 3
    assert binarySearch(["a", "b", "c", "d"], "c", system.cmp[string]) == 2
```

dryajov (changes requested):
Some minor comments, but overall LGTM

codex/blockexchange/engine/engine.nim
gmega: Ah yeah good point. I reasoned that I didn't want to "block" whatever came next in the delivery flow as cancellation doesn't have an ordering constraint with respect to anything else, really, but the exception handling is a problem. 

On your second point though: do we really want to issue multiple calls to sendWantList? I don't see how this helps: batching the send as a single call should be just as fast, probably lower overhead on the client (fewer messages), and the code is shorter?

codex/utils.nim
gmega: Binary search requires collections to be sorted. We could use [a simple `find`](https://nim-lang.org/docs/system.html#find%2CT%2CS), but that doesn't work either: if you look closely, [this is the place where I'm using `findIt`](https://github.com/codex-storage/nim-codex/pull/714/files#diff-dfaae75480f3156603a1a6c6bd492f6eaae100ce1fe4cc40aed4d6815e2db5e0R403), and its precisely replacing a `find` because we need to match against an attribute of the want list entry (the address), and not the entire want list entry. 

I therefore thought of putting out a `findIt`in the same spirit of [`filterIt`](https://nim-lang.org/docs/sequtils.html#filterIt.t%2Cuntyped%2Cuntyped), which seems to be the shortest path to a solution. Btw I'd argue `findIt` should be a part of sequtils, not sure why it's not there.

codex/blockexchange/engine/engine.nim
gmega: Btw, I'd probably not do much in terms of error handling here than logging the error. This is a pure-push protocol, and guarantees are probably limited to what the underlying library can provide (i.e., I'm guessing libp2p probably has some retransmission policy if things fail, and that's probably the best we can do here). A full treatment of this would involve reconciling knowledge on first connection, and expiring very old requests.

dryajov: > do we really want to issue multiple calls to sendWantList?

No we definitely don't that and it isn't whats going on either. What my example does is, initiate transmission without waiting for the previous request to complete and it's doing so only once for each peer.

> I reasoned that I didn't want to "block" whatever came next in the delivery flow as cancellation doesn't have an ordering constraint with respect to anything else, really, but the exception handling is a problem.

Yes,  precisely, the snippet above accomplishes what you were trying to do with the `asyncSpawn`, but in a more reliable and less error prone way. 

Really, if you think about it, the `asyncSpawn` + `await` kinda does that as well, but in a bit of a more convoluted way, that `await` in the for loop, will simply serialize the sends to each peer, which isn't required in this case.

```nim
  for neighbor in b.peers:
    await b.network.request.sendWantList(
      id = neighbor.id,
      addresses = addrs,
      cancel = true
    )
```

> Btw, I'd probably not do much in terms of error handling here other than logging the error.

Yeah, but we could probably either disconnect the failed peers or simply try to retransmit it by pushing the same task on the queue.

> I'm guessing libp2p probably has some retransmission policy if things fail,

No, not at all and frankly it's a good thing.

codex/utils.nim
dryajov: OK, thats a good point, but this utils are getting out of hands, and we need to figure out a way of cleaning them up and organizing better. For now, it's probably ok.

codex/blockexchange/engine/engine.nim
gmega: > No we definitely don't that and it isn't whats going on either. What my example does is, initiate transmission without waiting for the previous request to complete and it's doing so only once for each peer.
> ...
> Really, if you think about it, the `asyncSpawn` + `await` kinda does that as well, but in a bit of a more convoluted way, that `await` in the for loop, will simply serialize the sends to each peer, which isn't required in this case.

Oh, right, I managed to completely misread the snippet you put there. :-) Yes, that's indeed much better, thank you.

> > Btw, I'd probably not do much in terms of error handling here other than logging the error.
> 
> Yeah, but we could probably either disconnect the failed peers or simply try to retransmit it by pushing the same task on the queue.

Hm... I think outright disconnecting because one single message didn't go through might be too aggressive, isn't it? OTOH retrying means you need a circuit-breaking strategy so you don't end up flooding your own queue with retries for a dead peer (which means tracking retry counters/timers, putting together retry logic, etc).

> > I'm guessing libp2p probably has some retransmission policy if things fail,
> 
> No, not at all and frankly it's a good thing.

Yeah, I'm a bit torn on this. OTOH retry magic may create surprising behavior, on the other having to implement possibly similar retry strategies for every send is also a huge burden.

codex/utils.nim
gmega: Yeah I feel you. Ideally we want to isolate those utilities into packages which slowly get either pushed into existing libraries or new ones, otherwise it starts to get very hard to manage.

gmega:
I believe I have addressed your comments, @dryajov. I also attempted to pull cancellations as a separate primitive in `BlockExcNetwork`, though I think a bigger refactor is needed, perhaps hiding the cancel option from clients completely so that sendWantLists always maps into cancel = true, and sendWantCancellations always map to cancel = false.

codex/blockexchange/engine/engine.nim
dryajov: > having to implement possibly similar retry strategies for every send is also a huge burden.

Sure, but that doesn't mean it belongs in the core p2p stack and the biggest challenge is to make it work well with all the backpressure mechanisms that libp2p relies on. In any case, this is not where I would put retries unless there are really strong arguments for it.

codex/utils.nim
```diff
@@ -35,6 +36,24 @@ proc orElse*[A](a, b: Option[A]): Option[A] =
...
+  runnableExamples:
+    type MyType = object
+      att: int
+
+    var s = @[MyType(att: 1), MyType(att: 2), MyType(att: 3)]
+    doAssert s.findIt(it.att == 2) == 1
+    doAssert s.findIt(it.att == 4) == -1
+
+  var index = -1
+  for i, it {.inject.} in enumerate(items(s)):
```
dryajov, line 51: meh, so lame that `enumerate`  only works on for loops, otherwise it could have been just chained in an expression: 

```nim
  arr.enumerate.map((i, e) => if e == pred: e else: -1).filterIt( it > -1 )
```

But this is then, the next best thing :)

slightly off-topic:

A consistent iterators/streams/collections with proper backpressure support, is something that I'd love to have in nim. I think @tbekas started something in `utils/asynciter.nim`, but still a long way to go.

codex/blockexchange/engine/engine.nim
```diff
@@ -280,6 +269,20 @@ proc scheduleTasks(b: BlockExcEngine, blocksDelivery: seq[BlockDelivery]) {.asyn
 
           break # do next peer
 
+proc issueCancellations(b: BlockExcEngine, addrs: seq[BlockAddress]) {.async.} =
```
dryajov, line 272: I would rename this to something like `cancelBlocks`, because `cancelation` is commonly used in the context of `Future`s and it can get confussing.

codex/blockexchange/engine/engine.nim
```diff
@@ -280,6 +269,20 @@ proc scheduleTasks(b: BlockExcEngine, blocksDelivery: seq[BlockDelivery]) {.asyn
...
+  ## Tells neighboring peers that we're no longer interested in a block.
+  trace "Sending block request cancellations to peers", addrs = addrs.len
+
+  let sends = b.peers.mapIt(
+    (peer: it, request: b.network.request.sendWantCancellations(peer = it.id,
+      addresses = addrs)))
+
+  discard await allFinished(sends.mapIt(it.request))
```
dryajov, line 280: Please don't use `discard` in situations like this, either use `allFutures`, which will return `void` or use `allFinished` and check for failed futures. 

In general, It's better to always check if some of the futures failed and do something with it - at the very least log it.

codex/blockexchange/engine/engine.nim
```diff
@@ -280,6 +269,20 @@ proc scheduleTasks(b: BlockExcEngine, blocksDelivery: seq[BlockDelivery]) {.asyn
 
           break # do next peer
 
+proc issueCancellations(b: BlockExcEngine, addrs: seq[BlockAddress]) {.async.} =
+  ## Tells neighboring peers that we're no longer interested in a block.
+  trace "Sending block request cancellations to peers", addrs = addrs.len
+
+  let sends = b.peers.mapIt(
+    (peer: it, request: b.network.request.sendWantCancellations(peer = it.id,
```
dryajov, line 277: Also, no reason to split this into several statements, better do something like:

```nim
let
  failed = (await allFinished(
    b.peers.mapIt(b.network.request.sendWantCancellations(
      peer = it.id,
      address = addrs
      ....
  ))))
  .filterIt( it.failed )

if failed.len > 0:
  trace "Couldn't cancel blocks for some peers", peers = failed.len
```

codex/blockexchange/engine/engine.nim
gmega: OK good, I was not aware of the proper way to do this. `allFutures`it is, then.

codex/blockexchange/engine/engine.nim
gmega: Yeah, that's more complicated than I hoped for, but I want to log the addresses of the peers for which the send failed, not just the length cause that doesn't really tell me much. The other option would've been to log it at the `BlockExcNetwork` level but felt off as no op is doing logging there.

Do you know of a shorter way to do it? Plus: way register errors as trc?

gmega:
OK, I've either performed the changes or attempted to justify why I didn't. Ready for the next round.

codex/blockexchange/engine/engine.nim
dryajov: you can break it up into two steps, with `let finished = await allFinished` and then `if finished.filterIt( it.failed ) > 0: ...`

gmega: Wait, but how can I recover the Peer ID from the Future? I had to build a tuple list with peer ids and futures, then filter on that list later on to recover the peer ids associated to failed futures so I can print them.

dryajov: don't bother printing the peer ids, we just care about the number of failed peers..s

gmega: I disagree - if I need to correlate log traces while debugging, then I will want to know which peers failed to receive messages.

dryajov: You can get a potentially very large list of peers, it's unreasonable to print them all. Otherwise, the future's index corresponds to the index in the `b.peers` array.

gmega: This is not something I expect will be printed often, and the size is proportional to the neighborhood of a peer which I don't see being larger than a hundred or so (probably much much smaller?). Plus you'll hardly ever fail delivery to all peers, so likely will usually print a subset of them. Anyhow I won't fight this. I can replace with len tmw morning.

tests/codex/blockexchange/engine/testblockexc.nim
```diff
@@ -233,9 +205,9 @@ asyncchecksuite "NetworkStore - multiple nodes":
       )
 
     for i in 0..15:
-      (await networkStore[i div 4].engine.localStore.putBlock(blocks[i])).tryGet()
+      (await nodes[i div 4].networkStore.engine.localStore.putBlock(blocks[i])).tryGet()
```
elcritch, line 208: I'm lost on this bit, why the switch from engine to networkStore here?

gmega: I'm creating nodes instead of individual parts now, the change is behavior-preserving.

gmega: We're accessing engine in both versions, the difference is that before we had a reference to an array of networkStore, now we have a reference to an array of nodes (hence need to do nodes[blabla].networkStore).

codex/blockexchange/engine/engine.nim
gmega: Done. I believe this makes the logs of lower quality, though, and less useful if we need them for debugging (i.e., we'll have to go back and add the peer IDs back to figure out what's actually going on). I also do not share your concern that those messages are going to grow large, as peer neighborhoods are likely to be small, and failed sends to be even smaller and more sparse (except perhaps when there is an issue, at which point you probably want to see it loud and clear).

gmega: Btw, if you're concerned with trace sizes, we [trace every message send that goes through the block exchange](https://github.com/codex-storage/nim-codex/blob/1eebaa4ce336c3dd00db97c2c2b5614665414dd7/codex/blockexchange/network/network.nim#L97), and log the peer ID. This generates orders of magnitude more noise than my sparse error condition trace.

gmega: OTOH that may mean that I just need to beef up the info on the trace error condition handler of the block exchange. I suppose that if it contained the actual message type, this would be enough for me to have the info I think we need in the logs.

dryajov (approved):
LGTM

PR #715, opened 2024-02-21
Adds instruction for rust and how-to for MSYS2

benbierens:


BUILDING.md
```diff
@@ -101,6 +101,13 @@ File: `C:/Users/<username>/AppData/Roaming/Code/User/settings.json`
 }
 ```
 
+### Rust
```
dryajov, line 104: I would move this above the VS code steps, as a subsection of `MSYS2`, since this is specific to msys2, other environments, like a pure win32 setup with a native Rust, will have different steps...

dryajov (approved):

PR #716, opened 2024-02-21
Scheduling erasure coding on another thread

tbekas:


tests/testthreading.nim
```diff
@@ -0,0 +1,305 @@
+# import ./helpers
```
tbekas, line 1: This test is a playground file. No need to review it.

build.nims
```diff
@@ -14,7 +14,7 @@ proc buildBinary(name: string, srcDir = "./", params = "", lang = "c") =
     for i in 2..<paramCount():
       extra_params &= " " & paramStr(i)
 
-  let cmd = "nim " & lang & " --out:build/" & name & " " & extra_params & " " & srcDir & name & ".nim"
+  let cmd = "nim " & lang & " --threads:on" & " --out:build/" & name & " " & extra_params & " " & srcDir & name & ".nim"
```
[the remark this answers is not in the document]
tbekas: Fixing it 👍

codex/erasure/asyncbackend.nim
```diff
@@ -0,0 +1,62 @@
...
+  else:
+    let signal = signalRes.value
+
+    var args = ThreadArgs(signal: signal, encoderArgs: args)
+
+    let handle = proxy(tp, args)
+
+    await wait(signal) # TODO consider adding timeout
+
+    let res = sync handle # should be completed
```
elcritch, line 57: Doing this will turn the current main thread into a worker with taskpools. It assumes that if you're doing a sync you're waiting and can do work. 

What you will want to do is to use `tryComplete(handle, res)` from https://github.com/status-im/nim-taskpools/blob/d4c43137c0590cb47f893a66ca8cb027fa6c217e/taskpools/flowvars.nim#L56C10-L56C21

codex/erasure/asyncbackend.nim
```diff
@@ -0,0 +1,62 @@
...
+import pkg/questionable
+import pkg/questionable/results
+
+import ./backend
+
+type
+  EncoderArgs* = object
+    backend*: ptr EncoderBackend
+    data*: ptr seq[seq[byte]]
+    parity*: ptr seq[seq[byte]]
```
elcritch, line 16: Even though you're passing this via a pointer, that the type is a seq of seq's may eventually cause GC corruption. The reason is that even if you pass outer seq by pointer and copy it, the inner seq's will still be tied to the main thread's GC. Even iterating the outer seq to copy the inner seq's to new ones may end linking the original inner seq's memory to the worker threads heap, causing the corruption.

I'm not 100% certain what will happen in this scenario since it's both refc and a seq of seq's. Though pretty much anytime I've tried passing seq[byte] or strings in refc it'd result in GC corruption. IIRC, @dryajov ran into similar crashes as well. The most reliable I found is to use `ptr UncheckedArray[byte]` and a len.

It may not occur in this instance if `encode` args are passed purely as pointers by the compiler internally and the pointers are used to copy the data. However, that may depend on the compiler flags and version as well as the encode parameter types.

Note, the GC corruption is only triggered on GC collection, but you can add a `GC_fullCollect` to force the GC to run. I did this in a loop and just ran tests several times to trigger the error. I think GC_fullCollect may need to run on the worker thread, maybe the main thread too. I can't recall that.

elcritch: Unfortunately I don't have an easy recommendations. 

The simplest would probably be to flatten it into a single `seq[byte]` and pass that as an `(ptr UncheckedArray[byte], int)` and use iterators to reconstruct the shape in the worker thread.

Another option could be to use something like the `DataBuffer` SharedPtrs I setup in datastore which can be passed safely between threads as they use atomics and sharedMemory.

The last option I see would be to create a sequence:

```nim
type
  ArrayHolder*[T] = object
    data*: ptr UncheckedArray[T]
    size*: int

  EncoderArgs* = object
     backend*: ptr EncoderBackend
     data*: ArrayHolder[ArrayHolder[byte]]

var data: seq[ArrayHolder[byte]] = origData.mapIt(ArrayHolder[byte](data: addr(it[0]), size: it.len))
var dataHolder = ArrayHolder[ArrayHolder[byte](data: addr(data[0]), size: data.len())
```

That seems pretty gross though. Flattening it with existing serialization primitives would be probably be easier.

codex/erasure/asyncbackend.nim
elcritch: If you're curious, here's where that occurs: https://github.com/status-im/nim-taskpools/blob/d4c43137c0590cb47f893a66ca8cb027fa6c217e/taskpools/taskpools.nim#L247

codex/erasure/asyncbackend.nim
tbekas: Thanks for the support on the task. For viewers: I discussed this issue with @elcritch and passing `seq[seq[byte]]` to a thread does a deep copy - so it should be safe to simply pass nested seqs (tested on Nim 1.6.16)

elcritch: > Thanks for the support on the task. For viewers: I discussed this issue with @elcritch and passing `seq[seq[byte]]` to a thread does a deep copy - so it should be safe to simply pass nested seqs (tested on Nim 1.6.16)

@tbekas note that Balazs was running into odd corruptions and crashes when passing seq's as arguments to taskpools, similar to this case. So it'll be something to keep an eye on. Moving to ARC/ORC resolved his issues.

codex/erasure/asyncbackend.nim
```diff
@@ -0,0 +1,191 @@
...
+    return failure(err)
+
+  try:
+    let
+      blockSize = data[0].len
+      args = EncodeTaskArgs(signal: signal, backend: unsafeAddr backend, blockSize: blockSize, ecM: ecM)
+      handle = proxySpawnEncodeTask(tp, args, data)
+
+    await wait(signal)
+    let res = sync handle # should be completed
```
elcritch, line 139:
```suggestion
    var res: EncodeTaskResult
    assert tryResult(handle, res) # should be completed
```

Otherwise `sync` will possibly allocate taskpool jobs on the main thread.

codex/erasure/asyncbackend.nim
```diff
@@ -0,0 +1,191 @@
...
+    return failure(err)
+
+  try:
+    let
+      ecK = data[].len
+      args = DecodeTaskArgs(signal: signal, backend: unsafeAddr backend, blockSize: blockSize, ecK: ecK)
+      handle = proxySpawnDecodeTask(tp, args, data, parity)
+
+    await wait(signal)
+    let res = sync handle # should be completed
```
elcritch, line 174:
```suggestion
    var res: DecodeTaskResult
    assert tryResult(handle, res) # should be completed
```

Same as above, `sync` will possibly allocated jobs on the main thread.

codex/erasure/asyncbackend.nim
elcritch, line 139:
```suggestion
    doAssert(handle.isReady())
    let res = sync handle # should be completed
```

Would ensure this remains true.

codex/erasure/asyncbackend.nim
```diff
@@ -0,0 +1,191 @@
...
+      blockSize = data[0].len
+      args = EncodeTaskArgs(signal: signal, backend: unsafeAddr backend, blockSize: blockSize, ecM: ecM)
+      handle = proxySpawnEncodeTask(tp, args, data)
+
+    await wait(signal)
+    let res = sync handle # should be completed
+
+    if res.isOk:
+      var parity = seq[seq[byte]].new()
+      parity[].setLen(ecM)
```
elcritch, line 143:
```suggestion
      var parity = newSeqOfCap[seq[byte]](ecM)
```

I'm not sure what `seq[x].new()` will do. Generally the stdlib uses `newXYQ` or `initXYZ` and not `new`.

codex/erasure/asyncbackend.nim
```diff
@@ -0,0 +1,191 @@
...
+      ecK = data[].len
+      args = DecodeTaskArgs(signal: signal, backend: unsafeAddr backend, blockSize: blockSize, ecK: ecK)
+      handle = proxySpawnDecodeTask(tp, args, data, parity)
+
+    await wait(signal)
+    let res = sync handle # should be completed
+
+    if res.isOk:
+      var recovered = seq[seq[byte]].new()
+      recovered[].setLen(ecK)
```
elcritch, line 178:
```suggestion
      var recovered = newSeqOfCap[seq[byte]](ecK)
```

elcritch (changes requested):
Looks good. Would be nice to cleanup the manually results array into a proper type, but that can wait. 

I think we should change the seq init's though.

codex/erasure/asyncbackend.nim
tbekas: Agree, fixing it 👍

codex/erasure/asyncbackend.nim
tbekas: As far as I see the difference is that `newSeqOfCap()` returns `seq[T]` while `seq[T].new()` returns `ref seq[T]` and the latter type is needed to avoid passing sequences by copy.

elcritch: Ah, yah you're right it'll create a `ref seq[T]`.

> and the latter type is needed to avoid passing sequences by copy

Hmmm, yah I see it's used in the erasure encoding bit. Ok that makes sense then.

elcritch (approved):
Ok, looks good. I read through the copyMem and indexing a couple of times to double check.

benbierens (approved):

benbierens:
Image of this branch passes dist-tests. 👍

dryajov (changes requested):
This is much closer to what we need, but it's still a bit broken. For one, the erasure coding backend isn't thread safe and it requires a) either locking per invocation or b) a new thread local instance each time.

The reason is because the internal work buffers are shared by the same thread, as are the internal tables (but those are read only, so that should be fine).

I also see a lot of unnecessary copying happening between the main thread and the execution thread, but this is a lesser issue.

The main point to keep in mind when working with threads under `refc` and it's memory model is that, objects in the owning thread with a lifetime greater than the execution thread can be safely shared, this means that when reading from an owning thread, it should be safe to share objects between threads provided that they are read only and unmodified after creation and they don't get de-allocated too soon. There are a few trics one can use to ensure this, for example holding an object in a closure, if the stack is going to vanish too soon.

In this case, when encoding/decoding, the original data can be passed by pointer without copying, but the encoded/decodec data, that is held by the thread gc, will have to be copied out on completion.

elcritch:
> The reason is because the internal work buffers are shared by the same thread, as are the internal tables (but those are read only, so that should be fine).

Ah that'd be a problem. That'd be for the Encoder and Decoder backend?

> There are a few trics one can use to ensure this, for example holding an object in a closure, if the stack is going to vanish too soon.

Chronos v4 has a nice `noCancel` which would prevent the async closure from being GCed in the event of a cancellation.

codex/erasure/asyncbackend.nim
dryajov: This is correct, I believe we might want to poll with`tryComplete` here.

codex/erasure/asyncbackend.nim
dryajov: It's a bit of a hack to make the seq mutable and you can't do that with a simple `var`, since nim cannot make sense of the lifetime of the seq in a closure at that point.

codex/erasure/asyncbackend.nim
```diff
@@ -0,0 +1,191 @@
...
+  EncodeTaskResult = Result[SharedArrayHolder[byte], cstring]
+  DecodeTaskResult = Result[SharedArrayHolder[byte], cstring]
+
+proc encodeTask(args: EncodeTaskArgs, data: seq[seq[byte]]): EncodeTaskResult =
+  var
+    data = data.unsafeAddr
+    parity = newSeqWith[seq[byte]](args.ecM, newSeq[byte](args.blockSize))
+
+  try:
+    let res = args.backend[].encode(data[], parity)
```
dryajov, line 54: I think this is probably the biggest issue right now, the backend isn't really thread safe and it uses internal work (read/write) buffers that will be shared by multiple threads.

codex/erasure/asyncbackend.nim
dryajov: > Thanks for the support on the task. For viewers: I discussed this issue with @elcritch and passing seq[seq[byte]] to a thread does a deep copy - so it should be safe to simply pass nested seqs (tested on Nim 1.6.16)

I would recommend against this, there is really no reason to do a deep copy at all and passing by pointer should work, there is also no reason to flatten anything, arrays of arrays of arrays (etc..) are already laid out in a flattened manned in memory, so just getting the pointer to the first element and moving it by the number of bytes of each subelement should work just fine.

codex/erasure/asyncbackend.nim
```diff
@@ -0,0 +1,191 @@
...
+        )
+
+      for i in 0..<parity.len:
+        copyMem(addr resData[i * args.blockSize], addr parity[i][0], args.blockSize)
+
+      return ok(arrHolder)
+    else:
+      return err(res.error)
+  except Exception as exception:
+    return err(exception.msg.cstring)
```
emizzle, line 72: We should try to avoid this as much as possible. Defects and CatchableError inherit from Exception and Defects are not Catchable.

codex/erasure/asyncbackend.nim
tbekas: Passing a pointer can lead to memory error as @elcritch showed in this example
https://github.com/codex-storage/apatheia/pull/9/files#diff-764eb5308e6de6c589e7bd145e1863ba9a52b789a75bdef770099e8a3eaf7807

codex/erasure/asyncbackend.nim
tbekas: Fixing it 👍

codex/erasure/asyncbackend.nim
tbekas: Right, I will change it to `tryComplete` and return `failure(...)` if it happens to not be completed - which it shouldn't.

codex/erasure/asyncbackend.nim
elcritch: > Passing a pointer can lead to memory error as @elcritch showed in this example (and other examples from this PR) https://github.com/codex-storage/apatheia/pull/9/files#diff-764eb5308e6de6c589e7bd145e1863ba9a52b789a75bdef770099e8a3eaf7807

Note that that failure requires the current future to be cancelled, which is done `      await runTest(tp, sig).wait(100.milliseconds)`.

codex/erasure/asyncbackend.nim
```diff
@@ -0,0 +1,191 @@
...
+  finally:
+    if err =? args.signal.fireSync().mapFailure.errorOption():
+      error "Error firing signal", msg = err.msg
+
+proc proxySpawnEncodeTask(
+  tp: Taskpool,
+  args: EncodeTaskArgs,
+  data: ref seq[seq[byte]]
+): Flowvar[EncodeTaskResult] =
+  tp.spawn encodeTask(args, data[])
```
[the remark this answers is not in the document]
elcritch: > So two things to keep in mind, a) we aren't really making a copy of the data just the seq descriptor and b) this is something to keep in mind when debugging potential concurrency issues.

Generally that's the case. Taskpools wrap arguments in `Isolate` which creates a deepcopy unless it can do a move. However, that may not alleviate the concurrency issues with `refc` since it's not clear if the *ownership* of the copied seq data is transferred to the other thread. @tbekas and I tested that it is doing a deep copy, but didn't determine who ends up owning the seq.

```nim
  let taskNode = new(TaskNode, workerContext.currentTask) do:
    type
      ScratchObj_838861196 = object
        data: seq[seq[char]]
        sig: ThreadSignalPtr

    let scratch_838861188 = cast[ptr ScratchObj_838861196](c_calloc(csize_t(1),
        csize_t(16)))
    if isNil(scratch_838861188):
      raise
        (ref OutOfMemDefect)(msg: "Could not allocate memory", parent: nil)
    block:
      var isoTemp_838861192 = isolate(data)
      scratch_838861188.data = extract(isoTemp_838861192)
      var isoTemp_838861194 = isolate(sig)
      scratch_838861188.sig = extract(isoTemp_838861194)
    proc worker_838861197(args`gensym27: pointer) {.gcsafe, nimcall, raises: [].} =
      let objTemp_838861191 = cast[ptr ScratchObj_838861196](args`gensym27)
      let data_838861193 = objTemp_838861191.data
      let sig_838861195 = objTemp_838861191.sig
      worker(data_838861193, sig_838861195)

    proc destroyScratch_838861198(args`gensym27: pointer) {.gcsafe, nimcall,
        raises: [].} =
      let obj_838861199 = cast[ptr ScratchObj_838861196](args`gensym27)
      `=destroy`(obj_838861199[])

    Task(callback: worker_838861197, args: scratch_838861188,
         destroy: destroyScratch_838861198)
  schedule(workerContext, taskNode)
```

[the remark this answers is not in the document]
elcritch: All of that to say that using `ptr seq[seq[byte]]` like @dryajov suggests shouldn't cause us any more issues when using `refc` than using `seq[seq[byte]]` since it's unclear how `refc` mixes with `move`. In the best case @tbekas setup using `seq[seq[byte]]` would move the data to the foreign thread like it would in `arc`.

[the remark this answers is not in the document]
elcritch: Though I'm still concerned of why we're not getting foreign GC heap contamination issue like we were previously @dryajov. Seems like seq's don't cause that corruption to occur anymore. Perhaps the compiler uses `allocShared` now instead? But I've run a bunch of tests with seqs and don't seem to run into the GC-heap contamination issue anymore.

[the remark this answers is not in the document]
tbekas: @dryajov 
> we aren't really making a copy of the data just the seq descriptor

If that would be the case, the following test would fail, but it doesn't.

```nim
var tp = Taskpool.new(num_threads = 2)

proc runTask(zeroes: seq[byte]) =
  var ones = @[byte 1, 1, 1, 1]
  copyMem(unsafeAddr zeroes[0], addr ones[0], 4)

test "some test":
  var zeroes = newSeq[byte](4)

  tp.spawn runTask(zeroes)

  tp.syncAll()

  check:
    zeroes == @[byte 0, 0, 0, 0]
```

codex/erasure/asyncbackend.nim
dryajov: Ah, I forgot about that one :)

dryajov:
I think we can still avoid the copies all together when passing data to encode/decode and only copy the results. Also, deep copy will be quite expensive and it isn't even enabled in 2.0 by default, so I would take a closer look at that.

codex/erasure/asyncbackend.nim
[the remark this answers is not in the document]
dryajov: It would be helpful to print out the expanded version of this, since this is all transformed with macros...

But specifically, with your example, I don't think it would fail either way because you're performing `raw` operations on a memory segment that is big enough to hold the values you're copying into it. 

So, to separate this things a bit, your example above, uses pointer operations, which the GC isn't aware off and since the allocated memory segment is big enough to hold the elements being copied onto it, it will just work.

If you would do a copy directly into `zeroes = @[1, 1, 1, 1]`, it would also work, provided that `zeroes` is big enough to hold all the `ones` being copied. If you would instead copy 5 ones, and zeroes can only hold 4, the GC __could__ move your seq elsewhere, because it would need to do a realloc or some combination of memmove/copy.

What the isolate adds to this picture, is that it simply moves the data onto the taskpools worker object, but the object still lives on the owning thread, not on the worker thread.

So, bottom line, the snippet above doesn't really demonstrate anything with regards to my description of the seq header. 

What I don't understand is why your claiming that it would fail at all and how is this relevant?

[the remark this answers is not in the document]
dryajov: @elcritch 

> All of that to say that using ptr seq[seq[byte]] like @dryajov suggests shouldn't cause us any more issues when using refc than using seq[seq[byte]] since it's unclear how refc mixes with move. In the best case @tbekas setup using seq[seq[byte]] would move the data to the foreign thread like it would in arc.

I don't think the move happens across GCs/threads, this is a move on to the holder object, that still lives on the owning thread, not the worker thread?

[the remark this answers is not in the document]
dryajov: I'm curious to see where the deep copy happens?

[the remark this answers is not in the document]
tbekas: >What I don't understand is why your claiming that it would fail at all and how is this relevant?

Because you said that:

>we aren't really making a copy of the data just the seq descriptor

which implies that we refer the same memory region from both threads. My example shows that threads refer their own copies. If you meant something else by this statement, please explain what you meant exactly.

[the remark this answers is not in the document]
dryajov: > > What I don't understand is why your claiming that it would fail at all and how is this relevant?
> 
> Because you said that:
> 
> > we aren't really making a copy of the data just the seq descriptor
> 
> which implies that we refer the same memory region from both threads. My example shows that threads refer their own copies, both descriptor and the data. If you meant something else by this statement, please explain what you meant exactly.

No, your example doesn't show that at all, that is the point I'm making in https://github.com/codex-storage/nim-codex/pull/716#discussion_r1530745733. For reference, you can see the transformed version of the encode/decode tasks.

I annotated the encode task with comments at the relevant places, decode follows the same patter, but it's here for completenes

Encode:

```nim
  let fut = newFlowVar(typeof(EncodeTaskResult))
  proc taskpool_encodeTask(args: EncodeTaskArgs; data: seq[seq[byte]];
                           fut: Flowvar[EncodeTaskResult]) {.nimcall.} =
    let res`gensym105 = encodeTask(args, data)
    readyWith(fut, res`gensym105)

  let task =
    type
      ScratchObj_822084605 = object
        args: EncodeTaskArgs
        data: seq[seq[byte]]
        fut: Flowvar[EncodeTaskResult]

    let scratch_822084595 = cast[ptr ScratchObj_822084605](c_calloc(csize_t(1),
        csize_t(48)))
    if isNil(scratch_822084595):
      raise
        (ref OutOfMemDefect)(msg: "Could not allocate memory", parent: nil)
    block:
      var isoTemp_822084599 = isolate(args)
      scratch_822084595.args = extract(isoTemp_822084599)
      var isoTemp_822084601 = isolate(data[])
      scratch_822084595.data = extract(isoTemp_822084601) # <---- the seq gets moved over, `extract` does `move` on the contained isolate
      var isoTemp_822084603 = isolate(fut)
      scratch_822084595.fut = extract(isoTemp_822084603)
    proc taskpool_encodeTask_822084606(args`gensym129: pointer) {.gcsafe,
        nimcall.} =
      let objTemp_822084598 = cast[ptr ScratchObj_822084605](args`gensym129)
      let args_822084600 = objTemp_822084598.args
      let data_822084602 = objTemp_822084598.data
      let fut_822084604 = objTemp_822084598.fut
      taskpool_encodeTask(args_822084600, data_822084602, fut_822084604)

    proc destroyScratch_822084607(args`gensym129: pointer) {.gcsafe, nimcall.} =
      let obj_822084608 = cast[ptr ScratchObj_822084605](args`gensym129)
      `=destroy`(obj_822084608[])

    Task(callback: taskpool_encodeTask_822084606, args: scratch_822084595,
         destroy: destroyScratch_822084607)
  let taskNode = new(TaskNode, workerContext.currentTask, task)


  schedule(workerContext, taskNode) # this is where everything gets scheduled on the worker thread, everything else happens in the owning thread
```

Decode:

```nim
  let fut = newFlowVar(typeof(DecodeTaskResult))
  proc taskpool_decodeTask(args: DecodeTaskArgs; data: seq[seq[byte]];
                           parity: seq[seq[byte]];
                           fut: Flowvar[DecodeTaskResult]) {.nimcall.} =
    let res`gensym133 = decodeTask(args, data, parity)
    readyWith(fut, res`gensym133)

  let task =
    type
      ScratchObj_822084876 = object
        args: DecodeTaskArgs
        data: seq[seq[byte]]
        parity: seq[seq[byte]]
        fut: Flowvar[EncodeTaskResult]

    let scratch_822084862 = cast[ptr ScratchObj_822084876](c_calloc(csize_t(1),
        csize_t(56)))
    if isNil(scratch_822084862):
      raise
        (ref OutOfMemDefect)(msg: "Could not allocate memory", parent: nil)
    block:
      var isoTemp_822084868 = isolate(args)
      scratch_822084862.args = extract(isoTemp_822084868)
      var isoTemp_822084870 = isolate(data[])
      scratch_822084862.data = extract(isoTemp_822084870)
      var isoTemp_822084872 = isolate(parity[])
      scratch_822084862.parity = extract(isoTemp_822084872)
      var isoTemp_822084874 = isolate(fut)
      scratch_822084862.fut = extract(isoTemp_822084874)
    proc taskpool_decodeTask_822084877(args`gensym138: pointer) {.gcsafe,
        nimcall.} =
      let objTemp_822084867 = cast[ptr ScratchObj_822084876](args`gensym138)
      let args_822084869 = objTemp_822084867.args
      let data_822084871 = objTemp_822084867.data
      let parity_822084873 = objTemp_822084867.parity
      let fut_822084875 = objTemp_822084867.fut
      taskpool_decodeTask(args_822084869, data_822084871, parity_822084873, fut_822084875)

    proc destroyScratch_822084878(args`gensym138: pointer) {.gcsafe, nimcall.} =
      let obj_822084879 = cast[ptr ScratchObj_822084876](args`gensym138)
      `=destroy`(obj_822084879[])

    Task(callback: taskpool_decodeTask_822084877, args: scratch_822084862,
         destroy: destroyScratch_822084878)
  let taskNode = new(TaskNode, workerContext.currentTask, task)
  schedule(workerContext, taskNode)
```

dryajov:
There are some `sleepAsync` that can be removed now in https://github.com/codex-storage/nim-codex/blob/master/codex/erasure/erasure.nim#L291 and https://github.com/codex-storage/nim-codex/blob/master/codex/erasure/erasure.nim#L396

codex/node.nim
```diff
@@ -741,14 +746,18 @@ proc stop*(self: CodexNodeRef) {.async.} =
...
 proc new*(
   T: type CodexNodeRef,
   switch: Switch,
   networkStore: NetworkStore,
   engine: BlockExcEngine,
   discovery: Discovery,
   prover = Prover.none,
-  contracts = Contracts.default): CodexNodeRef =
+  contracts = Contracts.default,
+  taskpool = Taskpool.new(num_threads = countProcessors())): CodexNodeRef =
```
dryajov, line 760: I think this needs to be created in `codex.nim`, because it will be shared by more than one componnent.

codex/erasure/asyncbackend.nim
dryajov: I took another look at this and the way that the encode provider is being used right now is still not thread safe. 

Take a look at https://github.com/codex-storage/nim-codex/pull/716/files#diff-47e9a8d861097fbe82af91a1cb564c1c7d878b5aa46a553c610f75d26d7151efL275, it creates an instance per encoding/decoding session, but this instance is then shared by all the threads spawned by that session. We might be getting away with it because access is properly synchronized/serialized at the main thread level, but it isn't properly guarded at each individual thread level, perhaps an instance per thread is a better solution - this would require passing the encode/decode providers to the backend and having each thread create it's own instance. That or a lock per encoding/decoding session that is shared by all the workers?

For now, I think this is fine, but it seems a bit brittle.

dryajov (approved):
I think this is overall meargeable with the exception of the `sleepAsync` and the `tp` initialization. 

A TL;DR for this thread would be:

- The async backend is created per dataset encoding/decoding session and then shared by all the worker threads. This seems to be properly serialized at the main thread level, but not at the worker thread level and it might turn out to be a bit brittle - something to keep an eye on
- There doesn't seem to be any excessive copying going on. The data flow is as follows
  - The main thread's sets up a task object with all the data properly moved onto that object by way of an isolate
  - The main thread owns the task object and passes a pointer to the worker thread
  - The worker thread has to be careful not to modify this object directly as it is owned by a foreign GC, but read access should be fine
  - This doesn't eliminate potential issues with object lifetimes and cancellations, but it's probably OK in general. I suspect that the issue that @elcritch is talking about with passing the pointer around instead of a properly isolated object, is due to the fact that, the GC doesn't know that another thread is holding that object, in the case of the isolate, the lifetime of the containing object is bound to the task object, which outlives the cancelation. When using pointers, it might be required to handle the lifetime of the object with `GC_ref/GC_unref` to prevent this issues. All in all, this solution is better and safer than the pointer approach.

codex/erasure/asyncbackend.nim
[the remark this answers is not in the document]
tbekas: In your code sample:

```nim
var isoTemp_822084601 = isolate(data[])
scratch_822084595.data = extract(isoTemp_822084601) # <---- the seq gets moved over, `extract` does `move` on the contained isolate
```

The second line maybe does the move, but the first one does the copy.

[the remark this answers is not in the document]
elcritch: IIRC, `var isoTemp_822084601 = isolate(data[])` performs a deepcopy. I've used `Isolate` before and currently it's conservative and clones the data.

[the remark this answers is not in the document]
elcritch: > @elcritch
> 
> > All of that to say that using ptr seq[seq[byte]] like @dryajov suggests shouldn't cause us any more issues when using refc than using seq[seq[byte]] since it's unclear how refc mixes with move. In the best case @tbekas setup using seq[seq[byte]] would move the data to the foreign thread like it would in arc.
> 
> I don't think the move happens across GCs/threads, this is a move on to the holder object, that still lives on the owning thread, not the worker thread?

@dryajov for `refc` I believe you're correct, assuming move in `refc` *doesn't* move the data ownership of the seq to the new thread. I think that makes sense, but then in theory we should be getting "heap contamination".

I was digging into the seq memory handling in `refc` with moves but didn't get too far.

codex/node.nim
elcritch: Yah, I'd agree. The taskpools will probably want to be shared. Though it might make sense to overschedule threads and have multiple taskpools at some point, but I can't envision a good reason currently.

PR #717, opened 2024-02-21
Set environment variables from files in Docker

veaceslavdoina:
This PR adds a way to set environment variables from file(s) in Docker.

This is especially useful for Kubernetes when we run multiple replicas and need to set different variables for each replica.
For example
```shell
CODEX_LISTEN_ADDRS=/ipv4/0.0.0.0/30100
CODEX_LISTEN_ADDRS=/ipv4/0.0.0.0/30200
```

It works for a single or multiple files
```shell
# Single
/opt/env/env.sh

# Multiple
/opt/env/*
```

File is a simple shell file
`/opt/env/env.sh`
```shell
CODEX_ETH_PROVIDER=https://mainnet.infura.io/v3/...
CODEX_ETH_ACCOUNT=0x...
CODEX_ETH_PRIVATE_KEY=0x...
CODEX_MARKETPLACE_ADDRESS=0x...
```

### DIFF
--- a/docker/docker-entrypoint.sh
+++ b/docker/docker-entrypoint.sh
@@ -1,5 +1,10 @@
 #!/bin/bash
 
+# Environment variables from files
+if [[ -n "${ENV_PATH}" ]]; then
+  [[ -f ${ENV_PATH} ]] && source ${ENV_PATH} || for f in ${ENV_PATH}/*; do source $f; done
+fi
+
 # Parameters
 if [[ -z "${CODEX_NAT}" ]]; then
   if [[ "${NAT_IP_AUTO}" == "true" && -z "${NAT_PUBLIC_IP_AUTO}" ]]; then


veaceslavdoina:
There a different ways how we can deploy applications in Kubernetes
- StatefulSet
- Deployments

It is very simple when we have just a single node and thing goes harder when we would like to deploy more than one replica.
We for sure can make a separate installation for every replica, but we also can use another approach. Especially in case of unified configuration and big number of replicas.

In StatefulSet we will have the following pods in case of `replica=n`
```
codex-0
codex-1
codex-2
codex-3
codex-(n-1)
```
All pods will be managed by same StatefulSet and there is no currently way to set variables per specific replica.

One of the way is to use predefined data stored in `ConfigMap/Secrets` and then mount all that in all replicas. Then replica can pickup its configuration based on `POD_NAME` variable exposed automatically to the Pod
```yaml
- name: POD_NAME
  valueFrom:
    fieldRef:
      fieldPath: metadata.name
```

dryajov (approved):

PR #721, opened 2024-02-26
Feature/eth syncing

benbierens:


benbierens:
I've pulled in pkg/serde of the correct version.
But I guess the new version of nim-ethers also depends on chronos > 4.0.0?

benbierens:
Branch image passes basic dist-tests.

codex/codex.nim
```diff
@@ -56,6 +56,11 @@ type
   CodexPrivateKey* = libp2p.PrivateKey # alias
   EthWallet = ethers.Wallet
 
+proc waitForSync(provider: Provider): Future[void] {.async.} =
+  while await provider.isSyncing:
+    notice "Waiting for Ethereum provider to sync..."
+    await sleepAsync(10.seconds)
```
tbekas, line 62: 10 seconds is quite a large delay. How about making it 1 sec?

tbekas: Also it would be nice to have some timeout on the check

benbierens: This condition should happen only rarely: When your eth node is new and still pulling old blocks. This can easily take an hour in the case of our test net. In the event of a restart: It's likely this loop will not run at all or even more than twice.

Because of these reasons, I don't think we should make the loop faster, or give it an arbitrary time limit.

tbekas: Ok, I understand the issue better now. Still this await is interfering with the node startup time, so I would add some simple backoff - like adding 1s on each iteration, with maximum of 10 secs. So: 1, 2, 3, ... , 10, 10, 10, 10, ...

tbekas (approved):
LGTM

codex/codex.nim
```diff
@@ -56,6 +56,14 @@ type
   CodexPrivateKey* = libp2p.PrivateKey # alias
   EthWallet = ethers.Wallet
 
+proc waitForSync(provider: Provider): Future[void] {.async.} =
+  var sleepTime = 1
+  while await provider.isSyncing:
```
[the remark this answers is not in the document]
benbierens: If the node is synced, it returns immediately. This is a flag kept by geth and it shouldn't wait for a new block or anything.

PR #728, opened 2024-03-08
Bugfix/eth syncing

benbierens:


benbierens:
Turns out the issue was entirely in nim-ethers.

benbierens:
Picked up in update flow in other PRs

PR #731, opened 2024-03-08
Fix validator

markspanbroek:
Updates nim-ethers to latest chronosv3 branch.
Fixes validator, which uses transaction simulation to check whether a proof can be marked as missing.
Fixes gas estimations, so they are done on the pending block, instead of the latest block.

Extracted from #702


gmega:
The transaction simulation part looks fine, but I have one question on the gas estimates: why do we want to use pending? Is it to get a more current/accurate estimate?

benbierens (approved):

markspanbroek:
> why do we want to use pending? Is it to get a more current/accurate estimate?

Good question, the main reason is time accuracy. Gas estimation does a simulation of your transaction and then checks how much gas was consumed. If you use the 'latest' block, then this simulation is done using the timestamp of the latest block, which is in the past. Using the 'pending' block, the simulation is done using a newer timestamp.

This is especially important for the validator, which can only mark a proof as missing when the proof period has ended. If we use the old timestamp to estimate gas, then the gas estimation will fail with a 'period hasn't ended yet' exception, even though the real transaction would not have failed on this.

And in some cases the symptoms are less clear, and it won't fail with an obvious exception based on timing. If the amount of gas is dependent on time, then the estimation can be too low, leading to a "Transaction ran out of gas" exception on the real transactin. This unit test in `nim-ethers` captures this scenario: [testGasEstimation](https://github.com/codex-storage/nim-ethers/blob/942fe034fc96ec32942a1481379db5bb068717fd/testmodule/testGasEstimation.nim#L39).

gmega:
Got it - so basically it's because you have time-sensitive computations in your transactions, and those will fail their mock run during gas estimation if the timestamp taken to be the current time is too old.

PR #732, opened 2024-03-08
Fix on-chain clock

markspanbroek:
Fixes on-chain clock to deal with the weird behaviour that Hardhat shows. It appears that Hardhat is calling the 'newHeads'  callback for pending blocks. This works around the issue by only looking at the 'latest' block, and by only processing blocks once.

Extracted from #702
Depends on #731

### DIFF
--- a/codex/contracts/market.nim
+++ b/codex/contracts/market.nim
@@ -60,7 +60,8 @@ method proofDowntime*(market: OnChainMarket): Future[uint8] {.async.} =
   return config.proofs.downtime
 
 method getPointer*(market: OnChainMarket, slotId: SlotId): Future[uint8] {.async.} =
-  return await market.contract.getPointer(slotId)
+  let overrides = CallOverrides(blockTag: some BlockTag.pending)
+  return await market.contract.getPointer(slotId, overrides)
 
 method myRequests*(market: OnChainMarket): Future[seq[RequestId]] {.async.} =
   return await market.contract.myRequests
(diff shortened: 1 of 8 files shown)


gmega:
If I understand this right, updating the timestamp based on a pending block is not good cause you're basically using a future timestamp, correct? If so, this looks good to me. It also gets rid of the waitFor in `now` which should make my life easier. :-)

One question though: if this is a hardhat-specific issue, wouldn't we be better off by having an implementation with the workaround in testing only? Or do we get any benefits from using this in prod?

codex/contracts/market.nim
```diff
@@ -60,7 +60,8 @@ method proofDowntime*(market: OnChainMarket): Future[uint8] {.async.} =
   return config.proofs.downtime
 
 method getPointer*(market: OnChainMarket, slotId: SlotId): Future[uint8] {.async.} =
-  return await market.contract.getPointer(slotId)
+  let overrides = CallOverrides(blockTag: some BlockTag.pending)
```
benbierens, line 63: What does CallOverrides do, and why is it necessary?

benbierens (approved):
One comment, maybe a chance to clarify something.

markspanbroek:
> If I understand this right, updating the timestamp based on a pending block is not good cause you're basically using a future timestamp, which will `requestStorage` to fail, correct?

Yes, the behaviour that hardhat is displaying is quite bizarre. It looks as though every time we do a transaction simulation on the pending block, it actually creates a temporary block in the chain, and notifies subscribers of that temporary block. Since we do quite a lot of these simulations, we get notified about pending blocks a lot, often with the same block number. This confused the on-chain clock algorithm, so now we filter out these pending blocks and make sure that we only process real blocks once.

> One question though: if this is a hardhat-specific issue, wouldn't we be better off by having an implementation with the workaround in testing only? Or do we get any benefits from using this in prod?

I was weighing this option as well during implementation. But I'd rather not have two different implementations, where one is for testing and the other for real deployment, because then it's easy to miss problems in test scenario's. And the implementation as it is right now should be very robust against weird behaviour from the ethereum node, which can be a good thing when we deploy on a newer L2 whose behaviour we haven't tested yet either.

PR #733, opened 2024-03-08
Fix sales agent

markspanbroek:
Fixes a number of things in the sales agent:
- only act upon timeouts when really inside a new period (period start + 1 second)
- fixes issue with onCancelled and onFailed not being called
- fixes issue where the cancellation loop would continue running when request starts
- fixes the salesagent tests, which weren't being run 

Extracted from #702 
Depends on #732

### DIFF
--- a/codex/purchasing/states/started.nim
+++ b/codex/purchasing/states/started.nim
@@ -29,6 +29,7 @@ method run*(state: PurchaseStarted, machine: Machine): Future[?State] {.async.} 
     failed.complete()
   let subscription = await market.subscribeRequestFailed(purchase.requestId, callback)
 
+  # Ensure that we're past the request end by waiting an additional second
   let ended = clock.waitUntil((await market.getRequestEnd(purchase.requestId)) + 1)
   let fut = await one(ended, failed)
   await subscription.unsubscribe()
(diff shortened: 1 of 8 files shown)


codex/sales/states/proving.nim
```diff
@@ -62,24 +62,31 @@ proc proveLoop(
...
+    of SlotState.Filled:
+      debug "Proving for new period", period = currentPeriod
+      if (await market.isProofRequired(slotId)) or (await market.willProofBeRequired(slotId)):
+        let challenge = await market.getChallenge(slotId)
+        debug "Proof is required", period = currentPeriod, challenge = challenge
+        await state.prove(slot, challenge, onProve, market, currentPeriod)
+    of SlotState.Cancelled:
+      debug "Slot reached cancelled state"
+      # do nothing, let onCancelled callback take care of it
+    of SlotState.Failed:
```
benbierens, line 81: Does proving-loop get cleaned up correctly in case we leave this state via Cancelled or Failed callbacks?

codex/purchasing/states/started.nim
```diff
@@ -29,7 +29,7 @@ method run*(state: PurchaseStarted, machine: Machine): Future[?State] {.async.}
     failed.complete()
   let subscription = await market.subscribeRequestFailed(purchase.requestId, callback)
 
-  let ended = clock.waitUntil(await market.getRequestEnd(purchase.requestId))
+  let ended = clock.waitUntil((await market.getRequestEnd(purchase.requestId)) + 1)
```
benbierens, line 32: These, still todo.

markspanbroek: Fixed in de50950

dryajov (approved):

PR #734, opened 2024-03-08
Update contract

markspanbroek:
Updates the codex smart contracts to the latest version
- G2Point definition updated
- fixes integration tests
- works around hardhat issue where websocket subscriptions would be forgotten after 5 minutes

Extracted from #702 
Depends on #733

PR #735, opened 2024-03-08
Prover CLI updates

markspanbroek:
Updates the CLI parameters for the prover

Extracted from #702 
Depends on #734 
Closes #713


benbierens (approved):
Because of the increased time the integration-tests will now need, I propose we take some time and reconsider which tests run in which loop. Fast-feedback is important, and I'm not sure everything that's currently running is required to run this frequently.

codex.nim
```diff
@@ -54,91 +55,93 @@ when isMainModule:
...
 
-    trace "Repo dir initialized", dir = config.dataDir / "repo"
+  ## Ctrl+C handling
+  proc controlCHandler() {.noconv.} =
+    when defined(windows):
+      # workaround for https://github.com/nim-lang/Nim/issues/4057
+      try:
+        setupForeignThreadGc()
+      except Exception as exc: raiseAssert exc.msg # shouldn't happen
+    notice "Shutting down after having received SIGINT"
```
elcritch, line 104: @gmega this looks to conflict with your recent PR no?

markspanbroek:
> Because of the increased time the integration-tests will now need, I propose we take some time and reconsider which tests run in which loop. Fast-feedback is important, and I'm not sure everything that's currently running is required to run this frequently.

Quite. After this is all merged, I'd like to take some time with Eric and Adam to clean up the integration tests, and that includes taking a good look at what should be tested here, and what not.

dryajov (approved):
I think this looks good too

dryajov (commented):
> > Because of the increased time the integration-tests will now need, I propose we take some time and reconsider which tests run in which loop. Fast-feedback is important, and I'm not sure everything that's currently running is required to run this frequently.

Quite. After this is all merged, I'd like to take some time with Eric and Adam to clean up the integration tests, and that includes taking a good look at what should be tested here, and what not. 

A lot of it can be paralelized I bet.

PR #736, opened 2024-03-08
Wire up prover

markspanbroek:
Produces real ZK proofs when asked to do so by the sales module.

Extracted from #702 
Depends on #735

### DIFF
--- a/codex/blockexchange/engine/engine.nim
+++ b/codex/blockexchange/engine/engine.nim
@@ -129,7 +129,6 @@ proc sendWantHave(
   for p in peers:
     if p notin excluded:
       if address notin p.peerHave:
-        trace " wantHave > ", peer = p.id
         await b.network.request.sendWantList(
           p.id,
           @[address],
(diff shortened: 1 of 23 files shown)


codex/blockexchange/engine/engine.nim
```diff
@@ -167,38 +173,38 @@ proc monitorBlockHandle(b: BlockExcEngine, handle: Future[Block], address: Block
 proc requestBlock*(
   b: BlockExcEngine,
   address: BlockAddress,
-  timeout = DefaultBlockTimeout): Future[Block] {.async.} =
+  timeout = DefaultBlockTimeout): Future[?!Block] {.async.} =
```
benbierens, line 176: I would still very much like to know why this became a result[block]. It seems to make no difference here nor in the place where this is called. What's going on, and how does it relate to wiring in the prover?

dryajov: Handling results, rather than exceptions is a lot simpler and it doesn't break cancellations with `.catch` at the callsite. This should have been a result all along.

dryajov (approved):

codex/blockexchange/engine/engine.nim
benbierens: I agree, it should have been a result all along. If this change is not related to the prover, then this is some clean-up. It doesn't necessarily belong in this PR. But I don't insist it should be taken out. I defer to your judgement on this! :D. I am in favor of doing small clean-ups as we go. I wanted to be sure I wasn't missing something about the prover integration.

tests/codex/sales/states/testinitialproving.nim
```diff
@@ -47,30 +57,57 @@ asyncchecksuite "sales state 'initialproving'":
...
+
+  test "waits another period when the proof pointer is about to wrap around":
+    market.proofPointer = 250
+    let future = state.run(agent)
+    await allowProofToStart()
+    await sleepAsync(10.millis)
+    check not future.finished
+    market.proofPointer = 100
+    await allowProofToStart()
+    discard await future
```
emizzle, line 75: I feel like this needs a (~1ms) timeout, because this test is effectively testing that the state completes immediately given that the period has advanced.

markspanbroek: I don't understand? After advancing the period, we `await` the future. This is allowed to take a while (actually it is allowed to take forever, because we didn't specify a timeout).

PR #737, opened 2024-03-08
Use real verifier

markspanbroek:
Uses the real Groth16 verifier in the smart contracts for the integration tests.

Extracted from #702 
Depends on #736


benbierens (approved):

dryajov (approved):
LGTM

PR #738, opened 2024-03-08
Abort Codex when startup sequence fails

### LINKED ISSUE #640: Node "hangs" when `--persistence` but no provider runs on the expected address
When `--persistence` (or `--validator`) is turned on, the node needs Eth Provider. Codex expects it on the `ws://localhost:8545` by default. During the bootstraping there is step where we check if we have "well known contract address" for given chain id. We fetch Chain ID and compare it to our hardcoded addresses. 

If there is no provider running on the default/given provider URL, then the bootstrapping phase "hangs" and the node does not continue. 

We should add check that provider is present and responding, otherwise the node should exit.

### LINKED ISSUE #705: If no RPC endpoint on the expected address, error is swallowed
If you run Codex with flag that requires blockchain access (like `--persistence`) without having running RPC provider available on default/specified address, then Codex hangs without error. Only when enabling `TRACE` logging you can see the error:

```logs
TRC 2024-02-14 07:49:50.086+01:00 Error connecting to address                topics="JSONRPC-WS-CLIENT" tid=2795324 address=[::1]:8545 exc="(61) Connection refused"
TRC 2024-02-14 07:49:50.086+01:00 Error connecting to address                topics="JSONRPC-WS-CLIENT" tid=2795324 address=127.0.0.1:8545 exc="(61) Connection refused"
```

### LINKED ISSUE #725: [BUG] Ethereum private key is not checked
**Describe the bug**
Ethereum private key file is not checked at Codex run, when `--persistence` or `--validator` is enabled.

**To Reproduce**
1. [Build Codex](https://github.com/codex-storage/nim-codex/blob/master/BUILDING.md) from source
2. Run one of the following commands
    ```shell
     # App exits
     ./build/codex --persistence

     # App continues to run
     ./build/codex --persistence --eth-account=4242424242424242424242424242424242424242
     ./build/codex --persistence --eth-private-key=non-existing
     ./build/codex --persistence --eth-account=4242424242424242424242424242424242424242 --eth-private-key=non-existing
    ```

     Same is applicable for `--validator`.

**Expected behavior**
Codex should check existence of the Ethereum private key when that is required.

**Screenshots**
```shell
docker run --rm \
  -p 8545:8545 \
  codexstorage/codex-contracts-eth:latest

./build/codex \
  --log-level=trace \
  --persistence \
  --eth-provider=localhost:8545 \
  --eth-account=4242424242424242424242424242424242424242 \
  --eth-private-key=non-existing
```
Screenshot 2024-03-05 at 18 05 34

**Environment:**
```
Codex revision: 6ee1162
Nim Compiler Version 1.6.14 [Linux: amd64]
```

**Additional context**
1. Code that checks the key file - https://github.com/codex-storage/nim-codex/blob/master/codex/codex.nim#L96
2. Integration tests around that - https://github.com/codex-storage/nim-codex/blob/master/tests/integration/testcli.nim

### LINKED ISSUE #726: [BUG] Codex hangs on start if Geth node not running
**Describe the bug**

[`waitForSync`](https://github.com/codex-storage/nim-codex/blob/6ee116298929c5419f055abda5eccb7e25caef11/codex/codex.nim#L90) got recently introduced and it probes the provided Geth node in a loop to make sure it is synced before Codex starts.

The problem is that if the Geth node is not running, the `isSyncing` predicate seems to freeze. Either we have a long timeout there (which will add up with the backoff mechanism in `waitForSync`), or there's a bug. Either way, this should die faster (i.e. shouldn't take minutes) if connections are being rejected.

gmega:
fixes #726 
fixes #725 (in part)
fixes #705 
fixes #640 

This PR fixes Codex's startup sequence so it does proper error handling and doesn't simply hang when something goes wrong. It essentially deals with the errors in the main module, making it effectively the supervisor for these ops. As a sidenote, all service loops (and critical ops) should be supervised - there's still plenty of opportunity for us to go into zombie mode as internal services crash with no-one to see or deal with the failures.

### DIFF
--- a/codex.nim
+++ b/codex.nim
@@ -80,7 +80,7 @@ when isMainModule:
 
     var
       state: CodexStatus
-      pendingFuts: seq[Future[void]]
+      shutdown: Future[void]
 
     let
       keyPath =
@@ -93,6 +93,12 @@ when isMainModule:
       server = CodexServer.new(config, privateKey)
 
     ## Ctrl+C handling
+    proc doShutdown() =
+      shutdown = server.stop()
+      state = CodexStatus.Stopping
+
+      notice "Stopping Codex"
+
     proc controlCHandler() {.noconv.} =
       when defined(windows):
         # workaround for https://github.com/nim-lang/Nim/issues/4057
@@ -101,10 +107,7 @@ when isMainModule:
         except Exception as exc: raiseAssert exc.msg # shouldn't happen
       notice "Shutting down after having received SIGINT"
 
-      pendingFuts.add(server.stop())
-      state = CodexStatus.Stopping
-
-      notice "Stopping Codex"
+      doShutdown()
 
     try:
       setControlCHook(controlCHandler)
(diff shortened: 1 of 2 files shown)


gmega:
fixes #705 
fixes #640

elcritch:
This looks handy!

AuHau:
Nice 👍

benbierens (approved):
Nice improvement!

PR #739, opened 2024-03-12
fixes double lookups when block does not exist

### LINKED ISSUE #699: `POST /storage/request` hangs, when manifest can't be retrieved
The API endpoint `POST /storage/request` hangs, when the manifest can't be retrieved from the network. Most probably some time-out should be in place...

Related line: https://github.com/codex-storage/nim-codex/blob/2fc7c75fd25a1d62c2340832d3d01e5ef4578e01/codex/node.nim#L130

gmega:
fixes #699 

`CodexNodeRef.retrieve` always tried to refetch manifests for which `fetchManifest` fails, even if these failures were not due to decoding errors (which would indicate that the CID was pointing to a block which was not a manifest) but timeouts (which would indicate that the CID does not exist, and we should just return with a failure).

This means that, if the CID does not exist, we had to wait for twice the timeout (currently 20 minutes) before getting an error. This PR fixes that, so that we lookup only once and the actual timeout is respected.

### DIFF
--- a/codex/blockexchange/engine/engine.nim
+++ b/codex/blockexchange/engine/engine.nim
@@ -174,12 +174,10 @@ proc monitorBlockHandle(
 proc requestBlock*(
   b: BlockExcEngine,
   address: BlockAddress,
-): Future[Block] {.async.} =
+): Future[?!Block] {.async.} =
   let blockFuture = b.pendingBlocks.getWantHandle(address, b.blockFetchTimeout)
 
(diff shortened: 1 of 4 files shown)


dryajov (approved):
We should figure something out with retries and such, but this looks good for now 👍

AuHau:
While nice fix, I don't think it "solves" the issue I originally created. Having 10 minutes ongoing API call is IMHO still "hanging"... I mean beater 10 minutes then 20, but still...

PR #741, opened 2024-03-18
Remove `codex_use_hardhat` compilation flag

emizzle:
This flag was originally used to change OnChainClock behavior when using hardhat as an ethereum chain source due to a very strange bug which would mark the timestamp of new blocks as one second off the timestamp they should have been.

The issue has since been worked around in another manner, and thus this flag is no longer needed.


markspanbroek (approved):
Thanks for cleaning this up!

AuHau (approved):

PR #748, opened 2024-03-21
refactor: cleanup patch availabilities

AuHau:
Follow up PR to clean up some things after the early merge of https://github.com/codex-storage/nim-codex/pull/692

PR #752, opened 2024-03-26
feat[marketplace]: add slot queue pausing

### LINKED ISSUE #549: Introduce Slot Queue pausing and remove past events fetching
When the Slot Queue runs out of availabilities to fill, it should pause until new availability is added or some capacity is returned to existing availabilities.

Additionally, when new availability is added, the fetching of past events in the availability-added callback should be removed. This is because the queue will be paused when no availability is present, and slots will only be removed when they are filled or when there is availability available but non-matching.

emizzle:
Adds slot queue pausing based on the design in https://github.com/codex-storage/codex-research/pull/188.

@AuHau -- I took the changes in commit https://github.com/codex-storage/nim-codex/pull/748/commits/1425bf5c3f42a80f16821789af7957d4757f1e17 and added to this PR as I needed to update that routine.

Close #549

### DIFF
--- a/codex/sales.nim
+++ b/codex/sales.nim
@@ -257,13 +257,13 @@ proc load*(sales: Sales) {.async.} =
     agent.onCleanUp = proc(returnBytes = false, reprocessSlot = false) {.async.} =
       # since workers are not being dispatched, this future has not been created
       # by a worker. Create a dummy one here so we can call sales.cleanUp
-      let done = newFuture[void]("onCleanUp_Dummy")
+      let done: Future[void] = nil
       await sales.cleanUp(agent, returnBytes, reprocessSlot, done)
 
     # TODO: this seemed to be missing during load, but unknown if it was
     # intentionally left out for some unknown reason?
     agent.onFilled = some proc(request: StorageRequest, slotIndex: UInt256) =
-      let done = newFuture[void]("onFilled_Dummy")
+      let done: Future[void] = nil
       sales.filled(request, slotIndex, done)
 
     agent.start(SaleUnknown())
--- a/codex/sales.nim
+++ b/codex/sales.nim
@@ -283,6 +283,7 @@ proc onAvailabilityAdded(sales: Sales, availability: Availability) {.async.} =
 
   queue.clearSeenFlags()
   if queue.paused:
+    trace "unpausing queue after new availability added"
     queue.unpause()
 
 proc onStorageRequested(sales: Sales,
--- a/codex/sales.nim
+++ b/codex/sales.nim
@@ -159,7 +159,7 @@ proc cleanUp(sales: Sales,
                                       seen = true)
     if err =? queue.push(seenItem).errorOption:
       error "failed to readd slot to queue",
-        errorType = type err, error = err.msg
+        errorType = $(type err), error = err.msg
 
   # signal back to the slot queue to cycle a worker
   if not processing.isNil and not processing.finished():
@@ -267,6 +267,12 @@ proc load*(sales: Sales) {.async.} =
 
     agent.start(SaleUnknown())
     sales.agents.add agent
+
+proc onAvailabilitiesEmptied(sales: Sales, availability: Availability) {.async.} =
+  ## When there are no availabilities remaining, the queue should be paused
+  let queue = sales.context.slotQueue
+
+  queue.pause()
 
 proc onAvailabilityAdded(sales: Sales, availability: Availability) {.async.} =
   ## When availabilities are modified or added, the queue should be unpaused if
@@ -474,7 +480,11 @@ proc startSlotQueue(sales: Sales) {.async.} =
   proc onAvailabilityAdded(availability: Availability) {.async.} =
     await sales.onAvailabilityAdded(availability)
 
+  proc onAvailabilitiesEmptied(availability: Availability) {.async.} =
+    await sales.onAvailabilitiesEmptied(availability)
+
   reservations.onAvailabilityAdded = onAvailabilityAdded
+  reservations.onAvailabilitiesEmptied = onAvailabilitiesEmptied
 
 proc subscribe(sales: Sales) {.async.} =
   await sales.subscribeRequested()
--- a/codex/sales.nim
+++ b/codex/sales.nim
@@ -78,13 +78,13 @@ proc onProve*(sales: Sales): ?OnProve = sales.context.onProve
 
 proc onExpiryUpdate*(sales: Sales): ?OnExpiryUpdate = sales.context.onExpiryUpdate
 
-func new*(_: type Sales,
+proc new*(_: type Sales,
           market: Market,
           clock: Clock,
           repo: RepoStore): Sales =
   Sales.new(market, clock, repo, 0)
 
-func new*(_: type Sales,
+proc new*(_: type Sales,
           market: Market,
           clock: Clock,
           repo: RepoStore,
@@ -146,8 +146,6 @@ proc cleanUp(sales: Sales,
                   )).errorOption:
       error "failure deleting reservation", error = deleteErr.msg
 
-  await sales.remove(agent)
-
   # Re-add items back into the queue to prevent small availabilities from
   # draining the queue. Seen items will be ordered last.
   if reprocessSlot and request =? data.request:
@@ -157,9 +155,12 @@ proc cleanUp(sales: Sales,
                                       data.ask,
                                       request.expiry,
                                       seen = true)
+    trace "pushing ignored item to queue, marked as seen"
     if err =? queue.push(seenItem).errorOption:
       error "failed to readd slot to queue",
         errorType = $(type err), error = err.msg
+
+  await sales.remove(agent)
 
   # signal back to the slot queue to cycle a worker
   if not processing.isNil and not processing.finished():
@@ -473,6 +474,7 @@ proc startSlotQueue(sales: Sales) {.async.} =
 
   slotQueue.onProcessSlot =
     proc(item: SlotQueueItem, done: Future[void]) {.async.} =
+      trace "processing slot queue item", reqId = item.requestId, slotIdx = item.slotIndex
       sales.processSlot(item, done)
 
   asyncSpawn slotQueue.start()
--- a/codex/sales.nim
+++ b/codex/sales.nim
@@ -111,15 +111,19 @@ proc remove(sales: Sales, agent: SalesAgent) {.async.} =
 proc cleanUp(sales: Sales,
              agent: SalesAgent,
              returnBytes: bool,
+             reprocessSlot: bool,
              processing: Future[void]) {.async.} =
 
   let data = agent.data
 
-  trace "cleaning up sales agent",
-    requestId = data.requestId,
-    slotIndex = data.slotIndex,
-    reservationId = data.reservation.?id |? ReservationId.default,
+  logScope:
+    topics = "sales cleanUp"
+    requestId = data.requestId
+    slotIndex = data.slotIndex
+    reservationId = data.reservation.?id |? ReservationId.default
     availabilityId = data.reservation.?availabilityId |? AvailabilityId.default
+
+  trace "cleaning up sales agent"
 
   # if reservation for the SalesAgent was not created, then it means
   # that the cleanUp was called before the sales process really started, so
@@ -132,7 +136,6 @@ proc cleanUp(sales: Sales,
                       )).errorOption:
           error "failure returning bytes",
             error = returnErr.msg,
-            availabilityId = reservation.availabilityId,
             bytes = request.ask.slotSize
 
   # delete reservation and return reservation bytes back to the availability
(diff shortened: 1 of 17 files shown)


codex/sales/slotqueue.nim
```diff
@@ -245,6 +263,13 @@ proc push*(self: SlotQueue, item: SlotQueueItem): ?!void =
     self.queue.del(self.queue.size - 1)
 
   doAssert self.queue.len <= self.queue.size - 1
+
+  # when slots are pushed to the queue, the queue should be unpaused if it was
+  # paused
+  if self.paused:
+    trace "unpausing queue after new slots pushed"
+  self.unpause()
```
AuHau, line 271: While I guess it can be fired everytime, shouldn't it be intended under the `if` above?
```suggestion
    self.unpause()
```

codex/sales/slotqueue.nim
```diff
@@ -332,6 +357,17 @@ proc dispatch(self: SlotQueue,
       # throw because it is caller-defined
       warn "Unknown error processing slot in worker", error = e.msg
 
+proc clearSeenFlags*(self: SlotQueue) =
+  # To avoid issues with new queue items being pushed to the queue while all
+  # items are being iterated (eg if a new storage request comes in and pushes
+  # new slots to the queue), first pop all items in the queue off and
```
AuHau, line 363: Not finished comment ;-)

> first pop all items

Hmm I don't think this is what is happening bellow right? `self[i]` only access `i-th` element and not "pop" it... Right?

AuHau (commented):
Nice work! 👍

However, I am a bit confused about one part of your PR. In your proposal at https://github.com/codex-storage/codex-research/pull/188, there were no mentions of the concept of "empty availabilities". I remember we discussed this before and I think reaching the definition of what "empty availability" is was bit tough.

I noticed that you have two ways to pause the queue. The first is based on the `seen` flag as per your original proposal. The second is triggered when "all availabilities are empty". I am unsure why you added the second part. Wouldn't the `seen` flag mechanism already cover this scenario? It seems to introduce unnecessary complexity and complicates understanding and reasoning about the codebase.

What do you think?

codex/sales/slotqueue.nim
emizzle: Yep i figured out a better way and forgot to update the comment. Will update.

codex/sales/slotqueue.nim
emizzle: Yep, it doesn't matter but can update for optics.

codex/sales/reservations.nim
```diff
@@ -90,6 +95,8 @@ const
   SalesKey = (CodexMetaKey / "sales").tryGet # TODO: move to sales module
   ReservationsKey = (SalesKey / "reservations").tryGet
 
+proc all*(self: Reservations, T: type SomeStorableObject): Future[?!seq[T]] {.async.}
```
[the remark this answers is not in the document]
emizzle: It's a forward declaration. Allows calls to the routine in the module when the routine is defined further down the module.

codex/sales/states/errored.nim
```diff
@@ -30,5 +30,5 @@ method run*(state: SaleErrored, machine: Machine): Future[?State] {.async.} =
     onClear(request, data.slotIndex)
 
   if onCleanUp =? agent.onCleanUp:
-    await onCleanUp(returnBytes = true)
```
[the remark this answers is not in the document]
emizzle: A could have happened in any of the states, eg downloading.

emizzle:
That's a great point -- the queue passing should work without the added complexity of pausing when availabilities are "empty". I'll try to remove and write a test to assert that the queue pauses once availabilities are "emptied".

codex/sales/states/errored.nim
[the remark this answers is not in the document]
emizzle: Thinking about this again, if an error occurred in preparing, downloading, or filling, then it stands to reason that the node should take another stab at filling the slot. If an occurred in filled or proving, then the node should not try to reprocess the slot. Will update to reflect this.

emizzle:
@AuHau the PR has been updated with your feedback. Thanks for the review! 🙌

codex/sales/states/errored.nim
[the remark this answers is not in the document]
emizzle: I updated it so that only when `SaleErrored` is reached from `SaleDownloading` will the slot be reprocessed. `SalePreparing`. I was thinking that perhaps we could extend this same behaviour to SalePreparing and SaleInitialProving, however I'm not entirely sure if we would want to reprocess a slot if there's an exception while preparing (availability checking and reservation creation), or initial proving (generating the proof).

tests/codex/sales/testsales.nim
```diff
@@ -272,19 +272,39 @@ asyncchecksuite "Sales":
...
+    # onProcessSlot will prevent the queue working as expected in the Sales
+    # module.
     check queue.len == 3
+    for i in 0..<queue.len:
+      check queue[i].seen
+
+  test "queue is paused once availability is insufficient to service slots in queue":
+    createAvailability() # enough to fill a single slot
+    await market.requestStorage(request)
+    # await sleepAsync(10.millis)
```
AuHau, line 295: Dead code? Can be deleted?

tests/codex/sales/testsales.nim
```diff
@@ -272,19 +272,39 @@ asyncchecksuite "Sales":
     let expected = SlotQueueItem.init(request, 2.uint16)
     check eventually itemsProcessed.contains(expected)
 
-  test "items in queue are readded once ignored":
+  test "items in queue are readded (and marked seen) once ignored":
     await market.requestStorage(request)
-    await sleepAsync(10.millis)
+    # await sleepAsync(10.millis)
```
AuHau, line 277: Another dead code?

codex/sales/states/errored.nim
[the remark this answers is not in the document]
AuHau: Nice!
Yeah, lets discuss this on call today. I would maybe incline toward reprocessing for Preparing but not InitialProving?

AuHau (approved):
Pretty great work! There are some cleanup items, but generally LGTM. 

Before merging lets discuss on our call about the reprocessing of error states.

emizzle:
This PR should be ready to go, however I've some (small) changes to the way the queue pauses, so maybe it might be prudent to review again @AuHau ?

AuHau (approved):
Just few comments but generally LGTM! Great work! 👍

codex/sales/slotqueue.nim
```diff
@@ -205,6 +215,14 @@ proc activeWorkers*(self: SlotQueue): int =
...
   self.queue.contains(item)
 
+proc pause*(self: SlotQueue) =
+  # set unpaused flag to false -- unblocks coroutines waiting on unpaused.wait()
+  self.unpaused.clear()
+
+proc unpause*(self: SlotQueue) =
+  # set unpaused flag to true -- coroutines will block on unpaused.wait()
+  self.unpaused.fire()
+
```
markspanbroek, line 225: It's the other way around; `clear()` ensures that `wait()`s are blocked, `fire()` ensures that `wait()`s go through. The documentation of `AsyncEvent` is not very good on this point, but the description of `fire()` is clear: https://status-im.github.io/nim-chronos/api/chronos/asyncsync.html#fire%2CAsyncEvent

```suggestion
proc pause*(self: SlotQueue) =
  # set unpaused flag to false -- coroutines will block on unpaused.wait()
  self.unpaused.clear()

proc unpause*(self: SlotQueue) =
  # set unpaused flag to true -- unblocks coroutines waiting on unpaused.wait()
  self.unpaused.fire()

```

codex/sales/slotqueue.nim
```diff
@@ -332,6 +363,23 @@ proc dispatch(self: SlotQueue,
...
+  # new slots to the queue), this routine must remain synchronous.
+
+  if self.queue.empty:
+    return
+
+  for item in self.queue.mitems:
+    item.seen = false # does not maintain the heap invariant
+
+  # force heap reshuffling to maintain the heap invariant
+  doAssert self.queue.update(self.queue[0]), "slot queue failed to reshuffle"
```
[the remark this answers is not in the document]
markspanbroek: `update()` only returns false when the argument is not in the queue. Since we're supplying `self.queue[0]` as the arguments, this will never happen. And if it does happen, then it's a programmer error and the node should crash, I think.

codex/sales/slotqueue.nim
```diff
@@ -351,21 +399,47 @@ proc start*(self: SlotQueue) {.async.} =
 
   while self.running:
     try:
+      if self.paused:
+        trace "Queue is paused, waiting for new slots or availabilities to be modified/added"
+
+      # block until unpaused is false/cleared, ie wait for queue to be unpaused
```
markspanbroek, line 405:
```suggestion
      # block until unpaused is true/fired, ie wait for queue to be unpaused
```

codex/sales/states/downloading.nim
```diff
@@ -31,6 +31,11 @@ method onSlotFilled*(state: SaleDownloading, requestId: RequestId,
                      slotIndex: UInt256): ?State =
   return some State(SaleFilled())
 
+# override onError so that the slot is reprocess in the case of an unhandled
```
markspanbroek, line 34:
```suggestion
# override onError so that the slot is reprocessed in the case of an unhandled
```

codex/sales/states/downloading.nim
```diff
@@ -31,6 +31,11 @@ method onSlotFilled*(state: SaleDownloading, requestId: RequestId,
...
+# exception
+method onError*(state: SaleDownloading, error: ref CatchableError): ?State =
+  return some State(SaleErrored(error: error, reprocessSlot: true))
```
markspanbroek, line 37: I know we discussed this before, but I can't remember the conclusion. Did we want to try a slot again if it failed to download the first time? I would think that we don't want to do this, because there's no reason to assume that it would work a second time around, so it's better to focus on other slots?

codex/sales/states/downloading.nim
```diff
@@ -69,7 +74,7 @@ method run*(state: SaleDownloading, machine: Machine): Future[?State] {.async.}
   if err =? (await onStore(request,
                            data.slotIndex,
                            onBlocks)).errorOption:
-    return some State(SaleErrored(error: err))
+    return some State(SaleErrored(error: err, reprocessSlot: true))
```
markspanbroek, line 77: Same as above

markspanbroek (dismissed):
Great work @emizzle, thanks! I have a few small comments, but other than that good to go!

dryajov:
Where are we with this? Should it be merged?

codex/sales/slotqueue.nim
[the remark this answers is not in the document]
emizzle: Yes, exactly. In addition, we know that self.queue[0] exists because there is an early return
when `if self.queue.empty` is true. If the queue is empty when we don't expect
it to be, there is no real way to recover.

codex/sales.nim
```diff
@@ -241,59 +254,31 @@ proc load*(sales: Sales) {.async.} =
...
+      # since workers are not being dispatched, this future has not been created
+      # by a worker. Create a dummy one here so we can call sales.cleanUp
+      let done: Future[void] = nil
+      await sales.cleanUp(agent, returnBytes, reprocessSlot, done)
+
+    # TODO: this seemed to be missing during load, but unknown if it was
+    # intentionally left out for some unknown reason?
+    agent.onFilled = some proc(request: StorageRequest, slotIndex: UInt256) =
+      let done: Future[void] = nil
+      sales.filled(request, slotIndex, done)
```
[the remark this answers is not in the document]
emizzle: Removed and updated the comments to reflect.

codex/sales/slotqueue.nim
emizzle: Thanks for the clarification. I had forgotten to update the comments, but got it
right in the logic 😅

codex/sales/states/downloading.nim
emizzle: Yes, that's correct. We did discuss it and I have it in my notes that it should
be updated so that errors are not reprocessed. But it was so long ago, I forgot
to do it. Great catch, thank you!

codex/sales/states/downloading.nim
emizzle: Good catch, will remove the entire override

emizzle:
> Where are we with this? Should it be merged?

Will merge once green. Had to address a few small comments and rebase on master.

markspanbroek (approved):

codex/sales/slotqueue.nim
emizzle: I know where I got this mixed up. Look at the [documentation for the constructor for `AsyncEvent`](https://status-im.github.io/nim-chronos/api/chronos/asyncsync.html#AsyncEvent): 
> An event manages a flag that can be set to true with the fire() procedure and reset to false with the clear() procedure. _The wait() coroutine blocks until the flag is false._ (emphasis mine)

This is actually wrong, or at least I seemed to keep reading it in way that says the flag must be false to unblock coroutines.

AuHau (approved):

Issue #746, opened 2024-03-21
Integration tests improvements


markspanbroek:
The integration tests can use some work. They are slow to run and the tests use different abstractions to setup nodes.
Here are some of the things we can do to improve this, as discussed with @emizzle  and @AuHau:

 - try to parallelize execution of integration tests
   - [ ] use multiple hardhat instances, start them using multinodes
   - [ ] implement test parallelization, either using separate threads or processes, or using async/await
- use the same abstractions for test setup
  - [ ] clean up multinodes and use it in every test
- `testIntegration.nim`:
  - [ ] split this file into more specific tests "integration tests" is too generic
  - [ ] introduce REST unit tests that can replace a lot of the integration tests here
- `testMarketplace`
  - [ ] check out why this test fails every now and then, and whether it's a bug that we haven't discovered yet
  - [ ] try to replace with unit tests:
    - a test in codex-contracts-eth that tests partially paid out amounts
    - a test in sales that checks that we call freeSlot() to get paid

cskiraly:
What I see on current master is that we run lots of marketplace setup operations in a series, each taking I think 200 seconds. This is the timing measured (after adding [this patch](https://github.com/codex-storage/nim-codex/pull/1074/commits/73220d91e886bf285c41d429256c62ad1c47dcec)):

```
[Suite] Marketplace
  [OK] nodes negotiate contracts on the marketplace
      time = 203.03657794 sec
  [OK] node slots gets paid out and rest of tokens are returned to client
      time = 204.34756780 sec

[Suite] Marketplace payouts
  [OK] expired request partially pays out for stored time
      time = 203.91078019 sec

[Suite] Hosts submit regular proofs
  [OK] hosts submit periodic proofs for slots they fill
      time = 208.24815512 sec

[Suite] Simulate invalid proofs
  [OK] slot is freed after too many invalid proofs submitted
      time = 716.75327349 sec
  [OK] slot is not freed when not enough invalid proofs submitted
      time = 289.75894070 sec

[Suite] Validation
  [OK] validator marks proofs as missing when using validation groups
      time = 725.99719334 sec
  [OK] validator uses historical state to mark missing proofs
      time = 714.45928788 sec
```

I'm not sure whether these should indeed run on separate marketplaces, but IMHO we should:
- from those that take around 200 seconds, merge those that can be merged, to run in the same Suite, with only 1 init.
- run Suites in parallel in the CI
[...]

PR #757, opened 2024-03-26
Cleanup integration tests

markspanbroek:
Moves the tests from the venerable `testIntegration.nim` into several new test modules.
- `testrestapi.nim`: tests that exercise and test the REST API
- `testupdownload.nim`: tests that exercise uploading and downloading content
- `testpurchasing.nim`: tests for buying storage
- `testsales.nim`: tests for selling storage
- `testmarketplace.nim`: tests that exercise a combination of buying and selling storage

Part of #746


emizzle (approved):
LGTM, except oddly there is a failing test 🤔

AuHau (approved):

markspanbroek:
Failing test should be fixed in 3d07bbe. Timing between test runner ( provider.getCurrentTime() ) and nodes ( clock.now ) was sometimes off by more than 10 seconds, failing the test.

dryajov (approved):

dryajov (approved):
LGTM

emizzle (approved):
What are your thoughts on moving some of these tests over from `twonodesuite` to `marketplacesuite`?

markspanbroek:
The intermittently failing test should be fixed in 97a57a1. The test would fail when the storage request was cancelled while a proof was being generated. The async state machine would then cancel the proving state, which triggers a chronos cancellation of future that calculates the proof. But the code that calculates the proof would catch and ignore the CancelledError (by catching all CatchableErrors). This pattern of catching CatchableError and not dealing with CancelledError properly happens in more places in the codebase. I tried to fix them all.

emizzle:
> The intermittently failing test should be fixed in [97a57a1](https://github.com/codex-storage/nim-codex/commit/97a57a1703a8d60cd7e1a888f1ba2b9aeb59d4bd). The test would fail when the storage request was cancelled while a proof was being generated. The async state machine would then cancel the proving state, which triggers a chronos cancellation of future that calculates the proof. But the code that calculates the proof would catch and ignore the CancelledError (by catching all CatchableErrors). This pattern of catching CatchableError and not dealing with CancelledError properly happens in more places in the codebase. I tried to fix them all.

Right ok, this is a great catch! Just to clarify, you mean the `initialproving` state, correct? This is where the initial proof is generated, in the `node.onProve` callback, and if cancelled, would be caught here: https://github.com/codex-storage/nim-codex/commit/97a57a1703a8d60cd7e1a888f1ba2b9aeb59d4bd#diff-1809bdd70926fdcad64f23f0974984916e79885ea5316ed2f2ac1457583bb3e5R701-R702

markspanbroek:
> Just to clarify, you mean the `initialproving` state, correct?

No, if I remember correctly it was in the proving state. It had already filled the slot, but the other slots in the contract do not get filled, so eventually the contract is cancelled. This is also why the test was failing intermittently; the odds of it having to provide a proof at the same time that the storage contract expires is small.

codex/blockexchange/network/network.nim
```diff
@@ -244,7 +246,8 @@ proc rpcHandler(
 
     if payment =? SignedState.init(msg.payment):
       asyncSpawn b.handlePayment(peer, payment)
-
+  except CancelledError as error:
```
dryajov, line 249: If we're going to `asyncSpawn handlePayment`, there is no point in `try/catch` around it. Any escaping exception is converted to a `Defect` by `asyncSpawn`.

codex/blockexchange/network/networkpeer.nim
```diff
@@ -72,6 +74,8 @@ proc broadcast*(b: NetworkPeer, msg: Message) =
   proc sendAwaiter() {.async.} =
     try:
       await b.send(msg)
+    except CancelledError as error:
+      raise error
```
dryajov, line 78: Don't re-raise in a proc that is `asyncSpawn` it will be terminate the application because it is converted to a `Defect`. Anything that is `asyncSpawn` should not raise exceptions.

codex/chunker.nim
```diff
@@ -90,6 +90,8 @@ proc new*(
         res += await stream.readOnce(addr data[res], len - res)
     except LPStreamEOFError as exc:
       trace "LPStreamChunker stream Eof", exc = exc.msg
+    except CancelledError as error:
+      raise error
     except CatchableError as exc:
       trace "CatchableError exception", exc = exc.msg
       raise newException(Defect, exc.msg)
```
dryajov, line 97: Not sure why we're converting this to a `Defect`, should maybe be simply re-raised?

codex/blockexchange/network/network.nim
markspanbroek: Good point, I have no idea why the error handling was here in the first place. I just added the extra case for `CancelledError`, but in the current code everything is `asyncSpawn`ed, and no errors will be raised anyway.

codex/blockexchange/network/networkpeer.nim
markspanbroek: Good catch, I misread the `except` as being part of `broadcast` instead of `sendAwaiter`.

By the way, it seems that `broadcast` is unused? Should we remove it?

AuHau (approved):
Great catch! LGTM 👍

dryajov (approved):

PR #760, opened 2024-03-27
Add Node SPR endpoint

elcritch:
Adds a text only API endpoint `curl http://127.0.0.1:8080/api/codex/v1/spr` to simplify curl commands.

Question, when `node.discovery.dhtRecord.isSome` is empty what should we do? I followed the debug info and return empty string 204 status code. Though perhaps it should be an error, maybe a 503?

Implements https://github.com/orgs/codex-storage/projects/3/views/9?pane=issue&itemId=57116903

### DIFF
--- a/codex/rest/api.nim
+++ b/codex/rest/api.nim
@@ -492,6 +492,78 @@ proc initPurchasingApi(node: CodexNodeRef, router: var RestRouter) =
         trace "Excepting processing request", exc = exc.msg
         return RestApiResponse.error(Http500)
 
+proc initNodeApi(node: CodexNodeRef, conf: CodexConf, router: var RestRouter) =
+  ## various node management api's
+  ## 
+  router.api(
+    MethodGet,
+    "/api/codex/v1/spr") do () -> RestApiResponse:
+      ## Returns node SPR in requested format, json or text.
+      ##
+      try:
+        without spr =? node.discovery.dhtRecord:
+          return RestApiResponse.response("", status=Http503, contentType="application/json")
+
+        if $preferredContentType().get() == "text/plain":
+            return RestApiResponse.response(spr.toURI, contentType="text/plain")
+        else:
+            return RestApiResponse.response($ %* {"spr": spr.toURI}, contentType="application/json")
+      except CatchableError as exc:
+        trace "Excepting processing request", exc = exc.msg
+        return RestApiResponse.error(Http500)
+
+  router.api(
+    MethodGet,
+    "/api/codex/v1/peerid") do () -> RestApiResponse:
+      ## Returns node's peerId in requested format, json or text.
+      ##
+      try:
+        let id = $node.switch.peerInfo.peerId
+
+        if $preferredContentType().get() == "text/plain":
+            return RestApiResponse.response(id, contentType="text/plain")
+        else:
+            return RestApiResponse.response($ %* {"id": id}, contentType="application/json")
+      except CatchableError as exc:
+        trace "Excepting processing request", exc = exc.msg
+        return RestApiResponse.error(Http500)
+
+  router.api(
+    MethodGet,
+    "/api/codex/v1/connect/{peerId}") do (
+      peerId: PeerId,
+      addrs: seq[MultiAddress]) -> RestApiResponse:
+      ## Connect to a peer
+      ##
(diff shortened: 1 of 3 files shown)


AuHau (commented):
I am not a fan of the "text endpoint" approach. IMHO we should soon show some love to the API by, for example, unifying all the endpoints to return the same content type (IMHO JSON).

Regarding "what to do if no SPR is defined," it depends on whether it is a valid state. If I understand it correctly, it is not a valid state for the node to not have an SPR. Therefore, I would suggest returning an error instead. Additionally, I would use a simple `500` status code instead of `503`. The `503` code is more of a temporary error state (such as when the server is overloaded or undergoing maintenance).

elcritch:
> I am not a fan of the "text endpoint" approach. IMHO we should soon show some love to the API by, for example, unifying all the endpoints to return the same content type (IMHO JSON).

This endpoint is primarily for simplifying the tutorials for how to run nodes, which is why I initially went with text. Though I'm not opposed to JSON, I'd like to get @benbierens feedback on that as well. 

> Regarding "what to do if no SPR is defined," it depends on whether it is a valid state. If I understand it correctly, it is not a valid state for the node to not have an SPR. 

Ok that's what I was wondering. The only reason I could think would be while the DHT is initializing and perhaps the SPR wouldn't have been configured yet.

benbierens:
> > I am not a fan of the "text endpoint" approach. IMHO we should soon show some love to the API by, for example, unifying all the endpoints to return the same content type (IMHO JSON).
> 
> This endpoint is primarily for simplifying the tutorials for how to run nodes, which is why I initially went with text. Though I'm not opposed to JSON, I'd like to get @benbierens feedback on that as well.

I am also not a fan of a text endpoint and would prefer a JSON object. Still, in this case I think we should make an exception: If people start having fun with shell scripts and curl commands, having the node return a plain SPR that you can just throw in a var and use, is very handy. I'm sure JSON-parsing exists in shells on lots of platforms. But why make it more complicated?

> > Regarding "what to do if no SPR is defined," it depends on whether it is a valid state. If I understand it correctly, it is not a valid state for the node to not have an SPR.
> 
> Ok that's what I was wondering. The only reason I could think would be while the DHT is initializing and perhaps the SPR wouldn't have been configured yet.

Maybe during the earliest second, there is no SPR and returning and error code would be good, I think. 500? 503? I don't speak HTTP codes. Probably the one that means "Try again in a few seconds, please."

codex/rest/api.nim
```diff
@@ -577,42 +631,7 @@ proc initRestApi*(node: CodexNodeRef, conf: CodexConf, repoStore: RepoStore): Re
   initDataApi(node, repoStore, router)
   initSalesApi(node, router)
   initPurchasingApi(node, router)
+  initNodeApi(node, conf, router)
```
benbierens, line 634: Thanks for adding this one. Thanks for moving /connect to it! 👍

AuHau (approved):
Sweet! LGTM 👍

elcritch:
Ah, looks like I'll have to go back and sign my commits.

elcritch:
Oops, I forgot `Content-Type` http header is for the uploaded data, while `Accept` is the header for a client to specify their desired response type [see MSDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept).

elcritch:
Okay, fixed things and updated the twoclients docs. I'd appreciate if you guys could check the two client doc updates. 

I also added a `curl -X GET -H "Accept: text/plain" http://127.0.0.1:8081/api/codex/v1/peerid` endpoint.

benbierens:
Ran the updated openapi.yaml through the dist-test code generator: Checks out. 👍

elcritch:
> Wouldn't that suffice for your case @elcritch ? IMHO I would avoid information being duplicated between endpoints.

Normally I'd agree with you, but the `debug/info` endpoint is a debugging endpoint and these new endpoints are mainly to help simplify the guides. Also, I think it's info we should commit to offering them directly in the API rather in the debugging section since they're core to our current DHT.

Mainly these new endpoints make it simpler on parts of the user guides. You don't need a tool like `jq` to automate anything either. We discussed it in a client meeting and everyone seemed ok with adding the extra endpoints if it could simplify the guides.

PR #762, opened 2024-03-29
Replace 'number' with 'integer' in openapi spec

benbierens:
'number' can be a float or double.

This would sometimes cause trouble for serialization/deserialization of API calls.


tbekas (approved):
LGTM

PR #763, opened 2024-03-30
enabling hole punching and other libp2p goodies

dryajov:


codex/codex.nim
```diff
@@ -196,24 +200,73 @@ proc stop*(s: CodexServer) {.async.} =
...
+      .withRng(Rng.instance())
+      .withNoise()
+      .withYamux()
+      .withMplex(5.minutes, 5.minutes)
+      .withMaxConnections(config.maxPeers)
+      .withAgentVersion(config.agentString)
+      .withSignedPeerRecord(true)
+      .withTcpTransport({ServerFlags.ReuseAddr})
+      .withAutonat()
+      .withRendezVous()
```
elcritch, line 252: Oh interesting, libp2p uses rendezvous? That's handy.

emizzle:
What's the status on this? This is needed 100% for the demo at ethcc. If we can't get holepunching working in time, then we need to come up with workarounds...

gmega:
Superseded by https://github.com/logos-storage/logos-storage-pm/issues/5

PR #790, opened 2024-04-24
Pr add prover benchmark tool

elcritch:


benchmarks/utils.nim
```diff
@@ -0,0 +1,54 @@
...
+template runit*(cmd: string) =
+  ## run shell commands and verify it runs without an error code
+  echo "RUNNING: ", cmd
+  let cmdRes = execShellCmd(cmd)
+  echo "STATUS: ", cmdRes
+  assert cmdRes == 0
+
+template benchmark*(benchmarkName: string, blk: untyped) =
+  ## simple benchmarking of a block of code
+  let nn = 5
```
dryajov, line 19: Can we make this a parameter?

benchmarks/utils.nim
```diff
@@ -0,0 +1,54 @@
...
+      let t0 = epochTime()
+      `blk`
+      let elapsed = epochTime() - t0
+      vals.add elapsed
+
+  var elapsedStr = ""
+  for v in vals:
+    elapsedStr &= ", " & v.formatFloat(format = ffDecimal, precision = 3)
+  stdout.styledWriteLine(
+    fgGreen, "CPU Time [", benchmarkName, "] ", "avg(", $nn, "): ", elapsedStr, " s"
```
dryajov, line 32: Can we actually derive an average for the number of runs here? in fact, I would print this on it's own line at the end of the bench run.

Something like:

```sh
"CPU Time ["... "], ...s"
...

"Average for ${nn} runs is ...."
```

benchmarks/codex_ark_prover_cli.nim
```diff
@@ -0,0 +1,210 @@
+import std/[sequtils, strformat, os, options, importutils]
```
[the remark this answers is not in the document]
elcritch: Certainly

benchmarks/utils.nim
elcritch: Definitely, I was thinking about doing that. The current form is because I wanted to print out each run so I could do a simple linear regression on it. It's a pain to work with though in this form.

dryajov (dismissed):

elcritch:
@dryajov @gmega can you guys re-approve this? I had to fix a merge issue yesterday.

benchmarks/README.md
```diff
@@ -0,0 +1,33 @@
...
+```
+
+By default all circuit files for each combinations of circuit args will be generated in a unique folder named like:
+    nim-codex/benchmarks/circuit_bench_depth32_maxslots256_cellsize2048_blocksize65536_nsamples9_entropy1234567_seed12345_nslots11_ncells512_index3
+
+Generating the circuit files often takes longer than running benchmarks, so caching the results allows re-running the benchmark as needed.
+
+You can modify the `CircuitArgs` and `CircuitEnv` objects in `runAllBenchMarks` to suite your needs. See `create_circuits.nim` for their definition.
+
+By it runs all commands relative to the `nim-codex` repo. This simplifies finding the correct circuit includes paths, etc. `CircuitEnv` sets all of this.
```
dryajov, line 19:
```suggestion
The runner executes all commands relative to the `nim-codex` repo. This simplifies finding the correct circuit includes paths, etc. `CircuitEnv` sets all of this.
```

PR #792, opened 2024-04-25
Reduce logging during file upload

benbierens:
For the upload control flow:
Removes traces where they contribute little to nothing.
Turns traces into infos/warns where appropriate.


dryajov (approved):
LGTM. 

Although I would keep as many traces and would resort to more fine grained filtering through topics. These trace messages are very useful for debugging, even if they don't look so right now.

benbierens:
I checked the runtime performance of topic filtering and it seems to be actually really good. We're currently using topic filtering in the dist-tests to make life much easier.

PR #793, opened 2024-04-25
feat: expiry specified with number of seconds

### LINKED ISSUE #766: Small API improvement: Expiry
Requesting storage has a field 'expiry', which is a Unix timestamp.
It must be in the future, and it must be no further in the future than now + 'duration'.
It would be a very convenient improvement if we could change this to be like duration: a number of seconds.
Duration is a number of seconds and implicitly converts to a timestamp, x seconds from the contract-started moment.
Expiry could be a number of seconds that implicitly converts to a timestamp x seconds from the moment the contract was posted.

AuHau:
Closes #766

### DIFF
--- a/codex/market.nim
+++ b/codex/market.nim
@@ -84,6 +84,10 @@ method getRequestEnd*(market: Market,
                       id: RequestId): Future[SecondsSince1970] {.base, async.} =
   raiseAssert("not implemented")
 
+method getRequestExpiry*(market: Market,
+                      id: RequestId): Future[SecondsSince1970] {.base, async.} =
+  raiseAssert("not implemented")
+
 method getHost*(market: Market,
                 requestId: RequestId,
                 slotIndex: UInt256): Future[?Address] {.base, async.} =
--- a/codex/market.nim
+++ b/codex/market.nim
@@ -84,7 +84,7 @@ method getRequestEnd*(market: Market,
                       id: RequestId): Future[SecondsSince1970] {.base, async.} =
   raiseAssert("not implemented")
 
-method getRequestExpiry*(market: Market,
+method requestExpiresAt*(market: Market,
                       id: RequestId): Future[SecondsSince1970] {.base, async.} =
   raiseAssert("not implemented")
 
--- a/codex/purchasing.nim
+++ b/codex/purchasing.nim
@@ -53,7 +53,7 @@ proc populate*(purchasing: Purchasing,
   if result.ask.proofProbability == 0.u256:
     result.ask.proofProbability = purchasing.proofProbability
   if result.expiry == 0.u256:
-    result.expiry = (purchasing.clock.now().u256 + purchasing.requestExpiryInterval)
+    result.expiry = purchasing.requestExpiryInterval
   if result.nonce == Nonce.default:
     var id = result.nonce.toArray
     doAssert randomBytes(id) == 32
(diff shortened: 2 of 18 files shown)


codex/purchasing.nim
```diff
@@ -53,7 +53,7 @@ proc populate*(purchasing: Purchasing,
   if result.ask.proofProbability == 0.u256:
     result.ask.proofProbability = purchasing.proofProbability
   if result.expiry == 0.u256:
-    result.expiry = (purchasing.clock.now().u256 + purchasing.requestExpiryInterval)
+    result.expiry = purchasing.requestExpiryInterval
```
emizzle, line 56:
```suggestion
    result.expiry = await purchasing.market.getRequestExpiry(request.id)
```
I don't think we really need that default interval anymore. Not sure what it was for to be honest.

codex/purchasing/states/submitted.nim
```diff
@@ -34,7 +34,7 @@ method run*(state: PurchaseSubmitted, machine: Machine): Future[?State] {.async.
     await subscription.unsubscribe()
 
   proc withTimeout(future: Future[void]) {.async.} =
-    let expiry = request.expiry.truncate(int64) + 1
+    let expiry = (await market.getRequestExpiry(request.id)) + 1
```
emizzle, line 37: This value should already be on the `purchase.request`,  so no need to go to the chain for this.

emizzle: Actually, this makes me think of [my earlier point](https://github.com/codex-storage/codex-contracts-eth/pull/99/files#r1575703499), where we shouldn't be using expiry as a duration here, simply because the node's `now` is going to be different from the chain's `now`, meaning the time we wait in this state won't be the same as the event that occurs on chain. So I think a better option would be to:
> We could probably update the `purchasing/states/pending` state to listen for the `StorageRequested` event before going to the `submitted` state, passing in the `expiresAt` value from the event to the `submitted` state. In the `submitted` state, we can use `await clock.waitUntil(expiresAt)` and cancel that future if the request is fulfilled.

That will allow to wait for the right time from the chain.

codex/purchasing.nim
emizzle: Just a comment: `request.expiry` should never really be `0` as we checking for that in the rest api. I think this maybe isn't necessary at all, or actually we should be raising an error.

codex/market.nim
```diff
@@ -84,6 +84,10 @@ method getRequestEnd*(market: Market,
                       id: RequestId): Future[SecondsSince1970] {.base, async.} =
   raiseAssert("not implemented")
 
+method getRequestExpiry*(market: Market,
```
emizzle, line 87: I'm always confused as to whether or not expiry is a duration or a timestamp because it's both: on the request object it's a duration, and when returned from this routine it's a timestamp. Perhaps changing this to `requestExpiresAt` would be a bit less confusing at all the callsites?

codex/purchasing.nim
emizzle, line 56: Just a comment: `request.expiry` should never really be `0` as we checking for that in the rest api. I think this maybe isn't necessary at all, or actually we should be raising an error.

emizzle (approved):
LGTM!

codex/market.nim
AuHau: Great suggestion! Thanks!

codex/purchasing.nim
AuHau: You are right, but this is "older" than my original work on API side of the expiry. Mark, I guess originally envisioned the default value for expiry and hence this "populating" mechanism. I want to discuss it with him today to see if we can remove it in this PR...

PR #797, opened 2024-04-29
Advertise manifests and trees

benbierens:
This updates the advertising loop in the discovery engine to advertise only manifest CIDs and tree CIDs.
Current implementation advertises every block, and does NOT advertise trees.


dryajov (approved):
LGTM

PR #798, opened 2024-04-30
feat: persistant availabilties

### LINKED ISSUE #467: Persistent availabilities

AuHau:
WIP

Closes #467

### DIFF
--- a/codex/rest/api.nim
+++ b/codex/rest/api.nim
@@ -46,6 +46,27 @@ logScope:
 declareCounter(codex_api_uploads, "codex API uploads")
 declareCounter(codex_api_downloads, "codex API downloads")
 
+proc getLongestRequestEnd(node: CodexNodeRef, availabilityId: AvailabilityId): ?!SecondsSince1970 =
+  without contracts =? node.contracts.host:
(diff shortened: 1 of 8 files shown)


emizzle:
Closed in favour of https://github.com/codex-storage/nim-codex/pull/1099.

PR #799, opened 2024-04-30
Update logging for download

benbierens:
Same as https://github.com/codex-storage/nim-codex/pull/792
except for download flow.

### DIFF
--- a/codex/blockexchange/engine/discovery.nim
+++ b/codex/blockexchange/engine/discovery.nim
@@ -128,15 +128,12 @@ proc discoveryTaskLoop(b: DiscoveryEngine) {.async.} =
         cid = await b.discoveryQueue.get()
 
       if cid in b.inFlightDiscReqs:
-        trace "Discovery request already in progress", cid
         continue
 
       let
         haves = b.peers.peersHave(cid)
 
-      trace "Current number of peers for block", cid, peers = haves.len
       if haves.len < b.minPeersPerBlock:
-        trace "Discovering block", cid
         try:
           let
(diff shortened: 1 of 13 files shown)


codex/blockexchange/engine/discovery.nim
```diff
@@ -128,15 +128,12 @@ proc discoveryTaskLoop(b: DiscoveryEngine) {.async.} =
         cid = await b.discoveryQueue.get()
 
       if cid in b.inFlightDiscReqs:
-        trace "Discovery request already in progress", cid
```
dryajov, line 131: Lets not remove this, but put it behind a topic for proper filtering. This trace messages exist so that we debug issues when they appear, I assure you that this will be re-added at some later point again as we run into issues.

Also, trace is really not meant to be on all the time, only for extreme cases where we want to get as full of a picture of whats going on as possible, ideally we should be either on debug or info and only enable trace when required, we also have runtime control of the log levels, so we can enable/disable without recompiling it.

dryajov (changes requested):
please see comment.

codex/blockexchange/engine/discovery.nim
benbierens: This one is restored.

Yes, we can pretty nicely set the log levels for different topics, and zoom in on whatever area we're interested in. Much of these log PRs are also for changing some trace statements that should be info, warn, or errors.

dryajov (approved):
LGTM. But something tells me we'll end up re-adding most of this at some point ;-)

PR #800, opened 2024-04-30
Bug: Large manifests encoding

benbierens:
Something strange is happening to manifests that describe large datasets when they are encoded/decoded into blocks.
The datasetSize parameter never comes out right when the dataset is 5GB or larger.

Traced the issue to: uint32 is used in protobuf-encoding for datasetSize parameters.
5GB in bytes = 5.368.709.120
uint32 max = 4.294.967.295

This PR bumps those to uint64s.


dryajov (approved):
Great find! LGTM!

Issue #770, opened 2024-04-03
[Validator] Error adding new slot while iterating existing slots


emizzle:
From @benbierens:
Seen in the wild, just now:
```
2024-04-01 14:43:27 DBG 2024-04-01 12:43:27.445+00:00 Submitting proof                           topics="marketplace sales proving" tid=1 currentPeriod=28532923 slotId=0xdd25..0457 count=6621
2024-04-01 14:43:27 TRC 2024-04-01 12:43:27.446+00:00 updated clock                              topics="contracts clock" tid=1 blockTime=1711975407 blockNumber=616 offset="-446 milliseconds, -681 microseconds, and -680 nanoseconds" count=6622
2024-04-01 14:43:27 TRC 2024-04-01 12:43:27.446+00:00 Proof not missing                          topics="codex validator" tid=1 checkedPeriod=28532922 inDowntime=false currentPeriod=28532923 count=6623
2024-04-01 14:43:27 ERR 2024-04-01 12:43:27.446+00:00 Unhandled exception in async proc, aborting topics="codex" tid=1 msg="/src/vendor/nimbus-build-system/vendor/Nim/lib/pure/collections/sets.nim(266, 13) `len(s) == length` the length of the HashSet changed while iterating over it" count=6624
```

emizzle:
this is likely a slot to the validator while at the same, existing slots to validate are being iterated. We can potentially solve this using random access (by index) instead of iteration, however there's still a potential for missing a slot that way if new slots are added while iterating existing slots. But, imho, it's better than erroring. We will see more of these incidences as the network grows. 

eg, in `codex/validation.nim`:
```diff
- for slotId in validation.slots
+ for i in 0..<validation.slots.len:
+   let slotId = validation.slots[i]
```

emizzle:
It's probably better to use an `AsyncLock` when iterating. Other option is to use a copy.

benbierens:
I assume the asyncLock does not halt the app, but simply yields control until the lock can be acquired. If that's true then it would be the most efficient solution.

PR #801, opened 2024-05-02
Use a copy of slots hashset for iteration.

benbierens:
For https://github.com/codex-storage/nim-codex/issues/770

### DIFF
--- a/codex/validation.nim
+++ b/codex/validation.nim
@@ -34,6 +34,12 @@ proc new*(
 proc slots*(validation: Validation): seq[SlotId] =
   validation.slots.toSeq
 
+proc iterateSlots(validation: Validation, action: proc(s: SlotId): Future[void] {.async.}) {.async.} =
+  # Copy of hashSet, for iteration.
+  let slots = validation.slots
+  for slotId in slots:
+    await action(slotId)
+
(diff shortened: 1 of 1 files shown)


AuHau (commented):
Can't you simply copy the slots before the `for`? I am not much of fan of this "callback iteration", IMHO `for` is much better for readability...

benbierens:
Can change it back, I just figured this would be less bug-prone. Whatever you want.

emizzle (approved):

codex/validation.nim
```diff
@@ -34,6 +34,12 @@ proc new*(
 proc slots*(validation: Validation): seq[SlotId] =
   validation.slots.toSeq
 
+proc iterateSlots(validation: Validation, action: proc(s: SlotId): Future[void] {.async.}) {.async.} =
```
dryajov, line 37: This pattern doesn't make sense and it will lead to even harder to track concurrency issue due to the introduction of async calls. 

The safer pattern is the most straightforward one, where you copy the `iterable` at the place of ussage.

benbierens: This will patch it: https://github.com/codex-storage/nim-codex/pull/809

PR #802, opened 2024-05-07
feat: one confirmation for all transactions

AuHau:
This is a follow-up fix PR for #793.

The `purchasing` module had an issue because `requestStorage` was called, and a transaction to the blockchain was created. Right away, the state machine transitioned from `pending` to `submitted` state. However, the last PR added a call to the blockchain in the `submitted` state, which requires the transaction to be already mined. This PR ensures that with a requiring confirmation of one block.

After a discussion with Mark, we agreed that we should add confirmation of one block to all transaction creation calls. Mainly because the `Market` abstraction sort of communicates that "when a call to its method finishes, then it is final". Having created a transaction that might not be included in the blockchain is not "final", hence the one confirmation.

### DIFF
--- a/codex/contracts/market.nim
+++ b/codex/contracts/market.nim
@@ -92,7 +92,7 @@ method requestStorage(market: OnChainMarket, request: StorageRequest){.async.} =
   convertEthersError:
     debug "Requesting storage"
     await market.approveFunds(request.price())
-    await market.contract.requestStorage(request)
+    discard await market.contract.requestStorage(request).confirm(1)
 
 method getRequest(market: OnChainMarket,
                   id: RequestId): Future[?StorageRequest] {.async.} =
--- a/codex/contracts/market.nim
+++ b/codex/contracts/market.nim
@@ -159,16 +159,16 @@ method fillSlot(market: OnChainMarket,
                 collateral: UInt256) {.async.} =
   convertEthersError:
     await market.approveFunds(collateral)
-    await market.contract.fillSlot(requestId, slotIndex, proof)
+    discard await market.contract.fillSlot(requestId, slotIndex, proof).confirm(1)
 
 method freeSlot*(market: OnChainMarket, slotId: SlotId) {.async.} =
   convertEthersError:
-    await market.contract.freeSlot(slotId)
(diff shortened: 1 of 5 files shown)


benbierens:
Current master commit: `1a0d2d4 - feat: expiry specified with number of seconds (#793)`
Fails the marketplace dist-test with error: `[2024-05-10T07:47:18.2957273Z] System.Exception : Contract errored: {"State":"errored","Error":"Estimate gas failed Inner exception: execution reverted: Invalid client address"}`

This branch commit: `5a24258 - test: fixing contracts tests`
Test running, will update....

emizzle:
This is likely a needed change, but it makes me think about chain reorgs. Is one confirmation enough?

codex/contracts/market.nim
```diff
@@ -92,7 +92,7 @@ method requestStorage(market: OnChainMarket, request: StorageRequest){.async.} =
   convertEthersError:
     debug "Requesting storage"
     await market.approveFunds(request.price())
-    await market.contract.requestStorage(request)
+    discard await market.contract.requestStorage(request).confirm(1)
```
markspanbroek, line 95: It is probably ok to do `confirm(0)` everywhere; we only want to need sure that it ended up in a block before moving on, so that we can be sure that any queries to the contract state take this transaction into account.

markspanbroek (approved):

markspanbroek:
> This is likely a needed change, but it makes me think about chain reorgs. Is one confirmation enough?

It most likely is when we deploy to an L2, but it does need a bit more investigation: https://github.com/codex-storage/nim-codex/issues/804

codex/contracts/market.nim
emizzle: Good point.

AuHau: I have hard time understanding what `confirm(0)` should do and if it should even work 😳 Can you please comment on the expected behavior?

AuHau: Never mind, found the implementation and understood it.

Hmm, okey I will make it zero for now and lets revisit it later with the reorg research.

PR #806, opened 2024-05-14
Metadata in LevelDB

benbierens:
This sets LevelDB as the default store for both repo and discovery metadata.
Adds a CLI option to (fs, sqlite) to use LevelDB as primary datastore.


benbierens:
This branch passes basic dist-tests. (Marketplace skipped, confirm-fix not merged yet.)
I bet you find this is more interesting though:

Simple 1-node 1GB upload/download test:

Image of this PR branch `4cdc052 - Removes wildcard from metadata query key` = 31.8 seconds
Master image `9379c7c - Advertise manifests and trees (#797)` = 2.7 minutes

dryajov:
We should run the tests with `leveldb` if we're going to be switching to it as our main backend.

benbierens:
> We should run the tests with `leveldb` if we're going to be switching to it as our main backend.

I ran the 1Node-1GB upload/download:
repoKind=fs -> 31.3 seconds
repoKind=leveldb -> 1 minute
(same image, `4cdc052`)

It seems LevelDB is not faster than FS when it comes to storing block data. I'm not entirely surprised.
Would you like to omit leveldb from the repoKind option for this reason? Should we drop the sqlite as well and remove the option entirely?

emizzle:
all tests passed locally on my ARM machine (M2)

dryajov:
> all tests passed locally on my ARM machine (M2)

The tests aren't using leveldb, they still rely on the sqlite backend.

tbekas (approved):
LGTM

dryajov:
> > > all tests passed locally on my ARM machine (M2)
> > 
> > 
> > The tests aren't using leveldb, they still rely on the sqlite backend.
> 
> How so? Does `codex.nim:259` not switch the metadata to LevelDB?

It does for integration tests, but the unit tests are still using sqlite. It would be much safer to switch those over as well.

dryajov (approved):

benbierens:
@dryajov 
I remember we talked about the "/*" that trails our query keys, created here: https://github.com/codex-storage/nim-codex/blob/97b538961ca50909c77af298d628096e77f80a71/codex/stores/keyutils.nim#L45

I had removed this in Codex and you had pointed out that the LevelDB-DS should be made to handle this rather than change it. Which made sense, until I looked into this: https://github.com/codex-storage/nim-datastore/blob/master/tests/datastore/querycommontests.nim

It seems that we test all the datastore implementations to return nested keys without using /*. I think removing it from Codex even when using the existing datastores won't change the behavior. So if using /* is not according to datastore spec, I would propose to remove it from Codex.

elcritch (approved):
Looks good to me!

PR #810, opened 2024-05-15
Debug/Delete API

benbierens:
Enabled by the compile flag `-d:codex_enable_api_debug_delete`, this API lets you delete specific blocks from a dataset.
This API will be used to test the detection of missing/corrupted data and the repair mechanism.


codex/node.nim
```diff
@@ -514,6 +514,32 @@ proc requestStorage*(
...
+  let
+    totalBlocks = manifest.blocksCount
+    treeCid = manifest.treeCid
+
+  var
+    deleted = 0
+
+  for index in blockIndices:
+    if index >= totalBlocks:
+      return failure("Index out of range. Index: " & $index & " totalBlocks: " & $totalBlocks)
```
gmega, line 531: Hm... this leads to a strange API semantics in which if you provide an invalid block ID within a block of valid IDs, you'll delete the blocks that come before the invalid one but then get an error back. Maybe check that all of them are valid before starting to delete?

codex/node.nim
```diff
@@ -514,6 +514,32 @@ proc requestStorage*(
...
+    without hasBlock =? (await self.networkStore.hasBlock(treeCid, index)), err:
+      return failure(err)
+    if not hasBlock:
+      return failure("Tree does not have this index: " & $index & " totalBlocks: " & $totalBlocks)
+    else:
+      if err =? (await self.networkStore.delBlock(treeCid, index)).errorOption:
+        return failure(err)
+      inc deleted
```
gmega, line 539: Why count the deletes if you only return success when all deletes succeed?

codex/rest/api.nim
```diff
@@ -636,6 +636,35 @@ proc initDebugApi(node: CodexNodeRef, conf: CodexConf, router: var RestRouter) =
...
+            return RestApiResponse.error(Http400, error.msg)
+
+          let body = await request.getBody()
+          without params =? DebugDeleteParams.fromJson(body), error:
+            return RestApiResponse.error(Http400, error.msg)
+
+          warn "debug/delete is deleting blocks", cid = $cid, nBlocks = params.blockIndices.len
+
+          without deleted =? (await node.debugDelete(cid, params.blockIndices)), err:
+            return RestApiResponse.error(Http500, err.msg)
```
gmega, line 658: You'll get a 500 if you supply an invalid block index which is just a bad request. I think this is OK for a debug API but not in general.

gmega (approved):
Approved with some suggestions/comments.

codex/node.nim
dryajov: agree

codex/rest/api.nim
dryajov: Yeah, `400` is more appropriate.

codex/node.nim
benbierens: Sure, will fix.

I do want to point out that this is designed for test scenarios. Recovery from this failure is not the plan. There should be zero tolerance for failures for the methods being used to control the scenarios. Should you get this failure, the node is in an odd state, the test has failed, and the test scenario probably needs fixing.

codex/node.nim
benbierens: It's not strictly necessary. If it's success, you know it should have worked. I was thinking the test-code should assert the number that is returned here. Just to have an extra check on the state of the node. What do you think?

codex/rest/api.nim
benbierens: 400 it is!

codex/node.nim
gmega: Sure, I can agree with that. Still, since you've already changed it, all the better - less opportunity for behavior that could be puzzling.

codex/node.nim
gmega: Hm... you mean to assert that the delete code is doing what it's supposed to? I mean, I can live with that if you think it'll make you feel safer using it, but this means you're delegating testing the API to the client in some sense. For a debugging API could be fine.

codex/rest/api.nim
gmega: The error codes should make it easier for API users to understand what they are doing wrong - so 400 for bad parameters (i.e., debug your parameters!) or 500 for actual errors (i.e., debug the code!). We're now mixing validation errors with actual failure to delete stuff, which could lead someone to waste time tweaking parameters when there's an actual bug in code. At that point I'd rather have the 500.

gmega (approved):
Approved with non-critical comments. :-)

codex/node.nim
benbierens: Well yes, the whole thing is behind a compile flag. it's really only there for testing.
When I call this thing with X indices, I will always expect an error OR exactly X in the response. I don't think this will ever behave differently. *BUT* in case it does, in case some code gets changed in the future, in case the laws of physics no longer apply... I *REALLY* want to know. Please permit me this paranoia. :D

codex/rest/api.nim
benbierens: You're right. node.debugDelete's failure should be a 500 in some cases and 400 in others. However, I think it's not really worth it to make this code any more complicated to achieve this. Instead, I added two error log statements to debugDelete. I hope that's an acceptable compromise.

codex/rest/api.nim
```diff
@@ -664,6 +664,34 @@ proc initDebugApi(node: CodexNodeRef, conf: CodexConf, router: var RestRouter) =
...
         return RestApiResponse.error(Http500)
 
+  when codex_enable_api_debug_delete:
+    type
+      DebugDeleteParams = object
+        blockIndices* {.serialize.}: seq[int]
+
+    router.rawApi(
+      MethodPost,
+      "/api/codex/v1/debug/delete/{cid}") do (cid: Cid) -> RestApiResponse:
```
tbekas, line 674: I would prefer using HTTP method `DELETE` on a `block` resource, so basically

```
DELETE /api/codex/v1/block/{cid}
```

benbierens: I think there is currently no way to get the cid of a block. All I have is the cid of the manifest and the number of blocks in it, so with this I can pick and index.

dryajov: `DELETE /api/codex/v1/block/{cid}` - this should delete the manifest
`DELETE /api/codex/v1/block/{cid}?block=1&block=50..100` - should delete the first block and `..` would delete a range.

tbekas: Actually that's a decent idea @dryajov to use query params. Maybe then just `/api/codex/v1/data/...` instead of `/api/codex/v1/block/...` to be consistent with current endpoints?

PR #811, opened 2024-05-15
Rework AsyncIter

tbekas:
* Separate modules for `AsyncIter[T]` and `Iter[T]`
* Tests for both `asynciter.nim` and `iter.nim`
* Helper methods for converting from `QueryIter` to `AsyncIter`

### DIFF
--- a/codex/erasure/erasure.nim
+++ b/codex/erasure/erasure.nim
@@ -120,7 +120,7 @@ proc getPendingBlocks(
         CatchableError,
         "Future for block id not found, tree cid: " & $manifest.treeCid & ", index: " & $index)
 
-  newAsyncIter[(?!bt.Block, int)](genNext, isFinished)
+  AsyncIter[(?!bt.Block, int)].new(genNext, isFinished)
 
 proc prepareEncodingData(
   self: Erasure,
@@ -440,7 +440,7 @@ proc decode*(
   if treeCid != encoded.originalTreeCid:
     return failure("Original tree root differs from the tree root computed out of recovered data")
 
-  let idxIter = newIter(recoveredIndices)
+  let idxIter = Iter[Natural].new(recoveredIndices)
     .filter((i: Natural) => i < tree.leavesCount)
 
   if err =? (await self.store.putSomeProofs(tree, idxIter)).errorOption:
--- a/codex/erasure/erasure.nim
+++ b/codex/erasure/erasure.nim
@@ -120,7 +120,7 @@ proc getPendingBlocks(
         CatchableError,
         "Future for block id not found, tree cid: " & $manifest.treeCid & ", index: " & $index)
 
-  Iter.new(genNext, isFinished)
+  newAsyncIter[(?!bt.Block, int)](genNext, isFinished)
 
 proc prepareEncodingData(
   self: Erasure,
@@ -440,8 +440,7 @@ proc decode*(
   if treeCid != encoded.originalTreeCid:
     return failure("Original tree root differs from the tree root computed out of recovered data")
 
-  let idxIter = Iter
-    .fromItems(recoveredIndices)
+  let idxIter = newIter(recoveredIndices)
     .filter((i: Natural) => i < tree.leavesCount)
 
   if err =? (await self.store.putSomeProofs(tree, idxIter)).errorOption:
--- a/codex/indexingstrategy.nim
+++ b/codex/indexingstrategy.nim
@@ -39,22 +39,9 @@ func checkIteration(self: IndexingStrategy, iteration: int): void {.raises: [Ind
       IndexingError,
       "Indexing iteration can't be greater than or equal to iterations.")
 
-proc getIter(first, last, step: int): Iter[int] =
-  var
-    finish = false
-    cur = first
-
-  func get(): int =
-    result = cur
-    cur += step
-
-    if cur > last:
-      finish = true
-
-  func isFinished(): bool =
-    finish
-
-  Iter.new(get, isFinished)
+func getIter(first, last, step: int): Iter[int] =
+  {.cast(noSideEffect).}:
+    newIter(first, last, step)
 
 func getLinearIndicies(
   self: IndexingStrategy,
--- a/codex/indexingstrategy.nim
+++ b/codex/indexingstrategy.nim
@@ -41,7 +41,7 @@ func checkIteration(self: IndexingStrategy, iteration: int): void {.raises: [Ind
 
 func getIter(first, last, step: int): Iter[int] =
   {.cast(noSideEffect).}:
-    newIter(first, last, step)
+    Iter[int].new(first, last, step)
 
 func getLinearIndicies(
   self: IndexingStrategy,
--- a/codex/stores/queryiterhelper.nim
+++ b/codex/stores/queryiterhelper.nim
@@ -1,0 +1,66 @@
+import pkg/questionable
+import pkg/questionable/results
+import pkg/chronos
+import pkg/chronicles
+import pkg/datastore/typedds
+
+import ../utils/asynciter
+
+type KeyVal[T] = tuple[key: Key, value: T]
+
+proc toAsyncIter*[T](
+  queryIter: QueryIter[T],
+  finishOnErr: bool = true
+  ): Future[?!AsyncIter[?!QueryResponse[T]]] {.async.} =
+  ## Converts `QueryIter[T]` to `AsyncIter[?!QueryResponse[T]]` and automatically
+  ## runs dispose whenever `QueryIter` finishes or whenever an error occurs (only
+  ## if the flag finishOnErr is set to true)
+  ##
+
+  if queryIter.finished:
+    trace "Disposing iterator"
+    if error =? (await queryIter.dispose()).errorOption:
+      return failure(error)
+    return success(emptyAsyncIter[?!QueryResponse[T]]())
+
+  var errOccurred = false
+
+  proc genNext: Future[?!QueryResponse[T]] {.async.} =
+    let queryResOrErr = await queryIter.next()
+
+    if queryResOrErr.isErr:
+      errOccurred = true
+
+    if queryIter.finished or (errOccurred and finishOnErr):
+      trace "Disposing iterator"
+      if error =? (await queryIter.dispose()).errorOption:
+        return failure(error)
+
+    return queryResOrErr
+
+  proc isFinished(): bool =
+    queryIter.finished or (errOccurred and finishOnErr)
+
+  newAsyncIter[?!QueryResponse[T]](genNext, isFinished).success
+
+proc filterSuccess*[T](
+  iter: AsyncIter[?!QueryResponse[T]]
+  ): Future[AsyncIter[tuple[key: Key, value: T]]] {.async.} =
+  ## Filters out any items that are not success
+
+  proc mapping(resOrErr: ?!QueryResponse[T]): Future[?KeyVal[T]] {.async.} =
+    without res =? resOrErr, error:
+      error "Error occurred when getting QueryResponse", msg = error.msg
+      return KeyVal[T].none
+
+    without key =? res.key:
+      warn "No key for a QueryResponse"
+      return KeyVal[T].none
+
+    without value =? res.value, error:
+      error "Error occurred when getting a value from QueryResponse", msg = error.msg
+      return KeyVal[T].none
+
+    (key: key, value: value).some
+
+  await mapFilter[?!QueryResponse[T], KeyVal[T]](iter, mapping)
--- a/codex/stores/queryiterhelper.nim
+++ b/codex/stores/queryiterhelper.nim
@@ -21,7 +21,7 @@ proc toAsyncIter*[T](
     trace "Disposing iterator"
     if error =? (await queryIter.dispose()).errorOption:
       return failure(error)
-    return success(emptyAsyncIter[?!QueryResponse[T]]())
+    return success(AsyncIter[?!QueryResponse[T]].empty())
 
   var errOccurred = false
 
@@ -41,7 +41,7 @@ proc toAsyncIter*[T](
   proc isFinished(): bool =
     queryIter.finished or (errOccurred and finishOnErr)
 
-  newAsyncIter[?!QueryResponse[T]](genNext, isFinished).success
+  AsyncIter[?!QueryResponse[T]].new(genNext, isFinished).success
 
 proc filterSuccess*[T](
   iter: AsyncIter[?!QueryResponse[T]]
--- a/codex/utils/asynciter.nim
+++ b/codex/utils/asynciter.nim
@@ -47,6 +47,9 @@ proc new*[T](_: type AsyncIter[T], genNext: GenNext[Future[T]], isFinished: IsFi
       var item: T
       try:
         item = await genNext()
+      except CancelledError as err:
+        iter.finish
+        raise err
       except CatchableError as err:
         if finishOnErr or isFinished():
           iter.finish
@@ -113,6 +116,8 @@ proc mapFilter*[T, U](iter: AsyncIter[T], mapPredicate: Function[T, Future[Optio
           futU.complete(u)
           nextFutU = some(futU)
           break
+      except CancelledError as err:
+        raise err
       except CatchableError as err:
         let errFut = newFuture[U]("AsyncIter.mapFilterAsync")
         errFut.fail(err)
--- a/codex/utils/asynciter.nim
+++ b/codex/utils/asynciter.nim
@@ -39,7 +39,7 @@ proc flatMap*[T, U](fut: Future[T], fn: Function[T, Future[U]]): Future[U] {.asy
   let t = await fut
   await fn(t)
 
-proc newAsyncIter*[T](genNext: GenNext[Future[T]], isFinished: IsFinished, finishOnErr: bool = true): AsyncIter[T] =
+proc new*[T](_: type AsyncIter[T], genNext: GenNext[Future[T]], isFinished: IsFinished, finishOnErr: bool = true): AsyncIter[T] =
   var iter = AsyncIter[T]()
 
   proc next(): Future[T] {.async.} =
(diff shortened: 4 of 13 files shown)


codex/indexingstrategy.nim
```diff
@@ -39,22 +39,11 @@ func checkIteration(self: IndexingStrategy, iteration: int): void {.raises: [Ind
...
-
-    if cur > last:
-      finish = true
-
-  func isFinished(): bool =
-    finish
-
-  Iter.new(get, isFinished)
+func getIter(first, last, step: int): Iter[int] =
+  (cast[proc (f, l, s: int): Iter[int] {.nimcall, noSideEffect.}](getIterSE)(first, last, step))
```
dryajov, line 46: `cast` is totally unsafe and its bypassing all the language safety checks, I wonder why we're using it here instead of a safer type conversion?

tbekas: This is inside of a `func` so it can't have side effects. The call to newAsyncIter is a `proc` and can't be turned into a `func`. Is there any other way to solve this mismatch other than casting?

tbekas: By your suggestion I used `{.cast(noSideEffect).}` to do a bit safer casting than previously.

codex/stores/queryiterhelper.nim
```diff
@@ -0,0 +1,66 @@
...
+
+type KeyVal[T] = tuple[key: Key, value: T]
+
+proc toAsyncIter*[T](
+  queryIter: QueryIter[T],
+  finishOnErr: bool = true
+  ): Future[?!AsyncIter[?!QueryResponse[T]]] {.async.} =
+  ## Converts `QueryIter[T]` to `AsyncIter[?!QueryResponse[T]]` and automatically
+  ## runs dispose whenever `QueryIter` finishes or whenever an error occurs (only
+  ## if the flag finishOnErr is set to true)
```
benbierens, line 17: There is another case in which a queryIter must be disposed. This is when an iteration is stopped early, before all items have been yielded. LevelDB showed that we were leaking unfinished iters. Sqlite would clean them up automatically on db.close. But LevelDB will throw instead.

I'm not sure this impacts your changes. But please make sure that the QueryIter is disposed correctly, even when the async iter does not error and does not run to completion.

See the TODO here `reservations.nim` in `proc findAvailability`: https://github.com/codex-storage/nim-codex/blob/3246c43174fd5c3be9ad422d6c35431cccafc6c0/codex/sales/reservations.nim#L629

codex/utils/asynciter.nim
```diff
@@ -49,98 +56,71 @@ proc new*[T](_: type Iter, genNext: GenNext[T], isFinished: IsFinished, finishOn
...
-      if predicate(t):
-        nextT = some(t)
+      let futT = iter.next()
+      try:
+        if u =? await futT.flatMap(mapPredicate):
+          let futU = newFuture[U]("AsyncIter.mapFilterAsync")
+          futU.complete(u)
+          nextFutU = some(futU)
+          break
+      except CatchableError as err:
```
benbierens, line 96: It'd probably be good to have a test for the catchable error case too

codex/utils/iter.nim
```diff
@@ -0,0 +1,129 @@
...
+
+proc newIter*[T](genNext: GenNext[T], isFinished: IsFinished, finishOnErr: bool = true): Iter[T] =
+  var iter = Iter[T]()
+
+  proc next(): T {.raises: [CatchableError].} =
+    if not iter.finished:
+      var item: T
+      try:
+        item = genNext()
+      except CatchableError as err:
```
benbierens, line 37: Same here

tests/codex/stores/testqueryiterhelper.nim
```diff
@@ -0,0 +1,65 @@
...
+
+proc decode(T: type string, bytes: seq[byte]): ?!T =
+  success(string.fromBytes(bytes))
+
+asyncchecksuite "Test QueryIter helper":
+  var
+    tds: TypedDatastore
+
+  setupAll:
+    tds = TypedDatastore.init(SQLiteDatastore.new(Memory).tryGet())
```
benbierens, line 25: It might be useful to use the LevelDB one, just because it is more picky about disposing iterators. It'll help find issues.
There is no in-memory LevelDB datastore. There's a helper called `TempLevelDb` that is used in tests for this. Examples everywhere! :D

tests/codex/stores/testqueryiterhelper.nim
```diff
@@ -0,0 +1,65 @@
...
+
+  teardownAll:
+    (await tds.close()).tryGet
+
+  test "Should auto-dispose when QueryIter finishes":
```
benbierens, line 30: If I'm reading this right, this test runs the iter to completion. It'd be really nice to have a test that does not. So it iterates only the first few items of a longer list.

benbierens (commented):
Nice improvements all around.

codex/stores/queryiterhelper.nim
tbekas: Thanks for this comment. I realise that there's a flow like that, but I didn't realise we can address this flow when we have ARC. Current use cases are not finishing iteration prematurely, so either all items have been traversed or an error occurred - both cases are handled.

codex/utils/asynciter.nim
tbekas: You're right, I will put something together.

codex/utils/iter.nim
tbekas: You're right, I will put something together.

tests/codex/stores/testqueryiterhelper.nim
tbekas: I prefer such tests to be done in memory only.

tests/codex/stores/testqueryiterhelper.nim
tbekas: Not sure what such test would suppose to check?

codex/utils/asynciter.nim
```diff
@@ -49,98 +56,71 @@ proc new*[T](_: type Iter, genNext: GenNext[T], isFinished: IsFinished, finishOn
...
-    nextT = T.none
+  proc tryFetch(): Future[void] {.async.} =
+    nextFutU = Future[U].none
     while not iter.finished:
-      let t = iter.next()
-      if predicate(t):
-        nextT = some(t)
+      let futT = iter.next()
+      try:
+        if u =? await futT.flatMap(mapPredicate):
```
gmega, line 91: Maybe there's no way around it, but this does look somewhat redundant: you're copying the content (or error) of the future into another future exactly as is. Wouldn't it make sense to return the actual future as it is?

codex/utils/asynciter.nim
```diff
@@ -49,98 +56,71 @@ proc new*[T](_: type Iter, genNext: GenNext[T], isFinished: IsFinished, finishOn
...
 
-  proc tryFetch(j: int): void =
-    if not iter.finished:
-      let item = iter.next()
-      ringBuf[j mod n] = item
-      if iter.finished:
-        iterLen = min(j + 1, iterLen)
+proc filter*[T](iter: AsyncIter[T], predicate: Function[T, Future[bool]]): Future[AsyncIter[T]] {.async.} =
+  proc wrappedPredicate(t: T): Future[Option[T]] {.async.} =
+    if await predicate(t):
```
gmega, line 115: Why do we care about handling errors when doing mapFilter but not when doing filter?

codex/utils/asynciter.nim
```diff
@@ -49,98 +56,71 @@ proc new*[T](_: type Iter, genNext: GenNext[T], isFinished: IsFinished, finishOn
...
-    genNext    = () => fn(iter.next()),
+proc map*[T, U](iter: AsyncIter[T], fn: Function[T, Future[U]]): AsyncIter[U] =
+  newAsyncIter[U](
+    genNext    = () => iter.next().flatMap(fn),
     isFinished = () => iter.finished
   )
 
-proc filter*[T](iter: Iter[T], predicate: Function[T, bool]): Iter[T] =
-  var nextT: Option[T]
+proc mapFilter*[T, U](iter: AsyncIter[T], mapPredicate: Function[T, Future[Option[U]]]): Future[AsyncIter[U]] {.async.} =
```
gmega, line 83: I guess I was expecting a map and a filter async equivalents as you had before. Why do we have a combined version as well, which appears to have slightly different semantics than combining map and filter (e.g. wrt error handling)?

codex/utils/iter.nim
```diff
@@ -0,0 +1,129 @@
...
+
+  newIter(genNext, isFinished)
+
+proc map*[T, U](iter: Iter[T], fn: Function[T, U]): Iter[U] =
+  newIter(
+    genNext    = () => fn(iter.next()),
+    isFinished = () => iter.finished
+  )
+
+proc mapFilter*[T, U](iter: Iter[T], mapPredicate: Function[T, Option[U]]): Iter[U] =
```
gmega, line 100: Why do we use result objects and handle errors in mapPredicate/iter.next in the async version, but here we don't?

tests/codex/stores/testqueryiterhelper.nim
gmega: I believe @benbierens wants to guarantee that the query iterator is disposed even if it's not run to completion (which, I think, currently it won't be).

tests/codex/utils/testasynciter.nim
```diff
@@ -0,0 +1,77 @@
...
+
+    var items: seq[int]
+
+    for fut in iter3:
+      items.add(await fut)
+
+    check:
+      items == @[1, 3]
+
+  test "Should leave only odd items using `mapFilter`":
```
gmega, line 55: I don't see a test path for the error handling.

gmega (commented):
I am puzzled by some things, but maybe because I'm lacking context. I won't block the approval of this PR though as you have enough people asking for things there already. :-)

codex/erasure/erasure.nim
```diff
@@ -120,7 +120,7 @@ proc getPendingBlocks(
         CatchableError,
         "Future for block id not found, tree cid: " & $manifest.treeCid & ", index: " & $index)
 
-  Iter.new(genNext, isFinished)
+  newAsyncIter[(?!bt.Block, int)](genNext, isFinished)
```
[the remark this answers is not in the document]
tbekas: The new style will not work well with generics to my knowledge. This call is not going to work: `AsyncIter.new[T](...)` hence the old style used here and to keep it consistent it's used in both `Iter` and `AsyncIter`.

codex/indexingstrategy.nim
```diff
@@ -39,22 +39,9 @@ func checkIteration(self: IndexingStrategy, iteration: int): void {.raises: [Ind
...
-    cur += step
-
-    if cur > last:
-      finish = true
-
-  func isFinished(): bool =
-    finish
-
-  Iter.new(get, isFinished)
+func getIter(first, last, step: int): Iter[int] =
```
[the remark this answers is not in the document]
tbekas: The purpose of this change is to refactor **only** the Iter/AsyncIter API. I can create a new change for renaming this function if you like, but I would like to keep this PR clean of any other unrelated refactoring.

codex/utils/asynciter.nim
tbekas: I cannot simply return the future as I need to check if what's returned is `none` or `some` and unwrap it from `Option` if it's `some`. Errors however should be propagated.

codex/utils/asynciter.nim
tbekas: It should have exactly the same semantics. I can write a test for it if you want. The combined version is provided because it's a convenient API.

tests/codex/stores/testqueryiterhelper.nim
tbekas: Ok, not sure if that's possible.

tests/codex/utils/testasynciter.nim
tbekas: In line 79 there's a test that checks if iter finishes on err.

tests/codex/stores/testqueryiterhelper.nim
benbierens: I haven't looked into this. I don't mind either way. It's just that I noticed using the levelDB for real changes the timing behavior, which has already in the past revealed issues we didn't see. So you can keep the test Sqlite if you want. I'd still suggest you at least run it a few times with the levelDB thing instead, just to see nothing mysterious shows up.

tests/codex/stores/testqueryiterhelper.nim
benbierens: Ah yes!
You see, if you use the leveldb helper, then in the setup you create the DB and in teardown you dispose it. If your code does not correctly release the iterators, then your teardown will throw. This way you know there's a problem. (Sqlite doesn't do this, it quietly ignores undisposed iterators.)
So then simply having a test that partially iterates a list would reveal the problem if it's there.

codex/indexingstrategy.nim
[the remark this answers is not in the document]
gmega: Alright, fair.

codex/utils/asynciter.nim
gmega: Yeah OK, true. It's Future[Option[U]] to Option[Future[U]] :face_with_spiral_eyes:

codex/utils/asynciter.nim
gmega: Had missed that.

tests/codex/utils/testasynciter.nim
gmega: This file only goes until line 77 :joy:

gmega (approved):
Nothing major on my end so provided you address the rest LGTM.

codex/utils/asynciter.nim
gmega: Only if you feel you need to. I otherwise believe you.

tests/codex/stores/testqueryiterhelper.nim
tbekas: >So then simply having a test that partially iterates a list would reveal the problem if it's there.

I mean we know that the iterator will not be disposed in this case. If I would add such test it would always fail on the teardown. Also such test would be impossible to fix currently according to the `TODO` mentioned in [this comment](https://github.com/codex-storage/nim-codex/pull/811#discussion_r1627262366).

We could try to assert that iter isn't disposed in such scenario, but that's even worse, because we would specify (tests are specs) that not disposing an iter is expected behavior. So I prefer to leave it unspecified.

If I'm missing something please provide me an example how such test should look like.

codex/erasure/erasure.nim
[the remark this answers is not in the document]
tbekas: I used `AsyncIter[T].new(...)` and it works! Thanks for the suggestion @gmega  👍

codex/utils/iter.nim
tbekas: Fixed. Last item before error should no longer get lost.

tests/codex/stores/testqueryiterhelper.nim
benbierens: It sounds to me like using the query-iter-helper will guarantee we leak the iterator whenever a query is not run to completion. This can be solved by the other GC in the future. But I suppose in the meantime, we need a way to manually dispose the iter when the user of the iter knows that it's not going to complete.
Either asyncIter has to support manually disposing (for now), or we shouldn't use it instead of queryIter. :| This sucks, but crashes due to leaks or other memory issues are a nightmare to even detect.

tbekas: Please elaborate on how using this module will guarantee that "we leak the iterator", maybe an example?

tests/codex/utils/testasynciter.nim
```diff
@@ -0,0 +1,130 @@
...
+
+  test "Should be finished":
+    let iter = AsyncIter[int].empty()
+
+    check:
+      iter.finished == true
+
+  test "Should map each item using `map`":
+    let
+      iter1 = AsyncIter[int].new(0..<5).delayBy(10.millis)
```
AuHau, line 20: Interesting, why the `delayBy()` part there?

AuHau (commented):
Generally LGTM, I would only like to point out the "catch" of `CancelledError` being inherited from `CatchableError` which should be then reraised and not swallowed. For more see Mark's post about it: https://discord.com/channels/895609329053474826/1175098439504244767/1242444718231523338

I think I found one potential place where this should be handled (see the other comment), but generally please have a look on your changes from this POV if maybe there are some other non-handled cases.

benbierens (approved):

tests/codex/utils/testasynciter.nim
tbekas: Just to make sure that we are doing proper async flow, so yielding control back to chronos scheduler for the time of the sleep. Without this it would be synchronous calls only and essentially the same as non-async iter.

AuHau: Hmm, I guess it makes sense, although I don't think it is necessary? Because every time you call `await`, you are yielding control, and there are `await` calls in the `flatMap` and then in collecting, so you should be good in the "yielding control" part. In the same time, I guess it does not really hurt anything to do it like this.

tbekas: I might be wrong, but my impression was that Chronos will not yield control back to the scheduler if the future that we want to await for is already completed, which I think is the case for simple proc with `{.async.}` pragma returning a value immediately (without any await for IO or sleep).

gmega: Yep, I didn't follow the whole flow but you need to await on a non-completed Future [otherwise Chronos will not add a callback](https://github.com/codex-storage/nim-chronos/blob/master/chronos/internal/asyncfutures.nim#L379); i.e., it will run your code to completion.

codex/utils/asynciter.nim
```diff
@@ -49,98 +59,97 @@ proc new*[T](_: type Iter, genNext: GenNext[T], isFinished: IsFinished, finishOn
...
+proc mapAsync*[T, U](iter: Iter[T], fn: Function[T, Future[U]]): AsyncIter[U] =
+  AsyncIter[U].new(
+    genNext = () => fn(iter.next()),
+    isFinished = () => iter.finished()
+  )
 
-proc fromSlice*[U, V: Ordinal](_: type Iter, slice: HSlice[U, V]): Iter[U] =
-  ## Creates new iterator from slice
-  ##
+proc new*[U, V: Ordinal](_: type AsyncIter[U], slice: HSlice[U, V]): AsyncIter[U] =
```
dryajov, line 76: It's not clear anymore what any of this constructors do by themself. Lets add some comments to document it please.

dryajov (approved):
LGTM, except for the documentation comment. But we can address in a separate PR.

tests/codex/utils/testasynciter.nim
AuHau: Interesting! TIL - today I learned ;-)

AuHau (approved):

PR #812, opened 2024-05-15
Generic binary codecs

tbekas:


dryajov:
I don't get this at all... This is attempting to produce a protobuf encoding but in a totally broken way, why?

tbekas:
After an offline discussion with @dryajov we've decided to drop the binary codecs in favour of https://github.com/codex-storage/nim-serde

dryajov:
> After an offline discussion with @dryajov we've decided to drop the binary codecs in favour of https://github.com/codex-storage/nim-serde

Note that we most likely want to use a binary codec in the future, but this encoding wasn't neither here nor there, it was using protobuf in a non-standard way which made little sense. The idea is that we add support for multiple encodings to `nim-serde` in the future, but for now we'll stick with json as this is what is currently supported.

PR #813, opened 2024-05-16
Scheduling prover on another thread

elcritch:


codex/slots/proofs/backends/asynccircoms.nim
```diff
@@ -0,0 +1,82 @@
...
+  AsyncCircomCompat* = object
+    params*: CircomCompatParams
+    tp*: Taskpool
+
+  # Args objects are missing seq[seq[byte]] field, to avoid unnecessary data copy
+  ProveTaskArgs* = object
+    signal: ThreadSignalPtr
+    params: CircomCompatParams
+
+var circomBackend {.threadvar.}: Option[CircomCompat]
```
dryajov, line 24: Each thread is gonna get it's own instance? I think the backend is already thread safe as it is?

codex/slots/proofs/backends/asynccircoms.nim
```diff
@@ -0,0 +1,82 @@
...
+
+proc proveTask[H](
+    args: ProveTaskArgs, data: ProofInputs[H]
+): Result[CircomProof, string] =
+
+  try:
+    if circomBackend.isNone:
+      circomBackend = some CircomCompat.init(args.params)
```
dryajov, line 32: Not a big fan of this pattern, I would leave initialization to the caller.

codex/slots/proofs/backends/asynccircoms.nim
```diff
@@ -0,0 +1,82 @@
...
+): Future[?!CircomProof] {.async.} =
+  ## Generates proof using circom-compat asynchronously
+  ##
+
+  without signal =? ThreadSignalPtr.new().mapFailure, err:
+    return failure(err)
+
+  let args = ProveTaskArgs(signal: signal, params: self.params)
+  proc spawnTask(): Flowvar[Result[CircomProof, string]] =
+    self.tp.spawn proveTask(args, input)
```
dryajov, line 58: This will effectively block the caller

dryajov: In fact, this is also the case with the erasure coding :(

dryajov: Ok, I spoke too soon, `spawn` wont block, you have to call `sync` on the `Flowvar` to block it.

codex/utils/asyncthreads.nim
```diff
@@ -0,0 +1,33 @@
...
+import pkg/taskpools/flowvars
+import pkg/chronos
+import pkg/chronos/threadsync
+import pkg/questionable/results
+
+const
+  CompletionRetryDelay* = 10.millis
+  CompletionTimeout* = 1.seconds # Maximum await time for completition after receiving a signal
+
+proc awaitThreadResult*[T](signal: ThreadSignalPtr, handle: Flowvar[T]): Future[?!T] {.async.} =
```
dryajov, line 12: I know this exists in the erasure async backend right now, but imp this is way too convoluted. 

Maybe something like this:

```suggestion
proc awaitThreadResult*[T](signal: ThreadSignalPtr, handle: Flowvar[T]): Future[?!T] {.async.} =
  await wait(signal)
  success(handle.sync())
```

elcritch: >   `await wait(signal)`
>   `success(handle.sync())`

Unfortunately, this doesn't work properly because it can start running taskpool jobs on the current thread.

What happens is that the tasks have to call `fireSync` before they return and the taskpools worker can put the result into the `Flowvar` and mark it as finished. There's a chance here that the Chronos thread will receive the signal and proceed to call `sync` *before* the Flowvar is marked as completed. If, or probably when, this occurs it'll block the Chronos thread and cause lots of headaches.

That *shouldn't* happen normally, but it's a common race condition with this kinda worker pattern. With high load on a system one I'd expect to happen once in a while.

My preferred way to handle this case is to have our own queue to put the results in it and then call the signal. This way you avoid the race condition entirely.

codex/slots/proofs/backends/asynccircoms.nim
elcritch: It is pretty awkward. I was trying to avoid having to make a new instance on each call as that seems overkill. 

Though, the backend *should* be threadsafe so I could switch it to use one global config since the Rust code clones it anyways.

I'll switch it out with one global config.

elcritch: Though one reason to do this on each worker thread would be to avoid blocking the Chronos thread while CircomCompat loads the zkey. Though that currently runs pretty quickly right?

codex/utils/asyncthreads.nim
dryajov: This only happens for nested tasks, but it might be a problem due to task stealing. 

I would avoid flowvar altogether and just rely on the chronos signal, otherwise we're mixing messaging primitives, even if we're doing `tryComplete/tryRecv`.

> My preferred way to handle this case is to have our own queue to put the results in it and then call the signal. This way you avoid the race condition entirely.

No need, taskpools is already a queue why add yet another just to pass around a (dubiously useful) message?

What I would do is:

- Remove the flowvar
- Add a simple bool flag to the thread args
  - If more complex error conditions need to be communicated, use an enum
- Use chronos signaling
- Log the error in the task itself

codex/slots/proofs/backends/asynccircoms.nim
dryajov: We don't need an instance per thread, internally circom compat should be thread safe. So just create one instance and pass (a pointer to) it to the task on invocation.

> Though one reason to do this on each worker thread would be to avoid blocking the Chronos thread while CircomCompat loads the zkey. Though that currently runs pretty quickly right?

This only happens once on startup and it shouldn't be a big deal. If it becomes an issue, we can move initialization to is't own thread as well.

codex/utils/asyncthreads.nim
elcritch: > This only happens for nested tasks, but it might be a problem due to task stealing.

Yes looks like it'll try work stealing. @tbekas confirmed this previously in his testing for the erasure encoding threading.

> I would avoid flowvar altogether and just rely on the chronos signal, otherwise we're mixing messaging primitives, even if we're doing tryComplete/tryRecv.

The flowvar does provide convenience though. It's unforunate that just using it for returning data still requires checking `tryComplete`. 

> What I would do is:
> Remove the flowvar

If we don't use taskpools flowvar, then we need handle setting up the memory for where to place the proof results. It'd also mean we are doing threading in two different ways in the codebase. 

> Add a simple bool flag to the thread args
> If more complex error conditions need to be communicated, use an enum
> Log the error in the task itself

Using a result type seems simpler than setting up all of that. :)

> Use chronos signaling

This is using the Chronos signaling, same as the erasure coding threading. It's more about how we want to return the data.

elcritch: > No need, taskpools is already a queue why add yet another just to pass around a (dubiously useful) message?

To return the proof data (or erasure encoded data, etc) and a queue is just the canonical way to do that in worker pools. Taskpools uses the two queues pattern as well: the job queue, and the flowvar's. The flowvars are a `ChannelSPSCSingle` queue with fixed size.

My preferred approach would be to replace taskpool's flowvars (output queues) with a thread-async-aware queue that wraps ThreadSignalPtr and a standard locking queue so we don't need to re-invent the wheel each time. 

Then you just pass the results queue to the task and it avoids both the mixed signaling primitives and needing to decide how to allocate the results memory each time.

codex/slots/proofs/backends/asynccircoms.nim
elcritch: > This only happens once on startup and it shouldn't be a big deal. If it becomes an issue, we can move initialization to is't own thread as well.

Okay, yah that avoids a bit of complication.

codex/utils/asyncthreads.nim
tbekas: @dryajov Pattern with `tryComplete` and polling is introduced because you commented previously that it's a suitable solution
https://github.com/codex-storage/nim-codex/pull/716#discussion_r1526619271

I think however that we _can_ do it with `handle sync`, having a small tradeoff in mind that the main thread can be blocked now and then for a short period of time because chronos signal may arrive earlier than taskpool's flowar item on the internal queue. However it should be negligible. WDYT?

I just need to make sure that the main thread is not going to be a part of the worker threads.

PR #814, opened 2024-05-23
Chronos v4 Update (v3 Compat Mode)

gmega:
This PR uses Chronos v4 in v3 compat mode to get us there faster (i.e. without having to bump and resolve all the bugs in our deps). This is an alternative to #724 and would allow us to roll it out more slowly. For this to be properly merged, we need to:

- [x] merge the [v4 branch for nim-codex-dht and make a release](https://github.com/codex-storage/nim-codex-dht/pull/90);
- [x] have the [compat mode fixes merged](https://github.com/status-im/nim-chronos/pull/545) so we don't need to point to an unofficial branch.

We are also currently relying on an unreleased version of Nim 1.6.x, but this should change soon. We can in any case point to `version-1-6` and it will work, we can also pin to the current commit if using an unstable release looks scary.


codex/blockexchange/engine/discovery.nim
```diff
@@ -102,7 +102,7 @@ proc advertiseQueueLoop(b: DiscoveryEngine) {.async.} =
       trace "Begin iterating blocks..."
       for c in cids:
         if cid =? await c:
-          b.advertiseBlock(cid)
+          asyncSpawn b.advertiseBlock(cid)
```
dryajov, line 105: Why `asyncSpawn`?

codex/blocktype.nim
```diff
@@ -31,7 +31,7 @@ import ./codextypes
 export errors, logutils, units, codextypes
 
 type
-  Block* = object of RootObj
+  Block* = ref object of RootObj
```
dryajov, line 34: Why are we changing this to `ref`?

codex/manifest/manifest.nim
```diff
@@ -29,7 +29,7 @@ import ../logutils
 # TODO: Manifest should be reworked to more concrete types,
 # perhaps using inheritance
 type
-  Manifest* = object of RootObj
+  Manifest* = ref object of RootObj
```
dryajov, line 32: Again, why `ref`?

tests/codex/sales/testsales.nim
```diff
@@ -296,7 +296,7 @@ asyncchecksuite "Sales":
                          slot: UInt256,
                          onBatch: BatchProc): Future[?!void] {.async.} =
       let blk = bt.Block.new( @[1.byte] ).get
-      onBatch( blk.repeat(request.ask.slotSize.truncate(int)) )
+      discard await onBatch( blk.repeat(request.ask.slotSize.truncate(int)) )
```
dryajov, line 299: kinda weird that we're discarding the result here, why not handle/log it?

codex/blockexchange/engine/discovery.nim
gmega: If you check how the code was before, this call to `advertiseBlock` actually had nothing before it. I'm not even sure how this was compiling. `asyncspawn` is the closest thing in terms of semantics.

codex/blocktype.nim
gmega: Cause otherwise it crashes with a SIGSEV during testing. :-/ I can try to debug it, but that's probably not a quick one.

codex/manifest/manifest.nim
gmega: Same as before.

codex/blockexchange/engine/discovery.nim
dryajov: Hmm, that is weird... 

`advertiseBlock` pushes the blocks into an async queue with a limited number of entries, this is done for backpressure purposes. An `asyncSpawn` will effectively invalidate the queues backpressure. The correct thing to do is to await it and the `asyncSleep` is most likely unneeded, but probably harmless for now.

tests/codex/sales/testsales.nim
gmega: OK so a bit of context. This code is not even supposed to compile, as `onBatch` returns a `Future` and this is neither used nor discarded. The older versions of Nim seemed to let that pass somehow, but the newer ones will scream and refuse to compile. I then try to infer what this is supposed to do - since no-one seemed to care about the result of `onBatch` on this test, I simply `await discard` it.

codex/blockexchange/engine/discovery.nim
gmega: Well in this case let's fix it then! :-)

tests/codex/sales/testsales.nim
gmega: Btw the reason I `await`ed this one instead of doing an `asyncSpawn` is because I had a similar conversation with @emizzle in the past and he advised me to use `await`. So for marketplace tests that's my default. :-)

dryajov: The fact that it wasn't being detected in previous versions is definitely a Nim/chronos bug. But what I'm sayin here is that we should do proper error handling in case this is returning a `Result`, not simply discarding it. At the very least, the error should be logged.

And I do understand that there are a few of this weird cases or code that was broken/buggy in the original implementation, but we should try to fix this as much as possible as soon as we catch them, larger efforts should be captured as issues as well so that they don't slip through the cracks.

gmega: OK, in this particular case: since `onBatch` returns `Future[?!void]` as its last operation, I'll simply propagate that result upwards. This will make is so that if `onBatch` fails, then `onStore` will also failure, and hopefully the whole test will collapse at that point.

tests/codex/sales/testsales.nim
```diff
@@ -296,8 +296,7 @@ asyncchecksuite "Sales":
                          slot: UInt256,
                          onBatch: BatchProc): Future[?!void] {.async.} =
       let blk = bt.Block.new( @[1.byte] ).get
-      discard await onBatch( blk.repeat(request.ask.slotSize.truncate(int)) )
-      return success()
+      await onBatch( blk.repeat(request.ask.slotSize.truncate(int)) )
```
dryajov, line 299: I think this returns a `Result` object, which needs to be handled.

dryajov: with `if =? ...` or `without res =? ...`

gmega: Since this is a void result object, my thought is that I can just return it as it is from `onBatch` to the caller so they can see if there were errors, if any.

dryajov: Huh, I think this is still hitting the return error? `await onBatch` returns a `Result`, which isn't being handled and nim isn't picking that up?

gmega: Wait, no, not in that case: `await onBatch` will evaluate to `Return[void, ref CatchableError]`, which also happens to be the return type for this async proc (`Future[?!void]`). So what happens is that in case of an error, that error will simply bubble up to the caller of `onBatch` via the result object.

tests/codex/sales/testsales.nim
dryajov: ah, you're right, it's a closure 👍

tests/codex/sales/testsales.nim
dryajov: ah, you're right, it's a closure 👍

dryajov (approved):

markspanbroek (approved):

PR #816, opened 2024-05-28
Bug/sales race

benbierens:
I need help with this one. Please look.

OK, I'll explain.
`preparing.nim` does a `reservations.findAvailability` followed by a `reservations.createReservation`.
It assumes that if availability was found, then the reservation can be created. Unfortunately, both of these are async functions with a descent call-stack. So, plenty of time for interruptions. Additionally, we start 3 workers. These workers pop items from the slot queue and fire up a state machine for each. So, if we push 3 or more items to the queue at once, we can get:

(1) findAvailability -> query -> lots of free space
(2) findAvailability -> query -> lots of free space
(3) findAvailability -> query -> lots of free space
(1) createReservation -> put -> no more free space
(2) createReservation -> put -> BytesOutOfBoundsError `reservations.nim:391` -> transit to Errored state -> test failure
This occurs about 5% to 10% of the time on my machine while using leveldb. It's possible to reproduce with the sqlite backend by strategically placing a sleep.

This issue has always been there. The switch to LevelDB revealed it. The solution is: You can never assume `createReservation` will succeed until afterwards. There's always a chance it will be interrupted at the lowest point in the call-stack.

So the fix is that one if-statement I added to `preparing.nim`. My major issue with this PR is the test code. Because we don't have any convenient way to mock or stub anything, I basically once again had to bootstrap half of Codex in the test in order to even get `preparing`'s run method to run. After that, I realized there was no way to make the availability check pass while making the createReservation fail in the desired way. If we could somehow mock these out with objects we could control from the test, this would have been a 5-minute fix. I decided to add that horrible production-code hook, just so I could get this covered and in a state where I could present it to you. Now I hope you have a better idea for how to test this.

### DIFF
--- a/codex/sales/states/preparing.nim
+++ b/codex/sales/states/preparing.nim
@@ -15,6 +15,8 @@ import ./errored
 
 type
   SalePreparing* = ref object of ErrorHandlingState
+    # I really don't want to keep this. Someone, please help.
+    doThing*: proc(): Future[void] {.async, gcsafe.}
 
 logScope:
   topics = "marketplace sales preparing"
@@ -31,7 +33,7 @@ method onSlotFilled*(state: SalePreparing, requestId: RequestId,
                      slotIndex: UInt256): ?State =
   return some State(SaleFilled())
 
-method run*(state: SalePreparing, machine: Machine): Future[?State] {.async.} =
+method run*(prepareState: SalePreparing, machine: Machine): Future[?State] {.async.} =
   let agent = SalesAgent(machine)
   let data = agent.data
   let context = agent.context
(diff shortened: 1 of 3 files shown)


benbierens:
After this fix was applied, I ran all unit tests 72 times and found no further issues. Doesn't mean it's good. Just means it's probably less bad. :D

emizzle (commented):
One option is obviously to mock Reservations, but that's a decent effort. 

Another option would be to extend the `Reservations` module, to eg `TestReservations`, and override `createReservation`, so that it raises `BytesOutOfBoundsError` explicitly in the test. This would alleviate the need for a callback/hook.

AuHau:
Tackled in https://github.com/codex-storage/nim-codex/pull/825

PR #819, opened 2024-05-31
Disables testsales.nim

benbierens:
This test file is disabled while the race-condition in preparing.nim is being solved.
https://github.com/codex-storage/nim-codex/pull/816


AuHau:
Fixed in https://github.com/codex-storage/nim-codex/pull/825

PR #825, opened 2024-06-03
fix: createReservation lock

AuHau:
Supersedes https://github.com/codex-storage/nim-codex/pull/816/

### DIFF
--- a/codex/sales/reservations.nim
+++ b/codex/sales/reservations.nim
@@ -35,6 +35,7 @@ import std/sequtils
 import pkg/chronos
 import pkg/datastore
 import pkg/nimcrypto
+import pkg/metrics
 import pkg/questionable
 import pkg/questionable/results
 import pkg/stint
@@ -52,6 +53,8 @@ export logutils
 
 logScope:
   topics = "sales reservations"
+
+declareCounter(codex_reservations_availability_missmatch, "codex reservations availability_missmatch")
 
 type
   AvailabilityId* = distinct array[32, byte]
@@ -71,7 +74,8 @@ type
     size* {.serialize.}: UInt256
     requestId* {.serialize.}: RequestId
     slotIndex* {.serialize.}: UInt256
-  Reservations* = ref object
+  Reservations* = ref object of RootObj
+    locks: Table[AvailabilityId, AsyncLock]
     repo: RepoStore
     onAvailabilityAdded: ?OnAvailabilityAdded
   GetNext* = proc(): Future[?seq[byte]] {.upraises: [], gcsafe, closure.}
@@ -95,6 +99,7 @@ const
   SalesKey = (CodexMetaKey / "sales").tryGet # TODO: move to sales module
   ReservationsKey = (SalesKey / "reservations").tryGet
 
+proc hash*(x: AvailabilityId): Hash {.borrow.}
 proc all*(self: Reservations, T: type SomeStorableObject): Future[?!seq[T]] {.async.}
 
 proc new*(T: type Reservations,
@@ -370,54 +375,68 @@ proc createAvailability*(
 
   return success(availability)
 
-proc createReservation*(
+method createReservation*(
   self: Reservations,
   availabilityId: AvailabilityId,
   slotSize: UInt256,
   requestId: RequestId,
   slotIndex: UInt256
-): Future[?!Reservation] {.async.} =
+): Future[?!Reservation] {.async, base.} =
 
-  trace "creating reservation", availabilityId, slotSize, requestId, slotIndex
+  let lock = self.locks.mgetOrPut(availabilityId, newAsyncLock())
+  try:
+    trace "acquiring lock for availibility", availabilityId
+    await lock.acquire()
+    trace "got lock for availibility", availabilityId
 
-  let reservation = Reservation.init(availabilityId, slotSize, requestId, slotIndex)
+    without availabilityKey =? availabilityId.key, error:
+      return failure(error)
 
-  without availabilityKey =? availabilityId.key, error:
-    return failure(error)
+    without availability =? await self.get(availabilityKey, Availability), error:
+      return failure(error)
 
-  without var availability =? await self.get(availabilityKey, Availability), error:
-    return failure(error)
+    # We recheck that the found availability is still matching after we have lock acquired
+    if availability.freeSize < slotSize:
+       # Lets monitor how often this happen and if it is often we can do it more inteligent to handle it
+      codex_reservations_availability_missmatch.inc()
+      warn "we matched availability but then it changed and now it does not match anymore", id = availability.id
 
-  if availability.freeSize < slotSize:
-    let error = newException(
-      BytesOutOfBoundsError,
-      "trying to reserve an amount of bytes that is greater than the total size of the Availability")
-    return failure(error)
+      let error = newException(
+        BytesOutOfBoundsError,
+        "trying to reserve an amount of bytes that is greater than the total size of the Availability")
+      return failure(error)
 
-  if createResErr =? (await self.update(reservation)).errorOption:
-    return failure(createResErr)
+    trace "creating reservation", availabilityId, slotSize, requestId, slotIndex
 
-  # reduce availability freeSize by the slot size, which is now accounted for in
-  # the newly created Reservation
-  availability.freeSize -= slotSize
+    let reservation = Reservation.init(availabilityId, slotSize, requestId, slotIndex)
 
-  # update availability with reduced size
-  if updateErr =? (await self.update(availability)).errorOption:
+    if createResErr =? (await self.update(reservation)).errorOption:
+      return failure(createResErr)
 
-    trace "rolling back reservation creation"
+    # reduce availability freeSize by the slot size, which is now accounted for in
+    # the newly created Reservation
+    availability.freeSize -= slotSize
 
-    without key =? reservation.key, keyError:
-      keyError.parent = updateErr
-      return failure(keyError)
+    # update availability with reduced size
+    if updateErr =? (await self.update(availability)).errorOption:
 
-    # rollback the reservation creation
-    if rollbackErr =? (await self.delete(key)).errorOption:
-      rollbackErr.parent = updateErr
-      return failure(rollbackErr)
+      trace "rolling back reservation creation"
 
-    return failure(updateErr)
+      without key =? reservation.key, keyError:
+        keyError.parent = updateErr
+        return failure(keyError)
 
-  return success(reservation)
+      # rollback the reservation creation
+      if rollbackErr =? (await self.delete(key)).errorOption:
+        rollbackErr.parent = updateErr
+        return failure(rollbackErr)
+
+      return failure(updateErr)
+
+    return success(reservation)
+  finally:
+    if lock.locked():
+      lock.release()
 
 proc returnBytesToAvailability*(
   self: Reservations,
@@ -621,6 +640,7 @@ proc findAvailability*(
         minPrice >= availability.minPrice:
 
         trace "availability matched",
+          id = availability.id,
           size, availFreeSize = availability.freeSize,
           duration, availDuration = availability.duration,
           minPrice, availMinPrice = availability.minPrice,
@@ -635,6 +655,7 @@ proc findAvailability*(
         return some availability
 
       trace "availability did not match",
+        id = availability.id,
         size, availFreeSize = availability.freeSize,
         duration, availDuration = availability.duration,
         minPrice, availMinPrice = availability.minPrice,
--- a/codex/sales/reservations.nim
+++ b/codex/sales/reservations.nim
@@ -75,7 +75,7 @@ type
     requestId* {.serialize.}: RequestId
     slotIndex* {.serialize.}: UInt256
   Reservations* = ref object of RootObj
-    locks: Table[AvailabilityId, AsyncLock]
+    availabilityLock: AsyncLock # Lock for protecting assertions of availability's sizes when searching for matching availability
     repo: RepoStore
     onAvailabilityAdded: ?OnAvailabilityAdded
   GetNext* = proc(): Future[?seq[byte]] {.upraises: [], gcsafe, closure.}
@@ -102,10 +102,23 @@ const
 proc hash*(x: AvailabilityId): Hash {.borrow.}
 proc all*(self: Reservations, T: type SomeStorableObject): Future[?!seq[T]] {.async.}
 
+template withLock(lock, body) =
+  try:
+    trace "Acquiring lock"
+    await lock.acquire()
+    trace "Lock acquired"
+    body
+  finally:
+    if lock.locked:
+      trace "Releasing lock"
+      lock.release()
+      trace "Lock released"
+
+
 proc new*(T: type Reservations,
           repo: RepoStore): Reservations =
 
-  T(repo: repo)
+  T(availabilityLock: newAsyncLock(),repo: repo)
 
 proc init*(
   _: type Availability,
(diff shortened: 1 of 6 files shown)


codex/sales/reservations.nim
```diff
@@ -370,54 +375,68 @@ proc createAvailability*(
...
   self: Reservations,
   availabilityId: AvailabilityId,
   slotSize: UInt256,
   requestId: RequestId,
   slotIndex: UInt256
-): Future[?!Reservation] {.async.} =
+): Future[?!Reservation] {.async, base.} =
 
-  trace "creating reservation", availabilityId, slotSize, requestId, slotIndex
+  let lock = self.locks.mgetOrPut(availabilityId, newAsyncLock())
```
benbierens, line 386: How sure are we that this is the only function where we need to acquire/release a lock for an availabilityId? It seems there are other places in this module that interact with both the availabilities and the reservations.

tests/codex/sales/states/testpreparing.nim
```diff
@@ -27,3 +73,28 @@ suite "sales state 'preparing'":
...
+    check !next of SaleIgnored
+
+  test "run switches to downloading when reserved":
+    createAvailability()
+    let next = await state.run(agent)
+    check !next of SaleDownloading
+
+  test "run switches to ignored when reserve fails with BytesOutOfBounds":
+    createAvailability()
+    reservations.setCreateReservationThrowBytesOutOfBoundsError(true)
```
benbierens, line 97: Yes! Now we're talking.

codex/sales/reservations.nim
AuHau: Good point, will recheck.

AuHau: I have fixed those places that touch Availability's sizes. Please have a look.

codex/sales/reservations.nim
```diff
@@ -312,31 +322,36 @@ proc deleteReservation*(
   without key =? key(reservationId, availabilityId), error:
     return failure(error)
 
-  without reservation =? (await self.get(key, Reservation)), error:
-    if error of NotExistsError:
-      return success()
-    else:
-      return failure(error)
+  let lock = self.locks.mgetOrPut(availabilityId, newAsyncLock())
```
benbierens, line 325: same here, it looks like

codex/sales/reservations.nim
```diff
@@ -429,47 +458,51 @@ proc returnBytesToAvailability*(
     reservationId
     availabilityId
 
+  let lock = self.locks.mgetOrPut(availabilityId, newAsyncLock())
```
benbierens, line 461: One more

codex/sales/reservations.nim
```diff
@@ -429,47 +458,51 @@ proc returnBytesToAvailability*(
...
-    return failure(updateErr)
+      return failure(updateErr)
 
-  return success()
+    return success()
+  finally:
+    if lock.locked():
+      lock.release()
 
 proc release*(
```
benbierens, line 507: I suspect `release` should also have a lock. It seems then every function will have a lock if they do this flow:
- Read + deserialize
- Modify object
- Serialize and write/update

codex/sales/reservations.nim
```diff
@@ -233,10 +238,44 @@ proc update*(
   without key =? obj.key, error:
     return failure(error)
 
-  without oldAvailability =? await self.get(key, Availability), err:
-    if err of NotExistsError:
-      let res = await self.updateImpl(obj)
-      # inform subscribers that Availability has been added
+  let lock = self.locks.mgetOrPut(obj.id, newAsyncLock())
```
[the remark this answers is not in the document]
AuHau: lol, damn me 🙈 I am really blind 😅 Thanks for the catch!

codex/sales/reservations.nim
AuHau: Yes, but `release` is about Reservation and not Availabilities. The locks I am introducing here focus on locking for changes of sizes in Availabilities as that is the main thing that is changing. Reservation is always locked already to one Storage Request and one context, there should not be any concurrence here.

codex/sales/reservations.nim
```diff
@@ -53,6 +54,8 @@ export logutils
 logScope:
   topics = "sales reservations"
 
+declareCounter(codex_reservations_availability_missmatch, "codex reservations availability_missmatch")
```
markspanbroek, line 57:
```suggestion
declareCounter(codex_reservations_availability_mismatch, "codex reservations availability_mismatch")
```

codex/sales/reservations.nim
```diff
@@ -233,10 +238,47 @@ proc update*(
...
+        if onAvailabilityAdded =? self.onAvailabilityAdded:
+          # when chronos v4 is implemented, and OnAvailabilityAdded is annotated
+          # with async:(raises:[]), we can remove this try/catch as we know, with
+          # certainty, that nothing will be raised
+          try:
+            await onAvailabilityAdded(obj)
+          except CancelledError as e:
+            raise e
+          except CatchableError as e:
+            # we don't have any insight into types of exceptions that
```
markspanbroek, line 259: I know that this was already part of the original code, but I couldn't resist commenting on this.

I'm usually more in favor of a hard crash (raise Defect) than hiding the error in the logs. The intent here is that the callback should not raise any exceptions, so if it does then we'd like to know about it as soon as possible.

codex/sales/reservations.nim
```diff
@@ -370,54 +391,68 @@ proc createAvailability*(
...
-    return failure(error)
+    without availability =? await self.get(availabilityKey, Availability), error:
+      return failure(error)
 
-  without var availability =? await self.get(availabilityKey, Availability), error:
-    return failure(error)
+    # We recheck that the found availability is still matching after we have lock acquired
+    if availability.freeSize < slotSize:
+       # Lets monitor how often this happen and if it is often we can do it more inteligent to handle it
+      codex_reservations_availability_missmatch.inc()
```
markspanbroek, line 417:
```suggestion
      codex_reservations_availability_mismatch.inc()
```

tests/codex/sales/states/testpreparing.nim
```diff
@@ -27,3 +73,28 @@ suite "sales state 'preparing'":
   test "switches to filled state when slot is filled":
     let next = state.onSlotFilled(request.id, slotIndex)
     check !next of SaleFilled
+
+  proc createAvailability() =
+    let a = waitFor reservations.createAvailability(
```
markspanbroek, line 78: Why not `await` instead of `waitFor`?

tests/codex/sales/testreservations.nim
```diff
@@ -254,6 +254,17 @@ asyncchecksuite "Reservations module":
     check updated.isErr
     check updated.error of BytesOutOfBoundsError
 
+  test "cannot release more bytes than size of reservation - concurrency test":
```
markspanbroek, line 257: Where's the concurrency in this test?

codex/sales/reservations.nim
```diff
@@ -250,40 +292,11 @@ proc update*(
...
-        # `onAvailabilityAdded` can raise because it is caller-defined
-        warn "Unknown error during 'onAvailabilityAdded' callback",
-          availabilityId = obj.id, error = e.msg
-
-  return res
+
+    return res
+  finally:
+    if lock.locked():
+      lock.release()
```
markspanbroek, line 299: You can probably neatly wrap this in a template, e.g.:

```nim
template withLock(lockExpression, body) =
  let lock = lockExpression
  try:
    await lock.acquire()
    body
  finally:
    if lock.locked:
      lock.release()

```

codex/sales/reservations.nim
```diff
@@ -71,7 +74,8 @@ type
     size* {.serialize.}: UInt256
     requestId* {.serialize.}: RequestId
     slotIndex* {.serialize.}: UInt256
-  Reservations* = ref object
+  Reservations* = ref object of RootObj
+    locks: Table[AvailabilityId, AsyncLock]
```
markspanbroek, line 78: I'm not a big fan of this construction; this table will keep growing and elements are never removed. Wouldn't it make more sense to add a `lock` field to the `Availability` type?

markspanbroek (commented):
Thanks @AuHau, nice work! Besides a few small comments, I would like to see the lock table removed, if possible.

codex/sales/reservations.nim
AuHau: Yeah you are right that it will never be removed, that should be fixed, but I am not sure how the `lock` field would work here, because `Availability` is created from the JSON datastore backend... Every time you call `reservations.get()` you will get new object and unless we would persist the lock, which IMHO is bad idea, then you would get unsynced locks across the Availability's objects. Sooo... instead I would suggest removing the entry upon deleting availability, which BTW we still do not support 🤪 

Btw. this table will be very small as there will be generally very low number of Availabilities ;-)

tests/codex/sales/states/testpreparing.nim
AuHau: 🤷‍♂️ I "stole" this originally from @benbierens work 😅 @benbierens why?

codex/sales/reservations.nim
AuHau: Yeah, this is not the scope of the PR ;-) If you want create please issue and I will fix it up in follow up PR...

codex/sales/reservations.nim
markspanbroek: Yes, we should definitely not persist the lock :grinning: But I think that the need to introduce this table indicates that the architecture of Reservations is not working for this use case. The table feels like a workaround, and instead of making the workaround better (by deleting entries etc.) I'd rather we take a good look at the architecture and fix it.

The nim garbage collector is very good at removing things that are no longer needed, so that's why I was thinking of adding the lock to a garbage collected object. But as you pointed out, you get a fresh object (and therefore a fresh lock) every time you try to perform an operation on an availability. And I currently don't see a good way to have unique objects in memory for each availability.

So I'd like to suggest the following: remove the table of locks and replace it with a single lock. This makes it very simple to reason about. The only downside is that you cannot update two availabilities in parallel, but that is ok because they'll simply be processed one after the other.

AuHau: > So I'd like to suggest the following: remove the table of locks and replace it with a single lock. This makes it very simple to reason about. The only downside is that you cannot update two availabilities in parallel, but that is ok because they'll simply be processed one after the other.

When I started reading your first comment, I thought you would suggest this, so I already thought about it.

Honestly, I don't see why you consider this a bad workaround. From my POV this seems like a legitimate solution, that is trying to create the least obstruction to the concurrency that we have thanks to Eric's Slot Queue. And having a single lock for all the availabilities might hinder this quite a bit. I would understand your concern if we would be talking about a Table with potentially thousands of records, but realistically we will have tens of Availabilities and hence this Table will be very small.

tests/codex/sales/testreservations.nim
AuHau: Haha, yeah nowhere 🙈 That is what happens when you do not finish the work before the weekend... You forgot where you left 😅

tests/codex/sales/states/testpreparing.nim
benbierens: I did that?... I don't know. I'm sure I stole it from somewhere as well.

AuHau: The "copy&paste" mafia 😂

codex/sales/reservations.nim
```diff
@@ -221,20 +239,19 @@ proc updateImpl(
 
   return success()
 
-proc update*(
-  self: Reservations,
-  obj: Reservation): Future[?!void] {.async.} =
-  return await self.updateImpl(obj)
-
-proc update*(
+proc updateAvailability(
```
AuHau, line 242: I needed a non-locking way to update availability for functions that already have acquired a lock in order to avoid deadlocks. I thought about adding a flag "to lock or not" to the `update()` function, but since it is a public function, I suppose it should always lock externally. If you have any other suggestions on how to make this cleaner, please feel free to let me know.

tests/codex/helpers/mockreservations.nim
```diff
@@ -0,0 +1,34 @@
...
+type
+  MockReservations* = ref object of Reservations
+    createReservationThrowBytesOutOfBoundsError: bool
+
+func new*(
+    T: type MockReservations,
+    repo: RepoStore
+): MockReservations =
+  ## Create a mock clock instance
+  {.cast(noSideEffect).}:
```
AuHau, line 15: I am not sure why the compiler was complaining about `newAsyncLock` having side-effects here but not in the `Reservations.new()` 😳 I have solved this with this workaround, but not sure if this is really the solution... Can somebody let me know please?

emizzle: Chance `func` to `proc`.

`func` is an alias for `proc {.noSideEffect.}`

AuHau: Ohhhh fu** me 🙈 I missed that it is `func` 🙈 Thanks!

emizzle: ⌨️🧠🪑

AuHau: Haha yeah, today is that sort of day... 🙄

codex/sales/reservations.nim
emizzle: When is a lock not needed?

AuHau: Everywhere where `updateAvailability` is used 🤪 
But for example in `deleteReservation` and `createReservation`... These have already the lock acquired...

markspanbroek (approved):
:+1:

codex/sales/reservations.nim
```diff
@@ -95,12 +99,26 @@ const
...
+  try:
+    trace "Acquiring lock"
+    await lock.acquire()
+    trace "Lock acquired"
+    body
+  finally:
+    if lock.locked:
+      trace "Releasing lock"
+      lock.release()
+      trace "Lock released"
```
benbierens, line 115: Hmm, I think and I hope we don't really need these 4 trace statements. Since I'm frequently running lots of cluster tests and looking at GBs of log files already, I propose to remove them until I'm proven wrong.

benbierens (approved):
Small comment, still approve!

codex/sales/reservations.nim
AuHau: I see your point. Would scoping those lock logs to topic `locks` be enough for you to filter them out?

codex/sales/reservations.nim
```diff
@@ -370,54 +398,61 @@ proc createAvailability*(
...
+    without availabilityKey =? availabilityId.key, error:
+      return failure(error)
 
-  let reservation = Reservation.init(availabilityId, slotSize, requestId, slotIndex)
+    without availability =? await self.get(availabilityKey, Availability), error:
+      return failure(error)
 
-  without availabilityKey =? availabilityId.key, error:
-    return failure(error)
+    # We recheck that the found availability is still matching after we have lock acquired
```
emizzle, line 416: I think this is a bit misleading in the sense that it is only checked for size, and not "rechecked" for other qualifications (eg profitability). Plus this only makes sense if coming from `sales/states/pending`, which is currently the only callsite, but may be more in the future. Maybe more clear like, 
```suggestion
    # Check that the found availability has enough free space after the lock has been acquired, to prevent asynchronous Availiability modifications
```

codex/sales/reservations.nim
```diff
@@ -370,54 +398,61 @@ proc createAvailability*(
...
+    if availability.freeSize < slotSize:
+       # Lets monitor how often this happen and if it is often we can make it more inteligent to handle it
+      codex_reservations_availability_mismatch.inc()
+      warn "we matched availability but then it changed and now it does not match anymore", id = availability.id
```
emizzle, line 420: Why is this a warning? I don't think it's bad if the lock works as intended. The log should also be logged at the callsite as we don't know where it came from (technically). 

And don't forget to capitalize the start 😝

codex/sales/states/preparing.nim
```diff
@@ -78,7 +78,16 @@ method run*(state: SalePreparing, machine: Machine): Future[?State] {.async.} =
     request.id,
     data.slotIndex
   ), error:
+    trace "Creation of reservation failed"
+    # Race condition:
+    # reservations.findAvailability (line 64) is no guarantee. You can never know for certain that the reservation can be created until after you have it.
+    # Should createReservation fail because there's no space, we proceed to SaleIgnored.
+    if error of BytesOutOfBoundsError:
+      return some State(SaleIgnored())
```
emizzle, line 86: We should add the metric and log here, however again, I don't think it's bad if we hit this exception. The slot will go back in the queue and it'll get re-processed when more availability is added.

tests/codex/sales/testreservations.nim
```diff
@@ -148,6 +152,39 @@ asyncchecksuite "Reservations module":
...
+      )
+
+      let oneResult = await one
+      let twoResult = await two
+
+      check oneResult.isErr or twoResult.isErr
+      if oneResult.isErr:
+        check oneResult.error of BytesOutOfBoundsError
+      if twoResult.isErr:
+        check twoResult.error of BytesOutOfBoundsError
```
emizzle, line 179: i think chronos has a better primitive for this: `allFinished`

Not saying you should do this, but was writing out a different idea and thought it would be worth mentioning another way to achieve the same:
```nim
import std/sugar

let results = allFinished(one, two)
check results.any(r => r.error of BytesOutOfBoundsError)
check results.all(r => r.isOk or r.error of BytesOutOfBoundsError)
```

emizzle (approved):
LGTM! just a couple of small comments about the comments :)

tests/codex/sales/testreservations.nim
AuHau: Ohhh right, I was trying make it work with `allFinishedThrowing` and `allFinishedResult`, but it had its problems and kind of forgot the simplest of these `allFinished` 😅 Thanks! Will update...

codex/sales/reservations.nim
benbierens: That could work, but mostly when I'm using the logs I don't really need detail down the individual locks and operations. It's much more useful to see what is being done instead of the details of how.

codex/sales/reservations.nim
AuHau: Okey, lets keep the metric then. I want it in order to see how often this will be happening as we could optimize this further in the future...

codex/sales/reservations.nim
AuHau: I will remove it for now. These types of locks are usually very handy when you are trying to detect dead-locks though...

Also, I completely understand the pain of the logs, but IMHO this should be handled with proper log levels and separation into topics, that can be filtered out based on the need. We might not be using these appropriately that much. For example for "It's much more useful to see what is being done instead of the details of how." IMHO the `info` level should be used, while `trace` should be more low-level implementation notification (like doing locks).

tests/codex/sales/testreservations.nim
AuHau: Well, FYI this does not work. Mainly because `allFinished` will return sequence of futures and it does not "unwrap" the underlying result as you (and I was as well) expected. I have again tried to make this work for some time without success so I will just leave it as is.

emizzle: you need to use `.value` i think to get the underlying value, eg 

```nim
import std/sugar

let results = allFinished(one, two)
check results.any(r => r.value.error of BytesOutOfBoundsError)
check results.all(r => r.value.isOk or r.value.error of BytesOutOfBoundsError)
```

PR #827, opened 2024-06-04
Add Release workflow

### LINKED ISSUE #749: Add Release workflow for Codex
As a part of the Testnet launch we should publish early releases and it is required to prepare a releasee workflow for GitHub Actions.

We already discussed some options
1. [softprops/action-gh-release](https://github.com/softprops/action-gh-release)
     - https://github.com/status-im/nimbus-eth2/blob/stable/.github/workflows/release.yml
2. [googleapis/release-please-action](https://github.com/googleapis/release-please-action)
     -  https://github.com/ethersphere/bee-js/pull/674
     - https://github.com/ethersphere/bee-js/releases/tag/v4.0.0
3. [ncipollo/release-action](https://github.com/ncipollo/release-action)
4. [release-drafter/release-drafter](https://github.com/release-drafter/release-drafter)
5. [marvinpinto/action-automatic-releases](https://github.com/marvinpinto/action-automatic-releases)

We also should consider results from the [Branch for testnet #747](747).

After release we also should publish releases for supported architectures/platforms and GitHub support is the following
| OS/Architecture | amd64          | arm64                                                                                                                           |
| --------------- | -------------- | --------------------------------------------------------------------------------------------------------------------------- |
| Linux           | `ubuntu-22.04` | [public beta](https://github.blog/changelog/2024-06-03-actions-arm-based-linux-and-windows-runners-are-now-in-public-beta/) |
| Windows         | `windows-2022` | [public beta](https://github.blog/changelog/2024-06-03-actions-arm-based-linux-and-windows-runners-are-now-in-public-beta/) |
| macOS           | `macos-13`     | `macos-14`                                                                                                                  |

- We may consider to use [BuildJet](https://buildjet.com) for Linux arm64 builds.
- We may consider to use [self-hosted runners](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/about-self-hosted-runners) for Windows arm64 builds

veaceslavdoina:
PR adds a simple release workflow which will build binaries and attach them to the release
- Update Rust requirements in the Building guide
- It was required to adjust composite action [nimbus-build-system](https://github.com/codex-storage/nim-codex/tree/master/.github/actions/nimbus-build-system) to add Linux arm64 support
- In a composite action we also install specific Rust version, because BuildJet one is out of date
- Small refactoring of the composite action to make code more readable
- Release workflow added
   - Use all free available architectures on GitHub Actions
   - For Linux ARM use BuildJet
   - Include version in the binaries names
   - Pack binaries to minimize the size
   - Experimenting with the uploading binaries to the Codex Testnet nodes

> **Note:** Window arm64 runners will be available on GitHub for open source projects [by the end of the year](https://github.blog/changelog/2024-06-03-actions-arm-based-linux-and-windows-runners-are-now-in-public-beta/).

```shell
./codex-v0.0.1-darwin-arm64 --version
Codex version:  v0.0.1
Codex revision: 60e3e00
Nim Compiler Version 1.6.14 [MacOSX: arm64]
```

<details>
<summary>screenshots</summary>

**Workflow execution and timings**
Screenshot 2024-06-04 at 19 35 45

**Artifacts**
Screenshot 2024-06-04 at 19 36 52
</details>

Closes #749

### DIFF
--- a/.github/actions/nimbus-build-system/action.yml
+++ b/.github/actions/nimbus-build-system/action.yml
@@ -12,14 +12,23 @@ inputs:
   nim_version:
     description: "Nim version"
     default: "version-1-6"
+  rust_version:
+    description: "Rust version"
+    default: "1.78.0"
   shell:
     description: "Shell to run commands in"
     default: "bash --noprofile --norc -e -o pipefail"
 runs:
   using: "composite"
   steps:
-    - name: APT (Linux amd64)
-      if: inputs.os == 'linux' && inputs.cpu == 'amd64'
+    - name: Rust (Linux)
+      if: inputs.os == 'linux'
+      shell: ${{ inputs.shell }} {0}
+      run: |
+        curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs/ | sh -s -- --default-toolchain=${rust_version} -y
+
+    - name: APT (Linux amd64/arm64)
+      if: inputs.os == 'linux' && (inputs.cpu == 'amd64' || inputs.cpu == 'arm64')
       shell: ${{ inputs.shell }} {0}
       run: |
         sudo apt-fast update -qq
@@ -77,15 +86,15 @@ runs:
           printf "'%s'" "$quoted"
         }
 
-        if [[ '${{ inputs.cpu }}' == 'amd64' ]]; then
-          PLATFORM=x64
-        else
-          PLATFORM=x86
-        fi
+        case '${{ inputs.cpu }}' in
+          'amd64') PLATFORM=x64 ;;
+          'arm64') PLATFORM=x64 ;;
+          'i386')  PLATFORM=x86 ;;
+        esac
         echo "PLATFORM=${PLATFORM}" >> ${GITHUB_ENV}
 
-        # Stack usage on Linux amd64
-        if [[ '${{ inputs.os }}' == 'linux' && '${{ inputs.cpu }}' == 'amd64' ]]; then
+        # Stack usage on Linux amd64/arm64
+        if [[ '${{ inputs.os }}' == 'linux' && ('${{ inputs.cpu }}' == 'amd64' || '${{ inputs.cpu }}' == 'arm64')]]; then
           NIMFLAGS="${NIMFLAGS} -d:limitStackUsage"
           echo "NIMFLAGS=${NIMFLAGS}" >> $GITHUB_ENV
         fi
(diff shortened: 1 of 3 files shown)


.github/actions/nimbus-build-system/action.yml
```diff
@@ -77,15 +86,15 @@ runs:
...
         }
 
-        if [[ '${{ inputs.cpu }}' == 'amd64' ]]; then
-          PLATFORM=x64
-        else
-          PLATFORM=x86
-        fi
+        case '${{ inputs.cpu }}' in
+          'amd64') PLATFORM=x64 ;;
+          'arm64') PLATFORM=x64 ;;
```
markspanbroek, line 91: This looks a bit weird, setting the PLATFORM variable to `x64` on an ARM build. It is used to set ARCH_OVERRIDE in nimbus-build-system, and that only checks [whether it equals `x86` or not](https://github.com/search?q=repo%3Astatus-im%2Fnimbus-build-system%20arch_override&type=code). So setting it to `x64` works, but it's not great. Perhaps we can drop this PLATFORM variable all together, and only set ARCH_OVERRIDE when we're on i386?

BUILDING.md
```diff
@@ -34,6 +34,9 @@ On a bare bones installation of Debian (or a distribution derived from Debian, s
 
 ```shell
 apt-get update && apt-get install build-essential cmake curl git rustc cargo
+
+# curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs/ | sh -s -- --default-toolchain=1.76.0 -y
```
markspanbroek, line 38: Why is this commented out? 

Perhaps we can add an explanation here to say that if your apt doesn't install the correct version of rust, that you can then install it using this command? Or perhaps just point to the Rust documentation?

markspanbroek (approved):
Nicely done @veaceslavdoina! I have only a few minor comments.

BUILDING.md
veaceslavdoina: That story is started from the point that we relied to the distribution provided Rust packages and it worked well initially.

Recently we discovered that builds will fail if Rust version < 1.76.0. And I've just updated version constrains in the guide and added a way to install a specific. It is commented out, because is considered just a helper command when distribution package does not provide a required version.

Converted it to a part of the guide.

.github/actions/nimbus-build-system/action.yml
veaceslavdoina: I was not able to find why in general we need `ARCH_OVERRIDE` and who is using it, to set a right value for arm64. I saw just that distinguish between `i386/x64` and ended up with the moving arm64 to x64 side.

Now we set `ARCH_OVERRIDE` only for `i386` and otherwise it will be empty and not passed.

.github/workflows/release.yml
```diff
@@ -0,0 +1,145 @@
...
+            arc="${1}"
+            sha256sum "${arc}" >"${arc}.sha256"
+
+            # Upload to Codex
+            if [[ "${{ env.upload_to_codex }}" == "true" ]]; then
+              codex_endpoints="${{ secrets.CODEX_ENDPOINTS }}"
+              codex_username="${{ secrets.CODEX_USERNAME }}"
+              codex_password="${{ secrets.CODEX_PASSWORD }}"
+
+              for endpoint in ${codex_endpoints}; do
```
gmega, line 113: We're also pushing those into the Codex network? Why? Without purchasing storage those will be GC'ed as soon as the receiving node runs out of allowance or the blocks hit their expiry?

veaceslavdoina: This was an early experiment, like IPFS does - disabled for now
```yaml
upload_to_codex: false
```

gmega (approved):
LGTM

PR #829, opened 2024-06-06
Fix/spec intsizes

benbierens:
Uint64 doesn't work in every client. It doesn't produce an error either, which is why I didn't notice it earlier. Sorry we have to do this again.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -289,6 +289,7 @@ components:
           description: "Root hash of the content"
         originalBytes:
           type: integer
+          format: int64
           description: "Length of original content in bytes"
         blockSize:
           type: integer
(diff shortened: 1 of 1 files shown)


tbekas (approved):

openapi.yaml
```diff
@@ -289,7 +289,7 @@ components:
           description: "Root hash of the content"
         originalBytes:
           type: integer
-          format: uint64
+          format: int64
           description: "Length of original content in bytes"
         blockSize:
```
tbekas, line 294: This property doesn't have format specified. Should it stay this way?

benbierens: I did this on purpose. I don't think we will ever have a block size larger than int32.max.

markspanbroek (approved):

openapi.yaml
tbekas: Ok, but clients consuming API don't know that and they'll not learn it from this spec.

PR #832, opened 2024-06-13
Allow compiler version to be pinned consistently

gmega:
This PR introduces the ability to pin a compiler version (major or minor) so that we get consistent builds locally, in docker, and in CI.

The compiler version can be overridden by setting the `NIM_COMMIT` env variable as before to either a compiler version, branch, or commit hash, but we also introduce the following special semantics for `NIM_COMMIT`:

* if unset, uses the pinned version;
* if set to `pinned`, uses the pinned version (so that CI can specify it wants the pinned version in a matrix);
* if set to `nimbusbuild`, uses the version defined by the nimbus-build-system.

In case you need more convincing, the original reasoning for doing this [can be found here](https://hackmd.io/2dp6Q2tmQzWoKN-dIPckTQ).

### DIFF
--- a/.github/workflows/nim-matrix.yml
+++ b/.github/workflows/nim-matrix.yml
@@ -6,7 +6,7 @@ on:
 
 env:
   cache_nonce: 0 # Allows for easily busting actions/cache caches
-  nim_version: pinned, v1.6.14, v1.6.16, v1.6.18
+  nim_version: pinned, v1.6.16, v1.6.18
 
 jobs:
   matrix:
--- a/.github/workflows/nim-matrix.yml
+++ b/.github/workflows/nim-matrix.yml
@@ -6,7 +6,7 @@ on:
 
 env:
   cache_nonce: 0 # Allows for easily busting actions/cache caches
-  nim_version: v1.6.14, v1.6.16, v1.6.18
+  nim_version: pinned, v1.6.14, v1.6.16, v1.6.18
 
 jobs:
   matrix:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ on:
 
 env:
   cache_nonce: 0 # Allows for easily busting actions/cache caches
-  nim_version: v1.6.14
+  nim_version: pinned
 
 
 concurrency:
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@
 #
 # If NIM_COMMIT is set to "nimbusbuild", this will use the
 # version pinned by nimbus-build-system.
-PINNED_NIM_VERSION := v1.6.18
+PINNED_NIM_VERSION := v1.6.14
 
 ifeq ($(NIM_COMMIT),)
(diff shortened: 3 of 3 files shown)


veaceslavdoina:
So, we switched for CI from `v1.6.14` to the `v1.6.18` which is set in the _Makefile_.

And in case of the _nim-matrix.yml_ we have 4 versions where `v1.6.18` is used twice - pinned and explicitly
Screenshot 2024-06-13 at 20 37 26

gmega:
Hey, sorry, I should've said it: those are sort of put there as examples. I'm going to pin to `v1.6.14` so that it behaves as it does today. Assuming I do that, do you see any bad side effects from making this change?

veaceslavdoina:
It looks like a #680 successor.

I would say it should work and to be able to pass a custom version to the CI Docker builds, workflows should be adjusted - that is not required probably right now.

veaceslavdoina (approved):

.github/workflows/nim-matrix.yml
```diff
@@ -6,7 +6,7 @@ on:
 
 env:
   cache_nonce: 0 # Allows for easily busting actions/cache caches
-  nim_version: v1.6.14, v1.6.16, v1.6.18
+  nim_version: pinned, v1.6.16, v1.6.18
```
dryajov, line 9: I think the latest version in nimbus-build-system is 1.16.21, we probably want to use that one instead of 1.6.16.

dryajov: also, where is pinned defined?

gmega: Yeah sure, it's just that I think the current master won't even compile with 1.6.21. I'm going to bump it once we get the v4 merge in.

gmega: The actual pin is [defined here](https://github.com/codex-storage/nim-codex/pull/832/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R18).

veaceslavdoina: Quick experiment on `master` (`check-nim-v1.6.20`)
Screenshot 2024-06-14 at 21 19 38

Screenshot 2024-06-14 at 21 19 06

gmega: OK good. :-) Anyhow, I'm going to bump it when we merge the v4 PR.

gmega: Anyhow, this is somewhat beside the point: I'll bump it when we merge the Chronos v4 PR anyway or it won't work. This PR is about the pinning. :-)

veaceslavdoina: Just run a CI with v1.6.20 and looks like it works, but as you mentioned it is a different story

Screenshot 2024-06-14 at 23 56 27

markspanbroek (approved):

dryajov (approved):

PR #833, opened 2024-06-13
Remove stale and misleading dependency version metadata

gmega:
This PR removes dependency metadata which was spread in a number of locations but didn't reflect either:

1. the current list of dependencies;
2. their versions.

The lock files were completely stale/wrong, and the nimble file was even worse. The only thing that reflects our dependencies are the versions or commits set into the `/vendor` folder. If we want to have a more readable version of that, then we should probably write a script to extract them, and then perhaps run it before every commit.


benbierens (approved):
Though I have dreamt of one day building Codex with nimble or something similar, I do agree it's good to not have duplicate and/or misleading version information floating around in the repository.

markspanbroek (approved):

PR #834, opened 2024-06-14
config flag to enable CORS response headers for data downloads

emizzle:
This is needed when downloading data from a browser, eg the [ethcc-demo](https://github.com/codex-storage/ethcc-demo)

### DIFF
--- a/codex/rest/api.nim
+++ b/codex/rest/api.nim
@@ -107,6 +107,8 @@ proc retrieveCid(
       await stream.close()
 
 proc initDataApi(node: CodexNodeRef, repoStore: RepoStore, router: var RestRouter) =
+  let allowedOrigin = router.allowedOrigin # prevents capture inside of api defintion
+
   router.rawApi(
     MethodPost,
     "/api/codex/v1/data") do (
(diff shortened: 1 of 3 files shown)


codex/rest/api.nim
```diff
@@ -166,6 +168,12 @@ proc initDataApi(node: CodexNodeRef, repoStore: RepoStore, router: var RestRoute
           Http400,
           $cid.error())
 
+      if corsOrigin =? allowedOrigin:
```
gmega, line 171: OK for now, in the future we should probably think of a CORS config mechanism for presto so we don't have to manually set per-endpoint like this (maybe we can tackle [this issue](https://github.com/status-im/nim-presto/issues/11)).

gmega (approved):
Good for immediate needs, we should probably think of a follow up on presto to pull this from here in the long run.

codex/rest/api.nim
```diff
@@ -166,6 +168,12 @@ proc initDataApi(node: CodexNodeRef, repoStore: RepoStore, router: var RestRoute
...
+        resp.setHeader("Access-Control-Allow-Origin", corsOrigin)
+        resp.setHeader("Access-Control-Allow-Methods", "GET, OPTIONS")
+        resp.setHeader("Access-Control-Headers", "X-Requested-With")
+        resp.setHeader("Access-Control-Max-Age", "86400")
+
```
markspanbroek, line 176: This is repeated twice, can we extract this to a proc, or a template?

markspanbroek (approved):

PR #838, opened 2024-06-20
Fix StoreStream so it doesn't return parity bytes 

gmega:
This PR fixes `StoreStream` so it doesn't return parity bytes when one attempts to download a file by the CID of a protected of verifiable manifest. It also drops the `pad` option which was not used anywhere.


dryajov (approved):

emizzle (approved):
LGTM! 👏

benbierens (approved):

Issue #821, opened 2024-06-03
[BUG] Node crashes during erasure coding


tbekas:
**Describe the bug**
Bug found by @emizzle.

Found an issue that crashes the node, and it could be that I'm doing something wrong, but thought I'd ask anyway.

This is on commit `4312e5c` , the last commit before the expiry changes (from timestamp to duration) were made.

**To Reproduce**
1. Upload a file (CID = `zDvZRwzkvz9boDSsTb3YLoe2PEWtwSHpa7CiSe8YRtjXQTBUZtJy`)
2. Create a request for storage via rest api:
```json
{
  "reward": "400",
  "duration": "1000",
  "proofProbability": "1",
  "collateral": "200",
  "expiry": "1716430952", // in future
  "nodes": 3,
  "tolerance": 1
}
```
3. `StorageRequested` contract event is fired, with CID = `zDvZRwzm7iW2yfwcVv9br3NZFppmvYqwTLzd2GvH9PaTaCdS9VmS`

4. Attempt to download the dataset using REST API: `/data/zDvZRwzm7iW2yfwcVv9br3NZFppmvYqwTLzd2GvH9PaTaCdS9VmS/network`

Then codex crashes with: 
```shell
ERR 2024-05-23 15:18:26.863+10:00 Unhandled exception in async proc, aborting topics="codex" tid=11685603 msg="index -1 not in 0 .. 0"
```
I've isolated the issue to this line in the erasure decoding, but I'm not entirely sure WHY it went wrong (`ecK` is greater than `pos`):
```nim
if idx >= encoded.rounded:
      trace "Retrieved parity block"
      shallowCopy(parityData[pos - encoded.ecK], if blk.isEmpty: emptyBlock else: blk.data) # <== FAILS HERE with pos=1, encoded.ecK=2, parityData[].len=1
      parityPieces.inc
```
https://github.com/codex-storage/nim-codex/blob/4312e5ca361c91b99508b4d4968e641a57e33a02/codex/erasure/erasure.nim#L222

Could someone lend a hand with this please?

**Expected behavior**
Erasure coding succeeds and client app continues to operate normally.

**Environment:**
 - Tag/Commit 4312e5c

benbierens:
Captured in dist-test: https://github.com/codex-storage/cs-codex-dist-tests/blob/74ca512f92955ac914996b3194d8ac123fb5678a/Tests/CodexTests/BasicTests/MarketplaceTests.cs#L89

[...]

PR #839, opened 2024-06-20
Fix verifiable manifest initialization

gmega:
This PR ensures that the interleaving strategy for protected manifests gets passed onto its derived verifiable manifest. This is one of the reasons why we were seeing decoding failures with verifiable manifests in #821.


tests/codex/testmanifest.nim
```diff
@@ -74,3 +74,28 @@ checksuite "Manifest":
...
+      strategy = SteppedStrategy
+    )
+    verifiable = Manifest.new(
+      manifest = protected,
+      verifyRoot = Cid.example,
+      slotRoots = @[Cid.example, Cid.example]
+    ).tryGet()
+
+  test "Should preserve interleaving strategy for protected manifest in verifiable manifest":
+    check verifiable.protectedStrategy == SteppedStrategy
```
emizzle, line 101: Souls were also check the default is LinearStrategy in the case that the protected manifest has not set a strategy?

emizzle (approved):

benbierens (approved):

tbekas (approved):

tests/codex/testmanifest.nim
gmega: Good point

codex/node.nim
```diff
@@ -559,8 +561,8 @@ proc onStore(
 
     return success()
 
-  without indexer =? manifest.protectedStrategy.init(
-    0, manifest.numSlotBlocks() - 1, manifest.numSlots).catch, err:
+  without indexer =? manifest.verifiableStrategy.init(
```
dryajov, line 564: This only needs to download the slot's blocks, not all the blocks.

gmega: It's not downloading all the blocks, it's slicing the required blocks from the space of all blocks.

dryajov (approved):

PR #840, opened 2024-06-25
Add verify_circuit=true to the releases

veaceslavdoina:
As we discussed with @benbierens and @gmega, it make sense to add `-d:verify_circuit=true` to the releases
> This way we can check if proofs are invalid when they are created, or whether the validation fails on-chain

This is a quick PR to adjust release workflow for that.


gmega (approved):
LGTM

AuHau (approved):

PR #841, opened 2024-06-25
Provisional fix so EC errors do not crash the node on download

gmega:
This PR **hacks** Codex so that it catches defect on erasure code decode and stops the node from crashing. This is not an acceptable fix for the issue, and should be removed as soon as we fix the actual bug.

### DIFF
--- a/codex/node.nim
+++ b/codex/node.nim
@@ -240,14 +240,14 @@ proc streamEntireDataset(
   self: CodexNodeRef,
   manifest: Manifest,
   manifestCid: Cid,
-): ?!LPStream =
+): Future[?!LPStream] {.async.} =
   ## Streams the contents of the entire dataset described by the manifest.
   ##
   trace "Retrieving blocks from manifest", manifestCid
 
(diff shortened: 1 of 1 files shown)


veaceslavdoina (approved):

dryajov (approved):

codex/node.nim
```diff
@@ -240,14 +240,14 @@ proc streamEntireDataset(
...
-): ?!LPStream =
+): Future[?!LPStream] {.async.} =
   ## Streams the contents of the entire dataset described by the manifest.
   ##
   trace "Retrieving blocks from manifest", manifestCid
 
   if manifest.protected:
     # Retrieve, decode and save to the local store all EС groups
-    proc erasureJob(): Future[void] {.async.} =
+    proc erasureJob(): Future[?!void] {.async.} =
```
emizzle, line 250: Once we get the chronos v4 changes in, it'll pertinent to ensure that we raise nothing from this proc a la `{.async: (raises: []).}`

PR #844, opened 2024-06-26
Fix verifiable manifest constructor

tbekas:


emizzle:
The integration test also needs to be imported so it's called in `testIntegration`

gmega (approved):

tests/integration/testecbug.nim
```diff
@@ -10,7 +10,7 @@ marketplacesuite "EC bug":
     NodeConfigs(
       # Uncomment to start Hardhat automatically, typically so logs can be
       # inspected locally
-      hardhat: HardhatConfig().withLogFile().some,
+      hardhat: HardhatConfig.none,
```
dryajov, line 13: Probably just better to comment out, as it says above?

tbekas: I copied it from existing configuration in other test. If you want I can try with commenting it out.
https://github.com/codex-storage/nim-codex/blob/5b131611acb632e0ef4554996f9b774b534ff91b/tests/integration/testmarketplace.nim#L89

dryajov: Ok, I guess either way works. But when we enable this, we need to disable to starting hardhat from the CI, because I confirmed that its whats causing the failed tests - at least, locally.

dryajov (approved):

tests/integration/testecbug.nim
```diff
@@ -0,0 +1,65 @@
+from pkg/libp2p import Cid, init
+import ../examples
+import ./marketplacesuite
+import ./nodeconfigs
+import ./hardhatconfig
+
+marketplacesuite "EC bug":
```
dryajov, line 7: Hmm... Can we rename this something more appropriate? What EC bug, what does it actually test, etc?

tests/integration/testecbug.nim
```diff
@@ -0,0 +1,65 @@
...
+    let expiry = 5.periods
+    let data = await RandomChunker.example(blocks=8)
+    let client = clients()[0]
+    let clientApi = client.client
+
+    let cid = clientApi.upload(data).get
+
+    var requestId = none RequestId
+    proc onStorageRequested(event: StorageRequested) {.raises:[].} =
+      echo "storage requested event"
```
dryajov, line 41: looks like testing leftover.

dryajov:
Still failing 🤔

tests/integration/testecbug.nim
gmega: Removed

tests/integration/testecbug.nim
gmega: I just deleted the whole thing. The integration suite doesn't work with this right now, so I don't think it makes sense to keep it here.

tests/integration/testecbug.nim
```diff
@@ -4,14 +4,10 @@ import ./marketplacesuite
 import ./nodeconfigs
 import ./hardhatconfig
 
-marketplacesuite "EC bug":
+marketplacesuite "Bug #821 - node crashes during erasure coding":
```
dryajov, line 7: We should never name anything after a bug, this is most likely the wrong place to put things altogether and should be moved.

gmega: Yeah, I don't like it either, it completely breaks convention. This test probably belongs with API tests or whatnot. Will open an issue so we do it later.

dryajov: yeah, it should be part of the node tests in `node/testnode.nim` or similar.

gmega: #846

PR #848, opened 2024-06-27
feat(rest): adds erasure coding constraints when requesting storage

emizzle:
During our testing for the workshop, we noticed there were many instances arises from lack of constraints about dataset size and erasure coding parameters, that ultimately caused issues erasure decoding once the requests for storage had been created. Some of these constraints are due to limitations of Leopard, the underlying lib used for erasure coding. Attempting to erasure decode datasets that had been poorly constructed, but accepted as valid by the Codex node, causes quite a few crashes.

While this PR only addresses constraining the parameters for the REST API, further measures may be needed in the contracts.

### DIFF
--- a/codex/erasure/erasure.nim
+++ b/codex/erasure/erasure.nim
@@ -85,6 +85,10 @@ type
 
   ErasureError* = object of CodexError
   InsufficientBlocksError* = object of ErasureError
+    # Minimum size, in bytes, that the dataset must have had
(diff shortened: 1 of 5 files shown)


benbierens (approved):
Yes.

gmega (approved):
I've added some more feedback to the error message on small files, and will now approve my own changes :grin:

dryajov (approved):

PR #849, opened 2024-06-27
Workaround for SIGSEGV failures when using `Iter[Natural]`

tbekas:
Same issue is not happening when using `AsyncIter[Natural]`, however I've added a test case to `testasynciter.nim` anyway, to ensure that it won't start happening in the future.


benbierens (approved):
Yes.

tbekas:
Not relevant anymore. Bug can be fixed by bumping Nim version.

Issue #852, opened 2024-07-02
Prover fails to build valid proofs when `ecK` coincides with the number of blocks in the file.


gmega:
The prover will currently not generate valid proofs when `ecK` is equal to the number of blocks in the file. The following test case (to be patched onto [`testprover.nim`](https://github.com/codex-storage/nim-codex/blob/master/tests/codex/slots/testprover.nim)) illustrates the issue:

```nim
  test "Should generate valid proofs when k = ecK":
    let
      (_, _, verifiable) =
        await createVerifiableManifest(
          store,
          2,
          2, 1,
          DefaultBlockSize,
          DefaultCellSize)

    let
      r1cs = "tests/circuits/fixtures/proof_main.r1cs"
      wasm = "tests/circuits/fixtures/proof_main.wasm"
      circomBackend = CircomCompat.init(r1cs, wasm)
      prover = Prover.new(store, circomBackend, 5)
      challenge = 1234567.toF.toBytes.toArray32
      (inputs, proof) = (await prover.prove(1, verifiable, challenge)).tryGet

    check:
      (await prover.verify(proof, inputs)).tryGet == true
```

PR #853, opened 2024-07-02
Prover workshop band-aid

gmega:
Adds an API band-aid to avoid triggering #852.

### DIFF
--- a/codex/node.nim
+++ b/codex/node.nim
@@ -423,6 +423,15 @@ proc setupRequest(
     trace "Unable to fetch manifest for cid"
     return failure error
 
+  # ----------------------------------------------------------------------------
+  # FIXME this is a BAND-AID to address
+  #   https://github.com/codex-storage/nim-codex/issues/852 temporarily for the
+  #   workshop. Remove this once we get that fixed.
+  if manifest.blocksCount.uint == ecK:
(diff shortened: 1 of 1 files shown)


codex/node.nim
```diff
@@ -423,6 +423,15 @@ proc setupRequest(
     trace "Unable to fetch manifest for cid"
     return failure error
 
+  # ----------------------------------------------------------------------------
+  # FIXME this is a BAND-AID to address
+  #   https://github.com/codex-storage/nim-codex/issues/852 temporarily for the
+  #   workshop. Remove this once we get that fixed.
+  if manifest.blocksCount.uint == ecK:
+    return failure("Cannot setup slots for a dataset with ecK == numBlocks. Please use a larger file.")
```
emizzle, line 431:
```suggestion
    return failure("Cannot setup slots for a dataset with ecK == numBlocks. Please use a larger file or a different combination of `nodes` and `tolerance`.")
```

emizzle (approved):

tbekas (approved):

benbierens:
This is going to annoy a few people during the workshop. But, it would annoy them more of we don't have this.

benbierens (approved):

Issue #854, opened 2024-07-03
[BUG] Erasure decoding fails with `Original tree root differs from the tree root computed out of recovered data`


tbekas:
**Description**
Erasure coding fails non-deterministically to decode the data. It seems that the probability of failure is high when block size is ~16KiB or greater. In unit tests it doesn't fail because block size is hardcoded to 1KiB.

**To Reproduce**
1. Checkout branch [bug/ec-threading](https://github.com/codex-storage/nim-codex/tree/bug/ec-threading)
2. Generate a random dataset `./generate-mb.sh`. This will create a file `test_file.bin`.
3. Run `nim c --threads:on -r tests/codex/rndtesterasure.nim`

It will fail/succeed non-deterministically. So multiple execution of pt. 2 might be necessary.

From the investigation it seems like one of the blocks from the `parity` sequence is copied over to the `data`sequence when thread is spawned for decoding using `decodeTask` from `asyncbackend.nim`.

**Expected behavior**

**Screenshots**

**Environment:**
 - OS: [e.g. MacOs]
 - Version: [e.g. v0.1.2]
 - Arch [e.g. amd64]:

tbekas:
Bug reported in nim-lang
https://github.com/nim-lang/Nim/issues/23886

tbekas:
PR for nim-leopard
https://github.com/codex-storage/nim-leopard/pull/14

gmega:
This has been superseded by #1087

PR #855, opened 2024-07-03
Temporarily disable threading in erasure coding

tbekas:
This disables scheduling erasure-coding on another thread. More details in an issue #854

### DIFF
--- a/tests/codex/testerasure.nim
+++ b/tests/codex/testerasure.nim
@@ -252,3 +252,22 @@ suite "Erasure encode/decode":
       decoded.treeCid == manifest.treeCid
       decoded.treeCid == verifiable.originalTreeCid
       decoded.blocksCount == verifiable.originalBlocksCount
+
+  for i in 1..5:
+    test "Should encode/decode using various parameters " & $i & "/5":
+      let
+        blockSize   = rng.sample(@[1, 2, 4, 8, 16, 32, 64].mapIt(it.KiBs))
+        datasetSize = 1.MiBs
+        ecK         = 10.Natural
+        ecM         = 10.Natural
+
+      let
+        chunker = RandomChunker.new(rng, size = datasetSize, chunkSize = blockSize)
+        manifest = await storeDataGetManifest(store, chunker)
+        encoded = (await erasure.encode(manifest, ecK, ecM)).tryGet()
+        decoded = (await erasure.decode(encoded)).tryGet()
+
+      check:
+        decoded.treeCid == manifest.treeCid
+        decoded.treeCid == encoded.originalTreeCid
+        decoded.blocksCount == encoded.originalBlocksCount
--- a/codex/erasure/asyncbackend.nim
+++ b/codex/erasure/asyncbackend.nim
@@ -114,7 +114,14 @@ proc proxySpawnEncodeTask(
   args: EncodeTaskArgs,
   data: ref seq[seq[byte]]
 ): Flowvar[EncodeTaskResult] =
-  tp.spawn encodeTask(args, data[])
+  # FIXME Uncomment the code below after addressing an issue:
+  # https://github.com/codex-storage/nim-codex/issues/854
+
+  # tp.spawn encodeTask(args, data[])
+
+  let fv = EncodeTaskResult.newFlowVar
+  fv.readyWith(encodeTask(args, data[]))
+  return fv
 
 proc proxySpawnDecodeTask(
   tp: Taskpool,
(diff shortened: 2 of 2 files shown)


tests/codex/testerasure.nim
```diff
@@ -252,3 +252,22 @@ suite "Erasure encode/decode":
...
       decoded.treeCid == verifiable.originalTreeCid
       decoded.blocksCount == verifiable.originalBlocksCount
+
+  for i in 1..5:
+    test "Should encode/decode using various parameters " & $i & "/5":
+      let
+        blockSize   = rng.sample(@[1, 2, 4, 8, 16, 32, 64].mapIt(it.KiBs))
+        datasetSize = rng.sample(@[1, 2].mapIt(it.MiBs))
+        ecK         = rng.sample(@[20].mapIt(it.Natural))
+        ecM         = rng.sample(@[10].mapIt(it.Natural))
```
tbekas, line 262: Unfortunately when debugging issue with threading I discovered another issue with erasure coding (even if 0 blocks are deleted from the store).
```
 Unhandled exception: Unable to encode manifest, not enough blocks, ecK = 20, blocksCount = 16 [InsufficientBlocksError]
```
When using ecK=20 and ecM=10 params, I haven't observed such issue, therefore these collections have only these single elements.

tests/codex/testerasure.nim
```diff
@@ -252,3 +252,22 @@ suite "Erasure encode/decode":
       decoded.treeCid == manifest.treeCid
       decoded.treeCid == verifiable.originalTreeCid
       decoded.blocksCount == verifiable.originalBlocksCount
+
+  for i in 1..5:
+    test "Should encode/decode using various parameters " & $i & "/5":
+      let
+        blockSize   = rng.sample(@[1, 2, 4, 8, 16, 32, 64].mapIt(it.KiBs))
```
benbierens, line 259: I think blocksizes of anything other than 64kb will completely break the proving system. It's definitely OK to pin this one.

tests/codex/testerasure.nim
```diff
@@ -252,3 +252,22 @@ suite "Erasure encode/decode":
       decoded.treeCid == manifest.treeCid
       decoded.treeCid == verifiable.originalTreeCid
       decoded.blocksCount == verifiable.originalBlocksCount
+
+  for i in 1..5:
```
benbierens, line 256: I'm not exactly a fan of random in unit tests. Perhaps we can nest a few for-loops and test several combinations?
```
  for datasetSize in 1..x:
    for ecK in y..x:
```
etc
If nothing else, it'd be cool to see which points in this space are giving us trouble.

benbierens (approved):

tests/codex/testerasure.nim
tbekas: I think that fuzzy testing is a pragmatic approach to the problems with software failing on unexpected input (case that we do observe).

Of course there's more to it, like we should print rng seed whenever a failure is detected so that an error can be reproduced deterministically. Unfortunately this require more library kind of work to get it working properly.

tests/codex/testerasure.nim
tbekas: Ok, current test suite uses hardcoded 1kb block size. I think we should test at least a few different combinations in this unit test to make sure that EC works regardless of the block size.

gmega (approved):
LGTM!

tests/codex/testerasure.nim
bkomuves: Yes, the circuit is hardcoded for 64k blocks (we hash 2kb cells, then build a depth 5 Merkle tree to get the block hash. This was done so that the proving system is "compatible" with network-level structure. We then prove 2kb cells against the slot hash.

tests/codex/testerasure.nim
bkomuves: This is not very surprising, given that we do the network-level EC on the block level. If there are less blocks than `ecK`, it breaks. Not sure what's the flow here, but can't you include this constraints in the above randomly generated test parameters?

tests/codex/testerasure.nim
bkomuves: I think both "golden" test and randomized tests are important. But maybe they should be separated into two categories? (I have very good experience with randomized testing)

tests/codex/testerasure.nim
tbekas: The problem is that in this scenario there are 0 blocks deleted, so it should always bee N blocks available (ecK + ecM) for decoding. Regardless of this, an error shown above occurs sometimes.

bkomuves: It seems to me that the problem is with encoding, not decoding (that's what the quoted error message says?). Since there are less than K blocks, you cannot split it into K pieces. Though padding should solve that...

tbekas: You're right, I've been focused on decoding so much lately that I immediately assume it's about decoding. Indeed the error is trivial and expected.

PR #859, opened 2024-07-17
Fixes prover behavior with singleton proof trees

### LINKED ISSUE #852: Prover fails to build valid proofs when `ecK` coincides with the number of blocks in the file.

gmega:
fixes #852 

This PR adds a test and updated circuit assets which fix the bug in #852. The issue was that the convention for handling single-block trees was not entirely correct at the circuit, [which has now been fixed](https://github.com/codex-storage/codex-storage-proofs-circuits/commit/6e9fcc632c9599679bca8788c9c4221eabdb5e21).

The PR also contains updates that should make it easier to generate valid JSON Circom inputs from Codex tests, which should make the process of writing tests to debug circuits simpler.


tbekas (approved):
LGTM

codex/slots/proofs/backends/circomcompat.nim
```diff
@@ -51,25 +83,18 @@ proc release*(self: CircomCompat) =
 
 proc prove*[H](
```
dryajov, line 84: should this be public?

gmega: Guess it doesn't have to. Will make it private.

codex/slots/sampler/utils.nim
```diff
@@ -80,7 +69,7 @@ func cellIndices*(
   numCells: Natural, nSamples: Natural): seq[Natural] =
 
   var indices: seq[Natural]
-  while (indices.len < nSamples):
-    let idx = cellIndex(entropy, slotRoot, numCells, indices.len + 1)
-    indices.add(idx.Natural)
+  for i in 1..nSamples:
+    indices.add(cellIndex(entropy, slotRoot, numCells, i))
```
dryajov, line 73: This changes the logic by passing a different `counter` to `cellIndex`, this looks like a breaking change?

tests/codex/slots/testprover.nim
```diff
@@ -24,52 +24,69 @@ import ./backends/helpers
...
           store,
-          numDatasetBlocks,
-          ecK, ecM,
+          8, # number of blocks in the original dataset (before EC)
+          5, # ecK
+          3, # ecM
           blockSize,
           cellSize)
 
-  teardown:
```
dryajov, line 60: Why remove this?

gmega: This is just a weird diff, the teardown is still there.

codex/slots/proofs/backends/circomcompat.nim
gmega: Fixed!

codex/slots/sampler/utils.nim
dryajov: Ah, you're right. The previous logic is quite confusing indeed.

dryajov (approved):
LGTM

codex/slots/sampler/utils.nim
```diff
@@ -47,10 +38,8 @@ func floorLog2*(x : int) : int =
   return k
 
 func ceilingLog2*(x : int) : int =
-  if (x == 0):
-    return -1
-  else:
-    return (floorLog2(x-1) + 1)
+  doAssert ( x > 0 )
```
emizzle, line 41: Might be a good idea to provide a reason string here so if we do encounter this defect, we can see something that’ll give us a clue as to what went wrong. Otherwise, we might not even get a stack trace to help.

emizzle: Same with line 32

2-towns:
Do we still need https://github.com/codex-storage/nim-codex/blob/1f49f86131fbc9721fa1d136705a1aec4601572e/codex/node.nim#L432-L436 @gmega ?

gmega:
@2-towns yes :flushed:

2-towns:
And for this one (indirectly related), it was specific for the workshop: 

https://github.com/codex-storage/nim-codex/blob/1f49f86131fbc9721fa1d136705a1aec4601572e/codex/node.nim#L266-L270

@gmega

PR #863, opened 2024-07-30
Adds log to dialPeer

benbierens:
This will help us see who is initiating connection with who during our tests.
(We can already see when a connection is established. But we can't see who initiates it.)

### DIFF
--- a/codex/blockexchange/network/network.nim
+++ b/codex/blockexchange/network/network.nim
@@ -290,6 +290,7 @@ proc dialPeer*(b: BlockExcNetwork, peer: PeerRecord) {.async.} =
     trace "Skipping dialing self", peer = peer.peerId
     return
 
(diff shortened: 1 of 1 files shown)


gmega (approved):
Sounds good

codex/blockexchange/network/network.nim
```diff
@@ -290,6 +290,7 @@ proc dialPeer*(b: BlockExcNetwork, peer: PeerRecord) {.async.} =
     trace "Skipping dialing self", peer = peer.peerId
     return
 
+  trace "Dialing peer...", peer = peer.peerId
```
dryajov, line 293: libp2p already has the same exact log, but no harm I guess :)

benbierens: Yup found it after you mentioned it in the meeting. I'd rather use the libp2p one.

Issue #866, opened 2024-08-01
Upload huge files via chunks 


2-towns:
**Is your feature request related to a problem? Please describe.**
When attempting to upload a large file (6GB) using the SDK (and consequently the REST API) within the browser, the browser becomes unresponsive. To address this issue, I attempted to optimize the process by using a Web Worker, believing that it was caused by an overload on the UI thread; however, the problem persists.

**Describe the solution you'd like**
For the client, a viable solution would be to upload the file by segmenting it into smaller chunks (for instance, 10 MB per chunk), and sending individual requests for each chunk. 
The server would need to handle these separate chunks and merge them into the final file. 

Additionally, dividing the file into chunks for upload would enable resumable downloads.

**Describe alternatives you've considered**
Limit the file maximum size to 1.5 GB.   

**Additional context**
Here are some relevant links discussing the process of splitting large files into multiple chunks: 
- https://dev.to/codeparrot/how-google-drive-handles-huge-file-uploads-4c58
- https://docs.librephotos.com/docs/user-guide/mobile/librephotos-mobile/upload

dryajov:
Are you trying to load the entire file into the browser and then initiating the upload or are you using the streaming API?

2-towns:
I tried to load the file into browser, pass it to the web worker (no issue, no freezing) and then initiate the upload with `XMLHttpRequest`: 

```js
  const url = this.url + Api.config.prefix + "/data";

    const xhr = new XMLHttpRequest();

    const promise = new Promise<SafeValue<string>>(async (resolve) => {
      xhr.upload.onprogress = (evt) => {
        if (evt.lengthComputable) {
          onProgress?.(evt.loaded, evt.total);
        }
      };

      xhr.open("POST", url, true);

      xhr.send(new Blob(await Chunks.split(file)));

      xhr.onload = function () {
        if (xhr.status != 200) {
          resolve({
            error: true,
            data: {
              code: xhr.status,
              message: xhr.responseText,
            },
          });
        } else {
          resolve({ error: false, data: xhr.response });
        }
      };

      xhr.onerror = function () {
        resolve({
          error: true,
          data: {
            message: "Something went wrong during the file upload.",
          },
        });
      };
    });
```

The Chunks.split(file) is splitting the file into multiple chunks to avoid the browser complaining about the file size. 
The reason I was using XMLHttpRequest is to have the upload indicator and notify the UI about the progress. 
This way is causing a big freeze into the browser. 

[...]

PR #869, opened 2024-08-06
Add OPTIONS endpoint to allow the content-type header for the upload endpoint

2-towns:
This PR was created from the discussion in #866. 

The browser adds a content-type header when trying to upload the file, but the Codex client does not handle it. So this PR adds an OPTIONS endpoint to add this header and allow the preflight request.

### DIFF
--- a/codex/rest/api.nim
+++ b/codex/rest/api.nim
@@ -110,6 +110,20 @@ proc retrieveCid(
 proc initDataApi(node: CodexNodeRef, repoStore: RepoStore, router: var RestRouter) =
   let allowedOrigin = router.allowedOrigin # prevents capture inside of api defintion
 
+  router.api(
+    MethodOptions,
+    "/api/codex/v1/data") do (
+       resp: HttpResponseRef) -> RestApiResponse:
+
+      if corsOrigin =? allowedOrigin:
+        resp.setHeader("Access-Control-Allow-Origin", corsOrigin)
+        resp.setHeader("Access-Control-Allow-Methods", "POST, OPTIONS")
+        resp.setHeader("Access-Control-Headers", "X-Requested-With")
+        resp.setHeader("Access-Control-Allow-Headers", "content-type")
+
+      resp.status = Http204
+      await resp.sendBody("")
+
   router.rawApi(
     MethodPost,
     "/api/codex/v1/data") do (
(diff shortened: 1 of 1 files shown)


codex/rest/api.nim
```diff
@@ -110,6 +110,20 @@ proc retrieveCid(
...
+  router.api(
+    MethodOptions,
+    "/api/codex/v1/data") do (
+       resp: HttpResponseRef) -> RestApiResponse:
+
+      if corsOrigin =? allowedOrigin:
+        resp.setHeader("Access-Control-Allow-Origin", corsOrigin)
+        resp.setHeader("Access-Control-Allow-Methods", "POST, OPTIONS")
+        resp.setHeader("Access-Control-Headers", "X-Requested-With")
+        resp.setHeader("Access-Control-Allow-Headers", "content-type")
```
emizzle, line 122: We may also consider directing the browser to [cache the pre-flight response](https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request) so it doesn't need to continually re-issue it for some time:
> The preflight response can be optionally cached for the requests created in the same [URL](https://developer.mozilla.org/en-US/docs/Glossary/URL) using [Access-Control-Max-Age](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Max-Age) header like in the above example. To cache preflight responses, the browser uses a specific cache that is separate from the general HTTP cache that the browser manages. Preflight responses are never cached in the browser's general HTTP cache.

emizzle (commented):
LGTM, well done!

The title is a bit misleading, perhaps "Support CORS preflight requests for the upload endpoint" might be a bit more accurate.

emizzle (approved):
My apologies, I meant to approve the PR.

codex/rest/api.nim
AuHau: Good catch!

codex/rest/api.nim
```diff
@@ -110,6 +110,20 @@ proc retrieveCid(
...
 
+  router.api(
+    MethodOptions,
+    "/api/codex/v1/data") do (
+       resp: HttpResponseRef) -> RestApiResponse:
+
+      if corsOrigin =? allowedOrigin:
+        resp.setHeader("Access-Control-Allow-Origin", corsOrigin)
+        resp.setHeader("Access-Control-Allow-Methods", "POST, OPTIONS")
+        resp.setHeader("Access-Control-Headers", "X-Requested-With")
```
AuHau, line 121: Remove this line please. `Access-Control-Headers` does not do anything.

AuHau (approved):
One small tweak but other then that LGM

codex/rest/api.nim
2-towns: Yes Good catch, I added the "Access-Control-Max-Age" header.

codex/rest/api.nim
2-towns: Ok I removed it

gmega:
@2-towns I'm going to merge this as it looks good to go.

PR #870, opened 2024-08-07
feat: add `--payout-address`

emizzle:
Allows SPs to be paid out to a separate address, keeping their profits secure. Supports https://github.com/codex-storage/codex-contracts-eth/pull/144 in the nim-codex client.


AuHau (approved):
LGTM! Thanks and good work 💪

tests/contracts/testMarket.nim
```diff
@@ -1,30 +1,47 @@
...
+import std/importutils
 import pkg/chronos
+import pkg/ethers/erc20
 import codex/contracts
 import ../ethertest
 import ./examples
 import ./time
 import ./deployment
 
+privateAccess(OnChainMarket) # enable access to private fields
```
markspanbroek, line 11: 🟠 It doesn't look like you need access to private fields of `OnChainMarket` in this test? Am I missing something?

markspanbroek (approved):
:+1:

tests/contracts/testMarket.nim
emizzle: Good eye, Mark! `market.rewardRecipient` is used in `switchAccount`, which is not an exported field.

PR #872, opened 2024-08-08
refactor(marketplace): generic querying of historical marketplace events

emizzle:
Allows querying of past Marketplace events using a generic routine.

Also moves marketplace contract events to the Market abstraction so the types can be shared across all modules that call the Market abstraction.

### DIFF
--- a/codex/market.nim
+++ b/codex/market.nim
@@ -28,11 +28,27 @@ type
   OnRequestCancelled* = proc(requestId: RequestId) {.gcsafe, upraises:[].}
   OnRequestFailed* = proc(requestId: RequestId) {.gcsafe, upraises:[].}
   OnProofSubmitted* = proc(id: SlotId) {.gcsafe, upraises:[].}
-  PastStorageRequest* = object
+  ProofChallenge* = array[32, byte]
+
+  # Marketplace events -- located here due to the Market abstraction
+  StorageRequested* = object of Event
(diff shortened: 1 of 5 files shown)


AuHau (approved):
LGTM,thanks!

emizzle:
Going to make a small change to this PR, hence the chance to draft.

AuHau (approved):
LGTM

emizzle:
@AuHau Ready for another review, sorry about that!

codex/market.nim
```diff
@@ -31,6 +31,7 @@ type
   ProofChallenge* = array[32, byte]
 
   # Marketplace events -- located here due to the Market abstraction
+  MarketplaceEvent* = Event
```
AuHau, line 34: Shouldn't then all the events bellow be `of MarketplaceEvent`?

AuHau (approved):
Pretty great improvement with the generic! Nice!

Generally LGTM, only one question, but I will give you approve ;-)

codex/market.nim
emizzle: Good catch, yes, but it is an alias, so that's why the compiler didn't complain. The main reason for doing this was so that implementations using the abstraction don't also need to import `pkg/ethers`. Will fix.

markspanbroek (approved):
:+1:

PR #878, opened 2024-08-14
Support CORS preflight requests when the storage request api returns an error 

2-towns:
When trying to create a new storage request with the API, the request can fail if the parameters are invalid due to some rules (for example, parameters must satisfy 1 < (nodes - tolerance) ≥ tolerance).

When the error is returned with the following code, the CORS headers are not included:

```
return RestApiResponse.error(Http400, error.msg)
```

This PR fix this issue and returns the CORS headers when the storage requests api returns an error.


tbekas (approved):
LGTM

markspanbroek (approved):
Nice, thanks!

PR #879, opened 2024-08-15
GCC 14 update

benbierens:
Involves updating our dependencies.
And our dependency dependencies.


gmega:
OK thank you @2-towns and @benbierens.

PR #881, opened 2024-08-21
Workaround for erasure coding threading bug

tbekas:
Resolves: #854


gmega:
I'm closing this as we're going for a different approach in the end.

PR #882, opened 2024-08-22
Rework circuit downloader

benbierens:
We are moving the downloading and unpacking of the circuit files outside of codex into a separate tool. Codex will prompt the user to run the tool when necessary.


codex/slots/proofs/prover.nim
```diff
@@ -86,16 +87,18 @@ proc verify*(
...
 proc new*(
   _: type Prover,
   store: BlockStore,
-  backend: AnyBackend,
-  nSamples: int): Prover =
+  config: CodexConf): ?!Prover =
 
-  Prover(
-    backend: backend,
+  without backend =? initializeBackend(config), err:
```
dryajov, line 97: I'm really not a fan of this pattern, it violates the dependency injection principle, which has the nasty consequence of unnecessarily complicating testing. For example this usualy the main reasons for needing mocks.

It is perfectly fine to initialize the backend externally and then simply pass the required parameters to the constructor of the consumer class.

codex/codex.nim
```diff
@@ -272,32 +272,7 @@ proc new*(
...
-              msg: "zkey file not readable, doesn't exist or wrong extension (.zkey)")
-
-          $config.circomZkey
-        else: ""
-
-      some Prover.new(
-        store,
-        CircomCompat.init($config.circomR1cs, $config.circomWasm, zkey),
-        config.numProofSamples)
+      some Prover.new(store, config).expect("Unable to create prover.")
```
dryajov, line 275: Just use the factory to create the backend here and then pass it into the `Prover`, it will make your life much easier.

tools/cirdl/cirdl.nim
```diff
@@ -0,0 +1,91 @@
+import std/os
+import pkg/chronicles
+import pkg/chronos
+import pkg/ethers
+import pkg/questionable
+import pkg/questionable/results
+import pkg/zip/zipfiles
```
dryajov, line 7: You don't need this and their external dependencies, there is already https://github.com/status-im/nim-zlib which we link against either way already.

BUILDING.md
```diff
@@ -157,6 +157,22 @@ In Bash run
...
+
+Unix
+```shell
+$ apt-get update && apt-get install libzip-dev
+```
+
+MacOS
+```shell
+brew install libzip
+```
```
dryajov, line 174: there is no need for this with https://github.com/status-im/nim-zlib

dryajov (changes requested):
Overall a step in the right direction. Lets fix the `Prover` initialization and switch to using `nim-zlib`. Ideally, we should also use either `presto` or chronos `httpclient` instead of curl, but we can leave that for later.

tools/cirdl/cirdl.nim
```diff
@@ -0,0 +1,91 @@
...
+  "https://circuit.codex.storage/proving-key/" & hash
+
+proc downloadZipfile(url: string, filepath: string): ?!void =
+  try:
+    # Nim's default webclient does not support SSL on all platforms.
+    # Not without shipping additional binaries and cert-files... :(
+    # So we're using curl for now.
+    var rc = execShellCmd("curl -o " & filepath & " " & url)
+    if not rc == 0:
+      return failure("Download of '" & url & "' failed with return code: " & $rc)
```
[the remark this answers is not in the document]
benbierens: Makes sense. I'll look into switching out curl if there's time after the more critical issues.

codex/slots/proofs/prover.nim
benbierens: I agree the hard call to `initializeBackend` is a problem. Putting all the details about the backend in codex.nim is a problem too. I'll try attaching this proc to a type and pass it to prover.new that way. We can have both the dependency injected and we can confine the knowledge about the proving system inside the slots/proofs files.

codex/codex.nim
benbierens: Did that. Prover.new no longer needs to be a result-type.

tools/cirdl/cirdl.nim
benbierens: It would be great to drop this dependency, it's quite annoying... but nim-zlib (and zlib-gzip in general) doesn't do what we need here. The artifact we download is a zip with multiple files. nim-zlib doesn't handle those. So I suppose we can A) rewrite the artifact producing pipeline to a generate .tar.gz file and manually write the code to implement tar handling to reconstruct multiple files using nim-zlib. Or B) use zip/zipfiles.
Got any better ideas?

codex/slots/proofs/prover.nim
dryajov: I mean, just passing the backend is probably as good as it gets, I'm not sure why you'd want to pass in the factory instead, it doesn't seem to make anything "better"?

tools/cirdl/cirdl.nim
dryajov: How are this artifacts produced? The more straightforward thing to do is to use tar and gz, option A sounds more reasonable.

benbierens: Well, keep in mind that if we switch to a tar.gz file instead of a zip, nim-zlib doesn't support the tar format. It just does the gzip data streams. We would have to pull in something like https://github.com/guzba/zippy or implement the spec ourselves, or wrap something that implements it.

benbierens: (I also learned that "Nim Tar" is a star wars character.... :D )

AuHau: Switching to any other "bundling format" for the circuit assets is easy. One-line change here: https://github.com/codex-storage/codex-storage-proofs-circuits/blob/master/.github/workflows/generate.yml#L140

So pick what is best suited for your needs.

tools/cirdl/cirdl.nim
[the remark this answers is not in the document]
benbierens: Replaced curl with chronos httpsession. (Tried presto first, it didn't want to.)

benbierens:
This passes marketplace dist-test.

codex/slots/proofs/backendfactory.nim
```diff
@@ -0,0 +1,88 @@
...
+import pkg/questionable
+import pkg/confutils/defs
+import pkg/stew/io2
+import pkg/ethers
+
+import ../../conf
+import ./backends
+import ./backendutils
+
+type
```
dryajov, line 14: This is nitpicking a little bit, but you really don't need this placeholder, just attach the factory to the config object?

codex/slots/proofs/backendutils.nim
```diff
@@ -0,0 +1,12 @@
+import ./backends
+
+type
```
dryajov, line 3: Again, not sure why we need this placeholder types?

codex/slots/proofs/backendfactory.nim
benbierens: All right now I see what you mean. Yes this is better. 👍

benbierens:
Downloader is now reworked to use tar.gz instead of zip.
Using status-im/zippy library: https://github.com/status-im/nim-zippy
New circuit assets have been generated: https://github.com/codex-storage/codex-storage-proofs-circuits/actions/runs/10941033813
Codex Contracts repository PR for updated hash: https://github.com/codex-storage/codex-contracts-eth/pull/178

TODO:
- Once contracts update is merged, pull new master into nim-codex
- Redeploy marketplace for testnet with updated contract
- Adjust testnet scripts/docs with new marketplace address

codex/slots/proofs/prover.nim
```diff
@@ -86,16 +87,15 @@ proc verify*(
...
   ## Returns a future that resolves to a proof.
-
   self.backend.verify(proof, inputs)
 
 proc new*(
   _: type Prover,
   store: BlockStore,
   backend: AnyBackend,
-  nSamples: int): Prover =
+  config: CodexConf): Prover =
```
dryajov, line 96: Just pass the number of samples instead of the config object, there is no need for it here.

codex/conf.nim
```diff
@@ -126,6 +127,13 @@ type
       abbr: "d"
       name: "data-dir" }: OutDir
 
+    circuitDir* {.
```
dryajov, line 130: There are specific parametters to indicate where each circuit file is located, we should repoint the defaults to this directory and/or remove the individual circuit paths.

tools/cirdl/cirdl.nim
```diff
@@ -5,7 +5,7 @@ import pkg/chronos
 import pkg/ethers
 import pkg/questionable
 import pkg/questionable/results
-# import pkg/zippy/ziparchives
+import pkg/zippy/ziparchives
```
dryajov: cool, thats much better.

codex/slots/proofs/prover.nim
dryajov: in other words, revert back to original.

codex/conf.nim
benbierens: I have also moved this option into the `prover` switch. 👍

dryajov (approved):
Lgtm

codex.nim
```diff
@@ -67,6 +67,9 @@ when isMainModule:
     # permissions are insecure.
     quit QuitFailure
 
+  if config.prover() and not(checkAndCreateDataDir((config.circuitDir).string)):
```
gmega, line 70: Minor nitpick: it looks a bit weird that we create dir for circuit when we expect it to already be there (clouds intent a bit IMHO).

PR #886, opened 2024-08-26
API tweaks for OpenAPI, errors and endpoints

AuHau:
This is a bit of bundle PR with several smaller things. If you would really like, I could split these up, but... I don't want to 🙈

Changes:

 - Refactor variable naming `slot` -> `slotIndex`
 - Changing error `Sales unavailable`/`Purchasing unavailable` to `Persistence is not enabled` as that is better descriptive to what is actually going on to the user.
 - Tweaking the `docs/TwoClientTest.md` to be working again.
 - Extending `/sales/slots/{slotId}` endpoint to include `reservation` and `request` as well.
 - Fixing up `openapi.yaml` to be accurate again, with extending some validation enums


dryajov:
I like that we're adding more examples, but I wouldn't remove the simple no persistence example, because it adds more unnecessary steps to the flow. We also have the marketplace tutorial, which has a more sophisticated setup outlined.

codex/rest/api.nim
```diff
@@ -453,8 +456,11 @@ proc initPurchasingApi(node: CodexNodeRef, router: var RestRouter) =
         without params =? StorageRequestParams.fromJson(body), error:
           return RestApiResponse.error(Http400, error.msg, headers = headers)
 
-        let nodes = params.nodes |? 1
-        let tolerance = params.tolerance |? 0
+        let nodes = params.nodes |? 2
+        let tolerance = params.tolerance |? 1
+
+        if tolerance == 0:
+          return RestApiResponse.error(Http400, "Tolerance needs to be bigger then zero", headers = headers)
```
emizzle, line 463: It feels like this should be in its own PR. At the very least let's add some comments around it (maybe just copy pasta from the discord discussion).

emizzle (approved):
LGTM 👏

emizzle:
> but I wouldn't remove the simple no persistence example

Actually i think this is a good point, and maybe we could link to the marketplace example from this simple example instead, eg "This basic example explores connecting two client nodes, but a more in depth example with storage providers can be found in the Marketplace example (link)". Or, alternatively, add the "persistent" TwoNodesClient example after the simple example, but it may be a bit redundant with the Marketplace example...

emizzle:
Since @AuHau is away, I'll pull out the tutorial changes and keep the rest so we can get this merged for the release.

Issue #457, opened 2023-06-26
Parameter for Validator to validate only part of the Request-space


AuHau:
On our meetup, we discussed how to ensure that there is at least one validator "watching" every request. We have not come up with some decentralized structure for this for now, but in order to move on, we decided to do "centralized coordination" where we will provide validators that will cover the requirement for "at least one validator per request". 

But to ensure we can scale, we must introduce a parameter defining the part of "Request-space" that the validator should observe and validate.

marcinczenko:
Looking at the discussion in From the discussion in issue https://github.com/codex-storage/codex-contracts-eth/pull/123 and with helpful insights from Mark, here is some more context on how the issue could be attacked.

So, first, referring to https://github.com/codex-storage/codex-contracts-eth/pull/123, my understanding is that the current course of action is **Slot buckets, Local** and the most important reason is that we are currently looking how to save on gas, and keeping the validator buckets functionality off-chain has clear benefits in this aspect. Moreover, maybe also important is that the marketplace contract does not have to be extended and redeployed when tweaking the validator functionality wrt to the _stratification_ of slot-id space.

> Tried to find some verb for "bucketing". Found _stratification_. It is used in the context of logic programming, and maybe it matches what we are doing here: the buckets can overlap and they do not necessarily cover the entire slot-id space. Or just _partitioning_. Yet, it seems that partitioning is slightly better for non-overlapping subsets with complete coverage. Maybe you have some good ideas...

### Some More Context

A slot identifier, or a slot-id, being a `byte32` hash can be treated as `uint256` number. Thus, the size of the slot-id space is `2^256`.

 Having one central validator would not only create a single point of failure, but with the increasing number of active, filled slots, one validator may be not capable of serving all those slots, especially that the time given to the validator to validate the slots it observes is constrained by `config.proofs.timeout`.
 
 Thus, it is good for the Codex storage network to have many validators, run by the users, ideally in a _permissionless_ model.

Currently, by providing the `--validator` option when starting the Codex client, the Codex node operator indicates that the given Codex node should also perform as a validator. On top of that, the `--validator-max-slots` option lets the operator to constrain the number of slot that the validator will validate. 

Although the `--validator-max-slots` is helpful in limiting the load on the validator, it is somehow limiting in providing the node operator a more fine-grained control of which stratum of the whole space-id the validator wants to cover.

### Covering the slot-id space

Let's start with a simple example.
[...]

PR #890, opened 2024-09-02
Validator - support partitioning of the  slot id space

marcinczenko:
This PR, addresses issue https://github.com/codex-storage/nim-codex/issues/457 (please refer to the issue discussion for more context info so that we do not have to duplicate it all here).

To shortly summarise issue https://github.com/codex-storage/nim-codex/issues/457: we want to be able to tell the validator what partition of the whole slot id space it should observe and help validating.

To achieve that, we provide two new CLI/config parameters:

- `validator-partition-size` - to indicate number equal partitions dividing the whole slot id space,
- `validator-partition-index` - to indicate the index of the portion observed by the validator.

The validator will only consider a slot with slot id `slotId` for validation if it matches the following criteria:

```
slotId % validatorPartitionSize == validatorPartitionIndex
```

> We decided to keep the `--validator-max-slots` option to additionally limit the number of slots that will be observed by the validator for the chosen slot id partition.


codex/conf.nim
```diff
@@ -293,6 +293,22 @@ type
...
+          "The value must be greater than 0"
+        defaultValue: 1
+        name: "validator-partition-size"
+      .}: int
+      
+      validatorPartitionIndex* {.
+        desc: "Slot validation partition index"
+        longDesc: "The value provided must be in the range [0, partitionSize). " &
+          "Only slot ids satisfying condition [(slotId mod partitionSize) == partitionIndex] will be observed by the validator"
+        defaultValue: 0
```
emizzle, line 308: What happens if I want to validate all the slots in the slot Id space, and let the `--validator-max-slots` limit the number (ie the current behaviour of the node prior to this PR)? In my opinion, we should use an `Option[int]` (or `Option[uint16]`, see below) for this type with `int.none` being default which indicates the node will validate all slots (and be limited by `--validator-max-slots`).

emizzle (commented):
Overall, great job! 👏 

I have a few (detailed) comments, mostly because I had already gone through the mental hurdles of this task in the past, so I knew of some nuances.

codex/conf.nim
```diff
@@ -293,6 +293,22 @@ type
         name: "validator-max-slots"
       .}: int
 
+      validatorPartitionSize* {.
```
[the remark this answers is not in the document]
marcinczenko: I was trying to find some good name (you can check the original issue, where there is some discussion about it, maybe you did not see it). I am also fine with `bucket` or `group` (`bucket` sounds very Amazonish though). So, `group` or `bucket`? We can also discuss shortly tomorrow.

> I also think "size" is a bit misleading here.

Yes, same here. Indeed, `size` parameter is a bit indirect. By saying how many partitions will be considered, it indirectly indicates the size of the partition. What would be your recommendation for the name of this param?

> the value should be a power of 2

You're right. On the other hand, by limiting it to powers of two, we limit the number of options available to node operator.

codex/conf.nim
```diff
@@ -293,6 +293,22 @@ type
...
+        name: "validator-partition-size"
+      .}: int
+      
+      validatorPartitionIndex* {.
+        desc: "Slot validation partition index"
+        longDesc: "The value provided must be in the range [0, partitionSize). " &
+          "Only slot ids satisfying condition [(slotId mod partitionSize) == partitionIndex] will be observed by the validator"
+        defaultValue: 0
+        name: "validator-partition-index"
+      .}: int
```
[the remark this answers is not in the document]
marcinczenko: I was thinking about the same. Still, from confutes we get an `int` and it does not look like you can change. Thus, the user may still enter a negative number (say `-1`), which when casted to `uint16`, will be `65535`. Thus, type casting will take care about the ranges, but the user can be puzzled. So I would leave `int`s as the args to the initialiser, and use `uint16` in the `ValidationParams` type. Does it make sense?

tests/codex/testvalidation.nim
```diff
@@ -40,13 +60,45 @@ asyncchecksuite "validation":
...
   test "when a slot is filled on chain, it is added to the list":
     await market.fillSlot(slot.request.id, slot.slotIndex, proof, collateral)
     check validation.slots == @[slot.id]
+  
+  test "slot is not observed if it is not in the partition":
+    let validationParams = initValidationParams(maxSlots, partitionSize, partitionIndex + 1)
+    let validation = createValidation(clock, market, validationParams)
+    await validation.start()
+    await market.fillSlot(slot.request.id, slot.slotIndex, proof, collateral)
+    await validation.stop()
```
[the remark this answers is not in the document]
marcinczenko: I am creating a new instance of the validation here because it needs customised validation params. Thus I am overwriting the one already created in setup, and so the `stop` in `teardown` will not work on this one.

codex/conf.nim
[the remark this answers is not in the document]
emizzle: > from confutes we get an int

What do you mean by this? If the parameter type is uint16, then the compiler must enforce it.

tests/codex/testvalidation.nim
```diff
@@ -67,7 +119,11 @@ asyncchecksuite "validation":
...
 
   test "it does not monitor more than the maximum number of slots":
+    let validationParams = initValidationParams(maxSlots, partitionSize = 1, partitionIndex = 0)
+    let validation = createValidation(clock, market, validationParams)
+    await validation.start()
     for _ in 0..<maxSlots + 1:
       let slot = Slot.example
       await market.fillSlot(slot.request.id, slot.slotIndex, proof, collateral)
+    await validation.stop()
     check validation.slots.len == maxSlots
```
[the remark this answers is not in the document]
marcinczenko: It should be covered to some extent by `slot is not observed if it is not in the partition` + other other tests, but I agree it is easier to create a couple of more documented cases by directly testing the functions you mention.

By the way, can you explain this:

image

why does it complain about those two fields.

Also, are we measuring the coverage in the tests themselves?

tests/codex/testvalidation.nim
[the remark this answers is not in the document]
emizzle: Ah good point, didn't notice that.

codex/conf.nim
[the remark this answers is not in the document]
marcinczenko: Sorry, looks like a type checker - "confutils" I meant, but in you later comment it looks like you can specify types of the params in confutils - I will check that...

codex/validation.nim
```diff
@@ -43,13 +46,26 @@ proc waitUntilNextPeriod(validation: Validation) {.async.} =
   trace "Waiting until next period", currentPeriod = period
   await validation.clock.waitUntil(periodEnd.truncate(int64) + 1)
 
+func partitionIndexForSlotId(validation: Validation, slotId: SlotId): int =
+  if (validation.validationParams.partitionSize == 1):
+    0
+  else:
+    let slotIdUInt256 = UInt256.fromBytesBE(slotId.toArray)
```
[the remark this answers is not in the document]
marcinczenko: Yes, I believe it is good. EVM is big endian and here I am explicitly using `fromBytesBE`. The `fromBytes` uses the native (i.e. machine) endianness, which indeed would not be portable. By the way, `stint` v2 changes the default behaviour - it defaults to BE instead of native (machine) endianness, but we are still using the old version of `stint` package. `fromBytesBE` seems to be then a right option.

tests/codex/testvalidation.nim
```diff
@@ -13,18 +14,37 @@ asyncchecksuite "validation":
...
+  
+  func createValidation(clock: Clock, market: Market, validationParams: ValidationParams): Validation =
+    without validation =? Validation.new(clock, market, validationParams), error:
+      raiseAssert fmt"Creating Validation failed! Error msg: {error.msg}"
+    validation
+
+  func partitionIndexForPartitionSize(slot: Slot, partitionSize: int): int =
+    let slotId = slot.id
+    let slotIdUInt256 = UInt256.fromBytesBE(slotId.toArray)
+    (slotIdUInt256 mod partitionSize.u256).truncate(int)
```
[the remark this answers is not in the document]
marcinczenko: Sounds like a good idea.

codex/conf.nim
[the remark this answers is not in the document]
marcinczenko: > Something like --validator-groups for "size" and --validator-group-index for the assigned bucket.

Looks good to me.

codex/contracts/validation.nim
```diff
@@ -0,0 +1,3 @@
+import ./validation/validationparams
+
+export validationparams
```
emizzle, line 3: If you really think this is needed, it should probably be at the `/codex` level.

codex/validation.nim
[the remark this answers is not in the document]
emizzle: Sure, but the slotid isn't generated in the EVM if I'm not mistaken? All-in-all it's probably ok, but we may need to ensure that converting to an array using the right endianness. I could be wrong in that assumption as well.

tests/codex/testvalidation.nim
```diff
@@ -40,13 +60,45 @@ asyncchecksuite "validation":
...
+  
+  test "maxSlots in ValidationParams must be greater than 0":
+    let maxSlots = 0
+    let validationParams: ?!ValidationParams = ValidationParams.init(maxSlots = maxSlots, partitionSize = 1, partitionIndex = 0)
+    check validationParams.isFailure == true
+    check validationParams.error.msg == fmt"maxSlots must be greater than 0! (got: {maxSlots = })"
+
+  for partitionSize in [-100, -1, 0]:
+    test fmt"initializing ValidationParams fails when partitionSize has value our of range (testing for {partitionSize = })":
+      let validationParams: ?!ValidationParams = ValidationParams.init(maxSlots, partitionSize = partitionSize, partitionIndex = 0)
```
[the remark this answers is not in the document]
emizzle: Personal preference. No probs from me. We use a buffer width of 80, so this tends to limit how much can fit on one line.

codex/conf.nim
emizzle: I see, yes indeed using `1` for this parameter, which is default would satisfy this need. However, again my personal opinion (which absolutely does not mean it's right) is that it's more explicit in the code and easier to understand / follow if it's an `Option`, where `uint16.none` means the user supplied nothing, and we don't need to understand that `1` is "special". Personal preference again I guess.

codex/validation.nim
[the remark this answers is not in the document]
marcinczenko: I think it is done both in the client and in the contracts using the same kind of approach.

On-chain (https://github.com/codex-storage/codex-contracts-eth/blob/bba8736132007e52f560dd8f49e4a9b89b03eb6e/contracts/Requests.sol#L52):

```solidity
function slotId(
    RequestId requestId,
    uint256 slotIndex
  ) internal pure returns (SlotId) {
    return SlotId.wrap(keccak256(abi.encode(requestId, slotIndex)));
  }
```

In the client (https://github.com/codex-storage/nim-codex/blob/1e2ad956591d41ec93ede1dec7da228a4558e9d3/codex/contracts/requests.nim#L166):

```nim
func slotId*(requestId: RequestId, slot: UInt256): SlotId =
  let encoding = AbiEncoder.encode((requestId, slot))
  SlotId(keccak256.digest(encoding).data)
```

Because it is a digest in the end, the endianness should not matter that match as long as it is kept consistent on each architecture, right? I checked initially with Mark, he leans towards big endian as well.

codex/conf.nim
marcinczenko: But the user still can provide `1`, so it will remain special. But let's talk it quickly through tomorrow.

emizzle: Sure, but if the user provides `1`, they know what they're doing :)

codex/contracts/validation.nim
marcinczenko: I think it is not yet needed at this time, but while adding historical logging it may become handy. It it is still overkill after that I will just probably leave the `validation.nim` file at the contracts level. The reason I have put it on the `contracts` level is that `client` and `host` it is part of the node's contract (see https://github.com/codex-storage/nim-codex/blob/1e2ad956591d41ec93ede1dec7da228a4558e9d3/codex/codex.nim#L146):

```nim
s.codexNode.contracts = (client, host, validator)
```

codex/contracts/validation/validationparams.nim
```diff
@@ -0,0 +1,34 @@
...
+func init*(
+  _: type ValidationParams, 
+  maxSlots: int,
+  partitionSize: int,
+  partitionIndex: int
+): ?!ValidationParams =
+  if partitionSize <= 0:
+    return failure fmt"Partition size must be greater than 0! (got: {partitionSize = })"
+  if partitionIndex < 0:
+    return failure fmt"Partition index must be greater than or equal to 0! (got: {partitionIndex = })"
```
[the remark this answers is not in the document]
emizzle: (was referring to this part in a string interpolation: `{partitionIndex = }`)

codex/validation.nim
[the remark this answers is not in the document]
emizzle: I likely was not clear, my apologies. I wasn't trying to make a point that we should or shouldn't use BE, my point is that we should be consistent in *both conversion directions* since endianness is not portable. In this particular case, I was asking, "are you sure BE is the right solution" purely because it's not clear to me that we are using BE when converting from `SlotId` to `array[32, byte]`, which happens here: https://github.com/codex-storage/nim-codex/blob/4ab55c63c25d6dca29dfaea29487ce3ed23bd039/codex/contracts/requests.nim#L60-L61

codex/conf.nim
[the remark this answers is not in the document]
marcinczenko: > If the parameter type is uint16, then the compiler must enforce it.

I was not sure if we can tell confutils to use `uint16`... If confutils can handle the user input and disallow e.g. negatives then that would be great.... I will check it.

tests/codex/testvalidation.nim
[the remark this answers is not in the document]
emizzle: > can you explain this

Is this your editor? Is there a compiler error message?

> are we measuring the coverage in the tests themselves?

We are, but it's basic and we don't really use it because I don't think it's all that reliable. So yes, we have it as part of CI, but we don't really use it.

codex/conf.nim
marcinczenko: And if they will not provide anything, it will default to `1` thus the "old" behaviour, so also good ;).

codex/contracts/validation.nim
emizzle: Ah ok, good point 👍  Yeah, I don't know how necessary this file is right now, but I'll leave it up to you to decide.

codex/conf.nim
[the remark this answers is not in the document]
emizzle: > You're right. On the other hand, by limiting it to powers of two, we limit the number of options available to node operator.

We can probably keep this value open without limiting it to a power of two. We can add a description in the CLI help. Most likely the only users to set this value will be us :) After that, users who use it would have a good understanding of what they're doing (hopefull). IOW, don't spend the time on this, as I don't think it's needed.

tests/codex/testvalidation.nim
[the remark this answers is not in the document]
marcinczenko: > Is this your editor? Is there a compiler error message?

No, I just clicked the link somewhere the review (https://app.codecov.io/gh/codex-storage/nim-codex/pull/890/blob/codex/contracts/validation/validationparams.nim) and what I see does not make sense...

Or here: https://app.codecov.io/gh/codex-storage/nim-codex/pull/890

Also, I do not think it is good measuring coverage in the tests themselves... see this one: https://app.codecov.io/gh/codex-storage/nim-codex/pull/890/blob/tests/codex/testvalidation.nim

codex/validation.nim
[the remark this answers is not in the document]
marcinczenko: Apologies in advance if I am not getting it right - it is a lot of internals to parse :), but I tried to drill a bit...

The function you mentioned, is just getting around the fact that `SlotId` is already `array[32, byte]` but it is `distinct`: `SlotId* = distinct array[32, byte]` (same applies to `RequestId`) The function you probably want to be concerned with is this:

```nim
func slotId*(requestId: RequestId, slot: UInt256): SlotId =
  let encoding = AbiEncoder.encode((requestId, slot))
  SlotId(keccak256.digest(encoding).data)
```

and to be precise, the key seems to be here how the slot index - `slot: UInt256` above gets encoded when getting JSONs from RPC interface. It might be this one (`vendor/nim-ethers/ethers/providers/jsonrpc/conversions.nim`):

```nim
func fromJson*(json: JsonNode, name: string, result: var UInt256) =
  result = UInt256.fromHex(json.getStr())
```

And so, it may all get down to `UInt256.fromHex` in `vendor/stint/stint/io.nim`:

```nim
func fromHex*(T: typedesc[StUint|StInt], s: string): T {.inline.} =
  ## Convert an hex string to the corresponding unsigned integer
  parse(s, type result, radix = 16)
```

`parse` follows the input hex string, and convert each encountered byte to `Uint256` by doing: 

```nim
input[curr].readHexChar.stuint(bits)
```

we have `func readHexChar(c: char): int8 {.inline.}=` and thus the final conversion happens here:

```nim
func stuint*[T: SomeInteger](n: T, bits: static[int]): StUint[bits] {.inline.}=
```

So, it looks that the json hex will be encoded into UInt256 that follows the underlying architecture when it comes to endianness:

```nim
type
  # ### Private ### #
  UintImpl*[BaseUint] = object
    when system.cpuEndian == littleEndian:
      lo*, hi*: BaseUint
    else:
      hi*, lo*: BaseUint
```

It looks therefore, that the the result of `func slotId*(requestId: RequestId, slot: UInt256): SlotId =` is consistent across architectures (otherwise we would probably already be in trouble).

With all this in mind, I now start thinking that we should use the default version of `fromBytes` to keep the low level implementation correct:

Thus we probably need to use this:

```nim
let slotIdUInt256 = UInt256.fromBytes(slotId.toArray)
```

As you said running some experiment on a big endian machine would be nice thing to do in order to confirm that... otherwise it is quite hard to be sure...

> We may need to pay attention when updating Stint in the future.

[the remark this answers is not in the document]
emizzle: Yes, you're right, the conversion does start in the `AbiEncoder`, my apologies. It was getting late when I posted the type conversion snippet :)

> we should use the default version of fromBytes to keep the low level implementation correct

Yes, that is exactly what I meant. Because in that way, we are using the underlying UintImpl conversion in both directions, which should handle the endianness for us automatically.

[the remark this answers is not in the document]
marcinczenko: Hmm... I did some more digging and run some experiments on a separate repo. If we dig one step further than in ABI Encoder we will find this (`vendor/nim-contract-abi/contractabi/encoding.nim`):

```nim
func encode(encoder: var AbiEncoder, value: SomeUnsignedInt | StUint) =
  encoder.padleft(value.toBytesBE)
```

So, in functions `toBytesBE`/`fromBytesBE`, the endianness indicates the order in the output/input byte array.
If we use just `toBytes`/`fromBytes` it will apply the machine order to the byte arrays. Taken the above ` encoder.padleft(value.toBytesBE)`, we have to use `fromBytesBE` when converting to `StInt`s.

The rule seems to be the following:

`byte arrays must be always in network order, thus BE!`

I think when they were updating the Stint library, they realised that the byte arrays are best kept in network order and that's why there they use BE as default instead of machine endianness when working with byte arrays.

I will show you this later running. Now I need to switch off for a moment ;).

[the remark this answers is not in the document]
emizzle: Nice job, Marcin! Now we are sure that `toBytesBE` was indeed the right choice 👏

tests/codex/testvalidation.nim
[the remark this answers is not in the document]
emizzle: Ah ok, yeah it's complaining that those lines aren't covered, which is obviously not correct since those field are used in the tests. As I said, it's not very reliable and no one is looking at the coverage stats for now until we can get a better coverage story for nim.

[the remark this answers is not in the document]
marcinczenko: Good to hear that. Coverage may be helpful but is also often source of false perception of quality. So I am not worried at all that we are not looking dogmatically at the coverage stats 👍.

codex/validation.nim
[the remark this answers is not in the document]
marcinczenko: I am happy you triggered more attention to it because now I have better picture!

tests/codex/testvalidation.nim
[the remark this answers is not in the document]
marcinczenko: I would leave it for a moment. I assume someone will remove it when if it does not help in anything. BTW: my VSCode sometimes adds those "hints" but on our code base mostly not...

codex/conf.nim
marcinczenko: Using optionals 👍

README.md
```diff
@@ -109,6 +112,13 @@ The following options are available:
      --marketplace-address  Address of deployed Marketplace contract.
      --validator            Enables validator, requires an Ethereum node [=false].
      --validator-max-slots  Maximum number of slots that the validator monitors [=1000].
+     --validator-partition-size  Slot validation partition size [=1].
+                            A number indicating total number of partitions into which the whole slot id
+                            space will be divided. The value must be greater than 0.
+     --validator-partition-index  Slot validation partition index [=0].
```
emizzle, line 118: Don't forget to update these with the new names 🙏

codex/contracts/validation.nim
emizzle: I'm still thinking it might be best to keep everything "validation related" at the `/codex` level, or move everything to `/codex/contracts`. At the moment, prior to this PR, there is already a `validation` module at the `/codex`. I think introducing one in the `/codex/contracts` level will be confusing.

codex/validation.nim
```diff
@@ -1,35 +1,38 @@
...
 
 proc new*(
-    _: type Validation,
-    clock: Clock,
-    market: Market,
-    maxSlots: int
+  _: type Validation,
+  clock: Clock,
+  market: Market,
+  validationParams: ValidationParams
```
emizzle, line 33: 🟠 Just a suggestion, do you think possibly `config` or even `settings` might be better a more suitable name? That way, when called it would be, eg `validation.config.maxSlots`.

tests/codex/testvalidation.nim
```diff
@@ -12,19 +14,33 @@ import ./helpers
...
   var market: MockMarket
   var clock: MockClock
+  var groupIndex: uint16
+
+  proc initValidationParams(maxSlots: MaxSlots, validationGroups: ?ValidationGroups, groupIndex: uint16): ValidationParams =
+    without validationParams =? ValidationParams.init(maxSlots, groups=validationGroups, groupIndex), error:
+      raiseAssert fmt"Creating ValidationParams failed! Error msg: {error.msg}"
+    validationParams
+  
+  func createValidation(clock: Clock, market: Market, validationParams: ValidationParams): Validation =
```
emizzle, line 33: 🟠 Again, just a suggestion, I think the more "nim way" is to prefix a new call with `new`:
```suggestion
  func newValidation(clock: Clock, market: Market, validationParams: ValidationParams): Validation =
```

emizzle (approved):
Looking good, Marcin! I'll approve now as there is nothing blocking.

On a formatting note, it would be good if you could limit line length to 80 chars please 🙏

tests/codex/testvalidation.nim
marcinczenko: Agree!

codex/validation.nim
marcinczenko: I like `config`. Makes it more compact to read,

README.md
marcinczenko: Good catch 👍

codex/contracts/validation.nim
marcinczenko: I do not mind to move it to one level, but I would like to keep the `validationParams` (or from know most probably `validationConfig`) in a separate file (it will be then `validationconfig.nim`). The easiest would be to move that file to the top level next to the existing `validation.nim`. Would you agree?

marcinczenko:
> Looking good, Marcin! I'll approve now as there is nothing blocking.
> 
> On a formatting note, it would be good if you could limit line length to 80 chars please 🙏

Will be done.

codex/contracts/validation.nim
emizzle: Yep that sounds good 👍

codex/validation.nim
```diff
@@ -43,13 +46,26 @@ proc waitUntilNextPeriod(validation: Validation) {.async.} =
...
+    0
+  else:
+    let slotIdUInt256 = UInt256.fromBytesBE(slotId.toArray)
+    (slotIdUInt256 mod validation.validationParams.partitionSize.u256).truncate(int)
+
+func shouldValidateSlot(validation: Validation, slotId: SlotId): bool =
+  return (
+    validation.partitionIndexForSlotId(slotId) == validation.validationParams.partitionIndex and
+    slotId notin validation.slots and
+    validation.slots.len < validation.validationParams.maxSlots
```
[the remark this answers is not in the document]
emizzle: If `maxSlots == 0`, then it should validate all slots in the space, right? If so, then `shouldValidateSlot` would return `false` when `maxSlots` is `0`.

README.md
```diff
@@ -109,6 +112,13 @@ The following options are available:
      --marketplace-address  Address of deployed Marketplace contract.
      --validator            Enables validator, requires an Ethereum node [=false].
      --validator-max-slots  Maximum number of slots that the validator monitors [=1000].
```
emizzle, line 114: If we decide to allow `--validator-max-slots=0` means validate all slots, then we should add that to the help

README.md
```diff
@@ -109,6 +125,18 @@ The following options are available:
...
+                            monitors.
+     --validator-groups     Slot validation groups [=ValidationGroups.none].
+                            A number indicating total number of groups into which the whole slot id space
+                            will be divided. The value must be in the range [2, 65535]. If not provided, the
+                            validator will observe the whole slot id space and the value of the
+                            --validator-group-index parameter will be ignored. Powers of twos are advised
+                            for even distribution.
+     --validator-group-index  Slot validation group index [=0].
+                            The value provided must be in the range [0, validatorGroups). Ignored when
+                            --validator-groupsis not provided. Only slot ids satisfying condition [(slotId
```
emizzle, line 138:
```suggestion
                            --validator-groups is not provided. Only slot ids satisfying condition [(slotId
```

codex/validationconfig.nim
```diff
@@ -0,0 +1,36 @@
+import std/strformat
+import pkg/questionable
+import pkg/questionable/results
+
+type
+  ValidationGroups* = range[2..65535]
+  MaxSlots* = int
```
emizzle, line 7: 🟡 Is there a particular reason this alias was used?

tests/codex/testvalidation.nim
```diff
@@ -41,12 +56,170 @@ asyncchecksuite "validation":
...
+        fmt"be greater than or equal to 0! (got: {maxSlots})"
+  
+  test "initializing ValidationConfig fails when maxSlots is negative " &
+      "(validationGroups set)":
+    let maxSlots = -1
+    let validationConfig = ValidationConfig.init(
+        maxSlots = maxSlots, groups = validationGroups, groupIndex)
+    check validationConfig.isFailure == true
+    check validationConfig.error.msg == "The value of maxSlots must " &
+        fmt"be greater than or equal to 0! (got: {maxSlots})"
```
emizzle, line 86: 🟡 This test doesn't cover any additional lines, does it? Maybe it's not needed then?

tests/codex/testvalidation.nim
```diff
@@ -41,12 +56,170 @@ asyncchecksuite "validation":
...
+
+  test "group index is irrelevant if validation groups are not set":
+    randomize()
+    for _ in 0..<100:
+      let groupIndex = rand(1000).uint16
```
emizzle, line 91: 🟡 If we want to fuzz the input value, then perhaps we should cover the entire `uint16` space, eg use `rand(uint16.high.int).uint16` (bit ugly). 

Perhaps a more efficient test would be to test the boundaries (upper and lower) and then test a valid `uint16` number. Even that is a bit much honestly. Simply testing a valid `uint16` is probably enough since there are no "disallowed" exceptions.

tests/codex/testvalidation.nim
```diff
@@ -41,12 +56,170 @@ asyncchecksuite "validation":
...
+    let validationConfig = initValidationConfig(
+        maxSlots, ValidationGroups.none)
+    let validation = Validation.new(clock, market, validationConfig)
+    check validation.shouldValidateSlot(slot.id) == true
+  
+  test "slot should be observed if maxSlots is set to 0":
+    let validationConfig = initValidationConfig(
+        maxSlots = 0, ValidationGroups.none)
+    let validation = Validation.new(clock, market, validationConfig)
+    check validation.maxSlotsConstraintRespected
```
emizzle, line 112: I would expect this test to call `shouldValidateSlot` at the end. Also, imo, `maxSlotsConstraintRespected` shouldn't be exported from the Validation module as it's more an internal logic routine.

tests/codex/testvalidation.nim
```diff
@@ -12,19 +14,32 @@ import ./helpers
...
+    validationConfig
 
   setup:
+    groupIndex = groupIndexForSlotId(slot.id, !validationGroups)
     market = MockMarket.new()
     clock = MockClock.new()
-    validation = Validation.new(clock, market, maxSlots)
+    let validationConfig = initValidationConfig(
+        maxSlots, validationGroups, groupIndex)
+    validation = Validation.new(clock, market, validationConfig)
```
emizzle, line 42: Needed? From what I can see, all the tests create their own instances of `validationConfig` and `validation`.

emizzle: Ignore me, I see this is used in the existing tests.

tests/codex/testvalidation.nim
```diff
@@ -41,12 +56,170 @@ asyncchecksuite "validation":
...
+      check slots == toHashSet(validation.slots)
+
+  test "slot is not observed if it is not in the validation group":
+    let validationConfig = initValidationConfig(maxSlots, validationGroups,
+        (groupIndex + 1) mod uint16(!validationGroups))
+    let validation = Validation.new(clock, market, validationConfig)
+    await validation.start()
+    await market.fillSlot(slot.request.id, slot.slotIndex, proof, collateral)
+    await validation.stop()
+    check validation.slots.len == 0
```
emizzle, line 215: The main idea in these tests is to test that `fillSlot` causes the validator to add a slot to its list of slots to validate, right? The difference is that it is testing that sometimes slots are added and sometimes not, depending on the config. In my opinion, both of these two pieces of functionality (adding slots after `fillSlot`, and config dependent addition) are covered in isolation, and so I think maybe these tests aren't adding anything that isn't already covered in other tests.

marcinczenko: Yes, these tests test that some slots are sometimes added and sometimes not. The first one also tests that slots are effectively added to different groups and that the maxSlots parameter is working correctly in combination with other config options - this is not covered effectively by any other test (well, it is "kind-of" tested in some other tests, but those other tests are testing internals and are not really good tests, see my note above - in face of having the first bigger test above those other tests can and perhaps should be removed).

The second test that you mark above, is indeed somehow redundant in the face of the first test, but, is a bit easier to follow and therefore worth keeping as well - when things go wrong, having this test may help you to narrow down the issue slightly easier. Thus I would really prefer to keep them both (and especially the first one - this is the only test that makes me feel confident that the groups and maxSlots work well together).

tests/codex/testvalidation.nim
marcinczenko: This is one of those tests that I would not mind removing in the first place. I added them when I did not have an idea on how to test this properly. Those tests are testing internals - normally I do not like to add such tests, but sometimes I do when I do not have better idea... Thus, if you do not mind I would remove that one...

codex/validationconfig.nim
marcinczenko: Yes, I feel we still hesitate about the best type for `maxSlots`. This way it will be easier to change (otherwise you have to jump all over the places).

README.md
marcinczenko: I see I have the same typo is in the CLI. Fixed.

tests/codex/testvalidation.nim
marcinczenko: Using the "fuzz" is intended to communicate to the reader that the value really does not matter here. It is more to communicate to the reader the intended behaviour. I like more the option of covering the whole space, even though `rand(uint16.high.int).uint16` is a bit noisy. It is more what I intended.

tests/codex/testvalidation.nim
marcinczenko: Actually, having this test, I would be inclined to close hide both `maxSlotsConstraintRespected`
`shouldValidateSlot` and remove the corresponding tests that are just testing internals and thus they do not really add much confidence here.

tests/codex/testvalidation.nim
emizzle: The test before checks for negative maxSlots...

tests/codex/testvalidation.nim
emizzle: The compiler will only let us use valid `uint16` value in the `init` function, and there are no exceptions in the logic or logic branching that would result in a failed `init` function for specific `uint16` values. So I'm failing to understand how this is better than choosing a single valid `uint16` value to test.

tests/codex/testvalidation.nim
emizzle: I think it's still a valid test, as it checks a different branch in the logic. Perhaps the last line could be changed to `check validation.shouldValidateSlot(slot.id) == true`? Wdyt?

tests/codex/testvalidation.nim
emizzle: > the maxSlots parameter is working correctly in combination with other config options - this is not covered effectively by any other test (well, it is "kind-of" tested in some other tests

If we change and keep the test above (https://github.com/codex-storage/nim-codex/pull/890/files#r1772573230) then I think the logic for maxSlots is covered well enough. The `onSlotFilled` event handler only checks `shouldValidateSlot`, so really we should make sure that function is covered for all parameter options. And I think it's pretty well covered but you could bolster with some edge case or boundary tests if you really wanted.

AuHau (commented):
Adding just few comments from brief look I had.
 
Please minimize whitespace/line-break changes. Hopefully one day we will adopt nph that should unify the code style, but till then just please keep it as consistent as possible.

codex/validationconfig.nim
```diff
@@ -0,0 +1,36 @@
...
+      maxSlots: maxSlots, groups: groups, groupIndex: groupIndex)
+
+func maxSlots*(config: ValidationConfig): MaxSlots =
+  config.maxSlots
+
+func groups*(config: ValidationConfig): ?ValidationGroups =
+  config.groups
+
+func groupIndex*(config: ValidationConfig): uint16 =
+  config.groupIndex
```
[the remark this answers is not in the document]
marcinczenko: Because I want to be able to control how there are initialised. If I make them writable, anyone in any external module will be able to create an instance of ValidationConfig without relaying on the intended initialiser, which makes sure the provided values are correct. This validator does not need to re-validate the validation config and it also makes it easier for me to handle CLI. Do you feel there is a better approach of achieving that?

tests/codex/testvalidation.nim
marcinczenko: I am ok with using one value. I will change that.

tests/codex/testvalidation.nim
marcinczenko: Sorry, I put it wrong. I wanted to communicate that `maxSlots` are validated independently from `validationGroups`. But it is also true that it does not cover any additional lines, so if you both say it is not needed I will remove it. The less code the better...

tests/codex/testvalidation.nim
marcinczenko: I discussed this with Eric in a call. We agreed on the following:
- I will make `maxSlotsConstraintRespected` and `shouldValidateSlot`
- I will recheck the related tests to see if they are adding anything. What they cover is pretty much covered by other tests that do not rely on the internals. If there is anything missing I will cover that.

As much as I feel that the big test is very valuable, it does not seem to land, so I will remove it (and keep it for myself 😜.

tests/codex/testvalidation.nim
marcinczenko: Will be covered by https://github.com/codex-storage/nim-codex/pull/890#discussion_r1772975534.

marcinczenko:
> Adding just few comments from brief look I had.
> 
> Please minimize whitespace/line-break changes. Hopefully one day we will adopt nph that should unify the code style, but till then just please keep it as consistent as possible.

I tried to follow the guidelines and what I already found in the code. Some of the changes come from correcting the max line-width - following the earlier suggestion of Eric I made sure I keep in 80columns, but I also changed a couple of more places in the file I touched - I tried to be moderate though.

Maybe I missed some recommended formatting somewhere... Is there anything specific you would like me to change?

tests/codex/testvalidation.nim
marcinczenko: Done. Waiting on the last review round from Adam.

tests/codex/testvalidation.nim
AuHau: Unless there is a case where `validationGroups` is dependant on `maxSlots`, which as you mentioned is validated independently, so I would say please remove it.

tests/codex/testvalidation.nim
```diff
@@ -41,12 +56,76 @@ asyncchecksuite "validation":
...
+    let maxSlots = -1
+    let validationConfig = ValidationConfig.init(
+        maxSlots = maxSlots, groups = validationGroups, groupIndex)
+    check validationConfig.isFailure == true
+    check validationConfig.error.msg == "The value of maxSlots must " &
+        fmt"be greater than or equal to 0! (got: {maxSlots})"
+
+  test "group index is irrelevant if validation groups are not set":
+    randomize()
+    let groupIndex = rand(uint16.high.int).uint16
```
AuHau, line 90: Please, do not use random in unit tests. If you want to do fuzz testing, then please create a new category of testing (after "unit" and "integration" tests) and have a proper setup that would ensure reproducibility of the tests.

From my end, "random values" in tests make you sloppy because you will not think about the edge case values that should be tested.

For example, a writeup on why not to use random numbers in tests: https://scotthannen.org/blog/2024/05/20/dont-use-random-numbers-in-tests.html

AuHau (changes requested):
Almost there, but still two things to fix please.

tests/codex/testvalidation.nim
marcinczenko: Thanks for a nice article. It makes lots of sense in advising to focus on clear boundaries. Our case here is different though. There are no boundaries except for the limits of `uint16` type. The test should never fail, regardless of the value provided. You cannot really check boundaries, because there are no boundaries to check - value `0` value `65535` and value `1265` all have equal weight. No single value in the whole range of `uint16` is special. The alternative to not using random here would be to check the whole range of `uint16`, which I feel is an overkill. Thus, in my opinion, to communicate the intended behaviour, a random value feels like a good compromise. Even when there are boundaries, in some cases you could do both (checking boundaries and using random number) when the test is simple, the random value is reported and easy to interpret. Using random in a unit test is not by definition bad... "avoid" is what I would probably say.

So, I am not sure how to make it better... Maybe we should just discuss this shortly.

tests/codex/testvalidation.nim
marcinczenko: Indeed `validationGroups` **should** not depend on `maxSlots`. But we should test that. If we only have a test with negative `maxSlots` when `validationGroups` is not set, then it could still happen that there is a bug in the implementation causing the negative value of `maxSlots` to be accepted in the case when `validationGroups` is set. Saying that it is "validated independently" is relaying on the current implementation. If I change the implementation in such a way that negative `maxSlots` is accepted when `validationGroups` is set (which would be a bug), I would prefer to have a failing test.

For instance, if I remove this test, and I will make the implementation of the `ValidationConfig` initialiser the following:

```nim
func init*(
    _: type ValidationConfig,
    maxSlots: MaxSlots,
    groups: ?ValidationGroups,
    groupIndex: uint16 = 0): ?!ValidationConfig =

  if validationGroups =? groups and groupIndex >= uint16(validationGroups):
    return failure "The value of the group index must be less than " &
        fmt"validation groups! (got: {groupIndex = }, " &
        fmt"groups = {validationGroups})"
  
  if groups.isSome and maxSlots < 0:
    return success ValidationConfig(
      maxSlots: maxSlots, groups: groups, groupIndex: groupIndex)

  if maxSlots < 0:
    return failure "The value of maxSlots must be greater than " &
        fmt"or equal to 0! (got: {maxSlots})"
  
  success ValidationConfig(
      maxSlots: maxSlots, groups: groups, groupIndex: groupIndex)
```

then I would have all the tests pass even thought there is an obvious bug.

tests/codex/testvalidation.nim
marcinczenko: Maybe this: let's make `groupIndex` an optional. In this case, when `validationGroups` is not set **and** `groupIndex` is set (i.e. it has value different than `None`) then construction of the `ValidationConfig` should fail. This has stricter semantics can be tested without using any random things. What do you think?

marcinczenko: After meeting with Adam, we decided to settle down on some sort of a compromise: we remove this test, and the long complex one, while keeping the one discussed [here](https://github.com/codex-storage/nim-codex/pull/890#discussion_r1784291403).

tests/codex/testvalidation.nim
marcinczenko: Decided to keep that one for now.

AuHau (approved):

PR #908, opened 2024-09-20
Fix reservations endpoint documentation

2-towns:
Tiny PR to update the OpenAPI documentation for the reservations endpoint and change the PATCH method to GET.


AuHau:
This was in the end fixed in the https://github.com/codex-storage/nim-codex/pull/886

PR #910, opened 2024-09-23
feat(slot-reservations): Add expansionRate parameter to support contract changes

### LINKED ISSUE #909: feat(slot-reservations): Add `expansionRate` parameter to support contract changes
An `expansionRate` parameter was added to the contracts in https://github.com/codex-storage/codex-contracts-eth/issues/115. 

This parameter is part of the `StorageRequest` and needs to be supported in the Codex client.

emizzle:
Closes #909.

This is part 2 of a [longer list of tasks for slot reservations](https://github.com/codex-storage/codex-pm/issues/45) to prevent PRs from being too large.

Adds `expansionRate` field to the `StorageRequest` object, to mirror the object in the contracts.

The CodexClient api module uses a default value of 100, which means that all SPs will be available to reserve slots immediately, when the functionality is finally implemented. It was chosen to provide as little disruption to existing as possible, and may change in the future.

The default value for the REST API endpoint was 60, meaning 60% of the total time before expiry will have all addresses eligible to fill the slot.

### DIFF
--- a/codex/contracts/requests.nim
+++ b/codex/contracts/requests.nim
@@ -19,6 +19,7 @@ type
     content* {.serialize.}: StorageContent
     expiry* {.serialize.}: UInt256
     nonce*: Nonce
+    expansionRate* {.serialize.}: uint8
   StorageAsk* = object
     slots* {.serialize.}: uint64
     slotSize* {.serialize.}: UInt256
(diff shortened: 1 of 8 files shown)


emizzle:
Holding off on the expanding window work for now. Can resurrect this PR once the expanding window work is planned.

PR #917, opened 2024-09-25
feat(slot-reservations): Add SaleSlotReserving state

### LINKED ISSUE #916: Slot reservations: Add `SaleReserving` state
Add a state, `SaleReserving` to the sales state machine that reserves a slot before attempting to fill the slot.

This should be the first state after `SalePreparing`.

The state should:
1. Check if the SP is eligible to reserve the slot (`canReserveSlot` contract call).
2. If eligible, reserve the slot (`reserveSlot` contract call)
  - if errors are encountered, go to `SaleErrored`
3. If not eligible, go to `SaleIgnored`, but don't re-add to slot queue (`reprocessSlot = false`)
4. If there were no errors reserving the slot, proceed to `SaleDownload`

emizzle:
Closes #916

This is being implemented as part of a [longer list of tasks for slot reservations](https://github.com/codex-storage/codex-pm/issues/45) to prevent PRs from being too large.

Adds a new state, `SaleSlotReserving`, that attempts to reserve a slot before downloading.
If the slot cannot be reserved, the state moves to SaleIgnored.
On error, the state moves to `SaleErrored`.

`SaleIgnored` is also updated to pass in `reprocessSlot` and `returnBytes`, controlling the behaviour in the sales module after the slot is ignored. This is because previously it was assumed that SaleIgnored was only reached when there was no availability. This is no longer the case, since `SaleIgnored` can now be reached when a slot cannot be reserved.

### DIFF
--- a/codex/contracts/market.nim
+++ b/codex/contracts/market.nim
@@ -260,7 +260,8 @@ method canReserveSlot*(
   requestId: RequestId,
   slotIndex: UInt256): Future[bool] {.async.} =
 
-  await market.contract.canReserveSlot(requestId, slotIndex)
+  convertEthersError:
+    return await market.contract.canReserveSlot(requestId, slotIndex)
 
 method subscribeRequests*(market: OnChainMarket,
                          callback: OnRequest):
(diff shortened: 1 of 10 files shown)


markspanbroek (approved):
:+1:

AuHau (approved):
LGTM! Great work!

PR #926, opened 2024-09-30
feat(slot-reservations): support SlotReservationsFull event

### LINKED ISSUE #931: feat(slot-reservations): support `SlotReservationsFull` event
Supports the `SlotRervationsFull` contract event.

Subscribes to the `SlotReservationsFull` event in the sales module. When emitted, the slot is removed from the slot queue.

emizzle:
Closes #931.

This is being implemented as part of a [longer list of tasks for slot reservations](https://github.com/codex-storage/codex-pm/issues/45) to prevent PRs from being too large.

Subscribes to the `SlotReserverationsFull` contract event. Once the event is triggered, the slot is removed from the node's slot queue.

### DIFF
--- a/codex/contracts/market.nim
+++ b/codex/contracts/market.nim
@@ -307,6 +307,17 @@ method subscribeSlotFreed*(market: OnChainMarket,
     let subscription = await market.contract.subscribe(SlotFreed, onEvent)
     return OnChainMarketSubscription(eventSubscription: subscription)
 
+method subscribeSlotReservationsFull*(
+  market: OnChainMarket,
+  callback: OnSlotReservationsFull): Future[MarketSubscription] {.async.} =
+
+  proc onEvent(event: SlotReservationsFull) {.upraises:[].} =
+    callback(event.requestId, event.slotIndex)
+
+  convertEthersError:
+    let subscription = await market.contract.subscribe(SlotReservationsFull, onEvent)
+    return OnChainMarketSubscription(eventSubscription: subscription)
+
 method subscribeFulfillment(market: OnChainMarket,
                             callback: OnFulfillment):
                            Future[MarketSubscription] {.async.} =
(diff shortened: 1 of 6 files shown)


AuHau (approved):
LGTM, but have a question. 

Is it possible we have already running agent when this event is emitted? I guess it is possible but it should be "fast run" in a sence that the agent should reach fast either Downloading state (when it makes reservation), or Ignored state. I guess this would get more pressing issue once we implement expanding window, because then the node will have to wait before it will be able to make the reservation... WDYT?

emizzle:
I'm not sure I understand. What do you mean by "fast run"? Once the slot is reserved, the agent goes straight to downloading the data.

If the event is emitted while the agent is in the `SlotReserving` state, the call to `reserveSlot` should fail and it will go to `SaleErrored` with the error `Reservation not allowed`:
https://github.com/codex-storage/nim-codex/blob/1fb610da15a85e4f8ca0aa90a6ba7836ed82cf25/codex/sales/states/slotreserving.nim#L47-L51. Imo, it's not going to need to be reprocessed at this point as the reservations are full.

If the event is emitted while the agent is in the `Downloading` state, then it means the call to the `reserveSlot` was successful, so the SP would have been one of the ones to secure a reservation.

codex/contracts/market.nim
```diff
@@ -306,6 +306,17 @@ method subscribeSlotFreed*(market: OnChainMarket,
     let subscription = await market.contract.subscribe(SlotFreed, onEvent)
     return OnChainMarketSubscription(eventSubscription: subscription)
 
+method subscribeSlotReservationsFull*(
+  market: OnChainMarket,
+  callback: OnSlotReservationsFull): Future[MarketSubscription] {.async.} =
+
+  proc onEvent(event: SlotReservationsFull) {.upraises:[].} =
```
markspanbroek, line 313: Not relevant for this PR, but we should probably get rid of `upraises` and replace it with regular `raises` at some point.

markspanbroek (approved):
Looks good to me :+1:

I was also wondering whether the event should be passed into the state machine so that we can react to it, but as you explained that is unnecessary.

AuHau:
Yeah agree, all is good. I was thinking ahead of time when we gonna have expanding window in the game, but yeah lets ignore that.

PR #928, opened 2024-09-30
expose some metrics over HTTP (Prometheus)

cskiraly:
This is just to start exposing the metrics that are already there from libp2p.
Further changes are needed to make sure the right metrics are collected and exposed.


cskiraly:
closing, since metrics were already exposed with --metrics :-)

PR #929, opened 2024-10-01
feat: add REST endpoint to fetch manifest without fetching data

### LINKED ISSUE #920: REST API to fetch manifest
**Is your feature request related to a problem? Please describe.**
I'd like to altruistically cache data from different Codex nodes upon notification (e.g. a Waku message containing a CID). At the same time, I need to make sure that the data I am pulling is not too large - i.e. I want to limit the datasize to .e.g 5MB, but someone can easily send a CID pointing to blob of 50GB and fill my node. 

At the moment the REST API only allows me to directly request data download

**Describe the solution you'd like**

I'd like a REST API endpoint - e.g.

```
http://localhost:8080/api/codex/v1/manifest/{cid}/network
```

Which would only return the manifest obtained from the network for a given CID upon which I could analyze it and call the `data/{cid}/network` endpoint to actually pull the data.

**Describe alternatives you've considered**

I guess bindings to other languages would solve this without exposing it through REST API, but since it seems REST is the way to control the node at the moment, it seems to be the best choice.

vpavlin:
Closes #920 

I attempted to add the `manifest/{cid}/network` endpoint to fix #920 

Seems like `nph` reformated the whole file - do you use it? If so which version and with what config?:) If not, I'll try to disable it and push PR without this mess. The actual changes are on lines `104` and `220`


AuHau:
Yeah, we are unfortunately not yet using `nph` :-(

emizzle:
I'm am very much in favour of using it. We'd need a code base-wide effort to implement it initially, which may be a bit of an effort (worthwhile effort though).

benbierens:
I'm seeing a bunch of unrelated changes, I think. Formatting and ordering of procs. Makes the more tricky to review the functional changes. Mostly I'd like to ask you: when is it useful to fetch the manifest? I would like to have an API that gives a lot more info about the state of the node and much finer control.

vpavlin:
Disabled `nph` to make the PR look **sane**:D

@benbierens 

> when is it useful to fetch the manifest?

For me particularly, I'd like to leverage the altruistic data storing - but I don't want to store anything that people throw at me:) So I'd like to first know the size of the dataset they are sking me to store. Very specifically, I've built a tiny demo for Codex + Waku here https://app.radicle.xyz/nodes/seed.radicle.garden/rad:z47KBuinFp7miXRVPGz6cVHTTnr94 and part of it is a public node with a small Waku node "in front of it" which caches content based on `cid` passed via Waku message - I'd like to be able to verify the size of the data is < X MB before I actually pull the data to the node.

I assume there will be other use cases for knowing the manifest and data size before actually triggering the data pull in the future, but this is what I have at the moment:)

cskiraly:
> I'm am very much in favour of using it. We'd need a code base-wide effort to implement it initially, which may be a bit of an effort (worthwhile effort though).

If we decide to do some reformatting, please, please do it as a separate patch.
If merge commit is used, please also do it as a separate PR, so that it ends up being a separate patch.

AuHau:
@vpavlin so @benbierens added this feature in his API work and now we have released it, see here: https://api.codex.storage/#tag/Data/operation/downloadNetworkManifest

Does that satisfy your need? Can we close this PR?

vpavlin:
Yup, I think that's perfect:) I am only disapponted because I was hoping to have commit in Codex repo:-P

AuHau:
> Yup, I think that's perfect:) I am only disapponted because I was hoping to have commit in Codex repo:-P

Oh man, I am sorry. I will be watching for your next contribution ;-)

Issue #914, opened 2024-09-24
[BUG] Codex Storage node with prover crashes - Unable to get send connection for peer message not sent


veaceslavdoina:
**Describe the bug**
Codex Storage node crashes periodically.

**To Reproduce**
1. Run Codex Storage node
2. Make storage availability
3. Fill storage requests (?)
4. Node will crash periodically(?)

**Expected behavior**
Codex should not crash.

**Screenshots**
<details>
<summary>screenshots</summary>

### Logs
<img width="1048" alt="Screenshot 2024-09-24 at 22 57 04" src="https://github.com/user-attachments/assets/9d02a307-4fa0-4a81-89c9-eda404d6372f">

### All nodes
<img width="1510" alt="Screenshot 2024-09-24 at 22 52 55" src="https://github.com/user-attachments/assets/03590a17-af15-4bcc-8b15-675e6622c078">

### Node 10
<img width="1508" alt="Screenshot 2024-09-24 at 22 49 23" src="https://github.com/user-attachments/assets/68e728e0-7bf3-4b3c-ae08-26e8b566fbd9">
</details>

**Environment:**
 - OS: `24.04.1 LTS`
 - Version: `0.1.4`
 - Tag/Commit: `484124d`
 - Arch: `amd64`

**Additional context**
We are running 10 Codex Storage nodes and they all crashes periodically.
```shell
codex persistence prover
```

Please see attached full log of one of the crash - [codex-10-1-crash.txt](https://github.com/user-attachments/files/17120201/codex-10-1-crash.txt).

benbierens:
I've just now seen the same issue in my autoclient-node.
Important: This node did not run prover. This issue seems unrelated to prover.
Autoclient-node is run with persistence, and does create purchases. But it does not store and does not create proofs.
I think this one needs to be increased in priority...

benbierens:
I have a sneaky suspicion...
I think the bug might occur _sometimes_, when:
1. a number of hosts are downloading a slot from a client
2. a host completes the download first, submits proof and claims the slot
3. other hosts abort their downloads and drop the connection
I suspect the sudden drop triggers some code in the client that has a race condition, somewhere related to
`networkpeer.nim` line 35 = `proc connected*(b: NetworkPeer): bool`
My guess is that proc connected returns true right _before_ the connection is dropped. The client proceeds as if the connection is alive and runs into a null-ref exception.
I'm throwing tests at this to try to reproduce and gather some evidence.

[...]

PR #937, opened 2024-10-03
Fix: null-ref in networkPeer

benbierens:
Please read #914

### DIFF
--- a/codex/blockexchange/network/network.nim
+++ b/codex/blockexchange/network/network.nim
@@ -75,7 +75,6 @@ type
     handlers*: BlockExcHandlers
     request*: BlockExcRequest
     getConn: ConnProvider
-    inflightSema: AsyncSemaphore
 
 proc peerId*(b: BlockExcNetwork): PeerId =
   ## Return peer id
@@ -99,14 +98,11 @@ proc send*(b: BlockExcNetwork, id: PeerId, msg: pb.Message) {.async.} =
 
   let peer = b.peers[id]
   try:
-    await b.inflightSema.acquire()
     await peer.send(msg)
   except CancelledError as error:
     raise error
   except CatchableError as err:
     error "Error sending message", peer = id, msg = err.msg
-  finally:
-    b.inflightSema.release()
 
 proc handleWantList(
   b: BlockExcNetwork,
@@ -332,8 +328,7 @@ proc new*(
   let
     self = BlockExcNetwork(
       switch: switch,
-      getConn: connProvider,
-      inflightSema: newAsyncSemaphore(maxInflight))
+      getConn: connProvider)
 
   proc sendWantList(
     id: PeerId,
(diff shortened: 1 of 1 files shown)


benbierens:
I checked the other uses of .withValue and found no other situations where this same issue could occur.
Reran the dist-tests I had set up to reproduce the issue and indeed this fixes it.

codex/blockexchange/network/network.nim
```diff
@@ -93,18 +93,20 @@ proc send*(b: BlockExcNetwork, id: PeerId, msg: pb.Message) {.async.} =
...
-    except CatchableError as err:
-      error "Error sending message", peer = id, msg = err.msg
-    finally:
-      b.inflightSema.release()
-  do:
+  if not (id in b.peers):
     trace "Unable to send, peer not found", peerId = id
+    return
+
+  let peer = b.peers[id]
```
dryajov, line 100: Interesting, it makes sense that the object can get collected in between `acquire` and `send`. It would have been probably better if `withValue` would also increase the refcount with `GC_Ref/Unref` for reference types, so that this situations don't occur.

That said, I think the real culprit is the semaphore, I cannot remember exactly why I had it in there in the first place and it's only used here, but I think I wanted to avoid this type of situations by synchronizing access to to the peers table, but that isn't a great idea either way.

I would remove the semaphore entirely as it isn't doing much here either way and peers should be able to handle concurrent sends.

dryajov (dismissed):

gmega (approved):
Great find!

codex/blockexchange/network/network.nim
benbierens: I'm of two minds. I want the semaphore gone. I built an image without it and throw a bunch of tests at it. It seems to work all right. But the issue you found and linked has me worried.

So I propose we merge the fix by replacing the .withValue. And we make sure we never use withValue in an async context. (I checked all other uses of withValue and none of them are at risk!)

We can make a separate issue for the semaphore. I suspect a semaphore per peer might be a sensible approach: it allows more to go on simultaneously but still limits the outbound requests.

gmega:
Since Dmitriy will be away for a bit, I'm going to dismiss his review so we can unblock this, and then I'd ask someone else (e.g. @markspanbroek?) to re-review it in light of the current discussion on his behalf.

markspanbroek (approved):

codex/blockexchange/network/network.nim
emizzle: Might be a good idea to add a comment on `withValue` that instructs our future selves not to use it in within an async context, referencing this PR and https://github.com/codex-storage/nim-codex/issues/914.

dryajov: This. `withValue` needs manual refcount tracking with async due to raw pointers. Imo, I don't think the semaphore is needed in this context.

emizzle: Created an issue for this: https://github.com/codex-storage/nim-codex/issues/943

Issue #932, opened 2024-10-02
[BUG] Codex won't run with glibc 2.31


gmega:
Our binaries don't seem to support the [current version](https://launchpad.net/ubuntu/+source/glibc) for glibc (2.31) in Focal Fossa for some reason. I don't know why, out of all versions, this one seems to have been skipped.

```bash
➜  scripts git:(master) ✗ ldd -v codex-v0.1.4-linux-amd64 | grep 'libc.so.6 (GLIBC' | sed 's/.*GLIBC_\(.*\)).*$/\1/g' | sort | uniq | grep '2.3[0-9]*'
2.3
2.3.2
2.32
2.33
2.3.4
2.34
2.35
```

Shelpin:
Thanks a lot for opening the issue! hope to be able to run my codex node soon :))))....big congrats on the launch! and a big THANKS!

veaceslavdoina:
Just created a [new branch](https://github.com/codex-storage/nim-codex/tree/refs/heads/fix/use-ubuntu-20-04-for-linux-builds) and [run a release workflow](https://github.com/codex-storage/nim-codex/actions/runs/11152216422).

Let's wait for an artifacts and check linux binary.

PR #939, opened 2024-10-04
Use Ubuntu 20.04 for Linux amd64 releases

veaceslavdoina:
This PR address the issue #932 and we are using a workaround and run release builds for Linux amd64 on [Ubuntu 20.04](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2004-Readme.md).

For arm64 builds, we miss such an option on [BuildJet](https://buildjet.com/for-github-actions/docs/runners/hardware#arm), and by the end of the yer, we expect to switch to the [GitHub arm64 runners](https://github.blog/changelog/2024-06-03-actions-arm-based-linux-and-windows-runners-are-now-in-public-beta/). But it is not so clear now which Ubuntu versions will be available.
<details>
<summary>details</summary>

[About Ubuntu and Windows larger runners](https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners/about-larger-runners#about-ubuntu-and-windows-larger-runners)

Screenshot 2024-10-04 at 10 48 56
</details>

Other changes
- we also can handle branches with the slash in the name - `fix/use-ubuntu-20-for-linux-amd64-releases`, to generate intermediate/fixes releases for testing
- for testing purposes it might be sense to retain build artifacts for a longer period of time

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -71,8 +71,9 @@ jobs:
             macos*)   os_name="darwin"  ;;
             windows*) os_name="windows" ;;
           esac
-          codex_binary="${{ env.codex_binary_base }}-${{ github.ref_name }}-${os_name}-${{ matrix.cpu }}"
-          cirdl_binary="${{ env.cirdl_binary_base }}-${{ github.ref_name }}-${os_name}-${{ matrix.cpu }}"
+          github_ref_name="${GITHUB_REF_NAME/\//-}"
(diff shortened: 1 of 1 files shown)


benbierens (approved):

gmega (approved):
Yeah thanks @veaceslavdoina, this looks good. Since 20.04 is LTS and still active, I think this is good in more than one way.

PR #942, opened 2024-10-07
fix(restapi): Add cors headers when the request is returning errors

2-towns:
Currently, the REST endpoints don't return the CORS headers when they return an error.

One of the reasons is that Nim Presto has a weird behavior of adding the `Access-Control-Allow-Origin` header when a request is successful and omitting it when the request returns an error (see the method processRestRequest in nim-presto/presto/serverprivate.nim).

This PR fixes this behavior by adding the CORS headers whenever a response is returned. I was hesitant to create an `OPTIONS` endpoint for every request, as we did for the data upload, but it does not work with [simple requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#simple_requests) because no `OPTIONS` request is sent by the browser in that case.

Without the CORS headers returned when an endpoint returns an error, I am unable to display the error message to the user. Moreover, I am not able to know if the Marketplace is enabled or not (see https://github.com/codex-storage/codex-marketplace-ui/issues/29).

### DIFF
--- a/codex/rest/api.nim
+++ b/codex/rest/api.nim
@@ -134,7 +134,8 @@ proc initDataApi(node: CodexNodeRef, repoStore: RepoStore, router: var RestRoute
         resp.setCorsHeaders("POST", corsOrigin)
         resp.setHeader("Access-Control-Allow-Headers", "content-type")
 
-      return RestApiResponse.response("", Http204)
+      resp.status = Http204
+      await resp.sendBody("")
 
   router.rawApi(
     MethodPost,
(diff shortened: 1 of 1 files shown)


AuHau (approved):
Yeah it is pain that Presto does not handle this 😳 But yeah otherwise LGTM. Just fix the tests ;-)

2-towns:
> Just fix the tests ;-)

I just ran the testAll and I have three tests failing: 

```
[Suite] Marketplace contracts
  [FAILED] can be paid out at the end, specifying reward and collateral recipient

[Suite] On-Chain Market
  [FAILED] supports slot reservations full subscriptions
  [FAILED] pays rewards to reward recipient, collateral to host
```

But it doesn't seem related to my commit. Do you have specific tests that I should look into? @AuHau

2-towns:
> > Just fix the tests ;-)
> 
> I just ran the testAll and I have three tests failing:
> 
> ```
> [Suite] Marketplace contracts
>   [FAILED] can be paid out at the end, specifying reward and collateral recipient
> 
> [Suite] On-Chain Market
>   [FAILED] supports slot reservations full subscriptions
>   [FAILED] pays rewards to reward recipient, collateral to host
> ```
> 
> But it doesn't seem related to my commit. Do you have specific tests that I should look into? @AuHau

I just reran the tests and it worked

PR #950, opened 2024-10-10
docs: openapi node fix

### LINKED ISSUE #949: [BUG] node/peeris API call returns 404
**Describe the bug**
Based on https://api.codex.storage/#tag/Node/operation/getPeerId

I should be able to get peerId by calling

```
curl -v http://localhost:8080/api/codex/v1/node/peerid
```

Same for `/node/spr`

But I get `404 Not Found`
**To Reproduce**
1. Run 0.1.6
2. `curl -v http://localhost:8080/api/codex/v1/node/peerid`

**Expected behavior**
Return peerId

**Environment:**
 - OS: ubuntu
 - Version: 0.1.6

 - Arch amd64:

**Additional context**

AuHau:
Closes #949

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -726,7 +726,7 @@ paths:
         "503":
           description: Persistence is not enabled
 
-  "/node/spr":
+  "/spr":
     get:
       summary: "Get Node's SPR"
       operationId: getSPR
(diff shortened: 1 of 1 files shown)


vpavlin:
Just for record keeping, I think it would make sense to group these two under something - `node` made kinda sense.

IIRC I've seen `self` or `~` (i.e. shortcut for "homedir") in the past - i.e. `/self/spr` or `/~/spr`

Because, in theory in the future there could be something like `/peer/{peerId}/spr` added/needed

veaceslavdoina (approved):

Issue #940, opened 2024-10-07
Add Nix Flake with derivation, shell, service


jakubgs:
It would be helpful to have a Nix flake for this package. The most basic version would just include a build shell, but a a more advanced version would also include a full derivation and also a NixOS service module.

If possible we should also have CI job for our Nix builds.

jakubgs:
Here is an attempt by Corey to get something working, some parts of the derivation might be useful: [codex-flake.nix.txt](https://github.com/user-attachments/files/17275441/codex-flake.nix.txt)

veaceslavdoina:
@markoburcul, the following lines in [Makefile](https://github.com/codex-storage/nim-codex/pull/954/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52)
```diff
+ CXXFLAGS ?= -std=c++17 -mssse3
+ export CXXFLAGS
```
broke the builds on ARM64 with the error - `c++: error: unrecognized command-line option ‘-mssse3’`

<details>
<summary>more details</summary>

![Image](https://github.com/user-attachments/assets/c21e6b28-5c03-4c07-94c3-83ac5931b245)


```shell
CMake Error at /usr/share/cmake-3.22/Modules/CMakeTestCXXCompiler.cmake:62 (message):
  The C++ compiler

    "/usr/bin/c++"

  is not able to compile a simple test program.

  It fails with the following output:

    Change Dir: /opt/nim-codex/vendor/nim-leveldbstatic/build/CMakeFiles/CMakeTmp

    Run Build Command(s):/usr/bin/gmake -f Makefile cmTC_dc503/fast && gmake[1]: Entering directory '/opt/nim-codex/vendor/nim-leveldbstatic/build/CMakeFiles/CMakeTmp'
    /usr/bin/gmake  -f CMakeFiles/cmTC_dc503.dir/build.make CMakeFiles/cmTC_dc503.dir/build
    gmake[2]: Entering directory '/opt/nim-codex/vendor/nim-leveldbstatic/build/CMakeFiles/CMakeTmp'
    Building CXX object CMakeFiles/cmTC_dc503.dir/testCXXCompiler.cxx.o
[...]

PR #954, opened 2024-10-16
nix-flake: Add shell definition

markoburcul:
Initialized flake file and added development shell definition.

Referenced issue: https://github.com/codex-storage/nim-codex/issues/940

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,35 @@
+{
+  description = "Codex build flake";
+  
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-24.05";
+  };
+
+  outputs = { self, nixpkgs }:
+    let
+      supportedSystems = [
+        "x86_64-linux" "aarch64-linux"
+        "x86_64-darwin" "aarch64-darwin"
+      ];
+      forAllSystems = f: nixpkgs.lib.genAttrs supportedSystems (system: f system);
+      pkgsFor = forAllSystems (system: import nixpkgs { inherit system; });
+    in rec {
+      devShells = forAllSystems (system: let
+        pkgs = pkgsFor.${system};
+        inherit (pkgs) lib stdenv mkShell;
+      in {
+        default = mkShell.override { stdenv = pkgs.gcc11Stdenv; } {
+          buildInputs = with pkgs; [
+              # General
+              git pkg-config openssl lsb-release
+              # Build
+              rustc cargo nimble gcc11 cmake nim-unwrapped-1
+              # Libraries
+              gmp llvmPackages.openmp
+              # Tests
+              nodejs_18
+          ];
+        };
+      });
+    };
+}
--- a/Makefile
+++ b/Makefile
@@ -40,6 +40,9 @@ DOCKER_IMAGE_NIM_PARAMS ?= -d:chronicles_colors:none -d:insecure
 
 LINK_PCRE := 0
 
+CXXFLAGS ?= -std=c++17 -mssse3
+export CXXFLAGS
+
 # we don't want an error here, so we can handle things later, in the ".DEFAULT" target
 -include $(BUILD_SYSTEM_DIR)/makefiles/variables.mk
 
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,27 @@
+{
+  "nodes": {
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1729449015,
+        "narHash": "sha256-Gf04dXB0n4q0A9G5nTGH3zuMGr6jtJppqdeljxua1fo=",
+        "owner": "NixOS",
+        "repo": "nixpkgs",
+        "rev": "89172919243df199fe237ba0f776c3e3e3d72367",
+        "type": "github"
+      },
+      "original": {
+        "owner": "NixOS",
+        "ref": "nixos-24.05",
+        "repo": "nixpkgs",
+        "type": "github"
+      }
+    },
+    "root": {
+      "inputs": {
+        "nixpkgs": "nixpkgs"
+      }
+    }
+  },
+  "root": "root",
+  "version": 7
+}


markoburcul:
The shell example taken from @corpetty [comment](https://github.com/codex-storage/nim-codex/issues/940#issuecomment-2397079978). I've managed to reuse this shell.nix file with minimal modifications. To test the shell I have:
- built codex executable
- build/test and launch Hardhat node
- run `make testAll` with success
- run the [`Local two client test`](https://docs.codex.storage/learn/local-two-client-test) with success

Now, I am not 100% sure about this part:

> Here's how I updated the CMakeLists.txt file (from Claude.ai):
> 
>     1. Open the CMakeLists.txt file in a text editor:
>        `nano /home/petty/codex/repos/nim-codex/vendor/nim-leopard/vendor/leopard/CMakeLists.txt`

since we do set `CXXFLAGS` and `CFLAGS` for the shell and for the derivation in the `flake.nix` file you've sent. Maybe I didn't understand well..

flake.nix
```diff
@@ -0,0 +1,58 @@
...
+    flake-utils.url = "github:numtide/flake-utils";
+  };
+
+  outputs = { self, nixpkgs, flake-utils }:
+    flake-utils.lib.eachDefaultSystem (system:
+      let
+        pkgs = nixpkgs.legacyPackages.${system};
+
+        # Create a dummy lsb_release script
+        dummy-lsb-release = pkgs.writeScriptBin "lsb_release" ''
```
yakimant, line 15: Is it related?
https://github.com/NixOS/nixpkgs/tree/master/pkgs/os-specific/linux/lsb-release

yakimant (approved):
Looks good.
Althtough, I would add `.envrc` for nix flakes every time we add `flake.nix` - it's so fast and convinient!

flake.nix
jakubgs: Yes, I think we can use `lsb-release` from `nixpkgs` itself.

flake.nix
```diff
@@ -0,0 +1,58 @@
+{
+  description = "Codex build flake";
+
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
+    flake-utils.url = "github:numtide/flake-utils";
+  };
+
+  outputs = { self, nixpkgs, flake-utils }:
+    flake-utils.lib.eachDefaultSystem (system:
```
jakubgs, line 10: I think I prefer avoiding `flake-utils`, the same can be achieved with a few lines:
https://github.com/status-im/go-waku/blob/3c4e40c729a0ad5f1079ac1e3f31eeaae1b9cbdd/flake.nix#L8-L13

But I'm not saying you have to change it.

flake.nix
```diff
@@ -0,0 +1,58 @@
...
+                bash
+                pkg-config
+                openssl
+                gmp
+                dummy-lsb-release
+                nim
+                nimble
+                nodejs_18
+                llvmPackages.openmp
+            ];
```
jakubgs, line 39: I don't think separating it all like that into multiple lines is great. I'd do something more organized:
```suggestion
            buildInputs = with pkgs; [
              # General
              git pkg-config openssl lsb-release
              # Build
              rustc cargo nim nimble gcc11 cmake
              # Libraries
              gmp llvmPackages.openmp
              # What is this for?
              nodejs_18
            ];
```

flake.nix
```diff
@@ -0,0 +1,58 @@
...
+          #!${pkgs.stdenv.shell}
+          echo "Distributor ID: NixOS"
+          echo "Description:    NixOS"
+          echo "Release:        Unstable"
+          echo "Codename:       nixos"
+        '';
+      in {
+        devShell = pkgs.mkShell {
+            buildInputs = with pkgs; [
+                gcc11
```
jakubgs, line 25: I think the more kosher way of doing this would involve changing the GCC for the whole shell:
```nix
mkShell = mkShell.override { stdenv = pkgs.gcc11Stdenv; };
```
Something like that.

flake.nix
```diff
@@ -0,0 +1,58 @@
...
+                nim
+                nimble
+                nodejs_18
+                llvmPackages.openmp
+            ];
+
+            shellHook = ''
+                export PATH=${pkgs.gcc11}/bin:${dummy-lsb-release}/bin:$PATH
+                export CC=${pkgs.gcc11}/bin/gcc
+                export CXX=${pkgs.gcc11}/bin/g++
```
jakubgs, line 44: Should not be necessary if you adjust `stdenv` used.

jakubgs (changes requested):
Taking shape.

flake.nix
markoburcul: The nodejs18 is used for running tests https://docs.codex.storage/learn/local-two-client-test . I assumed if it is on their website listed as official test that it would be neat to have that possibility in the shell.

flake.nix
```diff
@@ -0,0 +1,45 @@
...
+              # Libraries
+              gmp llvmPackages.openmp
+              # Tests
+              nodejs_18
+          ];
+          shellHook = ''
+              export NIMBLE_DIR=$HOME/.nimble
+              export RUSTFLAGS="-C target-cpu=native"
+              export CXXFLAGS="-std=c++17 -march=native -mtune=native -msse -msse2 -msse3 -mssse3 -msse4.1 -msse4.2 -mavx -mavx2 -fopenmp"
+              export CFLAGS="-march=native -mtune=native -msse -msse2 -msse3 -mssse3 -msse4.1 -msse4.2 -mavx -mavx2"
```
jakubgs, line 36: Why do we need these? Were they recommended by devs? If it is it probably should be set in the `Makefile`.

flake.nix
```diff
@@ -0,0 +1,45 @@
...
+              mkdir -p $NIMBLE_DIR
+              echo "Codex build environment loaded."
+              echo "GCC path: $(which gcc)"
+              echo "G++ path: $(which g++)"
```
jakubgs, line 40: What's the utility of this?

jakubgs (approved):
Tested locally, works well.

flake.nix
markoburcul: Maybe just informative to the one who will be developing the code within the project

flake.nix
markoburcul: I didn't ask(I will) but without them build would fail

jakubgs: That's weird. If it's required for builds it should be in `Makefile`. It would be good to know what part of this is actually necessary. I'd rather avoid [cargo culting](https://en.wikipedia.org/wiki/Cargo_cult) stuff into config files that are already hard to follow for most people.

markoburcul: I've moved the CXXFLAGS env variable to makefile with **minimum** number of compiler flags in order for build to work. I'll add someone from codex team so they can see this change and comment on it.

markoburcul:
While testing the compiling with system nim compiler(`make USE_SYSTEM_NIM=1`), I realized the [nim](https://search.nixos.org/packages?channel=24.05&from=0&size=50&sort=relevance&type=packages&query=nim) from nixpkgs is version 2.0.4 and we can't compile the code with it. I also tried [nim1](https://search.nixos.org/packages?channel=24.05&from=0&size=50&sort=relevance&type=packages&query=nim) which is version 1.6.20 (much closer to the one we use in the Makefile which is a commit dating after official 1.6.20 release) and it failed. Only with [nim-unwrapped-1](https://search.nixos.org/packages?channel=24.05&from=0&size=50&sort=relevance&type=packages&query=nim) the compiling is successfull so I added it to the list of packages in the shell.

gmega:
I'm going to merge this as the people who use Nix seem to be happy with it.

flake.nix
gmega: I never had to set anything of the kind to build Codex, I wonder why this is needed here.

gmega:
@markoburcul I'm probably missing a lot of context as to how this works, but our build system should build the Nim version that's required by Codex. You shouldn't, indeed, use an external compiler as that might not work. Can't we just lean on that?

markoburcul:
> @markoburcul I'm probably missing a lot of context as to how this works and what the constraints are, but our build system should build the Nim version that's required by Codex. You shouldn't, indeed, use an external compiler as that might not work. Can't we just lean on that?

Using `nix shell` to build the code ensures consistency across environments by specifying **exact dependencies and versions**, including the Nim compiler. Adding a specific Nim compiler version in nix doesn’t force its use; it just makes that version available in the environment. You could use it by specifying `USE_SYSTEM_NIM=1` flag when running `make`. What I was more interested in was you're opinion on these CXXFLAGS I'm settng as default in the Makefile, because they are required to build `codex` without errors.

If you lack more context on this you can check out the task [here](https://github.com/codex-storage/nim-codex/issues/940).

jakubgs:
@gmega are you using a very specific version of Nim compiler because you need very specific patches. Or did you just copy what `nimbus-eth2` had like a year ago or more than that without thinking much about it and now are just continuing with it?

PR #955, opened 2024-10-16
Download API upgrade

benbierens:
- Download stream moved to `/data/CID/network/stream`
- Start download without stream: `/data/CID/network`
- Download only manifest: `/data/CID/network/manifest`

Updated openAPI.yaml and added tests.

### DIFF
--- a/codex/rest/api.nim
+++ b/codex/rest/api.nim
@@ -209,7 +209,7 @@ proc initDataApi(node: CodexNodeRef, repoStore: RepoStore, router: var RestRoute
       await node.retrieveCid(cid.get(), local = true, resp=resp)
 
   router.api(
-    MethodGet,
+    MethodPost,
     "/api/codex/v1/data/{cid}/network") do (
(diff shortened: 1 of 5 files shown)


codex/rest/api.nim
```diff
@@ -209,6 +211,43 @@ proc initDataApi(node: CodexNodeRef, repoStore: RepoStore, router: var RestRoute
   router.api(
     MethodGet,
     "/api/codex/v1/data/{cid}/network") do (
+      cid: Cid, resp: HttpResponseRef) -> RestApiResponse:
```
gmega, line 214: :yellow_circle: OK this is maybe nitpicking but since this generates side-effects at the server (and returns no data), shouldn't this be a POST?

gmega (approved):
LGTM, left minor comment.

codex/rest/api.nim
benbierens: A very good point. Fixing now!

PR #963, opened 2024-10-22
fix(slot-reservations): Avoid slot filled cancellations

emizzle:
This PR does two things which we have seen cause issues in the testnet:
1. Avoids cancelled `SaleReserving` and `SaleFilling` states when slots are filled. The cancellations were bubbling into `nim-ethers` and causing issues with nonce management.
2. Improves logging for situations where a sale should be ignored instead of being considered an error, including when a reservation is not allowed and when a slot was filled by another host.


AuHau (approved):
LGTM! Thanks a lot 🙏

PR #968, opened 2024-10-24
chore: update dependencies, especially nim-ethers to chronos v4 compatible version

emizzle:
This fixes the nonce issue that has been plaguing the testnet, causing banked up pending transactions and preventing requests from starting.

Bumps the following dependencies:
- nim-ethers to commit 6523e70eafecc3b9728a491058818e685bd384fb which contains a pinned nim-eth version. This is to be replaced by a versioned library, so it will be pinned to a particular version. There is a crucial fix in this version of ethers that fixes nonce management which is causing issues in the Codex testnet. It also contains a fixes for:
  - recreating subscription filters when not found
  - `JsonNode.items` symbol not found
- nim-json-rpc to v0.4.4
- nim-json-serialization to v0.2.8
- nim-serde to v1.2.2
- nim-serialization to v0.2.4
- codex-contracts-eth
  - currently on branch to support https://github.com/codex-storage/codex-contracts-eth/pull/198 (should be merged first)

### DIFF
--- a/tests/contracts/time.nim
+++ b/tests/contracts/time.nim
@@ -1,4 +1,5 @@
 import pkg/ethers
+import pkg/serde
 
 proc currentTime*(provider: Provider): Future[UInt256] {.async.} =
   return (!await provider.getBlock(BlockTag.pending)).timestamp
--- a/tests/contracts/time.nim
+++ b/tests/contracts/time.nim
@@ -1,5 +1,5 @@
 import pkg/ethers
-import pkg/serde
+import pkg/serde/json
 
 proc currentTime*(provider: Provider): Future[UInt256] {.async.} =
   return (!await provider.getBlock(BlockTag.pending)).timestamp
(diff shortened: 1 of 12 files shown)


AuHau:
Wuhuuu big congratz 🎉

AuHau (approved):
LGTM, thanks a lot for doing this! 🙏

tests/contracts/time.nim
```diff
@@ -1,4 +1,5 @@
 import pkg/ethers
+import pkg/serde/json
```
benbierens, line 2: Is this used?

tests/tools/cirdl/testcirdl.nim
```diff
@@ -2,10 +2,11 @@ import std/os
...
 import std/options
 import pkg/chronos
-import codex/contracts
-import ../../integration/marketplacesuite
+import pkg/codex/contracts
+import ../../asynctest
+import ../../contracts/deployment
 
-marketplacesuite "tools/cirdl":
+suite "tools/cirdl":
```
benbierens, line 9: The test starts circuit-downloader process, which tries to talk to an RPC endpoint at "ws://localhost8545".
Is this going to work without marketplacesuite?

PR #969, opened 2024-10-24
chore(openapi): add missing fields for debug info endpoint  

2-towns:


gmega (approved):
I've skimmed through the json serialization and this seems to match, so I'm going to trust that this actually describes what we're returning :roll_eyes:

PR #971, opened 2024-10-25
Remove duplicate cors header for download network endpoint

2-towns:
Same as #970 but for download network

### DIFF
--- a/codex/rest/api.nim
+++ b/codex/rest/api.nim
@@ -222,10 +222,6 @@ proc initDataApi(node: CodexNodeRef, repoStore: RepoStore, router: var RestRoute
           Http400,
           $cid.error(), headers = headers)
 
-      if corsOrigin =? allowedOrigin:
-        resp.setCorsHeaders("GET", corsOrigin)
-        resp.setHeader("Access-Control-Headers", "X-Requested-With")
-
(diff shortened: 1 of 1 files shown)


dryajov (approved):

dryajov:
Interesting, why did we add that in the first place and why are we removing it now?

2-towns:
> Interesting, why did we add that in the first place and why are we removing it now?

I guess it was an attempt to fix a potential issue with CORS, but nim-presto sometimes adds the Access-Control-Allow-Origin header itself. In that specific case, as well as in #970, the headers were duplicated, resulting in an error in the browser.

2-towns:
Example for error in the browser: 

```
Access to fetch at 'http://localhost:8080/api/codex/v1/data/zDvZRwzmB8eczXgrUfBVeh57TKVTRNSXStiEiGAKEzAaVe2Zgeui/network' from origin 'http://127.0.0.1:5173' has been blocked by CORS policy: The 'Access-Control-Allow-Origin' header contains multiple values '*, *', but only one is allowed. Have the server send the header with a valid value, or, if an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.
```

PR #972, opened 2024-10-25
chore(openapi): update the openapi spec for the manifest object 

2-towns:


gmega (approved):
LGTM

benbierens:
The newly added manifest fields (filename, etc) seem to some times return null, and crashing deserializers generated from the openAPI yaml. Can we fix this? should the new fields be optional? or have default empty-string values?

PR #975, opened 2024-10-28
Asyncify the Tree Building

benbierens:
Investigation into node stability shows that tree-building is blocking the main execution loop for long periods of time in some use-cases. This PR breaks up the execution time by asyncifying the tree-building call chain.

### DIFF
--- a/codex/merkletree/codex/codex.nim
+++ b/codex/merkletree/codex/codex.nim
@@ -145,16 +145,18 @@ func compress*(
   mhash.coder(@x & @y & @[ key.byte ], digest)
   success digest
 
-func init*(
+proc init*(
   _: type CodexTree,
   mcodec: MultiCodec = Sha256HashCodec,
-  leaves: openArray[ByteHash]): ?!CodexTree =
+  leaves: seq[ByteHash]): Future[?!CodexTree] {.async.} =
 
   if leaves.len == 0:
     return failure "Empty leaves"
 
+  without mhash =? mcodec.mhash(), error:
+    return failure error
+
   let
-    mhash = ? mcodec.mhash()
     compressor = proc(x, y: seq[byte], key: ByteTreeKey): ?!ByteHash {.noSideEffect.} =
       compress(x, y, key, mhash)
     Zero: ByteHash = newSeq[byte](mhash.size)
@@ -165,33 +167,42 @@ func init*(
   var
     self = CodexTree(mcodec: mcodec, compress: compressor, zero: Zero)
 
-  self.layers = ? merkleTreeWorker(self, leaves, isBottomLayer = true)
+  without layers =? (await merkleTreeWorker(self, leaves, isBottomLayer = true)), error:
+    return failure error
+
+  self.layers = layers
   success self
 
-func init*(
+proc init*(
   _: type CodexTree,
-  leaves: openArray[MultiHash]): ?!CodexTree =
+  leaves: seq[MultiHash]): Future[?!CodexTree] {.async.} =
 
   if leaves.len == 0:
     return failure "Empty leaves"
 
   let
     mcodec = leaves[0].mcodec
-    leaves = leaves.mapIt( it.digestBytes )
+    leafBytes = leaves.mapIt( it.digestBytes )
 
-  CodexTree.init(mcodec, leaves)
+  await CodexTree.init(mcodec, leafBytes)
 
-func init*(
+proc init*(
   _: type CodexTree,
-  leaves: openArray[Cid]): ?!CodexTree =
+  leaves: seq[Cid]): Future[?!CodexTree] {.async.} =
+
   if leaves.len == 0:
     return failure "Empty leaves"
 
-  let
-    mcodec = (? leaves[0].mhash.mapFailure).mcodec
-    leaves = leaves.mapIt( (? it.mhash.mapFailure).digestBytes )
+  without mhash =? leaves[0].mhash.mapFailure, error:
+    return failure error
 
-  CodexTree.init(mcodec, leaves)
+  var hashes = newSeq[seq[byte]]()
+  for leaf in leaves:
+    without hash =? leaf.mhash.mapFailure, error:
+      return failure error
+    hashes.add(hash.digestBytes)
+
+  await CodexTree.init(mhash.mcodec, hashes)
 
 proc fromNodes*(
   _: type CodexTree,
--- a/codex/slots/builder/builder.nim
+++ b/codex/slots/builder/builder.nim
@@ -161,8 +161,7 @@ proc buildBlockTree*[T, H](
   if blk.isEmpty:
     success (self.emptyBlock, self.emptyDigestTree)
   else:
-    without tree =?
-      T.digestTree(blk.data, self.cellSize.int), err:
+    without tree =? (await T.digestTree(blk.data, self.cellSize.int)), err:
       error "Failed to create digest for block", err = err.msg
       return failure(err)
 
@@ -213,7 +212,7 @@ proc buildSlotTree*[T, H](
     error "Failed to select slot blocks", err = err.msg
     return failure(err)
 
-  T.init(cellHashes)
+  await T.init(cellHashes)
 
 proc buildSlot*[T, H](
   self: SlotsBuilder[T, H],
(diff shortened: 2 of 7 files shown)


benbierens:
Image from this branch passes upload/download dist-tests.
This branch + async-profile with defect-assert on segmentTimes larger than 100 milliseconds, passes upload/download dist-tests: AKA: There are no operations in the basic upload/download flow with blocking execution times longer than 100 milliseconds.

This means that the cause of failing downloads cannot be explained by long-blocking operations.
Marketplace-flow with async-profile for detection of long-blocking operations is still pending.

tests/codex/node/testcontracts.nim
```diff
@@ -91,6 +91,8 @@ asyncchecksuite "Test Node - Host contracts":
 
     protected = (await erasure.encode(manifest, 3, 2)).tryGet()
     builder = Poseidon2Builder.new(localStore, protected).tryGet()
+    (await builder.init()).tryGet()
```
gmega, line 94: Hm... this looks quite weird (doing a get on a void) but OK, I don't have better ideas. :-)

gmega (approved):
OK, this looks fine, with the caveats that:

1. it's hackish, and IMHO we should revert it once we can run those easily on separate threads, which would be the "proper" solution;
2. I am not so sure about the new/init API. It doesn't seem to comply with the expected semantics;
3. we're doing some behavior-changing transformations by catching the exceptions and returning results.

Would be wise to get another pair of eyes on this as we've discussed earlier today, I'm nevertheless approving it.

codex/merkletree/codex/codex.nim
```diff
@@ -145,16 +145,18 @@ func compress*(
   mhash.coder(@x & @y & @[ key.byte ], digest)
   success digest
 
-func init*(
+proc init*(
```
emizzle, line 148: I think we might need to avoid making this `init` async so we can avoid using `waitFor` later on... instead this is a good pattern to follow: 
https://github.com/codex-storage/nim-ethers/blob/80b2ead97ce32cca74cef4a4501ab97106cb578b/ethers/providers/jsonrpc.nim#L50-L86

Then later, we can use `CodexTree.init` synchronously, and when needed, we can `let layers = await tree.layers`

emizzle (changes requested):
We've hit some significant issues using `waitFor` in the past, and we need to avoid it like the plague (cc @gmega). I outlined a pattern that might help avoid the usage.

codex/merkletree/codex/codex.nim
```diff
@@ -145,16 +145,18 @@ func compress*(
...
+  leaves: seq[ByteHash]): Future[?!CodexTree] {.async.} =
 
   if leaves.len == 0:
     return failure "Empty leaves"
 
+  without mhash =? mcodec.mhash(), error:
+    return failure error
+
   let
-    mhash = ? mcodec.mhash()
```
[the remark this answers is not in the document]
emizzle: > we could simply preserve this behavior and propagate the exception using a Chronos raises clause

That would actually change the behaviour. `= ?` will either assign a value to `mhash` or it will return a `Result` with an error to the closure. By raising an exception, and specifying that exception type in `{.async: (raises: [...]).}`, we're not returning a Result, but raising an exception. 

Imo, the proposed change here does not alter the behaviour.

codex/merkletree/codex/codex.nim
```diff
@@ -165,33 +167,42 @@ func init*(
...
-    mcodec = (? leaves[0].mhash.mapFailure).mcodec
-    leaves = leaves.mapIt( (? it.mhash.mapFailure).digestBytes )
+  without mhash =? leaves[0].mhash.mapFailure, error:
+    return failure error
+
+  var hashes = newSeq[seq[byte]]()
+  for leaf in leaves:
+    without hash =? leaf.mhash.mapFailure, error:
+      return failure error
+    hashes.add(hash.digestBytes)
```
emizzle, line 203: Did this really need a rewrite?

codex/slots/builder/builder.nim
```diff
@@ -305,6 +304,29 @@ proc buildManifest*[T, H](self: SlotsBuilder[T, H]): Future[?!Manifest] {.async.
     self.cellSize,
     self.strategy.strategyType)
 
+proc init*[T, H](self: SlotsBuilder[T, H]): Future[?!void] {.async.} =
```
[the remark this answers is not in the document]
emizzle: Agree that an `init` call can be confusing when not used for construction. In general, Codex follows the convention of using `init` for constructing object types and `new` for constructing `ref` types, as per @dryajov's comment: https://discord.com/channels/895609329053474826/895622391349272637/1086322836001534072

Maybe an easy fix is to call this `initialise` instead?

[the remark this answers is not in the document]
emizzle: A lot of this seems like it was replacing `let x = ? someResultFunc()...` with `without x =? someResultFunc()`, which doesn't seem necessary to me. In other words, it feels like we should make the switch to `async` without changing any other semantics if possible. It's easy to get into trouble modifying lines "while we're here" as there can be unintended consequences, especially in nim and especially with this many changes in a PR.

tests/codex/node/testcontracts.nim
emizzle: Yeah, it's odd... but it's a way to force an exception in case there is no value.

Another way (not sure if it's "better"):
```nim
if (await builder.init()).isErr: fail()
```
But this doesn't give much output if it does fail, so an even more explicit way is:
```nim
if error =? (await builder.init()).errorOption:
  echo error.msg
  fail()
```

emizzle (commented):
I'll remove my requested changes as I realise the call sites using `waitFor` are in the tests only. However, that is not to say that this pattern of using async `init` won't bite us later on. I just don't want to stand in the way of progress.

My overall comment for these changes is that because we are making so many changes here and it feels like we're trying to rush this in, we really should limit the scope of changes by changing as little as possible. If i understood correctly, the main idea of this PR was the make some things asynchronous. However, we've also replaced `let x = ? someResultFunc()...` with `without x =? someResultFunc()` in quite a few places. This is one of those things that seems innocuous, but can also bite us later on (especially in nim).

I am also not a big fan of "let's do this for now and revert later" kind of changes, mostly because there may be patches affecting this code in the meantime which are hard to merge, and also sometimes reversion are forgotten or partially done.

emizzle (approved):

emizzle:
Ok, so I didn't approve these changes. I merely wanted to *remove* my changes requested 🤦

benbierens:
All right. Let's put a break on this one. Here's why:
My plan with this PR was never to create a quick-and-dirt solution intended to be a temporary fix. I realized that threading might come along and make these change unnecessary. (I'm not entirely convinced of this.) But I was never counting on that. If you guys find these changes hacky and only suitable for temporarily hiding an issue, then let's _NOT_ merge this. I want to fix this correctly, as if it's going to be in production like this indefinitely.

The change I made is to make the tree-building async by adding a small async-sleep in the loop so the process can be interrupted for, for example, the handling of network traffic. That's it. All other changes I made are required downstream changes to make stuff compile and work again. Yes, there are a lot of those. There's not much we can do about that unless we are considering restructuring a lot more code, which is the opposite of making small changes.

The new `init` method is required because we don't have async `new` methods. The replacing of ` = ? ` with `without` is needed in methods that have become async.

codex/slots/builder/builder.nim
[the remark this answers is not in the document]
benbierens: I wouldn't have changed them if I didn't have to. ` = ? ` doesn't compile in async procs.

tests/codex/node/testcontracts.nim
benbierens: (await ...).tryGet() calls are all over the testing code. It fails the test when the result is an error. How would you like to see this changed?

codex/merkletree/codex/codex.nim
benbierens: I wouldn't figure out a cleaner way to unroll the results. If you've got one, I'm eager to learn.

gmega:
I don't think this is getting merged anymore especially with the threading work, so I'm thinking of closing it. Is it OK if I do that @benbierens @dryajov?

benbierens:
Please keep in mind that putting the erasure-coding and/or slot-building code on a different thread won't solve this everywhere. Our app will still freeze for longer than the network traffic will tolerate whenever a significantly large tree is built. This will happen and it will cause issues even if both of the above topics have their own threads.

PR #976, opened 2024-10-29
feat(integration): add ability to search logs in integration test nodes + slot reservations test

emizzle:
Adds the ability for integration tests to search and assert on the log output.

Adds an integration test that checks for "nonce too high" in the logs (which is the issue plaguing the testnet right now).


AuHau (approved):
Nice

dryajov (approved):
LGTM

PR #981, opened 2024-11-01
feat(rest): add default content-disposition header for stream download 

2-towns:
This PR adds `Content-Disposition` header, even if the filename does not exist in the metadata, to force the browser to download the file.


dryajov (approved):
looks useful :)

PR #982, opened 2024-11-01
Add ETH_PRIVATE_KEY to Docker entrypoint

veaceslavdoina:
This is a quick PR to add `ETH_PRIVATE_KEY` to the Docker entrypoint for variables names consistency.

During transition period we can use both, `PRIV_KEY` and `ETH_PRIVATE_KEY`, to the pass Ethereum private key.

After we will change all the code to use new variable, the old one will be removed.

### DIFF
--- a/docker/docker-entrypoint.sh
+++ b/docker/docker-entrypoint.sh
@@ -48,6 +48,7 @@ fi
 for key in PRIV_KEY ETH_PRIVATE_KEY; do
   keyfile="private.key"
   if [[ -n "${!key}" ]]; then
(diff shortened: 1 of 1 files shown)


emizzle (commented):
Should we maybe add a depreciation warning of PRIV_KEY is used?

veaceslavdoina:
Added, thanks
```
PRIV_KEY variable is deprecated and will be removed in the next releases, please use ETH_PRIVATE_KEY instead!
Private key set
```

emizzle (approved):
Nice!!

PR #985, opened 2024-11-04
feat: expose underlying nim-ethers errors to logs

AuHau:
Brings in a version of nim-ethers where subscription callbacks are returned Result, which will give an error when there is some underlying problem in the subscription mechanisms. This PR exposes these errors through logging.

### DIFF
--- a/tests/integration/testecbug.nim
+++ b/tests/integration/testecbug.nim
@@ -33,9 +33,9 @@ marketplacesuite "Bug #821 - node crashes during erasure coding":
     let cid = clientApi.upload(data).get
 
     var requestId = none RequestId
-    proc onStorageRequested(eventRes: ?!StorageRequested)=
-      if event =? eventRes:
-        requestId = event.requestId.some
+    proc onStorageRequested(eventResult: ?!StorageRequested)=
+      assert not eventResult.isErr
+      requestId = some (!eventResult).requestId
 
     let subscription = await marketplace.subscribe(StorageRequested, onStorageRequested)
 
--- a/tests/integration/testecbug.nim
+++ b/tests/integration/testecbug.nim
@@ -33,8 +33,9 @@ marketplacesuite "Bug #821 - node crashes during erasure coding":
     let cid = clientApi.upload(data).get
 
     var requestId = none RequestId
-    proc onStorageRequested(event: StorageRequested) {.raises:[].} =
-      requestId = event.requestId.some
+    proc onStorageRequested(eventRes: ?!StorageRequested)=
+      if event =? eventRes:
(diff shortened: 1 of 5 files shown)


tests/integration/testecbug.nim
```diff
@@ -33,8 +33,9 @@ marketplacesuite "Bug #821 - node crashes during erasure coding":
     let cid = clientApi.upload(data).get
 
     var requestId = none RequestId
-    proc onStorageRequested(event: StorageRequested) {.raises:[].} =
-      requestId = event.requestId.some
+    proc onStorageRequested(eventRes: ?!StorageRequested)=
+      if event =? eventRes:
+        requestId = event.requestId.some
```
emizzle, line 38: Wdyt about raising a Defect (and fail the test) if there's an error in the callback?

```suggestion
    proc onStorageRequested(event: ?!StorageRequested)=
      requestId = some event.get.requestId
```

emizzle (approved):
Nice work. My main question is more around should we do more than just log these errors? Are they bad enough to raise a Defect (and crash the node)? Or should we add metrics and monitor how often they happen? 

I'll approve now since the suggestions are not really blocked.

markspanbroek (approved):

AuHau:
@emizzle I am unsure about raising Defects. I would for now leave it simply on Error log level and we should be able to detect those if something is fishy. Depends if and what we will detect, then I would act accordingly.

AuHau:
> > @emizzle I am unsure about raising Defects. I would for now leave it simply on Error log level and we should be able to detect those if something is fishy. Depends if and what we will detect, then I would act accordingly.
> 
> I have to disagree that monitoring logs is sufficient. Manually combing through logs is tedious at best, especially when all log levels are enabled. We'd need to do some of regular search on kibana, but that's also not really practical. At the very least, a metric would be a better way to keep an eye this.

I have to disagree with your disagreeing 🤪 These errors will be logged as **error** level.  IMHO we should keep a tight eye on any errors popping up in logs (and generally, we should do logs monitoring) . You can easily set up a dashboard in Kibana that will show error logs occurrence (even related to these issues based on logs messages). IMHO this leads to the same result as metrics, but I agree that to some degree metrics give you an easier way to monitor these things as for example logs messages can change in the future (but then somebody can change the metrics key name as well). 

I am not solely opposing this, I just wanted to wrap this because this work was (again) going forever. Feel free to create an issue and we can tweak this later on.

emizzle:
> The main goal of this work was to expose in one way or other errors from nim-ether subscriptions which it does.

Yep, good point and it achieved the goal very well!

PR #988, opened 2024-11-07
limit amount of slot queue workers to 1

markspanbroek:
reason: proof generation does not happen in
separate threads yet, so to avoid taking too
long to provide initial storage proofs we only
pick up one slot at a time


AuHau (approved):

tests/integration/testmarketplace.nim
```diff
@@ -24,15 +24,15 @@ twonodessuite "Marketplace", debug1 = false, debug2 = false:
...
-      duration=10*60.u256,
+      duration=20*60.u256,
       reward=400.u256,
       proofProbability=3.u256,
-      expiry=5*60,
+      expiry=10*60,
       collateral=200.u256,
-      nodes = 5,
-      tolerance = 2).get
+      nodes = 3,
```
emizzle, line 32: Out of curiosity, is the motivation of limiting the number of nodes to hasten the completion of this test? In the past, having 5 nodes has surfaced a few issues. In saying that, I will eventually finish this PR which will have 5 or 6 nodes, so at least we'll have one test (in the future with 5/6 nodes): https://github.com/codex-storage/nim-codex/pull/976

emizzle (approved):
Thanks Mark!

emizzle:
Another option to speed up tests: pass number of workers as a cli param. More work of course...

tests/integration/testmarketplace.nim
markspanbroek: Yes, it now takes about 4 minutes to fill 3 slots. Something still is off that this takes so long (most likely the downloading)

markspanbroek:
> Another option to speed up tests: pass number of workers as a cli param. More work of course...

This is just a temporary fix. We should increase the number of workers as soon as we have threading for creating proofs.

tests/integration/testmarketplace.nim
emizzle: Hmm, even on the testnet I think downloading is quick... locally as an example, download takes from 12:01:02.236 to 12:01:02.346 in one of the integration tests (8 blocks)

emizzle: The proof generation might take some time. Also, I noticed on the testnet that we're doing local verification of the proofs, which is not needed and takes time on each node.

tests/ethertest.nim
```diff
@@ -16,7 +16,7 @@ template ethersuite*(name, body) =
     var snapshot: JsonNode
 
     setup:
-      ethProvider = JsonRpcProvider.new("ws://localhost:8545")
+      ethProvider = JsonRpcProvider.new("http://localhost:8545")
```
emizzle, line 19: I think this is the right change, but we also might run into the `poll` future failing... (something to be fixed in ethers)

tests/integration/testmarketplace.nim
markspanbroek: Yes, you're right. I'm running this test with some more logging on my machine and it spends most of its time in the SaleInitialProving state.

tests/ethertest.nim
markspanbroek: You're right, the contract tests started failing with this change. Strange that works in the integration tests, but not in the contract tests. I'll revert the change on this line for now.

emizzle: Hard to know if the reason they are failing is due to the `poll` failing issue though. In saying that, I'd like to get a fix in for the `poll()` fail. Trying to work on it now, but unsure if I'll get something in before my flight leaves in a few hours.

AuHau:
> I guess we really need to fix the issue with poll in nim-ethers before we can continue with this PR ☹️

Hmm now I am not sure which one exactly? The one @emizzle found related to Future being cancelled?

markspanbroek:
> Hmm now I am not sure which one exactly? The one @emizzle found related to Future being cancelled?

Yes, I think so. It seems that subscription handlers not being called is causing the unpredictable test failures. When using only websockets the tests actually work quite reliably up to the point that a test takes longer than 5 minutes. When using http and polling the tests fail at random points. My suspicion is that this is caused by what Eric noticed; that an exception is being raised in the `poll` function, that silently stops the polling.

AuHau:
I see, there is the https://github.com/codex-storage/nim-codex/pull/985 which might show the error. It could be worth rebase this PR on top of that to test your hypothesis.
It is unfortunately not finished, but it should be working, there are just some polishing needed in order to merge it. I will try to finish those tomorrow, I won't make it today unfortunately.

emizzle:
So I've got somewhere on a solution for this, but it's not complete. The first part of the solution is to use the `then` util to attach callbacks to `poll`, allowing us to catch exceptions and then propagate them to the synchronous closure. 

The second part, which I'm not sure about yet, is understanding whether or not we need to reconnect the `RpcHttpClient` when this exception is raised (this particular exception only or all exceptions)?

The third part is trying to understand if we need this reconnection logic from part 2 in any other areas of ethers (including websockets).

emizzle:
@markspanbroek here's my work on keeping the polling loop alive: https://github.com/codex-storage/nim-ethers/tree/fix/keep-polling-alive.

### NOTES
1. We cannot consume futures like we were without `asyncSpawn`ing or `await`ing them. This is where the `.then/.catch` comes in: it's effectively an `asyncSpawn` but does not raise a `Defect` and instead exposes the exceptions.
2. We only restart the polling loop when an `RpcPostError` is encountered
3. We raise a `Defect` on other errors, but i'm not so sure this is the right approach. We might need to handle all possible errors differently.
4. There may be other situations in ethers where this happens?
5. The `then.nim` addition was taken from codex and was simplified a bit. It needs to be pulled into its own lib and used in both codex and ethers. The version in this branch is the most recent and will be used for the lib.

markspanbroek:
> here's my work on keeping the polling loop alive

Thanks @emizzle, I think I've now got a version of http polling that works based on these ideas. I also found 4 other bugs that made the tests unstable.

I'll try to get PRs for all of this in tomorrow.

markspanbroek:
On Linux and MacOS the tests run stable now. The Windows tests are running so slow that they are even failing on the fairly simple contract tests that do not involve proofs.

I'm installing a Windows VM to see if I can reproduce this on my machine.

markspanbroek:
Moved most of fixed into #993, and rebased this PR on top of it.

I'm not sure whether it still makes sense to limit the amount of slot queue workers now that I've done some measurements on proof generation. It doesn't take that much time (~3 seconds).

gmega:
Should this be merged? Should this be closed? :slightly_smiling_face: @emizzle @markspanbroek

Issue #992, opened 2024-11-13
[BUG] intermittent leveldb iterator crash


markspanbroek:
Observed this crash several times while running the unit tests on windows in the CI:

```Assertion failed: dummy_versions_.next_ == &dummy_versions_```

A quick search seems to indicate that this is likely caused by a leveldb iterator not being properly closed.

Example CI run:
https://github.com/codex-storage/nim-codex/actions/runs/11731259014/job/32682716194?pr=988#step:4:946

benbierens:
I'm looking into this now.

benbierens:
While running the test a million times:
![Image](https://github.com/user-attachments/assets/c066c4d7-19c7-4b90-8f23-f896183553d0)

gmega:
Is this still happening @markspanbroek @benbierens?

benbierens:
I haven't seen it since #995 got merged. But I also haven't been running the tests in giant loops.
Are you suspecting some instability somewhere?

gmega:
We haven't seen this showing up in more than 6 months, so I'm closing it.

PR #995, opened 2024-11-21
Fixes race in testsales

benbierens:
Related to (but probably doesn't fix) #992


markspanbroek (approved):

emizzle (commented):
Thanks for doing this PR, Ben, much appreciate! 🙏  However, this feels a bit flaky to me. Wdyt about the suggestion?

tests/codex/sales/testsales.nim
```diff
@@ -190,7 +190,7 @@ asyncchecksuite "Sales":
 
   proc allowRequestToStart {.async.} =
     # wait until we're in initialproving state
-    await sleepAsync(10.millis)
+    check eventually clock.isWaiting
```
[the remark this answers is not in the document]
benbierens: I like this, it seems a lot more reusable. I think isWaiting for a mockClock makes perfect sense. I'm giving this a try.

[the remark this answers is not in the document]
emizzle: Be aware though that `allowRequestToStart` may not actually have been checking that the state was in the initial proving state, so if you update it, some tests may break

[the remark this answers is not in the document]
benbierens: I have applied your suggestion and I like it. The tests are all happy.
I know every keystroke I make reduces the likelihood that I can get this merged, BUT I did replace two more sleeps with check-eventually. Plz check if you agree with them.

tests/codex/sales/testsales.nim
```diff
@@ -188,9 +189,14 @@ asyncchecksuite "Sales":
     await repoTmp.destroyDb()
     await metaTmp.destroyDb()
 
+  proc isInState(idx: int, state: string): Future[bool] {.async.} =
```
emizzle, line 192: Does this need to be async?

emizzle (approved):
Other than the async comment, it looks great! 🙏

tests/codex/sales/testsales.nim
benbierens: Yes: check-eventually is needed to give sales time to load the agent.

emizzle: Right, but shouldn't we just early return if `sales.agent.len == 0`? I don't think there's a need for `check eventually` in the `isInState` proc, since its usage is used in a `check eventually isInState...`.

emizzle: Creating a PR for this: https://github.com/codex-storage/nim-codex/pull/1021

PR #1002, opened 2024-11-26
Fix race condition in testdiscoveryengine

benbierens:
Requesting the same CID sometimes causes a worker to discard the request while another worker has the same CID in-flight.

### DIFF
--- a/tests/codex/blockexchange/discovery/testdiscoveryengine.nim
+++ b/tests/codex/blockexchange/discovery/testdiscoveryengine.nim
@@ -111,10 +111,11 @@ asyncchecksuite "Test Discovery Engine":
         minPeersPerBlock = minPeers)
       want = newAsyncEvent()
 
+    var pendingCids = newSeq[Cid]()
     blockDiscovery.findBlockProvidersHandler =
       proc(d: MockDiscovery, cid: Cid): Future[seq[SignedPeerRecord]] {.async, gcsafe.} =
-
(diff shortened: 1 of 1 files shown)


tests/codex/blockexchange/discovery/testdiscoveryengine.nim
```diff
@@ -126,8 +127,12 @@ asyncchecksuite "Test Discovery Engine":
         want.fire()
 
     await discoveryEngine.start()
+    var idx = 0
     while peerStore.len < minPeers:
-      discoveryEngine.queueFindBlocksReq(@[blocks[0].cid])
```
gmega, line 130: Ugh, but can't this happen in production settings too? Like, us requesting the same block more than once? Will the requests then maybe get cancelled sometimes?

gmega (approved):
I think I understand the change, my only worry is that this is a bug that also happens in production? If this is due to our test `findBlocksProviderHandler`, then LGTM. Will approve and you decide if it needs more work or not.

tests/codex/blockexchange/discovery/testdiscoveryengine.nim
benbierens: If you request discovery for a CID that's already in progress, the discovery engine will ignore it. This is the behavior we want. It's just that the test was running into this sometimes when it didn't want to.

PR #1003, opened 2024-11-26
nix: make derivation and update shell

markoburcul:
Create a structure for nix files. Add the derivation file which is using system Nim to compile Codex. Move shell definition to a specific file.

Referenced issue: https://github.com/codex-storage/nim-codex/issues/940

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -3,32 +3,40 @@
   
   inputs = {
     nixpkgs.url = "github:NixOS/nixpkgs/nixos-24.05";
+    circom-compat.url = "github:codex-storage/circom-compat-ffi";
   };
 
-  outputs = { self, nixpkgs }:
+  outputs = { self, nixpkgs, circom-compat}:
     let
-      supportedSystems = [
+      stableSystems = [
         "x86_64-linux" "aarch64-linux"
         "x86_64-darwin" "aarch64-darwin"
       ];
-      forAllSystems = f: nixpkgs.lib.genAttrs supportedSystems (system: f system);
+      forAllSystems = f: nixpkgs.lib.genAttrs stableSystems (system: f system);
       pkgsFor = forAllSystems (system: import nixpkgs { inherit system; });
     in rec {
+      packages = forAllSystems (system: let
+        circomCompatPkg = circom-compat.packages.${system}.default;
+        buildTarget = pkgsFor.${system}.callPackage ./nix/default.nix {
+          inherit stableSystems circomCompatPkg;
+          src = self;
+        };
+        build = targets: buildTarget.override { inherit targets; };
+      in rec {
+        codex   = build ["all"];
+        default = codex;
+      });
+
       devShells = forAllSystems (system: let
         pkgs = pkgsFor.${system};
-        inherit (pkgs) lib stdenv mkShell;
       in {
-        default = mkShell.override { stdenv = pkgs.gcc11Stdenv; } {
-          buildInputs = with pkgs; [
-              # General
-              git pkg-config openssl lsb-release
-              # Build
-              rustc cargo nimble gcc11 cmake nim-unwrapped-1
-              # Libraries
-              gmp llvmPackages.openmp
-              # Tests
-              nodejs_18
+        default = pkgs.mkShell {
+          inputsFrom = [
+            packages.${system}.codex
+            circom-compat.packages.${system}.default
           ];
+          # Not using buildInputs to override fakeGit and fakeCargo.
+          nativeBuildInputs = with pkgs; [ git cargo nodejs_18 ];
         };
       });
     };
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -1,0 +1,84 @@
+{
+  pkgs ? import <nixpkgs> { },
+  src ? ../.,
+  targets ? ["all"],
+  # Options: 0,1,2
+  verbosity ? 0,
+  # Use system Nim compiler instead of building it with nimbus-build-system
+  useSystemNim ? true,
+  commit ? builtins.substring 0 7 (src.rev or "dirty"),
+  # These are the only platforms tested in CI and considered stable.
+  stableSystems ? [
+    "x86_64-linux" "aarch64-linux"
+    "x86_64-darwin" "aarch64-darwin"
+  ],
+  circomCompatPkg ? (
+    builtins.getFlake "github:codex-storage/circom-compat-ffi"
+  ).packages.${builtins.currentSystem}.default
+}:
+
+let
+  inherit (pkgs) stdenv lib writeScriptBin callPackage;
+  
+  revision = lib.substring 0 8 (src.rev or "dirty");
+
+  tools = callPackage ./tools.nix {};
+in pkgs.gcc11Stdenv.mkDerivation rec {
+  
+  pname = "codex";
+
+  version = "${tools.findKeyValue "version = \"([0-9]+\.[0-9]+\.[0-9]+)\"" ../codex.nimble}-${revision}";
+  
+  inherit src;
+
+  # Dependencies that should exist in the runtime environment.
+  buildInputs = with pkgs; [
+    openssl
+    gmp
+  ];
+
+  # Dependencies that should only exist in the build environment.
+  nativeBuildInputs = let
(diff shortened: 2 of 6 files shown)


nix/default.nix
```diff
@@ -0,0 +1,86 @@
...
+    fakeGit = writeScriptBin "git" "echo ${version}";
+    fakeLsbRelease = writeScriptBin "lsb_release" "echo nix";
+    # Fix for the nim-circom-compat-ffi package that is built with cargo.
+    fakeCargo = writeScriptBin "cargo" "echo ${version}";
+  in
+    with pkgs; [
+      gcc11
+      cmake
+      curl
+      bash
```
jakubgs, line 42: Bash is a part of `stdenv`, and `curl` makes no sense.

nix/default.nix
```diff
@@ -0,0 +1,86 @@
...
+      pkg-config
+      openssl
+      gmp
```
jakubgs, line 45: Aren't we already adding this in `buildInputs`?

nix/default.nix
```diff
@@ -0,0 +1,86 @@
...
+    "x86_64-darwin" "aarch64-darwin"
+  ],
+  circomCompatPkg
+}:
+
+let
+  inherit (pkgs) stdenv lib writeScriptBin callPackage;
+  
+  revision = lib.substring 0 8 (src.rev or "dirty");
+in pkgs.stdenv.mkDerivation rec {
```
jakubgs, line 22: Shouldn't we be enforcing GCC11?
```suggestion
in pkgs.gcc11Stdenv.mkDerivation rec {
```

nix/default.nix
```diff
@@ -0,0 +1,86 @@
...
+  # Dependencies that should only exist in the build environment.
+  nativeBuildInputs = let
+    # Fix for Nim compiler calling 'git rev-parse' and 'lsb_release'.
+    fakeGit = writeScriptBin "git" "echo ${version}";
+    fakeLsbRelease = writeScriptBin "lsb_release" "echo nix";
+    # Fix for the nim-circom-compat-ffi package that is built with cargo.
+    fakeCargo = writeScriptBin "cargo" "echo ${version}";
+  in
+    with pkgs; [
+      gcc11
```
jakubgs, line 39: This should be done using `gcc11Stdenv.mkDerivation`.

jakubgs (changes requested):
Good stuff.

flake.nix
```diff
@@ -3,33 +3,36 @@
   
   inputs = {
     nixpkgs.url = "github:NixOS/nixpkgs/nixos-24.05";
+    circom-compat.url = "github:codex-storage/circom-compat-ffi";
```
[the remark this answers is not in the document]
markoburcul: I did try it with relative path, but the problem is the `.?submodules=1` part when building the derivation. Since we declare the `circom-compat` flake as input when the submodules are not cloned yet, the build fails because the flake file is not present yet on the filesystem. Because of that I am using the GitHub URL here.

flake.nix
```diff
@@ -3,32 +3,42 @@
...
-              # Libraries
-              gmp llvmPackages.openmp
-              # Tests
-              nodejs_18
+        default = pkgs.mkShell {
+          inputsFrom = [
+            packages.${system}.codex
+            circom-compat.packages.${system}.default
           ];
+          buildInputs = with pkgs; [ git nodejs_18 ];
```
jakubgs, line 41: There is one issue with this approach, and that `buildInputs` dome in after `inputsFrom` in `PATH`:
```
[jakubgs@caspair:~/work/nim-codex]$ echo $PATH | tr ':' '\n' | grep git
/nix/store/cn0hd2vyrin50vsqqam6xy5jlqdgal8a-git/bin
/nix/store/vdd8ccp9x69v7kp5g9khq7j43n6dqf16-git-2.44.1/bin
```
So `fakeGit` actually wins due to the order of search. Which is annoying, but can be fixed by using `nativeBuildInputs`:
```suggestion
          # Not using buildInputs to override fakeGit and fakeCargo.
          nativeBuildInputs = with pkgs; [ git cargo nodejs_18 ];
```

nix/default.nix
```diff
@@ -0,0 +1,84 @@
...
+    "x86_64-linux" "aarch64-linux"
+    "x86_64-darwin" "aarch64-darwin"
+  ],
+  circomCompatPkg ? (
+    builtins.getFlake "github:codex-storage/circom-compat-ffi"
+  ).packages.${builtins.currentSystem}.default
+}:
+
+let
+  inherit (pkgs) stdenv lib writeScriptBin callPackage;
```
jakubgs, line 21: You're not actually using the `callPackage` from here.

flake.nix
```diff
@@ -3,33 +3,36 @@
...
         "x86_64-linux" "aarch64-linux"
         "x86_64-darwin" "aarch64-darwin"
       ];
-      forAllSystems = f: nixpkgs.lib.genAttrs supportedSystems (system: f system);
-      pkgsFor = forAllSystems (system: import nixpkgs { inherit system; });
+      forEach = nixpkgs.lib.genAttrs;
+      forAllSystems = forEach stableSystems;
+      pkgsFor = forEach stableSystems (
+        system: import nixpkgs { inherit system; }
+      );
```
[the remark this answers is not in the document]
jakubgs: Still don't get why this change happened.

nix/default.nix
```diff
@@ -0,0 +1,84 @@
...
+  # Avoid Nim cache permission errors.
+  XDG_CACHE_HOME = "/tmp";
+
+  makeFlags = targets ++ [
+    "V=${toString verbosity}"
+    "USE_SYSTEM_NIM=${if useSystemNim then "1" else "0"}"
+  ];
+
+  # Dependencies that should exist in the runtime environment.
+  buildInputs = with pkgs; [
```
jakubgs, line 64: I would move this above `nativeBuildInputs`.

jakubgs (approved):
Aside from a few nitpicks and some trailing whitespace this looks very good.

flake.nix
```diff
@@ -3,32 +3,42 @@
...
+      packages = forAllSystems (system: let
+        circomCompatPkg = circom-compat.packages.${system}.default;
+        buildTarget = pkgsFor.${system}.callPackage ./nix/default.nix {
+          inherit stableSystems circomCompatPkg;
+          src = self;
+        };
+        build = targets: buildTarget.override { inherit targets; };
+      in rec {
+        codex   = build ["all"];
+        default = codex;
```
jakubgs, line 30: One weird thing I found when testing was that if I specify target with `#codex` or `#default` it builds, but if I leave that out then it fails with:
```
 > nix build --print-build-logs --print-out-paths '.?submodules=1'
warning: Git tree '/home/jakubgs/work/nim-codex' is dirty
codex> Running phase: unpackPhase
codex> unpacking source archive /nix/store/k9ri8k88313z6ggkcm4k95aaapqw351i-source
codex> source root is source
codex> Running phase: patchPhase
codex> Running phase: updateAutotoolsGnuConfigScriptsPhase
codex> Running phase: configurePhase
codex> Running phase: buildPhase
codex> build flags: SHELL=/nix/store/516kai7nl5dxr792c0nzq0jp8m4zvxpi-bash-5.2p32/bin/bash all V=2 USE_SYSTEM_NIM=1
codex> Git submodules not found. Running 'git submodule update --init --recursive'.
codex> 0.1.0-dirty
codex> make: Nothing to be done for 'all'.
codex> Running phase: installPhase
codex> cp: cannot stat 'build/codex': No such file or directory
```
Surprisingly `.?submodules=1#` works fine, so that smells like a nix interpreter bug that ignores the submodules arg.

nix/default.nix
markoburcul: I am using it in line 25

jakubgs: Do you mean here?
```
        buildTarget = pkgsFor.${system}.callPackage ./nix/default.nix {
```
No you don't. That's a different `callPackage`.

markoburcul: I meant here:
```nix
tools = pkgs.callPackage ./tools.nix {};
```
but I should get rid of `pkgs` prefix here

PR #1007, opened 2024-11-27
Fixes race condition in initial-proving tests

benbierens:


gmega (approved):
If I get this right, you replaced a timing assumption with a deterministic assumption so yay approved!

PR #1012, opened 2024-11-28
Update nim-ethers

marcinczenko:
We released the new version of nim-ethers and as the changes applied by Adam are braking changes I am just kicking it out here.

Mark suggested to create a new release tag with version 1.0.0 (as we have a braking change and this is what we agreed to do on the first occasion).

I can create that one tomorrow.

To fix the compilation errors coming from the new API, I would need to better understand the intention of introducing the Result type.

Is the intention to do something like this

```nim
proc onEvent(event: ?!StorageRequested) {.upraises:[].} =
  if e =? event:
    callback(e.requestId, e.ask, e.expiry)
```

or

```nim
OnRequest* = proc(id: ?!RequestId,
                  ask: ?!StorageAsk,
                  expiry: ?!UInt256) {.gcsafe, upraises:[].}
```

or something else?


AuHau:
There is my PR open that brings the latest nim-ethers here: https://github.com/codex-storage/nim-codex/pull/985

So IMHO this one is not needed.

marcinczenko:
Closing in favour of https://github.com/codex-storage/nim-codex/pull/985.

PR #1016, opened 2024-12-02
Send pluralized wantBlock messages

benbierens:
Previously:
```
    discard await allFinished(
      wantCids.mapIt(b.sendWantBlock(it, peerCtx)))
```

This gives 1 wantBlock message per entry in wantCid blockAddress. But, message format supports multiple addresses.

New:
```
    await b.sendWantBlock(wantCids, peerCtx)
```

Todo, not in this PR: initial calls to send wantHave and wantBlock messages can be pluralized as well.

### DIFF
--- a/codex/blockexchange/engine/engine.nim
+++ b/codex/blockexchange/engine/engine.nim
@@ -127,7 +127,7 @@ proc stop*(b: BlockExcEngine) {.async.} =
 
   trace "NetworkStore stopped"
 
-proc `$`*(blkAddrs: seq[BlockAddress]): string =
+proc formatAddrs(blkAddrs: seq[BlockAddress]): string =
   result &= "["
   for addr in blkAddrs:
     result &= ($addr & " ")
@@ -142,7 +142,7 @@ proc sendWantHave(
   for p in peers:
     if p notin excluded:
       let toAsk = addresses.filterIt(it notin p.peerHave)
-      trace "Sending wantHave request", toAsk, peer = p.id
+      trace "Sending wantHave request", toAsk = formatAddrs(toAsk), peer = p.id
       await b.network.request.sendWantList(
         p.id,
         toAsk,
@@ -152,7 +152,7 @@ proc sendWantBlock(
   b: BlockExcEngine,
   addresses: seq[BlockAddress],
   blockPeer: BlockExcPeerCtx): Future[void] {.async.} =
-  trace "Sending wantBlock request to", addresses, peer = blockPeer.id
+  trace "Sending wantBlock request to", addrs = formatAddrs(addresses), peer = blockPeer.id
   await b.network.request.sendWantList(
     blockPeer.id,
     addresses,
--- a/codex/blockexchange/engine/engine.nim
+++ b/codex/blockexchange/engine/engine.nim
@@ -127,28 +127,35 @@ proc stop*(b: BlockExcEngine) {.async.} =
 
   trace "NetworkStore stopped"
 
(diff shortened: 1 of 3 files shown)


benbierens:
This passes dist-tests.

codex/logutils.nim
```diff
@@ -184,7 +184,7 @@ template formatIt*(format: LogFormat, T: typedesc, body: untyped) =
       let v = opts.map(opt => opt.formatJsonOption)
       setProperty(r, key, json.`%`(v))
 
-    proc setProperty*(r: var JsonRecord, key: string, val: seq[T]) =
+    proc setProperty*(r: var JsonRecord, key: string, val: seq[T]) {.upraises:[ValueError, IOError].} =
```
benbierens, line 187: I want someone to tell me whether this should be applied to all of these procs. Is it going to cause people confusion if at some point in the future they use this without knowing it and it'll crash compilation with an error that's unrelated to what they are doing?

benbierens: There's going to be a separate PR to fix the logutils raises issue. Then we'll rebase this, then we can review

gmega: You can start adding the raises clauses IMHO. And it's `raises` not `upraises`.

gmega (changes requested):
Seems harmless to me, just need to fix the raises annotations.

codex/logutils.nim
benbierens: Should I apply this same change to the other setProperty procs in the same file that already have exactly the same thing?

codex/blockexchange/engine/engine.nim
```diff
@@ -129,26 +129,27 @@ proc stop*(b: BlockExcEngine) {.async.} =
 
 proc sendWantHave(
   b: BlockExcEngine,
-  address: BlockAddress, # pluralize this entire call chain, please
+  addresses: seq[BlockAddress],
   excluded: seq[BlockExcPeerCtx],
   peers: seq[BlockExcPeerCtx]): Future[void] {.async.} =
-  trace "Sending wantHave request to peers", address
+  trace "Sending wantHave request to peers", numBlkAddrs = addresses.len, numPeers = peers.len - excluded.len
```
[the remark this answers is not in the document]
benbierens: You're right. I don't like this log. I'm taking it out. A few lines down is a trace that says exactly which addresses are sent to which peer. We don't need this one.

codex/blockexchange/engine/engine.nim
```diff
@@ -197,9 +198,10 @@ proc requestBlock*(
     if peer =? maybePeer:
       asyncSpawn b.monitorBlockHandle(blockFuture, address, peer.id)
       b.pendingBlocks.setInFlight(address)
-      await b.sendWantBlock(address, peer)
+      # TODO: Send more block addresses if at all sensible.
+      await b.sendWantBlock(@[address], peer)
       codex_block_exchange_want_block_lists_sent.inc()
-      await b.sendWantHave(address, @[peer], toSeq(b.peers))
+      await b.sendWantHave(@[address], @[peer], toSeq(b.peers))
```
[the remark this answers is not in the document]
benbierens: We could, but I'm trying to move away from single-address calls. In a follow-up PR I want to change the call site so that it makes use of the seq.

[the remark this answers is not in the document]
gmega: OK sounds good

gmega (approved):
LGTM

PR #1019, opened 2024-12-04
Fix sending of WantBlocks messages and tracking of peerWants

benbierens:
In the current situation when we initiate a block request, we send a wantBlock to one peer (even if they don't have the block, so it will be ignored) and we send wantHave messages to others, which are then discarded.

This PR fixes two issues:
- We only send wantBlock messages to peers who have the block. We use the same pseudo-random method for selecting a peer if multiple peers have the block.
- We track all peerWants instead of only the wantBlocks. This is according to spec, and solves an issue in the case where a block is received/detected after the want-list message.

I did not want to solve both issues in the same PR. But test `NetworkStore - multiple nodes` in testblockexc.nim failed because of the second issue after I fixed the first one. So instead of having to "break" a correct test, I figured it was better to fix the actual problem.

### DIFF
--- a/codex/blockexchange/engine/engine.nim
+++ b/codex/blockexchange/engine/engine.nim
@@ -412,8 +412,15 @@ proc wantListHandler*(
       if not e.cancel:
         peerCtx.peerWants.add(e)
 
+      # Important TODO! (See https://github.com/codex-storage/nim-codex/pull/1019#issuecomment-2525089803 )
+      # The if type==WantHave (below) is NOT nested behind the if not cancel (above) on purpose!
+      # This means we send presence-lists in response to cancel messages.
+      # Not doing so will degrade the performance by more than an order of magnitude.
+      # TODO: Investigate WHY the presence list has such a performance impact.
+      # (Sending a presence-list in response to a cancel is not according to spec and should be removed
+      # once the performance impact mystery is understood and solved.)
       if e.wantType == WantType.WantHave:
-        # does this happen for cancels?!
+        # Respond to wantHaves immediately.
         let
           have = await e.address in b.localStore
           price = @(
--- a/codex/blockexchange/engine/engine.nim
+++ b/codex/blockexchange/engine/engine.nim
@@ -270,7 +270,7 @@ proc scheduleTasks(b: BlockExcEngine, blocksDelivery: seq[BlockDelivery]) {.asyn
 
 proc cancelBlocks(b: BlockExcEngine, addrs: seq[BlockAddress]) {.async.} =
   ## Tells neighboring peers that we're no longer interested in a block.
-  trace "Sending block request cancellations to peers", addrs = addrs.len
+  trace "Sending block request cancellations to peers", addrs, peers = b.peers.mapIt($it.id)
 
   let failed = (await allFinished(
     b.peers.mapIt(
@@ -389,6 +389,7 @@ proc wantListHandler*(
 
   var
     presence: seq[BlockPresence]
+    schedulePeer = false
 
   for e in wantList.entries:
     let
@@ -398,16 +399,19 @@ proc wantListHandler*(
       peer      = peerCtx.id
       address   = e.address
       wantType  = $e.wantType
+      isCancel  = $e.cancel
 
     # Update metrics
     if e.wantType == WantType.WantHave:
       codex_block_exchange_want_have_lists_received.inc()
     elif e.wantType == WantType.WantBlock:
+      schedulePeer = true
       codex_block_exchange_want_block_lists_received.inc()
 
     # Update peerCtx
     if idx < 0: # new entry
-      peerCtx.peerWants.add(e)
+      if not e.cancel:
+        peerCtx.peerWants.add(e)
     else: # update existing entry
       # peer doesn't want this block anymore
       if e.cancel:
@@ -439,14 +443,13 @@ proc wantListHandler*(
             `type`: BlockPresenceType.Have,
             price: price))
 
-      # Schedule handling of wantBlocks
-      if e.wantType == WantType.WantBlock:
-        if not b.scheduleTask(peerCtx):
-          warn "Unable to schedule task for peer", peer
-
   if presence.len > 0:
     trace "Sending presence to remote", items = presence.mapIt($it).join(",")
     await b.network.request.sendPresence(peer, presence)
+
+  if schedulePeer:
+    if not b.scheduleTask(peerCtx):
+      warn "Unable to schedule task for peer", peer
 
 proc accountHandler*(
   engine: BlockExcEngine,
--- a/codex/blockexchange/engine/engine.nim
+++ b/codex/blockexchange/engine/engine.nim
@@ -240,7 +240,7 @@ proc blockPresenceHandler*(
     )
 
   if wantCids.len > 0:
-    trace "Peer has blocks in our wantList", peer, wantCount = wantCids.len
+    trace "Peer has blocks in our wantList", peer, wants = wantCids
     await b.sendWantBlock(wantCids, peerCtx)
 
   # if none of the connected peers report our wants in their have list,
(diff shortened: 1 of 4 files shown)


codex/blockexchange/engine/engine.nim
```diff
@@ -427,10 +425,8 @@ proc wantListHandler*(
           address: e.address,
           `type`: BlockPresenceType.Have,
           price: price))
-      elif e.wantType == WantType.WantBlock:
```
dryajov, line 430: There is a good reason why I separated `WantBlock` into it's own branch. We used to schedule a task for every single `WantList` update, but obviously this isn't necessary and worst, it was severely affecting performance. We should not attempt to schedule tasks on `WantHave`, since this is just a presence update.

dryajov (changes requested):
I gave this a review and there are a couple of things that need addressing.

codex/blockexchange/engine/engine.nim
dryajov: > We track all peerWants instead of only the wantBlocks. This is according to spec, and solves an issue in the case where a block is received/detected after the want-list message.

For more context, this isn't actually true, we do track `WantHave` messages, but we respond to them immediately with a presence message.

codex/blockexchange/engine/engine.nim
```diff
@@ -175,33 +175,27 @@ proc monitorBlockHandle(
...
+      asyncSpawn b.monitorBlockHandle(blockFuture, address, selected.id)
       b.pendingBlocks.setInFlight(address)
-      # TODO: Send more block addresses if at all sensible.
-      await b.sendWantBlock(@[address], peer)
-      codex_block_exchange_want_block_lists_sent.inc()
-      await b.sendWantHave(@[address], @[peer], toSeq(b.peers))
-      codex_block_exchange_want_have_lists_sent.inc()
+      await b.sendWantBlock(@[address], selected)
+      
+    await b.sendWantHave(@[address], peers.without)
```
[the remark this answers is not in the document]
benbierens: Peers are only reporting their haves if asked. We have to send a wantHave before we know if they have the block.

codex/blockexchange/peers/peerctxstore.nim
```diff
@@ -32,6 +32,9 @@ logScope:
 type
   PeerCtxStore* = ref object of RootObj
     peers*: OrderedTable[PeerId, BlockExcPeerCtx]
+  PeersForBlock* = ref object of RootObj
```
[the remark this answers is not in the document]
benbierens: I would prefer object over tuple. After compile it should be very similar, right?

codex/blockexchange/engine/engine.nim
benbierens: OK, I see the problem with the scheduling of tasks for every wantList. Makes perfect sense and I'll fix that right now.

Regarding the tracking of wantHaves: Responding immediately is good but not enough. We're expected to keep the want in the peerCtx. If we encounter the block later, we should still react to the want with a presence message. The tests in `NetworkStore - multiple nodes` in testblockexc.nim actually check this behavior perfectly. They were just previously passing because 1) we were sending wantBlocks when we shouldn't and 2) we were keeping the wantBlocks only. So fixing the send of wantBlock caused this correct test to reveal the other issue.

I would like to take a closer look at exactly how the stored wants will trigger messages when blocks are received. But I'm trying to keep the scope of this PR small.

benbierens: I've ran some more tests after applying these changes. The performance improvement is noticeable. I think reliability is also up, but that takes a bit longer to be sure about.

codex/blockexchange/engine/engine.nim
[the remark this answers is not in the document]
dryajov: Ah, you're right.

codex/blockexchange/peers/peerctxstore.nim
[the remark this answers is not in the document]
dryajov: sure, but let's not make it a ref for no reason.

codex/blockexchange/engine/engine.nim
dryajov: OK, I see. Then, what we need to do is to add these to the `BlockExcPeerCtx.peerWants` list in addition to responding to the request.

---------------

For a later PR, we should take a closer look at how these lists are kept, for example, the `peerWants` is a simple `seq`, which doesn't have any sort of caps, nor ordering. 

Some lists, like the `peerWants`, in addition to be capped, should also allow for ordering, for example `WantListEntry` has a `priority` field, which should be respected.

dryajov: So, what actually needs to happen is that when adding a `DontHave` to the `presence` seq, we also add it to `peerCtx.peerWants` list. We only need to remember the `DontHave`s, because the remote should keep track of the `Have`s we send back.

benbierens:
OK, I need to update the record with what I've been up to.
After commit "prevent duplicate scheduling. Fixes cancellation" I re-ran the dist-tests and noticed a huge performance deterioration. A transfer that takes 12 seconds on v0.1.9 took 1min-30secs. This is an unacceptable downgrade. So I've set out to discover the cause.

I isolated the slow-down to my commit "Review comment by Dmitriy". Reviewing the changes, the cause remained unexplained. Eventually through a process of reverting/retesting I was able to fix the performance in a late-evening commit fashionably titled "fast". (https://github.com/codex-storage/nim-codex/pull/1019/commits/dcf43d32277130bb9ae1223845e4df96535fb467) After this, performance picked back up, averaging between 8 and 9 seconds, slightly better than 0.1.9. (more tests needed!)

What I learned:
In the wantListHandler, if the want-type is WantHave, we must send a presence response *even if the cancel = true*. I figured if it is a cancel-message, no need to respond with a presence list. Indeed I later confirmed that these presence-lists sent in response to a cancel are ignored by the peer who receives them. Yet still, somehow, not sending them degrades the transfer performance hugely.

My plan is: Finalize this PR with as few changes as possible. That means leaving this puzzling behavior in place as-is.
After that, I'm going to be adding logs, making changes, and running tests until I can explain *why* these cancel-triggered presence messages have such an impact. I have a few ideas, but those will have to wait till Monday before I pick this back up.

benbierens:
Further delays/sidetracks:
tests/integration/testsales.nim fails. It is caused (at least sometimes) by a failure of the blockexchange engine. *But*, the integration tests for block exchange all pass. Enabling logging causes the test to pass. So I'm going to investigate what's wrong here, upgrade the block exchange integration tests if necessary. And hopefully solve the instability. (In a new PR, of course.) Whatever is failing this test might be the same cause as the instability we saw in the testnet: This means that the changes in this PR reveal the issue such that we can capture it in an integration test.
It's kind of annoying that this PR keeps spawning offshoots, but this is the road to more reliability.

codex/blockexchange/engine/engine.nim
benbierens: Yes: Upgrading the peerWants to a capped, prioritized list is definitely needed.

"what needs to happen is that when adding a DontHave to the presence seq, we also add it to peerCtx.peerWants list."
In the PR, yes we are adding it to the peerWants regardless of Have/DontHave

"We only need to remember the DontHaves, because the remote should keep track of the Haves we send back."
I understand this to mean:
- If we Have, don't add the want to the peerCtx, only send presence.
- If we dont Have, add want to peerCtx, and send dontHave presence if asked. (e.sendDontHave)
For the record, e.sendDontHave is has always been false up to this point except for some tests and that's a good thing because the peerContext peerHave proc doesn't consider the have-flag currently.

dryajov: > If we Have, don't add the want to the peerCtx, only send presence.
> If we dont Have, add want to peerCtx, and send dontHave presence if asked. (e.sendDontHave)

Yep, thats what I mean.

benbierens:
Spoke with Dmitriy about the following change:
Only add the want to the peerWant list if we don't have the block. If we *do* have the block, respond with a presence, but, no need to remember the want. This change was pushed in https://github.com/codex-storage/nim-codex/pull/1019/commits/23c0f30977eb6f1da5ac0ddb9d18204610c45c8d

I reran some dist-tests after. Performance and also the cancel-presence message performance impact is unaffected.

dryajov (approved):
I think this is good to go for now.

PR #1021, opened 2024-12-05
chore(tests): fix unneeded async

emizzle:
Removes an unneeded `check eventually` inside a helper proc, removing the need for the proc to be async.


emizzle:
I can run it 1000x, sure. But the main idea is that there is already a `check eventually` at the callsite so when there are not enough agents, `isInState` simply returns false.

emizzle:
@benbierens, I have successfully run this 1000x locally, recompiling every 100x.

benbierens (approved):
Thanks for checking :D

emizzle:
> Thanks for checking :D

Thanks for asking 😄

Issue #885, opened 2024-08-26
Data persistence for API


AuHau:
It's not the greatest title but I can't figure something better.

Currently, when you query endpoints like `/slots` or `/purchases` they are backed by the Marketplace's state-machines, and state-machines are loaded upon node's start from on-chain using the contract tracking `mySlots()` and `myRequests()`.

This mechanism does not really keep much of the data it represents around, mainly the Sales state-machine, because it cleans up after it finishes: https://github.com/codex-storage/nim-codex/blob/e6a387e8e81824b2fcd816f7bde25699bbe1a162/codex/sales.nim#L163

The purchasing machine I believe does not clean up so the "data" will stay there until the node shutdown, but then when it re-starts it won't "reload it" from the `myRequests()` because it is already gone.

This is not great API experience. Most probably for example UI will want to be able fetch history of purchases/sales to display to the user...

emizzle:
I definitely agree this is not the best API experience and should be unified. One of the more obvious options is to keep past slots / requests in a mapping in the contract. That mapping could be returned in a `myPastSlots`/`myPastRequests` function in the contract and a REST API request to get past slots/requests could  simply return a `myPastSlots`/`myPastRequests`. The purchasing module would then remove past purchases from memory. That way, both purchasing and sales don't keep past purchases / sales agents in memory, and the REST API endpoints would get the data from the contract.

This would allow us to get rid of the notion of returning `PurchaseIds` in the REST API, and rely on the `RequestId` instead. This will be more consistent with how we retrieve sales info by using the `SlotId` and not a `SalesAgentId` (which doesn't exist). On node startup, we can retrieve `myPastRequests` and reverse-map it to a `PurchaseState`, similar to how it is currently done on `Purchasing.load`, except the mapping is reversed. Likewise for `myPastRequests` in the sales module. I never liked how the REST API returned a `PurchaseId` after creating a storage request. Most users expected it to return a `RequestId`. This even tripped up a lot of our team during runs we did for the EthCC workshop.

AuHau:
Hmm, you are right that we already have the data in the contract. The question is how much we depend on the freeing gas refund of deleting the data from the mapping. If we are willing to take the hit, then it might be an interesting and more importantly simple solution. 

[...]

PR #1022, opened 2024-12-05
feat: repair is rewarded

AuHau:
Adds support for changes in the contract that introduce repair reward as described here: https://github.com/codex-storage/codex-contracts-eth/issues/154

There is one question to be answered about what the node should do when it restarts (after an outage for example) and finds his slot to be in `Repair` state. For more see bellow.

### DIFF
--- a/codex/sales/states/unknown.nim
+++ b/codex/sales/states/unknown.nim
@@ -51,7 +51,7 @@ method run*(state: SaleUnknown, machine: Machine): Future[?State] {.async.} =
   of SlotState.Cancelled:
     return some State(SaleCancelled())
   of SlotState.Repair:
-    # TODO: What to do? It could really mean that he was offline for that long that he was kicked out
-    # from the slot and lost all the collateral. Should we simply "give up" on the slot? Or try
-    # to "repair" (there is some chance that the dataset is still locally saved) the slot?
+    # This should never really happen as the Slot is removed from the node's onchain slots tracking
+    # upon freeing the slot. Hence upon node's restart the slot with the Repair state should not
+    # be loaded from the onchain.
     discard
--- a/codex/sales/states/unknown.nim
+++ b/codex/sales/states/unknown.nim
@@ -50,3 +50,8 @@ method run*(state: SaleUnknown, machine: Machine): Future[?State] {.async.} =
     return some State(SaleFailed())
   of SlotState.Cancelled:
     return some State(SaleCancelled())
+  of SlotState.Repair:
+    # TODO: What to do? It could really mean that he was offline for that long that he was kicked out
+    # from the slot and lost all the collateral. Should we simply "give up" on the slot? Or try
+    # to "repair" (there is some chance that the dataset is still locally saved) the slot?
+    discard
--- a/codex/sales/states/unknown.nim
+++ b/codex/sales/states/unknown.nim
@@ -5,6 +5,7 @@ import ./filled
 import ./finished
 import ./failed
 import ./errored
+import ./proving
 import ./cancelled
 import ./payout
 
@@ -38,7 +39,7 @@ method run*(state: SaleUnknown, machine: Machine): Future[?State] {.async.} =
   case slotState
   of SlotState.Free:
     let error = newException(UnexpectedSlotError,
-      "slot state on chain should not be 'free'")
+      "Slot state on chain should not be 'free'")
     return some State(SaleErrored(error: error))
   of SlotState.Filled:
     return some State(SaleFilled())
@@ -51,7 +52,6 @@ method run*(state: SaleUnknown, machine: Machine): Future[?State] {.async.} =
   of SlotState.Cancelled:
     return some State(SaleCancelled())
   of SlotState.Repair:
-    # This should never really happen as the Slot is removed from the node's onchain slots tracking
-    # upon freeing the slot. Hence upon node's restart the slot with the Repair state should not
-    # be loaded from the onchain.
-    discard
+    let error = newException(SlotFreedError,
+      "Slot was forcible freed and host was removed from its hosting")
+    return some State(SaleErrored(error: error))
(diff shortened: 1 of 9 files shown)


codex/sales/states/unknown.nim
```diff
@@ -50,3 +50,8 @@ method run*(state: SaleUnknown, machine: Machine): Future[?State] {.async.} =
     return some State(SaleFailed())
   of SlotState.Cancelled:
     return some State(SaleCancelled())
+  of SlotState.Repair:
+    # TODO: What to do? It could really mean that he was offline for that long that he was kicked out
+    # from the slot and lost all the collateral. Should we simply "give up" on the slot? Or try
+    # to "repair" (there is some chance that the dataset is still locally saved) the slot?
+    discard
```
AuHau, line 57: What should happen here? Any suggestions?

emizzle: In this case, we've seen that the slot state is in a Repair state for some time after the slot was freed, so it would have missed the `SlotFreed` event and not have re-added the slot to the slot queue. Maybe we could add the slot back into the slot queue here, and also add a `SaleSlotFreed` state that simply logs that the slot was freed?

We might want to consider that we cover the `Repair` case in `sales/states/proving.proveLoop`. This case is a bit different since the node will be running and presumably online so it will have received the `SlotFreed` event, putting the slot id back in the slot queue. But we could still move to a `SaleSlotFreed` state to log that a slot was freed.

Moving to a new state I think will help us when we try to tackle "preserving" state history for the UI in https://github.com/codex-storage/nim-codex/issues/885.

AuHau: Ooooh, I forgot about one thing: when a slot is forcibly freed, then it is also removed from the node's slot tracking, so when the node comes online and starts the "reloading of state" from this tracking, he will actually not find out that he was kicked out from a slot. So I guess the problem solved, but IMHO, this is bad UX 😅 And it circles back to the big task of keeping some metadata store of the marketplace and not relying only on contracts state. But that is another issue...

emizzle: Good point! Should we still handle the proveLoop case in this PR?

AuHau: > Good point! Should we still handle the proveLoop case in this PR?

I can't imagine that his case will happen (often), but yeah I will add handling for it in the prooving loop - good point!

AuHau: > But we could still move to a SaleSlotFreed state to log that a slot was freed.

I have chosen to not create another new state as it would simply transition to `SaleErrored` so instead, I am logging the case of repair and raising an exception that automatically transitions to `SaleErrored`. Maybe the question is what should happen regarding the `returnBytes` and `reprocessSlot`? I don't think we need to set the `reprocessSlot` but maybe we should prevent `returnBytes` as that would clean up the data from the node, preventing potentially "refilling"? 🤔 But at the same time if the refilling would not happen for any reason, then there would be a "data leak" as the bytes would be garbage collected only after the original slot's duration expiry...

emizzle: > it circles back to the big task of keeping some metadata store of the marketplace and not relying only on contracts state

This is discussed in #885. We both agreed to keep `myPastRequests` in the contract until we find time to develop storing it locally.

> I have chosen to not create another new state as it would simply transition to SaleErrored so instead, I am logging the case of repair and raising an exception that automatically transitions to SaleErrored

This is one of those situations where we know that this sequence of events can happen, but is it really an error? Imo, we should treat this as a known sequence, and not an unexpected error. As I said before, it will also help when we implement #885, so that some kind of persistence of the final state can happen. 

> Maybe the question is what should happen regarding the returnBytes and reprocessSlot?

`returnBytes` only returns bytes to the `Availability`, it does not clean up data from the node. The data will not be cleaned up until the end of the storage request (this happens in `SaleDownloading` > `onStore` > `updateExpiry`). Imo, we should `returnBytes` to the Availability so we can refill the slot. 

Should we `reprocessSlot`? Well, that's a bit trickier, because if we rely on the `proveLoop` (at the start of the next period) to realise the slot is in repair, we'll be later than other nodes who would have received the `SlotFreed` contract event earlier. However, we'll still have the data, so we have an advantage. And if we rely on the contract event only, then we bypass the state machine. So, we can combine the two (more work, I know):

1. Listen for the `SlotFreed` event and interrupt the `SaleProving` state with an `onSlotFreed` handler, moving to the `SaleRepairing` state.
2. (I'd still keep the check in the `proveLoop` for the repair case and move to the `SaleRepairing` state)
3. In `SaleRepairing`, **there are two options**:
    1. Call `onCleanUp(returnBytes: false, reprocessSlot: false)` and return to the `SaleInitialProving` state. Remove the logic that re-adds the slot to the queue on `SlotFreed`.
    2. Call `onCleanUp(returnBytes: true, reprocessSlot: true)`, relying on the `SlotFreed` contract event to re-add the slot to the queue. If there are spare workers, this re-add will be processed before the `SaleRepairing` step, so we would hope that there was enough Availability...
    
My vote would be option 1 as it's the easiest to understand (imo lol), removes hidden things happening (contract event reprocessing), and sets us up for #885. It's more work, but I think it will be much easier to follow logically. Likely should be done in a separate PR to keep changes small. In the meantime, for this PR, my vote is to have a `SlotRepairing` state that calls `onCleanUp(returnBytes: true, reprocessSlot: false)`.

codex/sales/states/unknown.nim
```diff
@@ -50,3 +50,8 @@ method run*(state: SaleUnknown, machine: Machine): Future[?State] {.async.} =
     return some State(SaleFailed())
   of SlotState.Cancelled:
     return some State(SaleCancelled())
+  of SlotState.Repair:
+    # This should never really happen as the Slot is removed from the node's onchain slots tracking
+    # upon freeing the slot. Hence upon node's restart the slot with the Repair state should not
+    # be loaded from the onchain.
+    discard
```
emizzle, line 57: Maybe log an error?

emizzle (approved):
Approved, as no blockers, but maybe we should discuss further what should happen when a slot is in the repair state. Imo, it should go to a `SlotRepairing` state (which can also call `onCleanUp`). Anyway, I commented more in the "conversation".

PR #1024, opened 2024-12-09
fix(trackedfutures): removes usage of `then` from tracked futures

emizzle:
- removes usage of `then`, simplifying the logic, and allowing `then` to be removed completely
- updates annotations to reflect that all procs (sync and async) raise no exceptions

This PR is part of a larger set of fixes to the codebase that removes usage of a faulty pattern of not awaiting or asyncSpawning futures.

### DIFF
--- a/codex/utils/trackedfutures.nim
+++ b/codex/utils/trackedfutures.nim
@@ -1,9 +1,9 @@
-import std/sugar
 import std/tables
 import pkg/chronos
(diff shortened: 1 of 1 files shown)


emizzle:
Closed in favour of https://github.com/codex-storage/nim-codex/pull/1032 (needed to rename remote branch).

PR #1025, opened 2024-12-09
fix(asyncstatemachine): fixes not awaiting or asyncSpawning futures

emizzle:
- adds a break in scheduler when CancelledError is caught
- tracks asyncSpawned state.run, so that it can be cancelled during stop
- removes usages of `then`
- ensures that no exceptions are leaked from async procs

### Notes
- Based on https://github.com/codex-storage/nim-codex/pull/1024
- This PR is part of a larger set of fixes to the codebase that removes usage of a faulty pattern of not awaiting or asyncSpawning futures.

### DIFF
--- a/codex/utils/asyncstatemachine.nim
+++ b/codex/utils/asyncstatemachine.nim
@@ -1,12 +1,11 @@
 import std/sugar
 import pkg/questionable
 import pkg/chronos
-import pkg/upraises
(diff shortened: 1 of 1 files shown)


emizzle:
Closing in favour of https://github.com/codex-storage/nim-codex/pull/1033 (changed remote branch name)

PR #1036, opened 2024-12-12
fix(sales): replaces `then` with `asyncSpawn`

emizzle:
- ensures `addSlotToQueue` does not raise exceptions as it is now asyncSpawned

### Notes
- Based on https://github.com/codex-storage/nim-codex/pull/1033
- This PR is part of a larger set of fixes to the codebase that removes usage of a faulty pattern of not awaiting or asyncSpawning futures.


markspanbroek (approved):
Thanks @emizzle!

PR #1039, opened 2024-12-13
fix(validation): asyncSpawns run

emizzle:
- annotates run with raises: []
- asyncSpawns run, to ensure there are no escaping exceptions

### DIFF
--- a/codex/validation.nim
+++ b/codex/validation.nim
@@ -109,7 +109,7 @@ proc markProofsAsMissing(validation: Validation) {.async.} =
     let previousPeriod = validation.getCurrentPeriod() - 1
     await validation.markProofAsMissing(slotId, previousPeriod)
 
-proc run(validation: Validation) {.async.} =
+proc run(validation: Validation) {.async: (raises: []).} =
   trace "Validation started"
   try:
     while true:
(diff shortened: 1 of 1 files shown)


codex/validation.nim
```diff
@@ -124,9 +124,11 @@ proc start*(validation: Validation) {.async.} =
   validation.proofTimeout = await validation.market.proofTimeout()
   await validation.subscribeSlotFilled()
   validation.running = validation.run()
+  asyncSpawn validation.running
 
 proc stop*(validation: Validation) {.async.} =
-  await validation.running.cancelAndWait()
+  if not validation.running.isNil and not validation.running.finished:
```
gmega, line 130: Ugh, when does this happen?

gmega (approved):
LGTM

marcinczenko (approved):

codex/validation.nim
marcinczenko: yes

emizzle: It's a good question. Th nil check prevents SIGSEGV and the `finished` check prevents Defects like "future was completed more than once". Both are a PITA to debug and these checks prevent that pain. However, if `validation.running` is nil, then we could raise a `Defect` imo, because something didn't quite go right and we're in a bad state. BUT, we're stopping the module, so do we really care if `validation.running` is already nil? Probably not.

PR #1047, opened 2024-12-17
refactor(then): removes then util

emizzle:
- removes then util as it is no longer being used in the codebase


benbierens (approved):
Only red lines! Best kind of PR.

PR #1048, opened 2024-12-18
fix(tests/integration): fix bootstrap node spr

emizzle:
After a change in PR #1031, bootstrap node sprs may not work when Hardhat nodes are started with the tests. This fixes it by appending all started client's and provider's SPR to a sequence, and using that sequence of SPRs to start the next node. This means all subsequently started nodes will be connected to its previously started peers.

This also fixes the case when bootstrap SPRs would not be present if no clients were started.

### DIFF
--- a/tests/integration/multinodes.nim
+++ b/tests/integration/multinodes.nim
@@ -83,7 +83,7 @@ template multinodesuite*(name: string, body: untyped) =
     #     ...
     let jsonRpcProviderUrl = "http://127.0.0.1:8545"
     var running {.inject, used.}: seq[RunningNode]
-    var bootstrap: string
+    var bootstrapNodes: seq[string]
     let starttime = now().format("yyyy-MM-dd'_'HH:mm:ss")
     var currentTestName = ""
     var nodeConfigs: NodeConfigs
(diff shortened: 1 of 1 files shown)


tests/integration/multinodes.nim
```diff
@@ -273,6 +273,13 @@ template multinodesuite*(name: string, body: untyped) =
         fail()
         quit(1)
 
+    proc updateBootstrapNodes(node: CodexProcess) =
+      without ninfo =? node.client.info():
+        # raise CatchableError instead of Defect (with .get or !) so we
+        # can gracefully shutdown and prevent zombies
```
AuHau, line 279: Graceful shutdown is for sure nice, but I think the OS would clean up the zombies, because the parent process (the test running process) will terminate anyhow. Just FYI.

AuHau (approved):

tests/integration/multinodes.nim
emizzle: Yeah zombies aren't necessarily an issue. What won't happen with a `Defect` is execution of the except in `failAndTeardownOnError`, which runs teardown routines, fails the tests, and quits. The teardown routines stop the nodes (gracefully) and remove the data dirs for a clean start next run.

benbierens (approved):

Issue #875, opened 2024-08-12
[BUG] - Compile Codex with GCC 14


veaceslavdoina:
**Describe the bug**
Codex currently can't be compiled with [GCC 14, which was released on Tue May 7 10:31:35 GMT 2024](https://gcc.gnu.org/pipermail/gcc-announce/2024/000180.html).

Based on the [numbering scheme](https://gcc.gnu.org/develop.html#num_scheme), this is a first `v14` public release, `14.0` was used in the development process.

**To Reproduce**
1. Follow [Building Codex](https://github.com/codex-storage/nim-codex/blob/master/BUILDING.md) guide
2. Install GCC 14
3. Run `make update` - ✅
4. Run `make` -  ❌

**Expected behavior**
We should be able to compile Codex with GCC 14 and older versions as well.

**Environment:**
 - OS: `Ubuntu, Windows`
 - Version: `latest`
 - Tag/Commit: `master`
 - Arch: `amd64`

**Additional context**
- Workaround for CI on Windows was merged - #874

Work in progress - [`nim-codex/.gitmodules`](https://github.com/codex-storage/nim-codex/blob/master/.gitmodules)
 - [ ] [`status-im/nim-bearssl`](https://github.com/status-im/nim-bearssl) - Update submodule
 - [ ] [`vacp2p/nim-libp2p`](https://github.com/vacp2p/nim-libp2p) - https://github.com/vacp2p/nim-libp2p/tree/multihash-poseidon2
 - [x] [`nim-leopard`](https://github.com/codex-storage/nim-leopard)  - https://github.com/codex-storage/nim-leopard/pull/15
 - [x] [`nim-codex-dht`](https://github.com/codex-storage/nim-codex-dht) - https://github.com/codex-storage/nim-codex-dht/pull/98
 - [ ] [`nim-chronos`](https://github.com/status-im/nim-chronos) - Update submodule
 - [ ] [`nim-datastore`](https://github.com/codex-storage/nim-datastore)
   - [ ] https://github.com/codex-storage/nim-datastore/pull/67
   - [x] https://github.com/codex-storage/nim-datastore/pull/68
   - [x] https://github.com/codex-storage/nim-datastore/pull/69
   - [ ] https://github.com/codex-storage/nim-datastore/pull/70
   - [x] https://github.com/codex-storage/nim-datastore/pull/73

benbierens:
[...]

PR #1052, opened 2024-12-18
Chore/update nim version

2-towns:
Update Nim to ~~2.0.12~~ 2.0.14


codex/rest/json.nim
```diff
@@ -90,7 +90,7 @@ proc init*(_: type RestNode, node: dn.Node): RestNode =
     peerId: node.record.data.peerId,
     record: node.record,
     address: node.address,
-    seen: node.seen
+    seen: node.alreadySeen()
```
2-towns, line 93: @cskiraly Is this the right method to call to verify that a node is seen?

codex/slots/proofs/backends/circomcompat.nim
```diff
@@ -102,9 +102,9 @@ proc prove[H](
 
   defer:
     if ctx != nil:
-      ctx.addr.releaseCircomCompat()
+      ctx.addr.release_circom_compat()
```
dryajov, line 105: Interesting, did Nim complain about the camel case here or are you just trying to match it to the wrapped exports? If it's the later, I would say camel case is more consistent with the codebase and it's a valid in Nim to use camel case instead of `_` and vice-versa.

2-towns: Yes Nim complained about the camel case: 

```Error: 'releaseCircomCompat' should be: 'release_circom_compat' ```

File: .github/actions/nimbus-build-system/action.yml [hunk 3]
```diff
@@ -168,6 +168,12 @@ runs:
       shell: ${{ inputs.shell }} {0}
       run: echo "NIM_COMMIT=${{ inputs.nim_version }}" >> ${GITHUB_ENV}
 
+    - name: MSYS2 (Windows All) - Disable git symbolic links (since miniupnp 2.2.5)
+      if: inputs.os == 'windows'
+      shell: ${{ inputs.shell }} {0}
+      run: |
+        git config --global core.symlinks false
```
2-towns, line 175: @veaceslavdoina Double check if it is okay for you

tests/codex/testchunking.nim
```diff
@@ -112,6 +120,11 @@ asyncchecksuite "Chunking":
     expect LPStreamError:
       await raiseStreamException(newException(LPStreamError, "test error"))
 
-  test "stream should convert other exceptions to defect":
-    expect Defect:
-      await raiseStreamException(newException(CatchableError, "test error"))
+  # This test cannot exist anymore.
+  # The signature of the method readOnce is explicitly listing the error raised:
```
dryajov, line 124: Cool, I think we can get rid of this test completely.

dryajov:
Great job! This is looking really good, lets try to merge this ASAP

dryajov (approved):

.github/actions/nimbus-build-system/action.yml
```diff
@@ -78,11 +78,21 @@ runs:
           mingw-w64-i686-ntldd-git
           mingw-w64-i686-rust
 
-    - name: MSYS2 (Windows All) - Downgrade to gcc 13
+    - name: MSYS2 (Windows All) - Update to gcc 14
       if: inputs.os == 'windows'
       shell: ${{ inputs.shell }} {0}
       run: |
-        pacman -U --noconfirm https://repo.msys2.org/mingw/ucrt64/mingw-w64-ucrt-x86_64-gcc-13.2.0-6-any.pkg.tar.zst https://repo.msys2.org/mingw/ucrt64/mingw-w64-ucrt-x86_64-gcc-libs-13.2.0-6-any.pkg.tar.zst
+        pacman -U --noconfirm https://repo.msys2.org/mingw/ucrt64/mingw-w64-ucrt-x86_64-gcc-14.2.0-2-any.pkg.tar.zst https://repo.msys2.org/mingw/ucrt64/mingw-w64-ucrt-x86_64-gcc-libs-14.2.0-2-any.pkg.tar.zst
```
veaceslavdoina, line 85: Windows is installing latest GCC version, which currently is 14 and 15 is on the way :)

We can remove this step, but later we might have an issue with version 15 (?).

.github/workflows/ci-reusable.yml
```diff
@@ -40,6 +40,7 @@ jobs:
           os: ${{ matrix.os }}
           shell: ${{ matrix.shell }}
           nim_version: ${{ matrix.nim_version }}
+          coverage: false
```
veaceslavdoina, line 43: We should add an input `coverage`  to the _.github/actions/nimbus-build-system/action.yml_ and set it `false` by default, than pass `true` just for coverage (we are doing it already).

https://github.com/codex-storage/nim-codex/actions/runs/12689785802
Screenshot 2025-01-09 at 16 12 51

.github/workflows/ci.yml
```diff
@@ -27,10 +27,10 @@ jobs:
         uses: fabiocaccamo/create-matrix-action@v4
         with:
           matrix: |
-            os {linux},   cpu {amd64}, builder {ubuntu-20.04},   tests {unittest},    nim_version {${{ env.nim_version }}}, shell {bash --noprofile --norc -e -o pipefail}
-            os {linux},   cpu {amd64}, builder {ubuntu-20.04},   tests {contract},    nim_version {${{ env.nim_version }}}, shell {bash --noprofile --norc -e -o pipefail}
-            os {linux},   cpu {amd64}, builder {ubuntu-20.04},   tests {integration}, nim_version {${{ env.nim_version }}}, shell {bash --noprofile --norc -e -o pipefail}
-            os {linux},   cpu {amd64}, builder {ubuntu-20.04},   tests {tools},       nim_version {${{ env.nim_version }}}, shell {bash --noprofile --norc -e -o pipefail}
+            os {linux},   cpu {amd64}, builder {ubuntu-latest},   tests {unittest},    nim_version {${{ env.nim_version }}}, shell {bash --noprofile --norc -e -o pipefail}
```
veaceslavdoina, line 30: We switched to `ubuntu-20.04` in #939 and because of the #932.

Probably we should stick with the 20.04?

Btw, we also have a [release workflow](https://github.com/codex-storage/nim-codex/blob/chore/update-nim-version/.github/workflows/release.yml) and `20.04` is required mostly for releases. But as I remember, there can be an issue when cache from `latest` (`24.04`) is restored on `20.04` and vice versa.

.github/actions/nimbus-build-system/action.yml
2-towns: For Windows 2025 right ? 

But the windows-latest is still pointing on windows-2022 which is using gcc 12.

.github/workflows/ci.yml
2-towns: > Probably we should stick with the 20.04? 

There are two problems: 1- 20.04 will be deprecated soon, 2- I am not sure I will be able to install GCC 14 on ubuntu 20.04.

Maybe #932 is fixed now with the update to Nim 2.

.github/actions/nimbus-build-system/action.yml
veaceslavdoina: We are not relying on the [preinstalled GCC on Windows](https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md). It is about [MSYS2](https://www.msys2.org/docs/updating/) and how it works
> MSYS2 is a rolling release distribution

At some point, it will just install latest GCC and version 15 will be released in March~April (https://github.com/codex-storage/nim-codex/issues/875#issuecomment-2495639242). It will appear in MSYS2 repository with a delay and we can expect an issue in August, as we had it with the version 14 😄.

So, if we remove that step, we can get in the the issue in August, but will have latest GCC releases all the time.

Let's leave it to have a stable CI?

2-towns: > We are not relying on the [preinstalled GCC on Windows](https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md). It is about [MSYS2](https://www.msys2.org/docs/updating/) and how it works  

Ah ok. 

> Let's leave it to have a stable CI? 

Yeah I agree. I prefer to leave it for now and after testing GCC 15, we could remove it.

.github/workflows/ci.yml
veaceslavdoina: It was related to the glibc version. So, new releases will not run on old distributions.

[The Ubuntu 20.04 Actions runner image will begin deprecation on 2025-02-01 and will be fully unsupported by 2025-04-01 #11101](https://github.com/actions/runner-images/issues/11101)
> Deprecation will begin on **2025-02-01** and the image will be fully unsupported by **2025-04-01**.

[The Ubuntu lifecycle and release cadence](https://ubuntu.com/about/release-cycle)
Screenshot 2025-01-09 at 17 05 20

benbierens:
Running make clean, make update, and make, works without any issues on my Win11 MSYS2-UCRT.
Commit `801f6b7 - Disable CI cache for coverage`.

codex/rest/json.nim
2-towns: Okay so if I want to detect the connection with a peer is "good enough", I could check that node.seen > 0.5 I guess ? 0.5 is totally arbitrary.

.github/workflows/ci.yml
2-towns: Yes, this is what we discussed yesterday.

Issue #1054, opened 2024-12-30
Outstanding cleanup items


AuHau:
In my cleanup work I mainly focused on removing unused imports, but there are 3 other types of issues that should get cleanuped:

<details>
<summary>Warning: use treeCid instead; cid is deprecated [Deprecated]</summary>

```
/Users/adam/Projects/status/nim-codex/codex/manifest/coders.nim(266, 16) Warning: use treeCid instead; cid is deprecated [Deprecated]
/Users/adam/Projects/status/nim-codex/tests/codex/blockexchange/discovery/testdiscovery.nim(111, 25) Warning: use treeCid instead; cid is deprecated [Deprecated]
/Users/adam/Projects/status/nim-codex/codex/streams/storestream.nim(107, 74) Warning: use treeCid instead; cid is deprecated [Deprecated]
/Users/adam/Projects/status/nim-codex/vendor/nim-chronicles/chronicles/scope_helpers.nim(10, 18) Warning: use treeCid instead; cid is deprecated [Deprecated]
/Users/adam/Projects/status/nim-codex/vendor/nim-chronicles/chronicles/scope_helpers.nim(10, 18) Warning: use treeCid instead; cid is deprecated [Deprecated]
/Users/adam/Projects/status/nim-codex/codex/erasure/erasure.nim(221, 19) Warning: use treeCid instead; cid is deprecated [Deprecated]
/Users/adam/Projects/status/nim-codex/vendor/nim-chronicles/chronicles/scope_helpers.nim(10, 18) Warning: use treeCid instead; cid is deprecated [Deprecated]
/Users/adam/Projects/status/nim-codex/codex/erasure/erasure.nim(221, 19) Warning: use treeCid instead; cid is deprecated [Deprecated]
/Users/adam/Projects/status/nim-codex/vendor/nim-chronicles/chronicles/scope_helpers.nim(10, 18) Warning: use treeCid instead; cid is deprecated [Deprecated]
/Users/adam/Projects/status/nim-codex/codex/node.nim(546, 27) Warning: use treeCid instead; cid is deprecated [Deprecated]
/Users/adam/Projects/status/nim-codex/vendor/nim-chronicles/chronicles/scope_helpers.nim(10, 18) Warning: use treeCid instead; cid is deprecated [Deprecated]
/Users/adam/Projects/status/nim-codex/codex/node.nim(546, 27) Warning: use treeCid instead; cid is deprecated [Deprecated]
[...]

PR #1055, opened 2024-12-30
chore: warning cleanup

AuHau:
I was fed up with so long build logs full of warnings, so did small cleanup mainly for `UnusedImports`, during which I discovered an actually bug in Nim: https://github.com/nim-lang/Nim/issues/24552

There are a few more areas that would be good to clean up that I listed here: https://github.com/codex-storage/nim-codex/issues/1054

Unfortunately this PR mainly helped the Test's build logs, where there were lot of unused imports. For normal compile logs the main source of the warnings is from Chronos's async-raises and deprected `cid` that I skipped for now.

### DIFF
--- a/codex/conf.nim
+++ b/codex/conf.nim
@@ -10,7 +10,11 @@
 {.push raises: [].}
 
 import std/os
-import std/terminal
+
+{.push warning[UnusedImport]: on.}
+import std/terminal # Is not used in tests
+{.pop.}
+
 import std/options
 import std/strutils
 import std/typetraits
--- a/codex/sales/states/filled.nim
+++ b/codex/sales/states/filled.nim
@@ -10,7 +10,8 @@ import ./errored
 import ./cancelled
 import ./failed
 import ./proving
-import ./provingsimulated
+import ./provingsimulated # used in tests
+{.push warning[UnusedImport]:off.}
 
 logScope:
   topics = "marketplace sales filled"
--- a/codex/sales/states/filled.nim
+++ b/codex/sales/states/filled.nim
@@ -10,8 +10,10 @@ import ./errored
 import ./cancelled
 import ./failed
 import ./proving
-import ./provingsimulated # used in tests
+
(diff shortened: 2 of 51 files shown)


codex/sales/states/filled.nim
```diff
@@ -10,7 +10,8 @@ import ./errored
 import ./cancelled
 import ./failed
 import ./proving
-import ./provingsimulated
+import ./provingsimulated # used in tests
+{.push warning[UnusedImport]:off.}
```
gmega, line 14: Hm... why do we have an import that's used only in tests inside of a production module? Shouldn't we keep this warning around and address that instead?

codex/conf.nim
```diff
@@ -10,7 +10,11 @@
 {.push raises: [].}
 
 import std/os
-import std/terminal
+
+{.push warning[UnusedImport]: on.}
+import std/terminal # Is not used in tests
+{.pop.}
```
gmega, line 16: Ugh... OK. I guess this points out that we should have a config that's specific to tests, but OK, maybe not worth the trouble for a single module.

gmega (approved):
Thanks for the cleanup! Left a couple of comments but otherwise LGTM.

codex/sales/states/filled.nim
AuHau: Because we need to have some way how to bypass the "normal proving" module, as generating normal proofs in tests would take too long, and we also need a way to control how and when proving should fail. This functionality is hidden behind compile flag (see here: https://github.com/codex-storage/nim-codex/pull/1055/files/c7074e42d82fe32fee7056eab1fd0e2e8cb9e942#diff-c422f0a891c99abb9c168acd1d6d88f2bd50b0571521e10fb5668a759a9e0e0dR56)  so it is not really shipped in production. 

I originally tried to move the import in "local import" fashion to the `when` below, but that does not work in Nim, but you gave me an idea to just add `when` to the import then ;-)

PR #1062, opened 2025-01-09
Add nix service definition for codex

markoburcul:
Added service definition for Nim Codex and tests.
The service can be instantiated by including it in flake inputs:
```nix
  inputs = {
    nixpkgs.url = "github:NixOS/nixpkgs/nixos-24.11";
    codex-flake.url = "git+https://github.com/codex-storage/nim-codex?submodules=1#";
  };
```

and configuring the service:
```nix
services.nim-codex = {
   enable = true;
   settings = {
       data-dir = "/var/lib/codex-test";
   };
};
```

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -7,17 +7,16 @@
         ]
       },
       "locked": {
-        "lastModified": 1736348751,
+        "lastModified": 1736521871,
         "narHash": "sha256-d34XNLg9NGPEOARHW+BIOAWalkHdEUAwsv3mpLZQxds=",
         "owner": "codex-storage",
         "repo": "circom-compat-ffi",
-        "rev": "afadf4d9a411ce0589f6b4c1858a9a5a4e7f4661",
+        "rev": "8cd4ed44fdafe59d4ec1184420639cae4c4dbab9",
         "type": "github"
       },
       "original": {
         "owner": "codex-storage",
         "repo": "circom-compat-ffi",
-        "rev": "afadf4d9a411ce0589f6b4c1858a9a5a4e7f4661",
         "type": "github"
       }
     },
--- a/flake.lock
+++ b/flake.lock
@@ -2,50 +2,37 @@
   "nodes": {
     "circom-compat": {
       "inputs": {
-        "nixpkgs": "nixpkgs"
+        "nixpkgs": [
+          "nixpkgs"
+        ]
       },
       "locked": {
-        "lastModified": 1732627240,
-        "narHash": "sha256-GvJTiBWBv799i5ZCCc4gF86bnQY/nZvx0vCPi1+OPD4=",
+        "lastModified": 1736348751,
+        "narHash": "sha256-d34XNLg9NGPEOARHW+BIOAWalkHdEUAwsv3mpLZQxds=",
         "owner": "codex-storage",
         "repo": "circom-compat-ffi",
-        "rev": "297c46fdc7d8a8fd53c8076b0be77334e4a54447",
+        "rev": "afadf4d9a411ce0589f6b4c1858a9a5a4e7f4661",
         "type": "github"
       },
       "original": {
         "owner": "codex-storage",
         "repo": "circom-compat-ffi",
+        "rev": "afadf4d9a411ce0589f6b4c1858a9a5a4e7f4661",
         "type": "github"
       }
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1731386116,
-        "narHash": "sha256-lKA770aUmjPHdTaJWnP3yQ9OI1TigenUqVC3wweqZuI=",
+        "lastModified": 1736200483,
+        "narHash": "sha256-JO+lFN2HsCwSLMUWXHeOad6QUxOuwe9UOAF/iSl1J4I=",
         "owner": "NixOS",
         "repo": "nixpkgs",
-        "rev": "689fed12a013f56d4c4d3f612489634267d86529",
+        "rev": "3f0a8ac25fb674611b98089ca3a5dd6480175751",
         "type": "github"
       },
       "original": {
         "owner": "NixOS",
-        "ref": "nixos-24.05",
-        "repo": "nixpkgs",
-        "type": "github"
-      }
-    },
-    "nixpkgs_2": {
-      "locked": {
-        "lastModified": 1729449015,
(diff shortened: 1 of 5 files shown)


nix/codex.nix
```diff
@@ -0,0 +1,380 @@
...
+    systemd.tmpfiles.rules = [
+      "d ${cfg.dataDir or "/var/lib/codex"} 0755 codex codex"
+    ];
+
+    systemd.services.codex = {
+      description = "Codex Node";
+      wantedBy = [ "multi-user.target" ];
+      serviceConfig = {
+        User = cfg.service.user;
+        Group = cfg.service.group;
```
jakubgs, line 324: Why not just use a `DynamicUser` and skip user creation altogether?
```suggestion
        DynamicUser = true;
```
We could also add some hardening options:
```
        PrivateTmp = true;
        ProtectHome = true;
        ProtectSystem = "full";
        NoNewPrivileges = true;
        PrivateDevices = true;
        MemoryDenyWriteExecute = true;
```

nix/codex.nix
```diff
@@ -0,0 +1,380 @@
...
+        default = null;
+        description = "Additional arguments to pass to the Codex binary.";
+      };
+
+    };
+  };
+
+  config = mkIf cfg.enable {
+
+    users.users = optionalAttrs (cfg.service.user == "codex") {
```
jakubgs, line 301: The `optionalAttrs` function seems to come from nowhere.

nix/codex.nix
```diff
@@ -0,0 +1,380 @@
...
+    systemd.services.codex = {
+      description = "Codex Node";
+      wantedBy = [ "multi-user.target" ];
+      requires = [ "network.target" ];
+      serviceConfig = {
+        User = cfg.service.user;
+        Group = cfg.service.group;
+        ExecStart = ''
+          ${cfg.package}/bin/codex \
+          ${optionalString (cfg.configFile != null) "--config-file=${cfg.configFile}"} \
```
jakubgs, line 327: I think adding flags to the service directly is fine. But if service supports a config file a reasonable thing would be to just generate the TOML file from a config object. You can define generic objects in options in Nix, for example:
```nix
tomlFormat = pkgs.formats.toml { };
```
```nix
    settings = lib.mkOption {
      default = { };
      type = tomlFormat.type;
      description = ''Structured settings object that will be used to generate a TOML config file.'';
    };
```
And you can easily create a TOML file using `pkgs.formats.toml`. Here's an [example](https://github.com/NixOS/nixpkgs/blob/3f0a8ac25fb674611b98089ca3a5dd6480175751/nixos/modules/virtualisation/containers.nix#L35-L39).

jakubgs (approved):
I think you've over-done it with defining every single setting, but it's not a bad thing :D.

Also, a README might benefit from an example of how to use the service.

nix/codex.nix
markoburcul: So if I understood correctly we would just expose configFile and configObject, where configObject is of type tomlFormat.type? We don't need then to define any other options right?

jakubgs: Essentially yeah.

There is some value in defining options explicitly in Nix, since they are more easily discoverable that way. But they are more rigid and don't allow users to expand config how they want. Most services define only a few core settings, for example listen address and port, while leaving the rest that's not relevant to most users to a generic object `settings` where they can put whatever they want.

nix/codex.nix
```diff
@@ -0,0 +1,58 @@
...
+  inherit (lib)
+    types mkEnableOption mkOption mkIf literalExpression;
+  
+  toml = pkgs.formats.toml { };
+
+  cfg = config.services.codex;
+in
+{
+  options = {
+    services.codex = {
```
jakubgs, line 13: I think it should be called `nim-codex`. That's the software we are running here, there's no generic `codex`, this is a specific implementation of Codex protocol which works in a specific way.

nix/codex.nix
```diff
@@ -0,0 +1,58 @@
...
+      requires = [ "network.target" ];
+      serviceConfig = {
+        DynamicUser = true;
+        PrivateTmp = true;
+        ProtectHome = true;
+        ProtectSystem = "full";
+        NoNewPrivileges = true;
+        PrivateDevices = true;
+        MemoryDenyWriteExecute = true;
+        ExecStart = ''${cfg.package}/bin/codex --config-file=${if cfg.configFile != null then cfg.configFile else "/etc/codex/config.toml"}'';
```
jakubgs, line 53: Not really something I've seen before. We should not use `/etc/codex/config.toml` as fallback since it probably won't exist. Instead we should simply provide a default `settings` object that includes some sane common default.

nix/codex.nix
```diff
@@ -0,0 +1,58 @@
...
+        description = lib.mdDoc "Package to use as Codex node.";
+      };
+
+      configFile = mkOption {
+        type = types.nullOr types.str;
+        default = null;
+        description = "Path to the Codex configuration file.";
+      };
+
+      configObject = lib.mkOption {
```
jakubgs, line 29: The usual name for this is `settings`. See https://search.nixos.org/options for examples.

nix/codex.nix
```diff
@@ -0,0 +1,58 @@
...
+        default = pkgs.callPackage ./default.nix { };
+        defaultText = literalExpression "pkgs.codex";
+        description = lib.mdDoc "Package to use as Codex node.";
+      };
+
+      configFile = mkOption {
+        type = types.nullOr types.str;
+        default = null;
+        description = "Path to the Codex configuration file.";
+      };
```
jakubgs, line 27: If we are providing `settings` object we don't need this. Also, if we did we should use `types.nullOr types.path`.

nix/codex.nix
```diff
@@ -0,0 +1,58 @@
...
+        type = toml.type;
+        description = ''Structured settings object that will be used to generate a TOML config file.'';
+      };
+    };
+  };
+
+  config = mkIf cfg.enable {
+    environment.etc = {
+      "codex/config.toml".source = toml.generate "config.toml" cfg.configObject;
+    };
```
jakubgs, line 40: If you want to do it like this, you will need to include `restartTriggers` for the service, because right now the service would not be restarted if the config file was changed. Normally NixOS services reference the generated config directly from `/nix/store`, which means the service would be restarted every time it changes, since the path would be different and service definition would have to be changed.

jakubgs:
Readme doesn't have an example of how to load the service module.

PR #1065, opened 2025-01-13
Log repostore timings

benbierens:
This adds a quick and easy way to keep an eye on repostore performance.
Hidden behind a compile flag, enabled in the dist-test images by default.

This can be used to automatically check on the disk IO performance in cluster-test scenarios.

### DIFF
--- a/codex/errors.nim
+++ b/codex/errors.nim
@@ -8,6 +8,8 @@
 ## those terms.
 
 import std/options
+import std/sugar
+import std/sequtils
 
 import pkg/results
 import pkg/chronos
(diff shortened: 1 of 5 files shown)


benbierens:
Following a chat with Giuliano, we agreed that information of this kind of much more useful as a metric than as log entry parameters. The log entries might be useful for testing/benchmarking. But from a node-operator point of view, the metrics are much more useful.

We found that the default collector types are not useful in this case, so a StatSummary type was set up following the example of custom collectors from the metrics library. I bet it'll be useful in other cases as well. (Block retrieval time comes to mind!)

Question is: Is it worth removing the log parameter? It doesn't really cost anything to have it.

gmega:
We'll be replacing the repostore, so for now this won't be needed.

PR #1067, opened 2025-01-14
style: nph formatting

AuHau:
This PR implements code formatting using `nph` of folders `codex/` and `tests/`. 

It is enforced on CI level using https://github.com/arnetheduck/nph-action. Nice thing it can suggest formatting changes in case somebody forgets to format its PR before submitting it.

More details on how to use it can be found in README: https://github.com/codex-storage/nim-codex/pull/1067/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R60

This work is based on `nwaku` implementation of `nph` - so kudos to them!


benbierens (approved):
It seems to have a precommit hook and a workflow checking step. Even if the style itself isn't perfect, I think this is still a very good thing to have.

gmega:
Yes we know you just want to have your name amongst the contributors with the most lines added/deleted, @AuHau ...

Makefile
```diff
@@ -199,4 +200,42 @@ ifneq ($(USE_LIBBACKTRACE), 0)
...
 
+############
+## Format ##
+############
+.PHONY: build-nph install-nph-hook clean-nph print-nph-path
+
+# Default location for nph binary shall be next to nim binary to make it available on the path.
+NPH:=$(shell dirname $(NIM_BINARY))/nph
+
+build-nph:
```
gmega, line 211: :yellow_circle: OK, I guess we're going through this trouble (having nph in our vendor folder and compiling it ourselves, instead of just requiring people to install it) to keep this self-contained and at a version we can control. But then in CI we use `latest` for `arnetheduck/nph-action`, which might end up being a different version than the one we have in our submodule? Can't this lead to head-scratching situations where you run nph but then it fails in CI because versions don't match?

gmega (commented):
On a more serious note: I'm happy to have nph, but I think we need to make sure versions match locally and in CI.

Makefile
emizzle: Agreed... instead of using `latest` in CI, maybe we could use the version that matches the vendor submodule?

AuHau: Yeah agree as well. I mean Waku guys use the `nph` from submodule, but since then there was the action introduced which seems like way easier and faster way to use it. Will see if I can pin those two together or will have to take the Waku road as well.

AuHau:
@emizzle and @gmega - CI action pinned. Please re-review ;-)

emizzle (approved):
LGTM

arnetheduck:
> committer

see https://github.com/arnetheduck/nph/blob/master/format-git-repo.sh#L12 for how to not become the target of future `git blame` - after all, the last to change the code is the one that is responsible for it's maintenance until the next sucker comes around ;)

AuHau:
> > committer
> 
> see [arnetheduck/nph@`master`/format-git-repo.sh#L12](https://github.com/arnetheduck/nph/blob/master/format-git-repo.sh?rgh-link-date=2025-01-15T08%3A14%3A02Z#L12) for how to not become the target of future `git blame` - after all, the last to change the code is the one that is responsible for it's maintenance until the next sucker comes around ;)

That is a super cool tip! Thanks will employ it ;-)

README.md
```diff
@@ -51,3 +52,15 @@ To get acquainted with Codex, consider:
...
+
+Feel free to dive in, contributions are welcomed! Open an issue or submit PRs.
+
+### Linting and formatting
+
+`nim-codex` uses [nph](https://github.com/arnetheduck/nph) for formatting our code and it is requrired to adhere to its styling.
+If you are setting up fresh setup, in order to get `nph` run `make build-nph`.
+In order to format files run `make nph/<file/folder you want to format>`. 
+If you want you can install Git pre-commit hook using `make install-nph-commit`, which will format modified files prior commiting them. 
+If you are using VSCode and the NimLang extension you can enable "Format On Save" that will format the files using `nph`.
```
marcinczenko, line 66: Is this "Format On Save" or is it `nim.lintOnSave`? There is also `editor.formatOnSave` option, and perhaps it also has to be set to `true`, but `formatOnSave` is just a generic VSCode option if I remember correctly.

image


Also, you could consider adding a link to the [NimLang](https://marketplace.visualstudio.com/items?itemName=NimLang.nimlang) to remove any doubts about the extension.

marcinczenko (changes requested):

marcinczenko (approved):

README.md
AuHau: image

Yeah it is, it is `nim.formatOnSave`. Will clarify though - thanks.

marcinczenko: Ok, I did some tests. So it looks like `nim.lintOnSave` is actually depreciated and `nim.formatOnSave` does exist on NimLang extension. Setting either `editor.formatOnSave` or `nim.formatOnSave` does the job (one of them need to be set to `true`, the order does not seem to matter).

PR #1073, opened 2025-01-16
chore(marketplace): cid should be bytes 

2-towns:
Related to https://github.com/codex-storage/codex-contracts-eth/pull/214

### DIFF
--- a/codex/contracts/requests.nim
+++ b/codex/contracts/requests.nim
@@ -29,7 +29,7 @@ type
     maxSlotLoss* {.serialize.}: uint64
 
   StorageContent* = object
-    cid* {.serialize.}: string
+    cid*: seq[byte]
     merkleRoot*: array[32, byte]
 
   Slot* = object
--- a/codex/contracts/requests.nim
+++ b/codex/contracts/requests.nim
@@ -6,8 +6,10 @@ import pkg/nimcrypto
 import pkg/ethers/fields
 import pkg/questionable/results
 import pkg/stew/byteutils
+import pkg/libp2p/[cid, multicodec]
 import ../logutils
 import ../utils/json
+from ../errors import mapFailure
 
 export contractabi
 
@@ -29,7 +31,7 @@ type
     maxSlotLoss* {.serialize.}: uint64
 
   StorageContent* = object
-    cid* {.serialize.}: seq[byte]
+    cid* {.serialize.}: Cid
     merkleRoot*: array[32, byte]
 
   Slot* = object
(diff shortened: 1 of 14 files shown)


tests/examples.nim
```diff
@@ -56,7 +56,7 @@ proc example*(_: type StorageRequest): StorageRequest =
       maxSlotLoss: 2 # 2 slots can be freed without data considered to be lost
     ),
     content: StorageContent(
-      cid: "zb2rhheVmk3bLks5MgzTqyznLu1zqGH5jrfTA1eAZXrjx7Vob",
+      cid: "zb2rhheVmk3bLks5MgzTqyznLu1zqGH5jrfTA1eAZXrjx7Vob".toBytes,
```
dryajov, line 59: We should initialize a cid object with the string (`Cid.init("zb2rhheVmk3bLks5MgzTqyznLu1zqGH5jrfTA1eAZXrjx7Vob")`) and the access the raw buffer as we do elsewhere (cid.data.buffer)

2-towns: Okay, done.

AuHau (changes requested):

codex/contracts/requests.nim
```diff
@@ -29,7 +29,7 @@ type
     maxSlotLoss* {.serialize.}: uint64
 
   StorageContent* = object
-    cid* {.serialize.}: string
+    cid*: seq[byte]
```
emizzle, line 32: Shouldn't this be a `Cid` type instead of `seq[byte]`?

emizzle (changes requested):
I may be wrong here, but shouldn't be using the `Cid` type instead of `seq[byte]`?

codex/contracts/requests.nim
```diff
@@ -29,7 +29,7 @@ type
     maxSlotLoss* {.serialize.}: uint64
 
   StorageContent* = object
-    cid* {.serialize.}: string
```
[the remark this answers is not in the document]
2-towns: Yes you are right, my bad.

codex/contracts/requests.nim
2-towns: No, I don't think so. The `Cid` is an object, while the previous type was a string. So I am not sure I understand why we would need to move from `string` to `Cid` rather than from `string` to `seq[byte]`.

codex/contracts/requests.nim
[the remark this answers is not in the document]
AuHau: No worries, that is what reviews are for ;-)

AuHau (approved):
LGTM 👍

codex/contracts/requests.nim
emizzle: From our discussion, we agreed that using a `Cid` would be best here.

2-towns: I applied the changes.

markspanbroek (approved):
Thanks @2-towns, looks good! Feel free to merge it once https://github.com/codex-storage/codex-contracts-eth/pull/214 is merged, so you can update the submodule in this PR.

emizzle (approved):

PR #1076, opened 2025-01-17
ci: create jenkins pipeline and enable build of nim compiler

markoburcul:
Because of the recent bump of Nim compiler for codex to >2 version, using the system Nim provided by nixpkgs became unusable since the nim-codex couldn't build. Because of this I've adapted the derivation to be similar to the one of [nimbus-eth2](https://github.com/status-im/nimbus-eth2) where we build Nim compiler using nimbus-build-system.  

The pipeline on Jenkins is currently set to build nim-codex through flake, but there is a possibility to run the nix tests in the pipeline with slight modifications of the ci agents([the error](https://ci.status.im/job/nim-codex/job/systems/job/linux/job/x86_64/job/package/58/console) when running nix tests).
```shell
error: a 'x86_64-linux' with features {kvm, nixos-test} is required to build '/nix/store/9avanzkwb8f25c0az0f3f8jv2mkm4gfq-vm-test-run-nim-codex-test.drv', but I am a 'x86_64-linux' with features {benchmark, big-parallel, nixos-test, uid-range}
```

P.S. Unfortunately it is not possible anymore(to my knowledge) to instantiate the nix shell without specifying additionally `'.?submodules=1#'`. Cause of this is that the `callPackage` from `preBuild` phase is being called and all three calls  require submodules to be present.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 
 !LICENSE*
 !Makefile
+!Jenkinsfile
 
 nimcache/
 
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,9 @@
 
 nimcache/
 
+# Executables when using nix will be stored in result/ directory
+result/
+
 # Executables shall be put in an ignored build/ directory
 build/
 
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,37 @@
+#!/usr/bin/env groovy
+library 'status-jenkins-lib@v1.9.13'
+
+pipeline {
+  agent { label 'linux && x86_64 && nix-2.24' }
+
+  options {
+    disableConcurrentBuilds()
+    /* manage how many builds we keep */
+    buildDiscarder(logRotator(
+      numToKeepStr: '20',
+      daysToKeepStr: '30',
+    ))
+  }
+
+  stages {
+    stage('Build') {
+      steps {
+        script {
+          nix.flake("default")
+        }
+      }
+    }
+
+    stage('Check') {
+      steps {
+        script {
+          sh './result/bin/codex --version'
+        }
+      }
+    }
+  }
+
+  post {
+    cleanup { cleanWs() }
+  }
+}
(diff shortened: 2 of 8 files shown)


markoburcul:
I've inspected the dependencies of the built derivation both manually with `nix derivation show` and with [nix-tree](https://github.com/utdemir/nix-tree). There was no hidden Nim compiler or anything similar among the dependecies that would explain the previous failures we had when debugging the build.

When building the package we can see in the output first:
```shell
codex> Nim Compiler Version 1.9.1 [Linux: amd64]
codex> Compiled at 2023-01-02
codex> Copyright (c) 2006-2022 by Andreas Rumpf
codex> git hash: 7f6681b4c4ccc0dc43fd256280be4c3ad3c773e5
```
but afterwards we can see:
```shell
codex> Building: build/all
codex> [using Nim version v2.0.14]
```
which states clearly that we are using the 2.0.14 Nim compiler.

Jenkinsfile
```diff
@@ -0,0 +1,29 @@
...
+      numToKeepStr: '20',
+      daysToKeepStr: '30',
+    ))
+  }
+
+  stages {
+    stage('Build') {
+      steps {
+        script {
+          nix.flake("default")
```
jakubgs, line 20: I would add a second stage where you run the built binary with `--version` like we do in `go-waku`:
https://github.com/status-im/go-waku/blob/4ef460cb951a72b2a0be2b3f23347ef8f569ea08/ci/Jenkinsfile.nix-flake#L33-L37

Jenkinsfile
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env groovy
+library 'status-jenkins-lib@v1.9.13'
+
+pipeline {
+  agent { label 'linux' }
```
jakubgs, line 5:
```suggestion
    label 'linux && x86_64 && nix-2.24'
```

.gitignore
```diff
@@ -5,6 +5,7 @@
 
 !LICENSE*
 !Makefile
+!Jenkinsfile
```
jakubgs, line 8: Maybe we should add `result/` to `.gitignore` too.

jakubgs (approved):
Looks mostly correct. Nice work.

PR #1080, opened 2025-01-20
chore(client): delete leftover code from old threaded erasure coding backend

gmega:
Description says it all: this is no longer used, and shouldn't be here.


2-towns (approved):
LGTM. Thanks Giuliano.

Issue #1072, opened 2025-01-16
TestNet issue: Slots not reserving


benbierens:
TLDR:
"Slot reservations full" events stopped occuring on the testnet after a few weeks of active use by autoclients.
Cloud node "codex-01" was the only node (of 10) still actively filling slots.
We should find out why the nodes stopped trying to reserve and fill slots.

Details:
See Discord entry: https://discord.com/channels/895609329053474826/1163106961064198144/1313086195160317983

Following Leo's plan to upload the entire pre-merge ethereum history into our testnet, I had modified and run the autoclient to perform this work. It ran for several days and got a considerable portion of the data into the testnet. Then we noticed in the chain-events channel that "slot reservations full" message was no longer appearing and that our cloud node "codex-01" was the only node still picking up slots. The reason why the other nodes stopped filling slots is so far unexplained.

markspanbroek:
This might have been caused by the fact that we don't have [persistent availabilties](https://github.com/codex-storage/nim-codex/issues/467) yet. Would be nice if we can reproduce this and check whether the nodes do not pick up slots because they don't have [matching availability](https://github.com/codex-storage/nim-codex/blob/39e8e6e6fa35d8ea78ffcba3901a5e7ebd356a6a/codex/sales/states/preparing.nim#L72).

veaceslavdoina:
1. Wings running storage nodes as well and we might ask him to check the logs.
 2. We run nodes with same configuration using templates, sometimes they might run on same IP - will collect that information.
 3. All our nodes were started after [latest release - v0.1.9 (Nov 26, 2024)](https://github.com/codex-storage/nim-codex/issues/1000).
   Every node log `started codex node` when Pod is started or container in a Pod is restarted. My assumption is the following
    - We have logs for just last 30 days - as for today it is `16.12.2024-16.01.2025`
    - Only container in a Pod codex-1 was restarted during that period - check screenshot
      <details>
      <summary>Codex restarts</summary>
       <img width="1181" alt="Image" src="https://github.com/user-attachments/assets/f7f711e8-206f-4b10-9a76-8ec2da93348a" />
      </summary>

    - That can be confirmed using previous container logs - the one before current start
      ```
      # Last line of the previous container - probably last line after the crash

      # Codex 1
[...]

PR #1081, opened 2025-01-20
fix slotqueue worker starvation

markspanbroek:
When an item in the slotqueue is already seen, then a worker got popped from the list of workers, and was not added back again. This meant that the number of workers steadily declined as the queue was paused and unpaused.

possible fix for #1072

### DIFF
--- a/tests/codex/sales/testslotqueue.nim
+++ b/tests/codex/sales/testslotqueue.nim
@@ -132,6 +132,11 @@ suite "Slot queue":
     check queue.len == 0
     check $queue == "[]"
 
+  test "starts with 0 active workers":
+    newSlotQueue(maxSize = 2, maxWorkers = 2)
+    check eventually queue.running
+    check queue.activeWorkers == 0
+
   test "reports correct size":
     newSlotQueue(maxSize = 2, maxWorkers = 2)
     check queue.size == 2
@@ -506,15 +511,6 @@ suite "Slot queue":
       ]
     )
 
-  test "processing a 'seen' item pauses the queue":
-    newSlotQueue(maxSize = 4, maxWorkers = 4)
-    let request = StorageRequest.example
-    let item =
-      SlotQueueItem.init(request.id, 0'u16, request.ask, request.expiry, seen = true)
-    check queue.push(item).isOk
-    check eventually queue.paused
-    check onProcessSlotCalledWith.len == 0
-
   test "pushing items to queue unpauses queue":
     newSlotQueue(maxSize = 4, maxWorkers = 4)
     queue.pause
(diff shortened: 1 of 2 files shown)


tests/codex/sales/testslotqueue.nim
```diff
@@ -562,6 +556,54 @@ suite "Slot queue":
...
+    )
+    # push seen item
+    check queue.push(seen).isSuccess
+    # unpause and pause a number of times
+    for _ in 0..<10:
+      check queue.push(unseen).isSuccess
+      check eventually queue.len == 1
+      check eventually queue.paused
+      check eventually queue.activeWorkers == 0
+
```
marcinczenko, line 606: I am wondering if we could not simplify this test and because there is lots of async processing that is not so easy to capture, I would suggest adding some comments:

```nim
test "processing a 'seen' item does not exhaust the worker queue":
  newSlotQueue(maxSize = 4, maxWorkers = 4)
  let request = StorageRequest.example
  let unseen = SlotQueueItem.init(
    request.id, 0'u16,
    request.ask,
    request.expiry,
    seen = false
  )
  let seen = SlotQueueItem.init(
    request.id, 1'u16,
    request.ask,
    request.expiry,
    seen = true
  )
  check queue.push(seen).isSuccess
  # push unseen item - it will unpause the queue that was initially paused
  # unseen item will be processed before seen item
  check queue.push(unseen).isSuccess

  # wait for queue to be paused because of seen item
  check eventually queue.paused
  # after marking queue as paused a new worker should immediately be added
  # thus the total number of workers should remain the same and
  # equal to maxWorkers (thus activeWorkers == 0)
  check queue.activeWorkers == 0
```

I tried it out - and it will fail if the worker is not re-added to the queue for "seen" items.

If we still like to have the "loop" version (which I do not mind), what would you say if we change the sequence a bit and we add some comments as well. I also replaced some `eventually` with regular checks as this makes the "loop" test to fail faster:

```nim
test "processing a 'seen' item does not decrease the number of workers":
  newSlotQueue(maxSize = 4, maxWorkers = 4)
  let request = StorageRequest.example
  let unseen = SlotQueueItem.init(
    request.id, 0'u16,
    request.ask,
    request.expiry,
    seen = false
  )
  let seen = SlotQueueItem.init(
    request.id, 1'u16,
    request.ask,
    request.expiry,
    seen = true
  )
  # push seen item
  check queue.push(seen).isSuccess

  # unpause and pause a number of times
  for _ in 0..<10:
    # wait for queue to be paused (notice that queue is initially paused
    # and processing seen item will pause the queue as well)
    check eventually queue.paused
    # after marking queue as paused a new worker should immediately be added
    # thus the total number of workers should remain the same and
    # equal to maxWorkers (thus activeWorkers == 0)
    check queue.activeWorkers == 0
    # push unseen item to unpause queue
    # the unseen item will be processed before seen item
    check queue.push(unseen).isSuccess
    check queue.len == 2
    # wait for unseen item to be processed
    check eventually queue.len == 1
    # after processing unseen item, new worker should be added
    check queue.activeWorkers == 0
```

marcinczenko: I would also consider adding the following test at the beginning:

```nim
suite "Slot queue start/stop":
  # ...  
  test "starts out paused":
    check queue.paused
  # ...
```

marcinczenko (changes requested):
Very good catch Mark. Perhaps you will consider some suggestions.

tests/codex/sales/testslotqueue.nim
markspanbroek: Thanks @marcinczenko, I applied your suggestions in ee8dacf. I kept the loop version, because it more accurately represents the scenario in which things went wrong in the testnet.

markspanbroek: By the way, I'm assuming that after the review you agreed with the changes but just had some suggestions on how to make it better. It's then probably better to use the "approve" option than the "request changes" option in Github, because then I don't have to wait for your approval again before I can merge this :smile:

marcinczenko: > Thanks @marcinczenko, I applied your suggestions in [ee8dacf](https://github.com/codex-storage/nim-codex/commit/ee8dacfc416c8a310fe3c2c2781d44b4cef53eb3). I kept the loop version, because it more accurately represents the scenario in which things went wrong in the testnet.

I would keep both tests, because the shorter test is more localized and it better documents the requirement. It also fails immediately, while the loop version needs almost 2 minutes to fail (around 1 minute if you do not use `eventually` in the last check - but perhaps you want to keep it there because you think it is testing an implementation detail).

marcinczenko (approved):

PR #1083, opened 2025-01-21
feat(marketplace): indicate that slot is being repaired when trying to download

### LINKED ISSUE #891: Indicate that slot is being repaired
Indicate to "download slot" function when the slot was freed and it is being repaired.

2-towns:
Fix #891 .


codex/sales/states/downloading.nim
```diff
@@ -65,10 +65,14 @@ method run*(state: SaleDownloading, machine: Machine): Future[?State] {.async.}
...
 
+  let slotId = slotId(request.id, data.slotIndex)
+  let isRepairing = (await context.market.slotState(slotId)) == SlotState.Repair
+
   trace "Starting download"
   if err =? (await onStore(request,
                            data.slotIndex,
-                           onBlocks)).errorOption:
+                           onBlocks,
+                           isRepairing)).errorOption:
```
2-towns, line 75: @emizzle @AuHau I know that we discussed about using a value from the context but I found it difficult to add it. I wanted to add this information into `SalesData` but it wasn't accessible from the event listener `onSlotFreed`.

AuHau (commented):
Good first approach, but I would have alternative suggestion. See comments.

codex/sales.nim
```diff
@@ -330,7 +336,15 @@ proc onSlotFreed(sales: Sales, requestId: RequestId, slotIndex: UInt256) =
           error "unknown request in contract"
           return
```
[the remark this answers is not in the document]
AuHau: `queue.populateItem` is actually used only here and nowhere else. IMHO I would rather replace it with caching on `getRequest()` level.

codex/sales/slotqueue.nim
```diff
@@ -36,6 +36,8 @@ type
     collateral: UInt256
     expiry: UInt256
     seen: bool
+    isRepairing: bool
+    repairRewardPercentage: UInt256
```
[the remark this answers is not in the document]
2-towns: Okay I applied this change.

codex/sales.nim
```diff
@@ -149,6 +149,10 @@ proc cleanUp(
       ).errorOption:
     error "failure deleting reservation", error = deleteErr.msg
 
+  let repairRewardPercentage = (await sales.context.market.repairRewardPercentage).u256
+  let slotId = slotId(data.requestId, data.slotIndex)
+  let isRepairing = (await sales.context.market.slotState(slotId)) == SlotState.Repair
+
```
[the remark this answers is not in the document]
2-towns: What do you think about passing the request collateral instead of the request id? It's more convenient because we already have the information everywhere when we need to calculate the repair collateral.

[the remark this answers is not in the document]
2-towns: So you would pass the slotId and the collateral to that function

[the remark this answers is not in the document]
AuHau: I don't understand. Passing in collateral? I thought that the helper function should calculate the collateral 😅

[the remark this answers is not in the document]
2-towns: I mean the collateral from the request and then it will calculate the repaired collateral it the slot state is being repaired. 
If we pass the request id and the slot index, as you suggested, we will need to retrieve the request data but this is not needed because we already have this data when we want to calculate the collateral in the `onSlotFreed` event and in the filling state.   

Or another option is to pass the StorageAsk so when the #1078 will be merge we could use the `collateralPerByte` and the `slotSize` to calculate the total slot collateral.

codex/sales/slotqueue.nim
```diff
@@ -34,6 +34,7 @@ type
     duration: UInt256
     pricePerBytePerSecond: UInt256
     collateralPerByte: UInt256
+    collateralPerSlot: UInt256
```
2-towns, line 37: @AuHau Since we have collateralPerByte now, I am not able to use the collateral field from the StorageAsk and I cannot apply the reward percent to the collateralPerByte because, as we discussed in the meeting, it could have some problem with float value. So I added this field `collateralPerSlot` which is a computed value. It preserves the SlotQueueItem immutability and the `collateralPerSlot` is not something `new` because we already have `collateralPerByte` now.

emizzle (commented):
Some suggestions:
1. pull out the LRU cache changes into a separate PR. It would be good to include calls to `marketplace.config` as well in that PR
2. `SlotQueueItem.collateralPerByte` > `SlotQueueItem.collateralPerSlot` (see comment)
3. Depending on the situation, we sometimes know the state of the slot (as you pointed out in one of the comments). For the times we don't know the state of the request, we need to get the state from the contract (network). So as an optimisation, maybe we could have two procs in the market abstraction: one that takes a SlotState, and one that doesn't. The one that takes the SlotState would do the calculation. The one that doesn't would need to get the SlotState from the contract before doing the calculation. Some of the known state situations where `market.slotCollateral` would be called that I can see would be: 
    - Sales module on storage requested: SlotState is Free
    - Sales module on slot freed: SlotState is Repair
    - filling state: SlotState is Free
    - downloading state: SlotState is Free
    - cancelled state onCleanUp: SlotState is Cancelled
    - errored state: pass in SlotState.Failed from where it originated
    - finished state onCleanUp: SlotState is Finished
    - ignored state onCleanUp: SlotState is Free

codex/sales/slotqueue.nim
```diff
@@ -145,14 +144,25 @@ proc init*(
     duration: ask.duration,
     pricePerBytePerSecond: ask.pricePerBytePerSecond,
     collateralPerByte: ask.collateralPerByte,
+    collateralPerSlot:
+      if collateralPerSlot.isZero: ask.collateralPerSlot else: collateralPerSlot,
```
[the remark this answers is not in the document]
2-towns: Yes ! I applied the changes.

codex/contracts/market.nim
```diff
@@ -483,3 +502,21 @@ method queryPastStorageRequestedEvents*(
     let fromBlock = await market.contract.provider.pastBlockTag(blocksAgo)
 
     return await market.queryPastStorageRequestedEvents(fromBlock)
+
+method slotCollateral*(
+    market: OnChainMarket, requestId: RequestId, slotIndex: UInt256
```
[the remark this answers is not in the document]
2-towns: Yes I applied the changes.

2-towns:
> Some suggestions:
> 
>     1. pull out the LRU cache changes into a separate PR. It would be good to include calls to `marketplace.config` as well in that PR
> 
>     2. `SlotQueueItem.collateralPerByte` > `SlotQueueItem.collateralPerSlot` (see comment)
> 
>     3. Depending on the situation, we sometimes know the state of the slot (as you pointed out in one of the comments). For the times we don't know the state of the request, we need to get the state from the contract (network). So as an optimisation, maybe we could have two procs in the market abstraction: one that takes a SlotState, and one that doesn't. The one that takes the SlotState would do the calculation. The one that doesn't would need to get the SlotState from the contract before doing the calculation. Some of the known state situations where `market.slotCollateral` would be called that I can see would be:
>        
>        * Sales module on storage requested: SlotState is Free
>        * Sales module on slot freed: SlotState is Repair
>        * filling state: SlotState is Free
>        * downloading state: SlotState is Free
>        * cancelled state onCleanUp: SlotState is Cancelled
>        * errored state: pass in SlotState.Failed from where it originated
>        * finished state onCleanUp: SlotState is Finished
>        * ignored state onCleanUp: SlotState is Free

1. Okay I will do it. But for `marketplace.config` the cache isn't existing already ? 
2. Okay did it. And I renamed to `collateral`. 
3. Thanks for those details. I tried to apply them on the different calls. But on the `cleanUp` function, I think we would need to pass the slot state in the function parameter (but I'm not sure we want to do that).

codex/sales.nim
```diff
@@ -284,7 +288,7 @@ proc onAvailabilityAdded(sales: Sales, availability: Availability) {.async.} =
 
 proc onStorageRequested(
     sales: Sales, requestId: RequestId, ask: StorageAsk, expiry: UInt256
-) =
+) {.async.} =
```
emizzle, line 291: We've had quite a few discussions and back and forth changes in the past to avoid async subscription handlers due to a number of pitfalls. So I think we have a few options, but I don't really like any of them 😂:
1. `asyncSpawn` and track `onStorageRequested` (see how it's being done with `addToSlotQueue`), which means this proc must not raise any exceptions (`{.async: (raises: []).}`). _This may have unforeseen consequences_
    - ```nim
        proc onStorageRequest(...) {.async: (raises: []).} =
          try:
            # ...
          except CancelledError as e:
            discard # do not propagate as onRequestStorage is asyncSpawned

        proc onStorageRequested(requestId: RequestId, ask: StorageAsk, expiry: UInt256) =
           let fut = sales.onStorageRequested(requestId, ask, expiry)
           sales.trackedFutures.track(fut)
           asyncSpawn fut
       ```
3. Keep it sync and add a new field to `SalesContext`: `config: ?MarketConfiguration`, populate it similar to how `retrieveRequest` is being done. This will allow you to access `sales.context.config.repairRewardPercentage`, passing it into a `slotCollateral` override that takes it as a parameter and therefore can be synchronous.
4. Keep it sync and export a new Market.configSync proc that gets the `market.configuration` value or raises an exception when `market.configuration.isNone`

emizzle (changes requested):
Hey sorry for requesting changes, but I think we need to address the async `onStorageRequested` proc. We've had many discussion around this, so we should be prudent about not introducing it here.

codex/contracts/market.nim
```diff
@@ -483,3 +483,25 @@ method queryPastStorageRequestedEvents*(
...
+): Future[UInt256] {.async.} =
+  let slotid = slotId(requestId, slotIndex)
+  let slotState = await market.slotState(slotid)
+
+  return await market.slotCollateral(requestId, slotState)
+
+method slotCollateral*(
+    market: OnChainMarket, requestId: RequestId, slotState: SlotState
+): Future[UInt256] {.async.} =
```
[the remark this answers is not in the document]
2-towns: Yes, I started to pay attention to this while working on the persistent availabilities, but I missed it for this PR. I will address it.

codex/sales.nim
2-towns: 1. I think we should not do that. 
2. I don't like the idea to add another field to `SalesContent` which is containing already lot of fields. 
3. This one looks not to bad. But it would be better to make sure that the configuration is already loaded by calling `Market.config` when registering the `onStorageRequested` event. 

Btw, don't we have the same problem with `onSlotFreed` which uses solutions 1. ?

emizzle: From our discussion, we decided to go with #3 but returning an Option (basically encapsulating `market.configuration`) instead of an raising exception

2-towns:
> Hey sorry for requesting changes, but I think we need to address the async `onStorageRequested` proc. We've had many discussion around this, so we should be prudent about not introducing it here.

No worries at all ! Feel free to request any changes you think are necessary.

codex/sales.nim
```diff
@@ -353,9 +377,14 @@ proc subscribeRequested(sales: Sales) {.async.} =
   let context = sales.context
   let market = context.market
 
-  proc onStorageRequested(requestId: RequestId, ask: StorageAsk, expiry: UInt256) =
+  proc onStorageRequested(
+      requestId: RequestId, ask: StorageAsk, expiry: UInt256
+  ) {.raises: [].} =
     sales.onStorageRequested(requestId, ask, expiry)
 
+  # Ensure that the config is loaded and repairRewardPercentage is available
```
2-towns, line 385: Not super happy with this but it seems the way to ensure that the configuration is loaded without exposing the `config` function. See https://github.com/codex-storage/nim-codex/blob/43f6a9ec0d0d1dcb94e0cc17792ff16ad130a5f1/tests/contracts/testMarket.nim#L78-L81.

codex/sales.nim
```diff
@@ -353,10 +377,15 @@ proc subscribeRequested(sales: Sales) {.async.} =
...
 
-  proc onStorageRequested(requestId: RequestId, ask: StorageAsk, expiry: UInt256) =
+  proc onStorageRequested(
+      requestId: RequestId, ask: StorageAsk, expiry: UInt256
+  ) {.raises: [].} =
     sales.onStorageRequested(requestId, ask, expiry)
 
   try:
+    # Ensure that the config is loaded in order to use the sync method slotCollateral
+    await market.loadConfig()
```
emizzle, line 387: Weren't we going to call this around here? 

https://github.com/codex-storage/nim-codex/blob/87590f43ce769bc964232e9450314d6047277f67/codex/codex.nim#L139-L140

codex/contracts/market.nim
```diff
@@ -55,46 +55,67 @@ template convertEthersError(body) =
...
-proc config(market: OnChainMarket): Future[MarketplaceConfig] {.async.} =
+proc config(
+    market: OnChainMarket
+): Future[MarketplaceConfig] {.async: (raises: [CatchableError]).} =
   without resolvedConfig =? market.configuration:
-    let fetchedConfig = await market.contract.configuration()
-    market.configuration = some fetchedConfig
-    return fetchedConfig
+    await market.loadConfig()
+    return !market.configuration
```
emizzle, line 63: > `(raises: [CatchableError])`
I try to avoid this wherever possible because it spreads everywhere and then we lose all ability to distinguish what types of errors are occurring where. If we wanted to simply inform callers what exception could leak out, we need to understand what exception is being raised inside each of these procedures and then either specify that exception type is "leaking" out, or convert it to a module-specific exception type, eg `MarketError`. 

> `!market.configuration`
In general, I would say it's probably best to avoid use the `!` here unless there is an assignment in the same scope and we know for sure there are no side effects occurring between the time of assignment and time of `!`. However, here i'm in two minds because if we can't fetch the marketplace config, we can't do much in the application, so may as well raise a Defect. 

All together, something like:

```nim
  method loadConfig*(
    market: OnChainMarket
): Future[void] {.async: (raises: [CancelledError, MarketError]).} =
  convertEthersError:
    without config =? market.configuration:
      let fetchedConfig = await market.contract.configuration()
      market.configuration = some fetchedConfig

proc config(
    market: OnChainMarket
): Future[MarketplaceConfig] {.async: (raises: [CancelledError, MarketError]).} =
  without resolvedConfig =? market.configuration:
      # don't need a try/except here bc we know `loadConfig` only raises `MarketError`
      await market.loadConfig()
      without config =? market.configuration:
         # we could all raise a Defect here with raiseAssert
         raiseMarketError("Failed to load config from the Marketplace contract") 
      return config
```

Most of the `(raises: [CatchableError])` changes here can be avoided by adding `convertEthersError`, which means they would only raise `MarketError` and `CancelledError`.

codex/contracts/market.nim
```diff
@@ -123,19 +144,21 @@ method requestStorage(market: OnChainMarket, request: StorageRequest) {.async.}
...
-      market.requestCache[key] = request
-      return some request
-    except Marketplace_UnknownRequest:
-      return none StorageRequest
+    let request = await market.contract.getRequest(id)
+    market.requestCache[key] = request
+    return some request
+  except Marketplace_UnknownRequest:
+    return none StorageRequest
+  except CatchableError as err:
```
emizzle, line 159: Because you removed `convertEthersError`, the exception type is likely `EthersError` here. By adding `except CatchableError`, we swallow `CancelledError`, which we don't want to do.

codex/contracts/market.nim
```diff
@@ -55,46 +55,67 @@ template convertEthersError(body) =
...
   return some config.proofs.zkeyHash
 
 method getSigner*(market: OnChainMarket): Future[Address] {.async.} =
   convertEthersError:
     return await market.signer.getAddress()
 
-method periodicity*(market: OnChainMarket): Future[Periodicity] {.async.} =
+method periodicity*(
+    market: OnChainMarket
+): Future[Periodicity] {.async: (raises: [CatchableError]).} =
```
emizzle, line 96: This and the next three can probably simply be `raises: [MarketError]`

codex/contracts/market.nim
```diff
@@ -502,3 +527,35 @@ method queryPastStorageRequestedEvents*(
...
+  let slotid = slotId(requestId, slotIndex)
+
+  try:
+    let slotState = await market.slotState(slotid)
+
+    without request =? await market.getRequest(requestId):
+      return UInt256.none
+
+    return market.slotCollateral(request.ask.collateralPerSlot, slotState)
+  except CatchableError as err:
```
emizzle, line 543:
```suggestion
  except MarketError as err:
```
This change will only work if the market abstraction changes i suggested above were made. Doing this will also not swallow CancelledErrors, stopping cancellation propagation.

codex/sales.nim
```diff
@@ -153,13 +153,19 @@ proc cleanUp(
   # Re-add items back into the queue to prevent small availabilities from
   # draining the queue. Seen items will be ordered last.
   if reprocessSlot and request =? data.request:
+    without collateral =?
+      await sales.context.market.slotCollateral(data.requestId, data.slotIndex):
+      error "Unable to calculate collateral; configuration data may not be retrievable."
```
emizzle, line 158:
```suggestion
      error "Failed to re-add item back to the slot queue: unable to calculate collateral; configuration data may not be retrievable."
```

codex/sales.nim
```diff
@@ -325,26 +337,38 @@ proc onSlotFreed(sales: Sales, requestId: RequestId, slotIndex: UInt256) =
...
+      return
+
+    # Take the repairing state into consideration to calculate the collateral.
+    # This is particularly needed because it will affect the priority in the queue
+    # and we want to give the user the ability to tweak the parameters.
+    # Adding the repairing state directly in the queue priority calculation
+    # would not allow this flexibility.
+    without collateral =?
+      market.slotCollateral(request.ask.collateralPerSlot, SlotState.Repair):
+      error "Unable to calculate collateral; configuration data may not be retrievable."
```
emizzle, line 351:
```suggestion
      error "Failed to add freed slot to queue: unable to calculate collateral; configuration data may not be retrievable."
```

codex/sales/states/filling.nim
```diff
@@ -35,18 +35,9 @@ method run(state: SaleFilling, machine: Machine): Future[?State] {.async.} =
...
-  if slotState == SlotState.Repair:
-    # When repairing the node gets "discount" on the collateral that it needs to
-    let repairRewardPercentage = (await market.repairRewardPercentage).u256
-    collateral =
-      requestedCollateral -
-      ((requestedCollateral * repairRewardPercentage)).div(100.u256)
-  else:
-    collateral = requestedCollateral
+  without collateral =? await market.slotCollateral(data.requestId, data.slotIndex):
+    error "Unable to calculate collateral; configuration data may not be retrievable."
```
emizzle, line 39:
```suggestion
    error "Failure attempting to fill slot: unable to calculate collateral; configuration data may not be retrievable."
```

tests/codex/helpers/mockmarket.nim
```diff
@@ -183,10 +196,18 @@ method requestState*(
...
+  if slotId notin market.slotState:
     return SlotState.Free
-  return market.slotState[slotId]
+
+  try:
+    return market.slotState[slotId]
+  except ref KeyError:
+    # Should never reach that case.
+    # Just returning a random slot state.
+    return SlotState.Repair
```
emizzle, line 210:
```suggestion
  except KeyError as e:
    raiseAssert "SlotId not found in known slots (MockMarket.slotState)
```

emizzle (commented):
Sorry for waiting so long to re-review! It's looking really good.  Just some minor changes that would help align it more to the exception tracking best practices.

codex/codex.nim
```diff
@@ -134,6 +134,12 @@ proc bootstrapInteractions(s: CodexServer): Future[void] {.async.} =
       if config.simulateProofFailures > 0:
         warn "Proof failure simulation is not enabled for this build! Configuration ignored"
 
+    try:
+      await market.loadConfig()
+    except MarketError:
+      error "Cannot load market configuration", error = getCurrentExceptionMsg()
+      quit QuitFailure
```
emizzle, line 141: 🟡  Just a suggestion that we can use the `fatal` log level in these instances. The other instances should probably also be fatal, but no need to cover that in this PR.
```suggestion
    except MarketError as e:
      fatal "Cannot load market configuration", error = e.msgDetail
```

codex/contracts/market.nim
```diff
@@ -53,43 +53,74 @@ template convertEthersError(body) =
   try:
     body
   except EthersError as error:
+    echo error.msgDetail
```
emizzle, line 56: leftover?

codex/contracts/market.nim
```diff
@@ -53,43 +53,74 @@ template convertEthersError(body) =
...
+method loadConfig*(
+    market: OnChainMarket
+): Future[void] {.async: (raises: [CancelledError, MarketError]).} =
+  convertEthersError:
+    try:
+      without config =? market.configuration:
+        let fetchedConfig = await market.contract.configuration()
+
+        market.configuration = some fetchedConfig
+    except AsyncLockError as err:
```
emizzle, line 91: Oh wow good find. This must be bubbling up from Ethers? If so, we should probably fix that in Ethers (in a different PR) so that it raises an `EthersError` instead.

emizzle: Great job isolating the exception here and not propagating CatchableError 👍

codex/contracts/market.nim
```diff
@@ -152,10 +188,17 @@ method requestState*(
...
   convertEthersError:
-    let overrides = CallOverrides(blockTag: some BlockTag.pending)
-    return await market.contract.slotState(slotId, overrides)
+    try:
+      let overrides = CallOverrides(blockTag: some BlockTag.pending)
+      return await market.contract.slotState(slotId, overrides)
+    except AsyncLockError as err:
+      raiseMarketError(
+        "Failed to fetch the slot state from the Marketplace contract: " & err.msg
+      )
```
emizzle, line 201: Nice work. When we make a fix in Ethers to convert AsyncLockError to EthersError, then this will be covered by `convertEthersError`. `convertEthersError` should also probably take a reason string parameter as well. I'll probably end up doing this soon 👍

codex/contracts/market.nim
```diff
@@ -507,3 +550,35 @@ method queryPastStorageRequestedEvents*(
...
+  try:
+    let slotState = await market.slotState(slotid)
+
+    without request =? await market.getRequest(requestId):
+      return UInt256.none
+
+    return market.slotCollateral(request.ask.collateralPerSlot, slotState)
+  except MarketError as err:
+    error "Cannot retrieve the slot collateral", error = err.msg
+    return UInt256.none
```
emizzle, line 568: 🟡  I don't see a particular reason to not let `MarketError` propagate here, letting the caller decide what to do with it. Certainly not a blocker, but i think it would make more sense that way

codex/sales.nim
```diff
@@ -153,13 +153,19 @@ proc cleanUp(
   # Re-add items back into the queue to prevent small availabilities from
   # draining the queue. Seen items will be ordered last.
   if reprocessSlot and request =? data.request:
+    without collateral =?
+      await sales.context.market.slotCollateral(data.requestId, data.slotIndex):
```
emizzle, line 157: If you decide to raise from `slotCollateral`, be careful using `.catch` with `await` (swallows CancelledError)

emizzle (approved):
Great job! I offered a few small suggestions, but there's nothing absolutely blocking, so I'll approve now 👍

codex/contracts/market.nim
2-towns: > Oh wow good find. This must be bubbling up from Ethers? If so, we should probably fix that in Ethers (in a different PR) so that it raises an `EthersError` instead.

Not sure, but will try to figure out

codex/contracts/market.nim
```diff
@@ -55,41 +55,71 @@ template convertEthersError(body) =
...
+
 proc approveFunds(market: OnChainMarket, amount: UInt256) {.async.} =
   debug "Approving tokens", amount
   convertEthersError:
     let tokenAddress = await market.contract.token()
     let token = Erc20Token.new(tokenAddress, market.signer)
     discard await token.increaseAllowance(market.contract.address(), amount).confirm(1)
 
-method getZkeyHash*(market: OnChainMarket): Future[?string] {.async.} =
+method loadConfig*(
```
marcinczenko, line 81: Just a thought...
Since this seems to be a new method, did you consider returning a `Result` instead of raising `MarketError`?

It also looks like this method does a lot the same that private `config` method is doing. What if we have just one:

```nim
proc loadConfig*(
  market: OnChainMarket
): Future[?!MarketplaceConfig] {.async: (raises: [CancelledError]).} =
  try:
    without config =? market.configuration:
      let fetchedConfig = await market.contract.configuration()

      market.configuration = some fetchedConfig
    return success(config)
  except AsyncLockError as err:
    return failure("Failed to fetch the config from the Marketplace contract: " & err.msg)
  except EthersError as error:
    return failure(error.msgDetail)
```

Then maybe we do not even need this private `config` that you have defined above.

Or if we really want to have a convenience proc, e.g. ignoring errors and just returning an optional:

```nim
proc config(
    market: OnChainMarket
): Future[?MarketplaceConfig] {.async: (raises: [CancelledError]).} =

  if config =? await market.loadConfig2():
    return some(config)
  
  # ignore errors
  return MarketplaceConfig.none
```

marcinczenko (approved):
Hey!!! I left some suggestions you may like to check. They do not correct any errors (rather my thoughts about simplifying a couple of things), and I still need to take a look quickly at the tests tomorrow, but since @emizzle already gave it a spin, I guess I can approve it already if you guys are in a hurry.

Still, I would appreciate if you comment on my comments, so that at least I know where I misunderstood something.

codex/codex.nim
```diff
@@ -134,6 +134,12 @@ proc bootstrapInteractions(s: CodexServer): Future[void] {.async.} =
       if config.simulateProofFailures > 0:
         warn "Proof failure simulation is not enabled for this build! Configuration ignored"
 
+    try:
+      await market.loadConfig()
+    except MarketError as e:
+      fatal "Cannot load market configuration", error = e.msg
+      quit QuitFailure
+
```
marcinczenko, line 142: If it does not make sense to proceed without a configuration and configuration is part of `OnChainMarket`, would it be nicer to move loading the config to `OnChainMarket.new` (and perhaps change its return type to return `Result`)?

codex/contracts/market.nim
```diff
@@ -128,19 +160,22 @@ method requestStorage(market: OnChainMarket, request: StorageRequest) {.async.}
...
-      return some request
-    except Marketplace_UnknownRequest:
-      return none StorageRequest
+    let request = await market.contract.getRequest(id)
+    market.requestCache[key] = request
+    return some request
+  except Marketplace_UnknownRequest, KeyError:
+    warn "Cannot retrieve the request", error = getCurrentExceptionMsg()
+    return none StorageRequest
+  except EthersError, AsyncLockError:
```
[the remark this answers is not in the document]
2-towns: Yes you are right `Result` is a better strategy here.

codex/contracts/market.nim
```diff
@@ -55,41 +55,71 @@ template convertEthersError(body) =
...
+    await market.loadConfig()
+
+    without config =? market.configuration:
+      raiseMarketError("Failed to access to config from the Marketplace contract")
+
+    return config
 
   return resolvedConfig
 
+proc configSync(market: OnChainMarket): ?MarketplaceConfig =
```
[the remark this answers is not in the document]
marcinczenko: But it just returns `market.configuration`, right? Why can't we use `market.configuration` directly in `slotCollateral`?

```nim
method slotCollateral*(
    market: OnChainMarket, collateralPerSlot: UInt256, slotState: SlotState
): ?UInt256 {.raises: [].} =
  if slotState == SlotState.Repair:
    without repairRewardPercentage =?
      market.configuration .? collateral .? repairRewardPercentage:
      return UInt256.none

    return (
      collateralPerSlot - (collateralPerSlot * repairRewardPercentage.u256).div(
        100.u256
      )
    ).some

  return collateralPerSlot.some
```

Perhaps I just miss something obvious...

codex/contracts/market.nim
2-towns: I applied the changes.

codex/contracts/market.nim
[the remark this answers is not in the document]
2-towns: I would address it in another PR

codex/codex.nim
2-towns: If you do that, you will need to make the `OnChainMarket.new` async or you will need to return a AsyncOnChainMarketm because the `loadConfig` is async.

codex/contracts/market.nim
2-towns: Thanks for the suggestion, I switched to `Result`. 

But I think we want to keep the both methods, as we discussed in our previous meeting, because `config` returns the configuration object while `loadConfig` is not returning anything. For what I remember, we didn't want to use the `config` method when we wanted to load only the configuration because of the confusion.

codex/contracts/market.nim
```diff
@@ -507,3 +549,30 @@ method queryPastStorageRequestedEvents*(
...
+  let slotState = await market.slotState(slotid)
+
+  without request =? await market.getRequest(requestId):
+    return UInt256.none
+
+  return market.slotCollateral(request.ask.collateralPerSlot, slotState)
+
+method slotCollateral*(
+    market: OnChainMarket, collateralPerSlot: UInt256, slotState: SlotState
+): ?UInt256 {.raises: [].} =
```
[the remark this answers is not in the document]
2-towns: Yes you are right. I switched to `Result`.

PR #1087, opened 2025-01-23
feat: multithreading support for erasure coding

munna0908:
This PR introduces the following changes:

- Utilizes `TaskPool` to spawn separate threads for the erasure encoding and decoding.
- Adds a new `num-threads` flag, enabling users to specify the number of CPU threads available for TaskPool workers.

**Docs [PR](https://github.com/codex-storage/codex-docs/pull/58)**

### DIFF
--- a/codex/codex.nim
+++ b/codex/codex.nim
@@ -11,8 +11,10 @@ import std/sequtils
 import std/strutils
 import std/os
 import std/tables
+import std/cpuinfo
 
 import pkg/chronos
+import pkg/taskpools
 import pkg/presto
 import pkg/libp2p
 import pkg/confutils
@@ -194,7 +196,21 @@ proc new*(
     .withTcpTransport({ServerFlags.ReuseAddr})
     .build()
 
-  var cache: CacheStore = nil
+  var
+    cache: CacheStore = nil
+    taskpool: Taskpool
+  try:
+    if config.numThreads < 0:
+      fatal "The number of threads --numThreads cannot be negative."
+      quit 1
+    elif config.numThreads == 0:
+      taskpool = Taskpool.new(numThreads = min(countProcessors(), 16))
+    else:
+      taskpool = Taskpool.new(numThreads = config.numThreads)
+
+      info "Threadpool started", numThreads = taskpool.numThreads
+  except Exception:
+    raise newException(Defect, "Failure in taskpool initialization.")
 
   if config.cacheSize > 0'nb:
     cache = CacheStore.new(cacheSize = config.cacheSize)
@@ -286,6 +302,7 @@ proc new*(
       engine = engine,
       discovery = discovery,
       prover = prover,
+      taskPool = taskpool,
     )
 
     restServer = RestServerRef
--- a/codex/codex.nim
+++ b/codex/codex.nim
@@ -199,6 +199,7 @@ proc new*(
   var
     cache: CacheStore = nil
     taskpool: Taskpool
+
   try:
     if config.numThreads < 0:
       fatal "The number of threads --numThreads cannot be negative."
@@ -207,8 +208,7 @@ proc new*(
       taskpool = Taskpool.new(numThreads = min(countProcessors(), 16))
     else:
       taskpool = Taskpool.new(numThreads = config.numThreads)
-
-      info "Threadpool started", numThreads = taskpool.numThreads
+    info "Threadpool started", numThreads = taskpool.numThreads
   except Exception:
     raise newException(Defect, "Failure in taskpool initialization.")
 
--- a/codex/codex.nim
+++ b/codex/codex.nim
@@ -201,18 +201,13 @@ proc new*(
     taskpool: Taskpool
 
   try:
-    if config.numThreads == 0:
+    if config.numThreads == ThreadCount(0):
       taskpool = Taskpool.new(numThreads = min(countProcessors(), 16))
-    elif config.numThreads < 2:
-      raise newException(
-        Defect, "The number of threads --numThreads cannot be less than two."
-      )
     else:
-      taskpool = Taskpool.new(numThreads = config.numThreads)
+      taskpool = Taskpool.new(numThreads = int(config.numThreads))
     info "Threadpool started", numThreads = taskpool.numThreads
-  except CatchableError as parent:
-    raise
-      newException(Defect, "Failure in taskpool initialization:" & parent.msg, parent)
+  except CatchableError as exc:
+    raiseAssert("Failure in taskpool initialization:" & exc.msg)
 
   if config.cacheSize > 0'nb:
     cache = CacheStore.new(cacheSize = config.cacheSize)
--- a/codex/codex.nim
+++ b/codex/codex.nim
@@ -201,16 +201,18 @@ proc new*(
     taskpool: Taskpool
 
   try:
-    if config.numThreads < 0:
-      fatal "The number of threads --numThreads cannot be negative."
-      quit 1
-    elif config.numThreads == 0:
+    if config.numThreads == 0:
       taskpool = Taskpool.new(numThreads = min(countProcessors(), 16))
+    elif config.numThreads < 2:
+      raise newException(
+        Defect, "The number of threads --numThreads cannot be less than two."
+      )
     else:
       taskpool = Taskpool.new(numThreads = config.numThreads)
     info "Threadpool started", numThreads = taskpool.numThreads
-  except Exception:
-    raise newException(Defect, "Failure in taskpool initialization.")
+  except CatchableError as parent:
+    raise
+      newException(Defect, "Failure in taskpool initialization:" & parent.msg, parent)
 
   if config.cacheSize > 0'nb:
     cache = CacheStore.new(cacheSize = config.cacheSize)
--- a/codex/conf.nim
+++ b/codex/conf.nim
@@ -53,6 +53,10 @@ export
   DefaultQuotaBytes, DefaultBlockTtl, DefaultBlockMaintenanceInterval,
   DefaultNumberOfBlocksToMaintainPerInterval
 
+type ThreadCount* = distinct Natural
+
+proc `==`*(a, b: ThreadCount): bool {.borrow.}
+
 proc defaultDataDir*(): string =
   let dataDir =
     when defined(windows):
@@ -71,6 +75,7 @@ const
 
   DefaultDataDir* = defaultDataDir()
   DefaultCircuitDir* = defaultDataDir() / "circuits"
+  DefaultThreadCount* = ThreadCount(0)
 
 type
   StartUpCmd* {.pure.} = enum
@@ -187,9 +192,9 @@ type
     numThreads* {.
       desc:
         "Number of worker threads (\"0\" = use as many threads as there are CPU cores available)",
-      defaultValue: 0,
+      defaultValue: DefaultThreadCount,
       name: "num-threads"
-    .}: int
+    .}: ThreadCount
 
     agentString* {.
       defaultValue: "Codex",
@@ -489,6 +494,13 @@ proc parseCmdArg*(
     quit QuitFailure
   ma
 
+proc parseCmdArg*(T: type ThreadCount, input: string): T {.upraises: [ValueError].} =
+  let count = parseInt(input)
+  if count != 0 and count < 2:
+    warn "Invalid number of threads", input = input
+    quit QuitFailure
+  ThreadCount(count)
+
 proc parseCmdArg*(T: type SignedPeerRecord, uri: string): T =
   var res: SignedPeerRecord
   try:
@@ -587,6 +599,15 @@ proc readValue*(
   val = NBytes(value)
 
 proc readValue*(
+    r: var TomlReader, val: var ThreadCount
+) {.upraises: [SerializationError, IOError].} =
+  var str = r.readValue(string)
+  try:
+    val = parseCmdArg(ThreadCount, str)
+  except CatchableError as err:
+    raise newException(SerializationError, err.msg)
+
+proc readValue*(
     r: var TomlReader, val: var Duration
 ) {.upraises: [SerializationError, IOError].} =
   var str = r.readValue(string)
@@ -614,6 +635,9 @@ proc completeCmdArg*(T: type NBytes, val: string): seq[string] =
   discard
 
 proc completeCmdArg*(T: type Duration, val: string): seq[string] =
+  discard
+
+proc completeCmdArg*(T: type ThreadCount, val: string): seq[string] =
   discard
 
 # silly chronicles, colors is a compile-time property
--- a/codex/conf.nim
+++ b/codex/conf.nim
@@ -184,6 +184,13 @@ type
       name: "max-peers"
     .}: int
 
+    numThreads* {.
+      desc:
+        "Number of worker threads (\"0\" = use as many threads as there are CPU cores available)",
+      defaultValue: 0,
+      name: "num-threads"
+    .}: int
+
     agentString* {.
       defaultValue: "Codex",
       desc: "Node agent string which is used as identifier in network",
--- a/codex/erasure/erasure.nim
+++ b/codex/erasure/erasure.nim
@@ -613,7 +613,6 @@ proc new*(
 ): Erasure =
   ## Create a new Erasure instance for encoding and decoding manifests
   ##
-
   Erasure(
     store: store,
     encoderProvider: encoderProvider,
--- a/codex/erasure/erasure.nim
+++ b/codex/erasure/erasure.nim
@@ -91,14 +91,14 @@ type
     minSize*: NBytes
 
   EncodeTask = object
-    result: Atomic[bool]
+    success: Atomic[bool]
     encoder: EncoderBackend
     blocks: ref seq[seq[byte]]
     parity: ptr seq[seq[byte]]
     signal: ThreadSignalPtr
 
   DecodeTask = object
-    result: Atomic[bool]
+    success: Atomic[bool]
     decoder: DecoderBackend
     data: ref seq[seq[byte]]
     parity: ref seq[seq[byte]]
@@ -291,22 +291,22 @@ proc leopardEncodeTask(tp: Taskpool, task: ptr EncodeTask) =
   # Task suitable for running in taskpools - look, no GC!
   if (let res = task[].encoder.encode(task[].blocks[], task[].parity[]); res.isErr):
     warn "Error from leopard encoder backend!", error = $res.error
-    task[].result.store(false)
+    task[].success.store(false)
   else:
-    task[].result.store(true)
+    task[].success.store(true)
 
   discard task[].signal.fireSync()
 
-proc asyncEncode(
+proc encodeAsync(
     self: Erasure,
     blockSize, ecK, ecM: int,
     data: ref seq[seq[byte]],
     parity: ptr seq[seq[byte]],
-): Future[Result[void, string]] {.async.} =
+): Future[?!void] {.async: (raises: []).} =
   ## Create an encoder backend instance
   let encoder = self.encoderProvider(blockSize, ecK, ecM)
   without threadPtr =? ThreadSignalPtr.new():
-    return err("Unable to create thread signal")
+    return failure("Unable to create thread signal")
 
   ## Create an ecode task with block data 
   var task =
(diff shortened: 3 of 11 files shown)


codex/codex.nim
```diff
@@ -194,7 +196,21 @@ proc new*(
...
+    if config.numThreads < 0:
+      fatal "The number of threads --numThreads cannot be negative."
+      quit 1
+    elif config.numThreads == 0:
+      taskpool = Taskpool.new(numThreads = min(countProcessors(), 16))
+    else:
+      taskpool = Taskpool.new(numThreads = config.numThreads)
+    info "Threadpool started", numThreads = taskpool.numThreads
+  except Exception:
+    raise newException(Defect, "Failure in taskpool initialization.")
```
gmega, line 213: :yellow_circle: Not sure we want to be casting everything as Defect and bubbling it up this way? I think we'd use a result type for this instead?

codex/conf.nim
```diff
@@ -184,6 +184,13 @@ type
       name: "max-peers"
     .}: int
 
+    numThreads* {.
+      desc:
+        "Number of worker threads (\"0\" = use as many threads as there are CPU cores available)",
```
gmega, line 189:
```suggestion
        "Number of worker threads for running erasure coding jobs (\"0\" = use as many threads as there are CPU cores available)",
```

I prefer to be specific about what threads are those.

codex/erasure/erasure.nim
```diff
@@ -87,6 +90,21 @@ type
     # provided.
     minSize*: NBytes
 
+  EncodeTask = object
+    result: Atomic[bool]
+    encoder: EncoderBackend
+    blocks: ref seq[seq[byte]]
+    parity: ptr seq[seq[byte]]
```
gmega, line 97: Curious: why do we use `ref` and not `ptr` as well for blocks?

codex/erasure/erasure.nim
```diff
@@ -87,6 +90,21 @@ type
...
+    signal: ThreadSignalPtr
+
+  DecodeTask = object
+    result: Atomic[bool]
+    decoder: DecoderBackend
+    data: ref seq[seq[byte]]
+    parity: ref seq[seq[byte]]
+    recovered: ptr seq[seq[byte]]
```
gmega, line 105: Same question here.

gmega (changes requested):
I think this is good, well done. I see one minor issue with the exception handling but otherwise looks great. We should wait for @dryajov's review on this one anyhow though.

codex/conf.nim
dryajov: This is not the only place where threads are going to be used, the taskpool is shared across several different places, so the old wording is totally accurate.

codex/codex.nim
emizzle: @munna0908 `Exception` is a parent type for `Defect` and `CatchableError`, the latter of which all other "catchable" error types are derived from. Defect is not catchable, and therefore we should not be using `except Expection` where we can avoid it (despite its usage in other places in the codebase).

In this particular case, using `except CatchableError` and converting to a `Defect` is a valid pattern, and it also means that if a Defect was raised in the try block, it would not be caught in the except, and would be bubbled up the stack.

See https://status-im.github.io/nim-style-guide/errors.exceptions.html for more info on exception handling best practices (TL;DR using Result is recommended)

@gmega I agree that we should probably return a `Result[void]` type here, read the Result in the layer above, and then quit on error. However, the layer above currently quits on exception, and there are a fair few instances of raised Defects in the new proc, so converting everything to Result would require a decent amount of cascading changes. For the scope of this PR, I think that Defect is suitable since there is no recovery opportunity here. 

So my suggestion would be something like:
```nim
try:
  # start taskpool
except CatchableError as parent:
  raise newException(Defect, "Failure in taskpool initialisation: " & parent.msg, parent)
```

codex/erasure/erasure.nim
```diff
@@ -269,6 +287,50 @@ proc init*(
...
+    return err("Unable to create thread signal")
+
+  ## Create an ecode task with block data 
+  var task =
+    EncodeTask(encoder: encoder, blocks: data, parity: parity, signal: threadPtr)
+
+  let t = addr task
+  doAssert self.taskPool.numThreads > 1,
+    "Must have at least one separate thread or signal will never be fired"
+  self.taskPool.spawn leopardEncodeTask(self.taskPool, t)
```
emizzle, line 318: If I'm reading this correctly, if the user configures `--num-threads=1`, then the node will crash with a defect. Is there a non-threading equivalent? Or should our default config be `--num-threads=2` (with validation at startup that threads must be >=2)?

codex/erasure/erasure.nim
```diff
@@ -269,6 +287,50 @@ proc init*(
...
+    task[].result.store(true)
+
+  discard task[].signal.fireSync()
+
+proc asyncEncode(
+    self: Erasure,
+    blockSize, ecK, ecM: int,
+    data: ref seq[seq[byte]],
+    parity: ptr seq[seq[byte]],
+): Future[Result[void, string]] {.async.} =
```
emizzle, line 305: 🟡 This is completely correct, just highlighting that another option would be to continue the pattern in this module of returning `?!void`, which is an alias for`Result[void, CatchableError]`, from the `questionable` lib. Then, you can return `success()` or `failure("something")`. Again, just a suggestion.

codex/erasure/erasure.nim
```diff
@@ -87,6 +90,21 @@ type
     # provided.
     minSize*: NBytes
 
+  EncodeTask = object
+    result: Atomic[bool]
```
emizzle, line 94: `result` is generally a reserved word. Not sure if we should use something else? @dryajov?

codex/erasure/erasure.nim
```diff
@@ -381,6 +442,57 @@ proc encode*(
...
+    task[].result.store(true)
+
+  discard task[].signal.fireSync()
+
+proc decodeAsync(
+    self: Erasure,
+    blockSize, ecK, ecM: int,
+    data, parity: ref seq[seq[byte]],
+    recovered: ptr seq[seq[byte]],
+): Future[Result[void, string]] {.async.} =
```
emizzle, line 463: Probably a good idea to tell the compiler we don't want any exceptions leaking out of this routine, so we don't have any issues with crashing thread tasks
```suggestion
): Future[Result[void, string]] {.async: (raises:[]).} =
```

emizzle: Same for `encodeAsync`

emizzle (commented):
Nice job Rahul! I left mostly suggestions, just one question around requiring threading to be turned on and the configuration requirements/validation.

codex/erasure/erasure.nim
munna0908: @emizzle Good point, I will change this to success

codex/erasure/erasure.nim
munna0908: @gmega `blocks` uses ref since it's read-only input data for the encoding. `parity` uses ptr since the C encoding function needs direct memory access to write the parity data

codex/erasure/erasure.nim
munna0908: @gmega Same here, Decode function updates the recovered sequence so we need a ptr

codex/conf.nim
gmega: I don't like this for two reasons:

1. "worker threads" means nothing. Someone familiar with the project might not be able to tell what those "worker threads" do;
2. the concurrency limits for different things might be completely different. I don't know, for instance, that you want to use the same number of worker threads for erasure coding vs. running the prover - if the prover has all-core internal parallelism, for instance, you might not want to run more than one proof at a time.

So either way I'd improve the wording and be specific about what those threads are doing. As to sharing the taskpool, I'd only do it if the concurrency requirements for all tasks sharing that taskpool are similar, otherwise you'll need to to extra bookkeeping to enforce concurrency limits.

codex/erasure/erasure.nim
gmega: OK fair enough. I guess we could've used ptr for everything but would've been pointless.

codex/erasure/erasure.nim
gmega: Yeah probably a good idea. Not sure why this assertion is here but it's also present in Nimbus (I wonder if the taskpool may end up losing threads somehow) so I'd probably keep it even with your suggestion above.

munna0908: @gmega , ThreadPool is common resource, it can be used by prover, erasure or any other component. For this reason we need to check if threads are available for this particular task.

munna0908: @emizzle , Our default value is `0`, When it is set to zero, we calculate the number of threads based on the core's available. I will add a check while creating threadpool to validate `config.numThreads`

codex/erasure/erasure.nim
munna0908: @emizzle Commit-[link](https://github.com/codex-storage/nim-codex/pull/1087/commits/9d2e0d386abf0ee2f5907caa81a751891dbeab8f)

codex/erasure/erasure.nim
munna0908: Commit-[link](https://github.com/codex-storage/nim-codex/pull/1087/commits/9d2e0d386abf0ee2f5907caa81a751891dbeab8f)

codex/erasure/erasure.nim
munna0908: Commit-[link](https://github.com/codex-storage/nim-codex/pull/1087/commits/9d2e0d386abf0ee2f5907caa81a751891dbeab8f)

codex/erasure/erasure.nim
dryajov: With one thread, the taskpool could potentially deadlock and not make any progress (I remember that from a conversation with @mratsim), but this check should be done sooner, when the threadpool is instantiated.

codex/erasure/erasure.nim
dryajov: It should be fine inside a struct/object, I don't think it will be a problem, but might be prudent to use a different name, `res`/`ok`?

codex/codex.nim
munna0908: @emizzle Commit-[link](https://github.com/codex-storage/nim-codex/pull/1087/commits/3d8d13356ef090bfd680c31fc9ce96d0baac5bf5)

codex/conf.nim
dryajov: > "worker threads" means nothing.

On the contrary, this is pretty common nomenclature.

> the concurrency limits for different things might be completely different. 

This isn't the case and balancing this properly is beyond the scope of what we need and this point or in the future. We aren't planning on having _different_ thread pools for different subsystems, simply because the complexity that this brings and the potential gains are not justified. 

> if the prover has all-core internal parallelism, for instance, you might not want to run more than one background instance of the prover at a time.

If the prover has it's own parallelism mechanism (which it does), it means that we need to contend with this resources regardless of who is using the cores, the erasure coding, or something else - the cores/cpus are shared by the entire process, so splitting the threadpool makes absolutely no sense.

gmega: > On the contrary, this is pretty common nomenclature.

Yes, it is common nomenclature for "background threads that do work", but in vacuum this means nothing. These worker threads could be doing anything, and you can't make an educated guess about how to size this properly without knowing what that anything is.

> If the prover has it's own parallelism mechanism (which it does), it means that we need to contend with this resources regardless of who is using the cores, the erasure coding, or something else - the cores/cpus are shared by the entire process, so splitting the threadpool makes absolutely no sense.

I understand it doesn't solve the problem of resource contention completely, but allowing such large number of provers to run concurrently might make things much worse, including causing Codex to crash/thrash. If you need 8-16GB of RAM per prover instance, for example, you'll need 4x as much in a 4-core machine.

dryajov: > I understand it doesn't solve the problem of resource contention completely, but allowing such large number of provers to run concurrently might make things much worse, including causing Codex to crash/thrash. If you need 8-16GB of RAM per prover instance, for example, you'll need 4x as much in a 4-core machine (and most machines these days have a lot more than that...).

That doesn't affect the number of dedicated cores/threads, this merely dictates how the threadpool is used by the different subsystems. For example, the erasure coding subsystem, is only attempting to offload work to a single thread at a time right now, without attempting to parallelize anything, later this might change. The same goes for the prover and anything else we use, how many instances of "something" are running at the same time is not directly related to the number of worker threads in the pool, rather to how the pool is used by the subsystem. So, if you can only run one or two instances of the prover, this should be properly synchronized at the invocation point.

> Yes, it is common nomenclature for "background threads that do work", but in a vacuum this means nothing.

Again, in the context of using a thread pool, this is pretty self explanatory? Usually, you have a single threadpool per process, very rarely do you would want a different parallelization mechanisms across the codebase.

codex/codex.nim
```diff
@@ -194,7 +196,23 @@ proc new*(
...
-  var cache: CacheStore = nil
+  var
+    cache: CacheStore = nil
+    taskpool: Taskpool
+
+  try:
+    if config.numThreads == 0:
+      taskpool = Taskpool.new(numThreads = min(countProcessors(), 16))
+    elif config.numThreads < 2:
+      raise newException(
```
dryajov, line 207: I would use a `raiseAssert` here instead.

codex/codex.nim
```diff
@@ -194,7 +196,23 @@ proc new*(
...
+        Defect, "The number of threads --numThreads cannot be less than two."
+      )
+    else:
+      taskpool = Taskpool.new(numThreads = config.numThreads)
+    info "Threadpool started", numThreads = taskpool.numThreads
+  except CatchableError as parent:
+    raise
+      newException(Defect, "Failure in taskpool initialization:" & parent.msg, parent)
```
dryajov, line 215: Probably do a `raiseAssert` as well.

emizzle (approved):
Thanks for making the changes, Rahul. Apart from the minor suggestions (`raiseAssert` instead of `raise newException(Defect...`), this LGTM. So, I'll approve now.

codex/erasure/erasure.nim
```diff
@@ -269,6 +287,52 @@ proc init*(
...
+    "Must have at least one separate thread or signal will never be fired"
+  self.taskPool.spawn leopardEncodeTask(self.taskPool, t)
+
+  try:
+    await threadPtr.wait()
+  except AsyncError as exc:
+    warn "Async thread error", err = exc.msg
+    return failure(exc.msg)
+  except CancelledError:
+    return failure("Thread cancelled")
```
emizzle, line 326: Here, do we want to return a failure? I thought about this a bit further... this proc executes in the master thread, correct? If so, maybe we should propagate `CancelledErrors`. So the raises annotation would be:

```nim
{.async: (raises: [CacncelledError]).}
```
and the except block would re-raise:
```nim
except CancelledError as e:
    raise e
```

This brings up another question -- should the `TaskPool` be gracefully terminated in `stop`? Or is there no point to wait for worker threads to join master? IOW, would stopping the master thread kill the worker at the OS-level and possibly produce an out-of-band error?

codex/erasure/erasure.nim
```diff
@@ -381,6 +446,61 @@ proc encode*(
...
+    "Must have at least one separate thread or signal will never be fired"
+  self.taskPool.spawn leopardDecodeTask(self.taskPool, t)
+
+  try:
+    await threadPtr.wait()
+  except AsyncError as exc:
+    warn "Async thread error", err = exc.msg
+    return failure(exc.msg)
+  except CancelledError:
+    return failure("Thread cancelled")
```
emizzle, line 494: Same comment as above.

emizzle (commented):
I just noticed the CancelledError changes, which brought to mind that maybe we need to consider gracefully terminate the `TaskPool` in the `stop` routine.

Sorry for the premature approval before!

dryajov (approved):

dryajov (changes requested):
I'm still reviewing this changes, so lets not merge this just yet.

dryajov (changes requested):

codex/erasure/erasure.nim
dryajov: So, this is quite unsafe (as we discussed), for both `ptr` (and more so) for `ref`.

The correct way is to pass around `ptr UncheckedArray[ptr UncheckedArray]` which point to the underlying seq heap location, which is retrieved with `addr buffers[0][0]` to get to the individual buffer, the first element in the seq points to the address of the contiguous heap memory.

The reason this is unsafe, is because with `refc`, the GC is not threadsafe, and any refcount increment from a foreign thread, would trip the GC. It's actually surprising that this works at all, but I guess we got lucky.

The problem however, is that converting `UncheckedArray` back to a seq is not safe either in this case. Essentially we end up with two GCs contending for the same memory region. Say for example, that we have a seq allocated in the main thread, and we pass a pointer to it to the child/worker thread which then (somehow) maps it back into it's own seq. What happens when the worker thread is finished and the second seq is collected? Most likely it will try to free that memory that is also owned by the main thread.

So, what to do? I think, there are two ways forward.

1. Copy the contents of `ptr UncheckedArray[ptr UncheckedArray[byte]]` onto a `seq[seq[byte]]`, on the threads local heap. This incurres 4 copies, one to copy the ptr onto the thread and then another by the nim leopard wrapper, which also has internal work buffers, and then back.
2. Change the interface of nim-leopard to work with `ptr`, so that we avoid the excessive copies.

I like option 2 because it's cheaper, plus we control the nim wrapper and can modify it as we see fit.

Option 1 is simpler, but not by much.

If we go with option two, I would make the following changes to `vendor/nim-leopard/leopard/leopard.nim`:

```nim
# Current
func encode*(
  self: var LeoEncoder,
  data,
  parity: var openArray[seq[byte]]): Result[void, cstring] =
  ...

func decode*(
  self: var LeoDecoder,
  data,
  parity,
  recovered: var openArray[seq[byte]]): Result[void, cstring] =
  ...

# New

func encode*(
  self: var LeoEncoder,
  data,
  parity: ptr UncheckedArray[ptr UncheckedArray[byte]]): Result[void, cstring] =


func decode*(
  self: var LeoDecoder,
  data,
  parity,
  recovered: ptr UncheckedArray[ptr UncheckedArray[byte]]): Result[void, cstring] =

```

I would leave the old `encode/decode` as overloads that convert the `openArray[seq[byte]]` to it's pointer version for backwards compatibility reasons.

codex/codex.nim
dryajov: Actually, I think this is better checked in the `conf.nim` so that the invalid value doesn't get passed the cmd validation.

dryajov (approved):

dryajov (changes requested):
Great job for a first pass @munna0908! 

Let me summarize what needs to be changed:

- Rework how buffers are passed arround from `ptr seq[seq[]]` to `ptr UncheckedArray[ptr UncheckedArray[byte]]`
- Dispose of the thread signal
- Handle cancelations
- Add tests to explicitely test the multithreaded functionality in `testerasure.nim`

codex/erasure/erasure.nim
emizzle: Hmm, couldn't that create a deadlock situation where the task is cancelled with `cancelAndWait` and but cancellation has wait on the thread to finish its operation before cancelling? Is there a way to forcefully kill the task and discard any errors or results?

munna0908:
@dryajov I have accommodated all suggested changes including the test cases

codex/erasure/erasure.nim
```diff
@@ -269,6 +292,80 @@ proc init*(
...
+  try:
+    await threadFut.join()
+  except CatchableError as exc:
+    try:
+      await threadFut
+    except AsyncError as asyncExc:
+      return failure(asyncExc.msg)
+    finally:
+      if exc of CancelledError:
+        raise (ref CancelledError) exc
```
dryajov, line 358: no need to cast it here, just re-raise

codex/erasure/erasure.nim
```diff
@@ -269,6 +292,80 @@ proc init*(
...
+      else:
+        return failure(exc.msg)
+  finally:
+    threadPtr.close().expect("closing once works")
```
dryajov, line 362: maybe, this can be in a `defer` block instead?

codex/erasure/erasure.nim
```diff
@@ -269,14 +291,88 @@ proc init*(
...
+    "Must have at least one separate thread or signal will never be fired"
+  self.taskPool.spawn leopardEncodeTask(self.taskPool, t)
+  let threadFut = threadPtr.wait()
+
+  try:
+    await threadFut.join()
+  except CatchableError as exc:
+    try:
+      await threadFut
+    except AsyncError as asyncExc:
```
dryajov, line 356: Why AsyncError here and not CatchableError?

dryajov (approved):
LGTM! Thanks!

codex/erasure/erasure.nim
munna0908: @dryajov AsyncError would be more appropriate as threadPtr.wait() raises an AsyncError, [ref](https://github.com/status-im/nim-chronos/blob/0646c444fce7c7ed08ef6f2c9a7abfd172ffe655/chronos/threadsync.nim#L367)

PR #1090, opened 2025-01-28
chore(marketplace): add a cache for storage requests

### LINKED ISSUE #1089: Use a cache for the storage requests
This task was originally part of the discussion in https://github.com/codex-storage/nim-codex/pull/1083 with @AuHau.
This will reduce the RPC calls when trying to get the storage request data.

2-towns:
Fix #1089

### DIFF
--- a/codex/conf.nim
+++ b/codex/conf.nim
@@ -354,7 +354,8 @@ type
           "Reduces fetching of StorageRequest data from the contract.",
         defaultValue: DefaultRequestCacheSize,
         defaultValueDesc: $DefaultRequestCacheSize,
-        name: "request-cache-size"
+        name: "request-cache-size",
+        hidden
       .}: uint16
 
       case persistenceCmd* {.defaultValue: noCmd, command.}: PersistenceCmd
--- a/codex/conf.nim
+++ b/codex/conf.nim
@@ -44,6 +44,7 @@ import ./utils
 import ./nat
 import ./utils/natutils
 
+from ./contracts/config import DefaultRequestCacheSize
 from ./validationconfig import MaxSlots, ValidationGroups
 
 export units, net, codextypes, logutils, completeCmdArg, parseCmdArg, NatConfig
@@ -51,7 +52,7 @@ export ValidationGroups, MaxSlots
 
 export
   DefaultQuotaBytes, DefaultBlockTtl, DefaultBlockMaintenanceInterval,
-  DefaultNumberOfBlocksToMaintainPerInterval
+  DefaultNumberOfBlocksToMaintainPerInterval, DefaultRequestCacheSize
 
 proc defaultDataDir*(): string =
   let dataDir =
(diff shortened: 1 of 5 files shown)


codex/contracts/market.nim
```diff
@@ -20,19 +21,30 @@ type
...
   EventSubscription = ethers.Subscription
   OnChainMarketSubscription = ref object of MarketSubscription
     eventSubscription: EventSubscription
 
 func new*(
-    _: type OnChainMarket, contract: Marketplace, rewardRecipient = Address.none
+    _: type OnChainMarket,
+    contract: Marketplace,
+    rewardRecipient = Address.none,
+    requestCacheSize: uint16 = 0,
```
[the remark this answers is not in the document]
emizzle: Another option is you could set a default value in the constructor and in the config similar to how it's being done with `DefaultBlockTtl`. This would remove the need to pass in a value to the constructor in the tests.

codex/conf.nim
```diff
@@ -347,6 +347,15 @@ type
         name: "reward-recipient"
       .}: Option[EthAddress]
 
+      marketplaceRequestCacheSize* {.
+        desc:
+          "The size of the request cache - " &
+          "reduce the contract calls to get the request data.",
```
emizzle, line 353:
```suggestion
          "Maximum number of StorageRequests kept in memory. " &
          "Reduces fetching of StorageRequest data from the contract.",
```

emizzle (commented):
Looking good 👍  Just a few small suggestions. Is this PR ready to be out of draft?

2-towns:
> Looking good 👍 Just a few small suggestions. Is this PR ready to be out of draft?

Okay I will check your suggestions. 

It is still in draft because you mentioned in #1083 that it would be good to include calls to `marketplace.config` but I think Adam already addressed it here: https://github.com/codex-storage/nim-codex/blob/58a962add8a5a46b774889320f1490ffc056eb08/codex/contracts/market.nim#L46-L52. 
Am I missing something ?

codex/contracts/market.nim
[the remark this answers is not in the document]
2-towns: I like the `DefaultBlockTtl` solution. I applied it.

codex/conf.nim
```diff
@@ -347,6 +348,15 @@ type
...
       .}: Option[EthAddress]
 
+      marketplaceRequestCacheSize* {.
+        desc:
+          "Maximum number of StorageRequests kept in memory." &
+          "Reduces fetching of StorageRequest data from the contract.",
+        defaultValue: DefaultRequestCacheSize,
+        defaultValueDesc: $DefaultRequestCacheSize,
+        name: "request-cache-size"
+      .}: uint16
```
AuHau, line 358: I am not sure if this really needs to be CLI flag 😅 IMHO this only will confuse users who do not need this sort of granular control over the node. I would either remove this, or make this flag `hidden`, which will make it not show on CLI help page and maybe add note to our documentation. Which btw. should be updated anyhow as you are adding new CLI flag - the configuration section.  aa

2-towns: I think we need to keep it because we would be in a situation where we are trying to debug something and we may want to change the value without building a new version. But I agree with you that it would confuse the users, so I think the best solution is to make it hidden.
For the documentation, yes I will update it after the PR is merged.

AuHau (approved):
LGTM, just fix the CI ;-)

PR #1092, opened 2025-01-30
Minor improvements to download API

gmega:
I had to debug a download failure and ended up adding slightly better error feedback. Also noted some dead code, which I have removed.

### DIFF
--- a/codex/rest/api.nim
+++ b/codex/rest/api.nim
@@ -234,9 +234,6 @@ proc initDataApi(node: CodexNodeRef, repoStore: RepoStore, router: var RestRoute
     finally:
       await reader.closeWait()
 
-    trace "Something went wrong error"
-    return RestApiResponse.error(Http500)
-
   router.api(MethodGet, "/api/codex/v1/data") do() -> RestApiResponse:
     let json = await formatManifestBlocks(node)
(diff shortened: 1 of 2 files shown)


codex/rest/api.nim
```diff
@@ -190,10 +190,13 @@ proc initDataApi(node: CodexNodeRef, repoStore: RepoStore, router: var RestRoute
...
 
     if mimetype.get() != "":
+      let mimetypeVal = mimetype.get()
       var m = newMimetypes()
-      let extension = m.getExt(mimetype.get(), "")
+      let extension = m.getExt(mimetypeVal, "")
       if extension == "":
-        return RestApiResponse.error(Http422, "The MIME type is not valid.")
+        return RestApiResponse.error(
+          Http422, "The MIME type '" & mimetypeVal & "' is not valid."
```
2-towns, line 198: Yes it is better. You just need to update the test case "upload fails if content type is invalid" to include the mimetype value.

2-towns (approved):
LGTM, just need to fix the integration test

codex/rest/api.nim
gmega: Yep, just noticed this is failing now. Thanks

2-towns (approved):
LGTM

Issue #974, opened 2024-10-28
BlockExchange Performance and reliability


benbierens:
Tests show that performance of the block exchange varies significantly in controlled networks.
In some cases, downloads fail when datasets are available both in the network, and in peers to which existing connections exist. Reliability must be improved.

Work done:

* Reliablity test write-up: https://hackmd.io/rwPtPJ7KTw6cGjhN0zNYig
* Block retry: https://github.com/codex-storage/nim-codex/pull/1119
* Block protocol write up: https://hackmd.io/ZAblhSiPTRahEz9w0mulcA
  * simple simulator for looking at graphs/sampling: https://github.com/gmega/block.protocol

benbierens:
Reliability Test write-up: https://hackmd.io/rwPtPJ7KTw6cGjhN0zNYig

gmega:
This is an important step: #1119

gmega:
Write-up on me trying to build a mental model of out how I think this should work: https://hackmd.io/ZAblhSiPTRahEz9w0mulcA

benbierens:
It's been a while since I've done anything for this topic. So I dusted off some of performance tracking tests. (They never made it into the official release tests because they don't really give a clear pass or fail.)

Block-retransmit was an issue where a node would receive the same block multiple times from multiple peers, wasting bandwidth and slowing down total transfer. It has gone down from 2% retransmit in the "normal" swarm test for 0.2.4 to 0% for latest master for the same test. I'll see about getting some reliability numbers.

PR #1093, opened 2025-01-31
Prevent duplicate block transmission

benbierens:
During reliability testing (https://github.com/codex-storage/nim-codex/issues/974) block retransmission was found to cause significant delay and download failure in larger networks.

This PR prevents wantBlocks being sent for open requests for which a wantBlock has already been sent, reducing duplicates to zero.

Open issue (that already existed and hasn't been seen in testing) is that when a wantBlock is sent and the peer *doesn't* respond with the block, the inFlight flag is never cleared, forcing the full block-retrieval-timeout before the block can be retried.

### DIFF
--- a/codex/blockexchange/engine/engine.nim
+++ b/codex/blockexchange/engine/engine.nim
@@ -152,10 +152,15 @@ proc sendWantHave(
 proc sendWantBlock(
     b: BlockExcEngine, addresses: seq[BlockAddress], blockPeer: BlockExcPeerCtx
 ): Future[void] {.async.} =
-  trace "Sending wantBlock request to", addresses, peer = blockPeer.id
+  let notInFlight = b.pendingBlocks.getNotInFlight(addresses)
+  if notInFlight.len == 0:
+    return
+  b.pendingBlocks.setInFlight(notInFlight)
+
+  trace "Sending wantBlock request to", addrs = notInFlight, peer = blockPeer.id
   await b.network.request.sendWantList(
-    blockPeer.id, addresses, wantType = WantType.WantBlock
-  ) # we want this remote to send us a block
+    blockPeer.id, notInFlight, wantType = WantType.WantBlock
+  )
   codex_block_exchange_want_block_lists_sent.inc()
 
 proc monitorBlockHandle(
@@ -197,7 +202,6 @@ proc requestBlock*(
     else:
       let selected = pickPseudoRandom(address, peers.with)
       asyncSpawn b.monitorBlockHandle(blockFuture, address, selected.id)
-      b.pendingBlocks.setInFlight(address)
       await b.sendWantBlock(@[address], selected)
 
     await b.sendWantHave(@[address], peers.without)
--- a/codex/blockexchange/engine/pendingblocks.nim
+++ b/codex/blockexchange/engine/pendingblocks.nim
@@ -113,12 +113,23 @@ proc setInFlight*(p: PendingBlocksManager, address: BlockAddress, inFlight = tru
   p.blocks.withValue(address, pending):
     pending[].inFlight = inFlight
 
+proc setInFlight*(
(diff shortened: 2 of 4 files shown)


emizzle (approved):
LGTM

codex/blockexchange/engine/engine.nim
```diff
@@ -152,10 +152,15 @@ proc sendWantHave(
 proc sendWantBlock(
     b: BlockExcEngine, addresses: seq[BlockAddress], blockPeer: BlockExcPeerCtx
 ): Future[void] {.async.} =
-  trace "Sending wantBlock request to", addresses, peer = blockPeer.id
+  let notInFlight = b.pendingBlocks.getNotInFlight(addresses)
+  if notInFlight.len == 0:
+    return
+  b.pendingBlocks.setInFlight(notInFlight)
```
marcinczenko, line 158: I am a bit puzzled here. `sendWantBlock` is called only from two places:

1. in `requestBlock`
2. in`blockPresenceHandler`

In `requestBlock`, it looks to me that we already do the filtering: we will only ever call `sendWantBlock` if the block is `not b.pendingBlocks.isInFlight(address)`. Thus in this case the filtering that you added here looks like duplication to me. It looks to me that you are trying to prevent duplicate `sendWantBlock` while calling it from `blockPresenceHandler` - isn't it better to add this filtering there. A disadvantage of course is that we will have the "not-in-flight" filtering in two places, but an advantage is that the `sendWantBlock` has single responsibility without triggering side-effects - if one calls `sendWantBlock` then one can be sure it will be sent and not filtered out.

marcinczenko: One more small thought. In the PR description you say:

> Open issue (that already existed and hasn't been seen in testing) is that when a wantBlock is sent and the peer doesn't respond with the block, the inFlight flag is never cleared, forcing the full block-retrieval-timeout before the block can be retried.

If you look into `getWantHandle`, there is this nice error message: 

```nim
  error "Pending WANT failed or expired", exc = exc.msg
  # no need to cancel, it is already cancelled by wait()
```

where this "wait" is to be found in `requestBlock`:

```nim
try:
  success await blockFuture
except AsyncTimeoutError as err:
  failure err
```

`PendingBlocksManager` is able to track pending requests, but only when called via `requestBlock`, which makes sure that the "requested" block is actually requested via `sendWantBlock`. As long as `requestBlock` is used, "pendingBlocks" and the "inFlight"  blocks will be in sync, but when we start calling `sendWantBlock` outside of `requestBlock` without updating pending blocks, those two go out of sync and I feel this is what makes managing block requests more challenging. A little bit like we have two mechanism to manage the same things: pending blocks, and inFlight blocks. Perhaps, when we want to _request_ a block, we should have one entry point where we (1) make sure that the block was actually requested (via `sendWantBlock`) (2) we `asyncSpawn` something similar to what `pendingBlocks.getWantHandle` is currently doing (reducing current `pendingBlocks.getWantHandle` to just returning the future so that the requester can still wait on it if the block is to be returned). Having this one entry point, the pending blocks would then always be "inFlight", and we could even maybe even get rid of "inFlight" at all.

We use a similar "inFlight" mechanism when sending, but there we do not have to wait for anything coming back, what makes "inFlight" working good there.

I am documenting all this stuff right now (I need it for BitTorrent integration), so this is why I was triggered by this PR. I think am have pretty good picture of the whole block exchange engine, and I think I could improve this a bit, or at least help.

benbierens: Ideally, I would have liked to rewrite `requiestBlock` after this, because the case where it calls `sendWantBlock` can never happen. It is only ever called in unit tests, but can't occur during app run. I'm always trying to make small PRs.

dryajov:
This is addressed in #1119

PR #1095, opened 2025-02-02
fix dataset and slot size calculations in integration tests

marcinczenko:
When working on price-per-byte feature, I added two functions to `tests/integration/marketplacesuite.nim`:

```nim
proc slotSize(blocks: int): UInt256 =
  (DefaultBlockSize * blocks.NBytes).Natural.u256

proc datasetSize(blocks, nodes, tolerance: int): UInt256 =
  (nodes + tolerance).u256 * slotSize(blocks)
```

They are not correct, and should be defined as follows:

```nim
proc slotSize(blocks, nodes, tolerance: int): UInt256 =
  let ecK = nodes - tolerance
  let blocksRounded = roundUp(blocks, ecK)
  let blocksPerSlot = divUp(blocksRounded, ecK)
  (DefaultBlockSize * blocksPerSlot.NBytes).Natural.u256

proc datasetSize(blocks, nodes, tolerance: int): UInt256 =
  return nodes.u256 * slotSize(blocks, nodes, tolerance)
```

The functions are used in the following integration tests:

- `ntegration/testmarketplace`
- `integration/testproofs`
- `integration/testvalidator`

While working on this, I discovered that `proc example*(_: type RandomChunker, blocks: int): Future[string] {.async.}`, returning `string`, was doing `byteutils.toHex(data)` to convert the `seq[bytes]` into a `string`, which means it was returning double of the requested size (so e.g. `16` blocks where `8` were requested). This was causing subtle errors in some integration tests, once the computation of the dataset size and the slot size has been corrected as indicated above.

To fix this, the `example` proc listed above has been changed so that it now returns `seq[byte]` with the correct size:

```nim
proc example*(_: type RandomChunker, blocks: int): Future[seq[byte]] {.async.} =
  let rng = Rng.instance()
  let chunker = RandomChunker.new(
    rng, size = DefaultBlockSize * blocks.NBytes, chunkSize = DefaultBlockSize
  )
  var data: seq[byte]
  while (let moar = await chunker.getBytes(); moar != []):
    data.add moar
  return data
```

This change, however, was causing a problem with `CodexClient.upload`, which expects passed `contents` to have type `string`. Because there are still a number of tests where we do actually use strings, in order to support both content types, I added the following overload for the `CodexClient.upload` proc:

```nim
proc upload*(client: CodexClient, bytes: seq[byte]): ?!Cid =
  proc toString(bytes: seq[byte]): string =
    result = newString(bytes.len)
    copyMem(result[0].addr, bytes[0].unsafeAddr, bytes.len)

  client.upload(bytes.toString)
```

The relevant tests has been reviewed and fixed.

### DIFF
--- a/tests/integration/codexclient.nim
+++ b/tests/integration/codexclient.nim
@@ -45,11 +45,7 @@ proc upload*(client: CodexClient, contents: string): ?!Cid =
   Cid.init(response.body).mapFailure
 
 proc upload*(client: CodexClient, bytes: seq[byte]): ?!Cid =
-  proc toString(bytes: seq[byte]): string =
-    result = newString(bytes.len)
-    copyMem(result[0].addr, bytes[0].unsafeAddr, bytes.len)
-
-  client.upload(bytes.toString)
+  client.upload(string.fromBytes(bytes))
 
 proc download*(client: CodexClient, cid: Cid, local = false): ?!string =
   let response = client.http.get(
--- a/tests/integration/codexclient.nim
+++ b/tests/integration/codexclient.nim
@@ -44,6 +44,13 @@ proc upload*(client: CodexClient, contents: string): ?!Cid =
   assert response.status == "200 OK"
   Cid.init(response.body).mapFailure
 
+proc upload*(client: CodexClient, bytes: seq[byte]): ?!Cid =
+  proc toString(bytes: seq[byte]): string =
+    result = newString(bytes.len)
+    copyMem(result[0].addr, bytes[0].unsafeAddr, bytes.len)
+
+  client.upload(bytes.toString)
+
 proc download*(client: CodexClient, cid: Cid, local = false): ?!string =
   let response = client.http.get(
     client.baseurl & "/data/" & $cid & (if local: "" else: "/network/stream")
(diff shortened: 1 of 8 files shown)


tests/integration/codexclient.nim
```diff
@@ -44,6 +44,13 @@ proc upload*(client: CodexClient, contents: string): ?!Cid =
   assert response.status == "200 OK"
   Cid.init(response.body).mapFailure
 
+proc upload*(client: CodexClient, bytes: seq[byte]): ?!Cid =
+  proc toString(bytes: seq[byte]): string =
+    result = newString(bytes.len)
+    copyMem(result[0].addr, bytes[0].unsafeAddr, bytes.len)
```
marcinczenko, line 50: I was also thinking of just using `cast[string](bytes)` to avoid copying, but after seeing https://forum.nim-lang.org/t/6477 and https://forum.nim-lang.org/t/1356 it looks it is not a good idea... Going other way around feels safer.

See https://play.nim-lang.org/#pasty=KVKaEQFx

emizzle (approved):

markspanbroek (approved):
Nice, thanks @marcinczenko!

PR #1098, opened 2025-02-04
docker: add BOOTSTRAP_NODE_URL to the Docker entrypoint

veaceslavdoina:
This is a quick and one more workaround for Docker images.

Now, we can pass `BOOTSTRAP_NODE_URL` and optionally `BOOTSTRAP_NODE_URL_WAIT` variables.

1. Entrypoint will loop for `BOOTSTRAP_NODE_URL_WAIT` or default for `300` seconds
2. Will try to get SPR from `BOOTSTRAP_NODE_URL/api/codex/v1/spr`
3. Will stop execution if SPR not received in time.

`BOOTSTRAP_NODE_URL` should be passed in form of base URL - `http://localhost:8080`.

### DIFF
--- a/docker/docker-entrypoint.sh
+++ b/docker/docker-entrypoint.sh
@@ -9,6 +9,34 @@ if [[ -n "${ENV_PATH}" ]]; then
   set +a
 fi
 
+# Bootstrap node from URL
+if [[ -n "${BOOTSTRAP_NODE_URL}" ]]; then
+  BOOTSTRAP_NODE_URL="${BOOTSTRAP_NODE_URL}/api/codex/v1/spr"
+  WAIT=${BOOTSTRAP_NODE_URL_WAIT:-300}
+  SECONDS=0
+  SLEEP=1
+  # Run and retry if fail
+  while (( SECONDS < WAIT )); do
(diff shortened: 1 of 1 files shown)


docker/docker-entrypoint.sh
```diff
@@ -9,6 +9,34 @@ if [[ -n "${ENV_PATH}" ]]; then
   set +a
 fi
 
+# Bootstrap node from URL
+if [[ -n "${BOOTSTRAP_URL}" ]]; then
+  BOOTSTRAP_URL="${BOOTSTRAP_URL}/api/codex/v1/spr"
+  WAIT=${BOOTSTRAP_URL_WAIT:-300}
+  SECONDS=0
+  SLEEP=5
```
gmega, line 17:
```suggestion
  SLEEP=1
```
Maybe a shorter sleep?

gmega (approved):
LGTM

docker/docker-entrypoint.sh
veaceslavdoina: Updated

dryajov (approved):

PR #1099, opened 2025-02-05
feat(marketplace): persistent availabilities

2-towns:
Closes https://github.com/codex-storage/nim-codex/issues/467


codex/rest/api.nim
```diff
@@ -49,6 +49,31 @@ logScope:
...
+    let requestEnds = allReservations.mapIt(await market.getRequestEnd(it.requestId))
+
+    if len(requestEnds) == 0:
+      return success(0.SecondsSince1970)
+
+    return success(requestEnds.max)
+  except CancelledError, CatchableError:
+    error "Error when trying to get longest request end",
+      error = getCurrentExceptionMsg()
+    return failure("Cannot retrieve the request dates")
```
emizzle, line 75: 🔴  Swallowing `CancelledError` here will prevent cancellation propagation.

```suggestion
  except CancelledError as e:
    raise e
  except CatchableError as e:
    error "Error when trying to get longest request end",
      error = e.msg
    return failure("Cannot retrieve the request dates: " & e.msg)
```

codex/rest/api.nim
```diff
@@ -556,6 +593,26 @@ proc initSalesApi(node: CodexNodeRef, router: var RestRouter) =
...
         availability.totalCollateral = totalCollateral
 
+      if until =? restAv.until:
+        if until < 0:
+          return RestApiResponse.error(
+            Http400, "Until parameter must be greater or equal 0. Got: " & $until
+          )
+
+        without longestRequestEnd =? (await node.getLongestRequestEnd(id)).catch, err:
+          return RestApiResponse.error(Http500, err.msg)
```
emizzle, line 603: 🔴  Avoid using `await` and `catch` together as this swallow cancellations. See the [discord message](https://discord.com/channels/895609329053474826/1168926478180810903/1339174136290873356) for reasoning and alternative options.
🟡  Just a suggestion, and more verbose, but you could try to distinguish known errors and return more accurate http error codes, eg 
```nim
try:
  without longestRequestEnd =? (await node.getLongestRequestEnd(id)), err:
    if err of NotExistError:
       return RestApiResponse.error(Http404, err.msg)
    else:
       return RestApiResponse.error(Http500, err.msg)
except CancelledError as e:
  raise e
```
(probably more trouble than it's worth)

emizzle (changes requested):
Looking good, but i think there's a few things that need to be addressed:

1. validate `until` value in `updateAvailability` instead of in rest api
2. do not swallow `CancelledError` in `getLongestRequestEnd` (might need to move this into reservations module due to #1 (above))
3. `await` + `catch`

codex/rest/api.nim
2-towns: Okay I will change this. 

So from what I understand, `CancelledError` is an error (or exception) like other errors in Nim, but it is used to cancel a `Future`. So when it is raised, it should be propagated to stop that `Future` but it should not be logged as an error because it is the result of a cancellation and not an error. Am I right ?

codex/sales/reservations.nim
```diff
@@ -268,18 +280,8 @@ proc updateAvailability(
...
-        # certainty, that nothing will be raised
-        try:
-          await onAvailabilityAdded(obj)
-        except CancelledError as e:
-          raise e
-        except CatchableError as e:
-          # we don't have any insight into types of exceptions that
-          # `onAvailabilityAdded` can raise because it is caller-defined
-          warn "Unknown error during 'onAvailabilityAdded' callback", error = e.msg
+      if obj.enabled and onAvailabilityAdded =? self.onAvailabilityAdded:
```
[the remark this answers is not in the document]
2-towns: The problem with the `until` validation is you need to check that a running request doesn't end after the `until` date. And this information has to be retrieved from the contracts with `getRequestEnd`.

codex/rest/api.nim
2-towns: Okay, I totally agree with you in principle.

But in that specific case, is it a problem?

In your example, we would not raise a CancelledError, we would still return a RestApiResponse.error(Http500, err.msg), wouldn't we?

codex/rest/api.nim
emizzle: Yes that's correct. In most cases, the error should be propagated. The most common case where it should not be propagated is when a Future is asyncSpawned. In the case, cancelling that Future will raise a Defect.

codex/rest/api.nim
emizzle: There may be a reason for a cancellation to be propagated in presto and presto may have a way of handling it in relation to current requests 🤷‍♂️ , so my assumption would be to propagate a cancellation here and not turn it into an error. I could be wrong. But if, for example, presto gets a stop signal externally and it cancels any ongoing operations, perhaps the cancellation should be propagated or else risk a non-graceful termination. In practice, I could see something like an integration test failing for some reason and then shutting down the nodes in the middle of a REST api call, you may see a "future operation cancelled" error in the logs AFTER shutdown has already occurred.

codex/sales/reservations.nim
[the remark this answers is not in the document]
emizzle: Good point. I think in this case, we could set a `requestEnd` or `validUntil` property on the Reservation during creation, and then use that value here. This is somewhat inline with the idea we discussed a while back about persisting Request info on the Reservation. Wdyt?

codex/rest/api.nim
2-towns: Yeah but I don't think presto is taking care of that. 

From what I saw, it is just re-raising the `CancelledError` hoping someone will manage it: 

https://github.com/status-im/nim-presto/blob/3ccb356220b70f7d9eb0fbd58b674c4080f78014/presto/route.nim#L82-L85

codex/sales/reservations.nim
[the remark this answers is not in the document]
2-towns: Okay so when the user will try to update the `until` date, we will loop over each reservations and compare the `until` date with the `validUntil`. Sounds good.

codex/rest/api.nim
emizzle: Right, so CancelledError is re-raised, effectively propagated the cancellation, and so we wouldn't want to interfere with the cancellation propagation by swallowing the CancelledError.

tests/codex/sales/testreservations.nim
```diff
@@ -323,6 +392,41 @@ asyncchecksuite "Reservations module":
...
+    let found = await reservations.findAvailability(
+      availability.freeSize, availability.duration,
+      availability.minPricePerBytePerSecond, collateralPerByte,
+    )
+
+    check found.isNone
+
+  test "finds an availability when the until date is after the duration":
+    let example = Availability.example(collateralPerByte)
+    let until = getTime().toUnix() + cast[int64](example.duration)
```
emizzle, line 407: 🟡  Just a suggestion
```suggestion
    let until = getTime().toUnix() + example.duration.int64
```

tests/codex/sales/testreservations.nim
```diff
@@ -259,6 +268,48 @@ asyncchecksuite "Reservations module":
...
+  test "update an availability fails when trying set until with a negative value":
+    let until = getTime().toUnix()
+    let availability = createAvailability(until = until)
+
+    availability.until = -1
+
+    let result = await reservations.update(availability)
+    check result.isErr
+    check result.error of UntilOutOfBoundsError
+
```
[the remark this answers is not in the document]
emizzle: Ok I see you added one in `testsales` 👍

tests/codex/sales/testsales.nim
```diff
@@ -416,6 +420,41 @@ asyncchecksuite "Sales":
...
+    check wasIgnored()
+
+  test "ignores request when availability was disabled after the request storage is created":
+    createAvailability(enabled = true)
+    await market.requestStorage(request)
+
+    availability.enabled = false
+    discard await reservations.update(availability)
+
+    check wasIgnored()
```
emizzle, line 435: I'm curious why this works. When the request is made, that should trigger subscription callbacks in the sales module, which should then trigger the Sales statemachine workflow where a Reservation is created. But that's not happening here. Is it a timing issue that makes it work where the availability is updated quick enough before the statemachine is run?

tests/integration/testmarketplace.nim
```diff
@@ -123,6 +125,52 @@ marketplacesuite "Marketplace":
       timeout = 10 * 1000, # give client a bit of time to withdraw its funds
     )
 
+  test "returns an error when trying to update the until date before an existing a request is finished",
```
emizzle, line 128: 🟡  It feels like this belongs in testsales, but not a problem either way/

tests/integration/testsales.nim
```diff
@@ -53,6 +56,22 @@ multinodesuite "Sales":
     ).get
     check availability in host.getAvailabilities().get
 
+  test "creating availability fails when until is negative", salesConfig:
```
emizzle, line 59: 🟡  I feel like this might fit better in the testrestapi module

emizzle (approved):
Great job, Arnaud! Nothing block at all, just some minor comments/suggestions that aren't at all blocking, so will approve 👍

codex/sales/reservations.nim
```diff
@@ -256,78 +282,91 @@ proc updateImpl(self: Reservations, obj: SomeStorableObject): Future[?!void] {.a
...
+      # inform subscribers that Availability has been modified (with increased
+      # size)
+      if onAvailabilityAdded =? self.onAvailabilityAdded:
+        await onAvailabilityAdded(obj)
+    return res
+  except CancelledError as e:
+    raise e
+  except CatchableError as e:
+    error "Error when trying to update availability", error = e.msg
+    return failure(e)
```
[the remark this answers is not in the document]
2-towns: `self.get` is raising `CatchableError`. Here is the compilation error: 

```Error: cast[type(get(self, key, Availability))](chronosInternalRetFuture.internalChild).internalError can raise an unlisted exception: ref CatchableError```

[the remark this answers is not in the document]
Comment (2-towns) [reply codex/sales/reservations.nim:356]: Maybe we can catch the `CatchableError` in the `get` proc and return a failure. Wdyt ?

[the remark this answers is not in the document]
Comment (emizzle) [reply codex/sales/reservations.nim:349]: I believe `get` can return `GetFailedError`, or `SerializationError`, both of which inherit from `ReservationsError` (which is inherited from `CatchableError`). Sometimes the compiler does not do a good job of telling us which exceptions are being raised when having a return type of `Result[T, ref CatchableError]` (equivalent to `?!T`), and then returning a derived exception type in the failure.

File: tests/codex/sales/testsales.nim (hunk 6 above)
Comment (emizzle) [reply tests/codex/sales/testsales.nim:436]: If this test purely relies on timing, maybe we should remove it? Timing-related tests have caused issues in the past in slow CI environments. Wdyt?

File: codex/sales/reservations.nim [hunk 10]
```diff
@@ -256,78 +282,91 @@ proc updateImpl(self: Reservations, obj: SomeStorableObject): Future[?!void] {.a
...
-          (obj.totalSize - oldAvailability.totalSize).truncate(uint).NBytes
+        return res
+      else:
+        return failure(err)
+
+    if obj.until > 0:
+      without allReservations =? await self.all(Reservation, obj.id):
+        let error = newException(
+          GetFailedError,
+          "Until parameter must be greater or equal the current longest request",
```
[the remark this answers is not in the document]
2-towns: Yes it does work, I applied it.

codex/sales/reservations.nim
```diff
@@ -438,9 +486,13 @@ method createReservation*(
       )
       return failure(error)
 
-    trace "Creating reservation", availabilityId, slotSize, requestId, slotIndex
+    let validUntil =
+      times.now().utc().toTime().toUnix() + duration.truncate(SecondsSince1970)
```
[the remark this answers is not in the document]
2-towns: Yes it is I applied the change.

tests/codex/sales/testreservations.nim
2-towns: Yes I applied it. Thanks

tests/codex/sales/testsales.nim
2-towns: I agree, I will remove it.

tests/integration/testsales.nim
2-towns: Yes agree, I moved it.

codex/sales/reservations.nim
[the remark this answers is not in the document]
2-towns: Well actually, `get` proc does not raise any error (it returns it as a Result) but the problem is in `nim-datastore`, the pragma `async` don't include the `raises` element with the errors, so the compiler return a false positive by telling us the `CatchableError` is raised. I will update `nim-datastore` for the methods used by reservations in this PR and create another PR for the other methods.

tests/integration/testmarketplace.nim
2-towns: I moved it to testsales

codex/sales/reservations.nim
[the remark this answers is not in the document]
2-towns: I updated `nim-datastore` to explicitly define the errors raised, which led to the removal of the try/except block.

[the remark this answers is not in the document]
emizzle: Ah yes good point! Great find. Thanks for doing that!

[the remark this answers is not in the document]
emizzle: The reason is because: 
> old-style functions annotated with plain `async` will behave as if they had been annotated with `async: (handleException: true)`.

This is functionally equivalent to `async: (handleException: true, raises: [CatchableError])` and will, as before, remap any `Exception` that is not `Defect` into `AsyncExceptionError`, while also allowing any `CatchableError` (including `AsyncExceptionError`) to get through without compilation errors.

From https://status-im.github.io/nim-chronos/error_handling.html

AuHau (changes requested):
Generally LGTM! Thanks for pushing to the finish line what I did not manage! 🎉 This will be great improvement for UX.

There are two blockers from my side though and few other minor comments.

codex/sales/reservations.nim
```diff
@@ -70,24 +71,35 @@ type
...
+    # If set to 0, there will be no restrictions.
+    until* {.serialize.}: SecondsSince1970
 
   Reservation* = ref object
     id* {.serialize.}: ReservationId
     availabilityId* {.serialize.}: AvailabilityId
     size* {.serialize.}: uint64
     requestId* {.serialize.}: RequestId
     slotIndex* {.serialize.}: uint64
+    validUntil* {.serialize.}: SecondsSince1970
```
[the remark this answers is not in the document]
2-towns: > I am looking into what this is used for, and I guess the main use-case is for validation of the until update of Availability (eq. making sure it is not lower than the "longest until"), right? 

Yes 

> Just curious what should happen if now is after validUntil? 

I am trying to understand your question. If the current timestamp `now` is after `validUntil` that means the reservation is completed. So the reservation should be deleted at this point.

AuHau (changes requested):
Thanks for implementing the feedback! It generally LGTM, but "unfortunatelly" I have discovered one more bug, that needs fixing before I can approve ;-)

The bug is related to quite an edge-case, but IMHO still valid. It is related to the fact that `reservations.deleteReservation()` call will return "undownloaded" bytes to the availability. The edge-case bug will be present if there will be an error during the `downloading` state which will lead to calling `cleanUp`, where we will return the bytes for the whole slot using the `reservations.returnBytesToAvailability()`, but then we will also return the remaning "undownloaded" bytes, which could lead to having `freeSize > totalSize` of availability. Most probably this bug is already present in the current master, so not really related to your work, I just realised about it thanks to your PR ;-)

My suggestion is therefore remove the "returning undownloaded bytes to the availability" in the `deleteReservations` as we do not need that functionality as those bytes will be returned in the `cleanUp` with the `reservations.returnBytesToAvailability()` call.

I would ask @emizzle to double-check my thinking here though ;-)

tests/codex/sales/states/testpreparing.nim
```diff
@@ -52,6 +55,9 @@ asyncchecksuite "sales state 'preparing'":
     context.reservations = reservations
     agent = newSalesAgent(context, request.id, slotIndex, request.some)
 
+    market.requestEnds[request.id] =
+      getTime().toUnix() + cast[int64](request.ask.duration)
+
```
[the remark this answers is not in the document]
2-towns: Great point thanks. I changed to use `clock.now()`.

emizzle:
> My suggestion is therefore remove the "returning undownloaded bytes to the availability" in the deleteReservations as we do not need that functionality as those bytes will be returned in the cleanUp with the reservations.returnBytesToAvailability() call.

Yes, this is a valid point and definitely a bug. Great find! I believe the only complete way to address this would be to store the amount of released bytes in the Reservation, and then upon deletion, re-reserve those bytes before returning them to the Availability. Even with that change, there would still need to be a reconciliation between `returnBytesToAvailability` and `deleteReservation`. My best guess would be to remove `returnBytesToAvailability` completely and add the re-reservation of released bytes logic into `deleteReservation`. 

However, this highlights the complexities of accounting wrt Reservations and Availabilities. With the new RepoStore changes that removes the need to reserve/release bytes externally, we will no longer need Reservations at all. So maybe we should create an issue for this, but I would avoid spending more time on it since it will not need to be addressed once Reservations are gone. In saying that, I think we should still get this PR across the line as it is the latest evolution of Reservations/Availabilities, thoughts?

PR #1102, opened 2025-02-06
fix: use ubuntu-24.04 runners for docker workflows

veaceslavdoina:
We recently observed that `amd64` Docker builds, which uses `ubuntu-22.04` runner, started to fail because they are running on the self-hosted runners.

And there two issues with that
- Self hosted runners are not configured properly to run Docker
- We have clashed labels - `ubuntu-22.04` is used by self-hosted runners and GitHub ones as well

PR contains a quick workaround to switch to the GutHub hosted runners with the labels `ubuntu-24.04` and `ubuntu-24.04-arm`.

### DIFF
--- a/.github/workflows/docker-reusable.yml
+++ b/.github/workflows/docker-reusable.yml
@@ -94,11 +94,11 @@ jobs:
           - target:
               os: linux
               arch: amd64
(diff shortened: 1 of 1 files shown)


gmega (approved):
YES

PR #1103, opened 2025-02-06
Purging of local datasets

gmega:
This PR adds an endpoint for deleting local datasets by their CID. Consistently with download API, passing a block CID will cause a block to be deleted, whereas passing a manifest CID will cause all the blocks under the dataset represented by that manifest to be deleted.

### DIFF
--- a/build.nims
+++ b/build.nims
@@ -2,25 +2,8 @@ mode = ScriptMode.Verbose
 
 import std/os except commandLineParams
 
-const VendorPath = "vendor/nim-nat-traversal/vendor/libnatpmp-upstream"
-let 
-  oldVersionFile = joinPath(VendorPath, "VERSION")
-  newVersionFile = joinPath(VendorPath, "VERSION_temp")
-
-proc renameFile(oldName, newName: string) =
-  if fileExists(oldName):
-    mvFile(oldName, newName)
-  else:
-    echo "File ", oldName, " does not exist"
-
-
 ### Helper functions
 proc buildBinary(name: string, srcDir = "./", params = "", lang = "c") =
-  # This is a quick workaround to avoid VERSION file conflict on macOS
-  # More details here: https://github.com/codex-storage/nim-codex/issues/1059
-  if defined(macosx):
-    renameFile(oldVersionFile, newVersionFile)
-
   if not dirExists "build":
     mkDir "build"
 
@@ -30,25 +13,25 @@ proc buildBinary(name: string, srcDir = "./", params = "", lang = "c") =
     for param in commandLineParams():
       extra_params &= " " & param
   else:
-    for i in 2..<paramCount():
+    for i in 2 ..< paramCount():
       extra_params &= " " & paramStr(i)
 
   let
     # Place build output in 'build' folder, even if name includes a longer path.
     outName = os.lastPathPart(name)
-    cmd = "nim " & lang & " --out:build/" & outName & " " & extra_params & " " & srcDir & name & ".nim"
-  try:
-    exec(cmd)
-  finally:
-    if defined(macosx):
-      renameFile(newVersionFile, oldVersionFile)
+    cmd =
+      "nim " & lang & " --out:build/" & outName & " " & extra_params & " " & srcDir &
+      name & ".nim"
+
+  exec(cmd)
 
 proc test(name: string, srcDir = "tests/", params = "", lang = "c") =
   buildBinary name, srcDir, params
   exec "build/" & name
 
 task codex, "build codex binary":
-  buildBinary "codex", params = "-d:chronicles_runtime_filtering -d:chronicles_log_level=TRACE"
+  buildBinary "codex",
+    params = "-d:chronicles_runtime_filtering -d:chronicles_log_level=TRACE"
 
 task toolsCirdl, "build tools/cirdl binary":
   buildBinary "tools/cirdl/cirdl"
@@ -60,7 +43,9 @@ task testContracts, "Build & run Codex Contract tests":
   test "testContracts"
 
 task testIntegration, "Run integration tests":
-  buildBinary "codex", params = "-d:chronicles_runtime_filtering -d:chronicles_log_level=TRACE -d:codex_enable_proof_failures=true"
+  buildBinary "codex",
+    params =
+      "-d:chronicles_runtime_filtering -d:chronicles_log_level=TRACE -d:codex_enable_proof_failures=true"
   test "testIntegration"
   # use params to enable logging from the integration test executable
   # test "testIntegration", params = "-d:chronicles_sinks=textlines[notimestamps,stdout],textlines[dynamic] " &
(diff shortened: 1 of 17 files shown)


codex/node.nim
```diff
@@ -267,6 +267,59 @@ proc retrieve*(
...
+    self: CodexNodeRef, cid: Cid
+): Future[?!void] {.async: (raises: [CatchableError]).} =
+  ## Deletes a whole dataset, if Cid is a Manifest Cid, or a single block, if Cid a block Cid,
+  ## from the underlying block store. This is a strictly local operation.
+  ##
+  ## Missing blocks in dataset deletes are ignored.
+  ##
+
+  without isManifest =? cid.isManifest, err:
+    trace "Bad content type for CID:", err = err.msg
```
2-towns, line 314: I think you missed to add the cid in the trace message no ?

codex/node.nim
```diff
@@ -267,6 +267,59 @@ proc retrieve*(
...
+    return failure(err)
+
+  if not isManifest:
+    return await self.deleteSingleBlock(cid)
+
+  echo "dispatch to entire dataset"
```
2-towns, line 320: You forgot to delete this debug message, or maybe you `trace`

codex/rest/api.nim
```diff
@@ -254,6 +263,27 @@ proc initDataApi(node: CodexNodeRef, repoStore: RepoStore, router: var RestRoute
...
     await node.retrieveCid(cid.get(), local = true, resp = resp)
 
+  router.api(MethodDelete, "/api/codex/v1/data/{cid}") do(
+    cid: Cid, resp: HttpResponseRef
+  ) -> RestApiResponse:
+    ## Deletes either a single block or an entire dataset
+    ## from the local node. Does nothing and returns 200
+    ## if the dataset is not locally available.
+    ##
+    var headers = buildCorsHeaders("DELETE", allowedOrigin)
```
2-towns, line 273: I don't think you need to take care of the CORS here. This is the purpose of the endpoint that you created before with MethodOptions. When a preflight request is sent, it will hit the method 'OPTIONS', and you allowed 'GET' and 'DELETE' methods. So the preflight request will be successful, and the 'DELETE' request will be executed.

codex/rest/api.nim
```diff
@@ -254,6 +263,27 @@ proc initDataApi(node: CodexNodeRef, repoStore: RepoStore, router: var RestRoute
...
+      resp.setCorsHeaders("DELETE", corsOrigin)
+      resp.setHeader("Access-Control-Headers", "X-Requested-With")
+
+    if cid.isErr:
+      return RestApiResponse.error(Http400, $cid.error(), headers = headers)
+
+    if err =? (await node.delete(cid.get())).errorOption:
+      return RestApiResponse.error(Http500, err.msg, headers = headers)
+
+    return RestApiResponse.response("", headers = headers)
```
2-towns, line 285: You could return 204 error code instead of 200 since the response is empty

codex/node.nim
```diff
@@ -267,6 +267,59 @@ proc retrieve*(
 
   await self.streamEntireDataset(manifest, cid)
 
+proc deleteSingleBlock(self: CodexNodeRef, cid: Cid): Future[?!void] {.async.} =
```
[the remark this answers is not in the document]
gmega: The problem is that `delBlock` doesn't have well-defined exception contracts, and I also cannot catch CancelledError. so I'd end up with something clumsy there:

```nim
try:
   ...
except CancelledError as err:
   raise err
except CatchableError as err:
   return failure(err)
```
And I'd still need to have `raises: [CancelledError]`. So it didn't seem worth it to me.

codex/node.nim
gmega: Ooops true.

codex/node.nim
gmega: :grimacing:

codex/rest/api.nim
gmega: Yeah OK. I thought you always had to check the CORS origin, even if you have a preflight, cause callers can still not use the preflight, or otherwise screw things up.

codex/rest/api.nim
gmega: Good point.

gmega:
I think I've addressed your comments @2-towns, either by fixing them or by continuing the discussion.

codex/node.nim
```diff
@@ -267,6 +267,58 @@ proc retrieve*(
...
+  if not (await cid in store):
+    return failure("Dataset's manifest is not available locally - nothing to delete")
+
+  without manifestBlock =? await store.getBlock(cid), err:
+    return failure(err)
+
+  without manifest =? Manifest.decode(manifestBlock), err:
+    return failure(err)
+
+  for i in 0 ..< manifest.blocksCount:
```
dryajov, line 291: This will very likely lockup the node for the duration of the delete, unless we yield some time back to the main thread. I think there is an `IdleAsync` method in chronos, that will schedule this after all the IO related work is done, otherwise, we can always do an async sleep after some number of blocks have been processed.

gmega: Yeah good catch - I forget that those are actually blocking all the way. I'll add one of those.

codex/rest/api.nim
2-towns: You could remove this line as well actually. The other endpoints include it because if there is an error, you still need to pass the CORS headers, yes. But in your case, your created a specific endpoint to handle this so you don't need to pass the headers even if there is an error. 

If think the only header we don't need is `X-Requested-With`. But `Access-Control-Allow-Origin` and `Access-Control-Allow-Methods` are definitely needed. `Access-Control-Max-Age` is for caching the previous header.

codex/node.nim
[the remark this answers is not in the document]
2-towns: OK gotcha, I don't have a strong opinion on it. I just feel strange that we return `Result` while the method can raise en `Error`.

codex/node.nim
gmega: OK I've added a fix for this, I ended up going with a time-based quota as I think that's a lot more predictable than counting iterations.

codex/rest/api.nim
gmega: OK so if we're to follow the standard, my understanding is that I should always send `Access-Control-Allow-Origin`, even in this endpoint. 

Then, for the preflight endpoint, I should send:

* `Access-Control-Allow-Origin`;
* `Access-Control-Allow-Methods`;
* `Access-Control-Max-Age`
* `Access-Control-Headers`

2-towns: Not sure why we need Access-Control-Allow-Origin in the endpoint itself.

2-towns: But yes you are right we need it

2-towns (approved):
I am approving it since you resolved my comments and you will address the CORS issues in another PR. 

Thanks @gmega for your work.

build.nims
```diff
@@ -4,7 +4,6 @@ import std/os except commandLineParams
```
gmega, line 4: Looks like I've ran nph on build.nims as part of some commit, for some reason, in this branch. Rolling this back would be more trouble than just leaving it here, so I'll "bundle" it with this PR.

dryajov (approved):

PR #1104, opened 2025-02-07
chore(marketplace) - fix the http error codes when validating the availability requests

### LINKED ISSUE #884: [BUG] `PATCH` Availability's `totalSize` above node's quota should not result with 500, but 400
Currently if updating `totalSize` above node's quota, the call correctly fails because it tries to reserve additional space above node's capacity. But it fails with `500 Internal Server Error`, while this is valid edge-case which should fail with `400 Bad Request` instead.

2-towns:
This PR fixes #884 and fixes the semantics regarding the validation error codes. 

400 should be returned when the request cannot be processed (malformatted for example) [rfc](https://www.rfc-editor.org/rfc/rfc9110.html#name-400-bad-request). 
422 should be returned when the request has semantic errors (validation error for example) [rfc](https://www.rfc-editor.org/rfc/rfc4918.html#section-11.2).

### DIFF
--- a/codex/rest/api.nim
+++ b/codex/rest/api.nim
@@ -430,7 +430,7 @@ proc initSalesApi(node: CodexNodeRef, router: var RestRouter) =
 
       if restAv.totalSize == 0:
         return RestApiResponse.error(
-          Http400, "Total size must be larger then zero", headers = headers
+          Http422, "Total size must be larger then zero", headers = headers
         )
 
       if not reservations.hasAvailable(restAv.totalSize.truncate(uint)):
@@ -503,16 +503,19 @@ proc initSalesApi(node: CodexNodeRef, router: var RestRouter) =
         return RestApiResponse.error(Http500, error.msg)
 
       if isSome restAv.freeSize:
-        return RestApiResponse.error(Http400, "Updating freeSize is not allowed")
+        return RestApiResponse.error(Http422, "Updating freeSize is not allowed")
 
       if size =? restAv.totalSize:
         # we don't allow lowering the totalSize bellow currently utilized size
         if size < (availability.totalSize - availability.freeSize):
           return RestApiResponse.error(
-            Http400,
+            Http422,
             "New totalSize must be larger then current totalSize - freeSize, which is currently: " &
               $(availability.totalSize - availability.freeSize),
           )
+
+        if not reservations.hasAvailable(size.truncate(uint)):
+          return RestApiResponse.error(Http422, "Not enough storage quota")
 
         availability.freeSize += size - availability.totalSize
         availability.totalSize = size
--- a/tests/integration/testsales.nim
+++ b/tests/integration/testsales.nim
@@ -1,5 +1,6 @@
 import std/httpclient
 import pkg/codex/contracts
+from pkg/codex/stores/repostore/types import DefaultQuotaBytes
 import ./twonodes
 import ../codex/examples
 import ../contracts/time
@@ -93,7 +94,7 @@ multinodesuite "Sales":
     ).get
     let freeSizeResponse =
       host.patchAvailabilityRaw(availability.id, freeSize = 110000.u256.some)
-    check freeSizeResponse.status == "400 Bad Request"
+    check freeSizeResponse.status == "422 Unprocessable Entity"
     check "not allowed" in freeSizeResponse.body
 
   test "updating availability - updating totalSize", salesConfig:
@@ -143,7 +144,7 @@ multinodesuite "Sales":
     let totalSizeResponse = host.patchAvailabilityRaw(
       availability.id, totalSize = (utilizedSize - 1.u256).some
     )
-    check totalSizeResponse.status == "400 Bad Request"
+    check totalSizeResponse.status == "422 Unprocessable Entity"
     check "totalSize must be larger then current totalSize" in totalSizeResponse.body
 
     host.patchAvailability(availability.id, totalSize = (originalSize + 20000).some)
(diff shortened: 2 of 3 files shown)


tests/integration/testsales.nim
```diff
@@ -142,11 +143,35 @@ multinodesuite "Sales":
...
+      totalSize = 24000000000.uint64,
+      duration = 200.uint64,
+      minPricePerBytePerSecond = 3.u256,
+      totalCollateral = 300.u256,
+    )
+
+    check response.status == "422 Unprocessable Entity"
+    check response.body == "Not enough storage quota"
+
+  test "updating availability above the node quota returns 422", salesConfig:
```
emizzle, line 166: 🟡 Just a suggestion: to get better coverage, maybe we could add a test that checks for 
1. `Updating freeSize is not allowed`
2. `New totalSize must be larger then current totalSize - freeSize, which is currently`
Wdyt?

tests/integration/testsales.nim
```diff
@@ -142,11 +143,35 @@ multinodesuite "Sales":
...
+    check totalSizeResponse.status == "422 Unprocessable Entity"
     check "totalSize must be larger then current totalSize" in totalSizeResponse.body
 
     host.patchAvailability(availability.id, totalSize = (originalSize + 20000).some)
     let newUpdatedAvailability =
       (host.getAvailabilities().get).findItem(availability).get
     check newUpdatedAvailability.totalSize == originalSize + 20000
     check newUpdatedAvailability.freeSize - updatedAvailability.freeSize == 20000
+
+  test "creating availability above the node quota returns 422", salesConfig:
```
emizzle, line 155: 🟡  Just a suggestion: to get better coverage, maybe we could add a test that checks for:
1. `Total size must be larger then zero`
Wdyt?

emizzle (approved):
Well done. Nothing blocking, but just a suggestion to add a bit more coverage. Will approve for now 👍

tests/integration/testsales.nim
2-towns: for 1. There is already a test `updating availability - freeSize is not allowed to be changed`

2-towns: and for 2. as well `updating availability - updating totalSize does not allow bellow utilized`

tests/integration/testsales.nim
2-towns: Done. And I found out that when trying to send a negative value, it is raising a `RangeDefect` which wasn't handled. I guess it could be the case for other endpoints. I will check later on.

codex/rest/api.nim
```diff
@@ -533,17 +538,23 @@ proc initSalesApi(node: CodexNodeRef, router: var RestRouter) =
...
         if size < (availability.totalSize - availability.freeSize):
           return RestApiResponse.error(
-            Http400,
+            Http422,
             "New totalSize must be larger then current totalSize - freeSize, which is currently: " &
               $(availability.totalSize - availability.freeSize),
           )
 
+        if not reservations.hasAvailable(size):
+          return RestApiResponse.error(Http422, "Not enough storage quota")
```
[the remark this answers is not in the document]
2-towns: I think it is `422` because the request is failing due to the user input attempting to update the size. A `507` status code would be used if the error originated internally.

tests/integration/testsales.nim
2-towns: I agree that the requests take a long time to run. However, I'm not a fan of using mocks for HTTP, as it adds significant complexity to reproducing the HTTP environment.
There are some tools to test the APIs, like `HURL`; we could use that for the API. The tests could be run very quickly and are easy to write, but it is a tool outside of the Nim ecosystem.
Another approach would be to move the validation into another layer, like a `service layer`, which would allow us to write unit tests easily.

codex/rest/api.nim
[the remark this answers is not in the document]
AuHau: Well, this is a philosophical question for sure :-P But I could also argue that it is actually an internal error because the node does not have the required space - hence coming from "inside"... But since we have the `/space` endpoint where the user can get the limits for this call, I guess we can leave it `422`...

codex/rest/api.nim
```diff
@@ -484,6 +484,11 @@ proc initSalesApi(node: CodexNodeRef, router: var RestRouter) =
     except CatchableError as exc:
       trace "Excepting processing request", exc = exc.msg
       return RestApiResponse.error(Http500, headers = headers)
+    except RangeDefect as exc:
+      trace "Out of range error", exc = exc.msg
+      return RestApiResponse.error(
+        Http422, "The values provided are out of range", headers = headers
+      )
```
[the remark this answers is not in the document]
2-towns: The error message is only `Unhandled exception in async proc, aborting topics="codex" tid=734662 msg="value out of range`.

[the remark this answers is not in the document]
emizzle: A `RangeDefect` should not be catchable, as it is a `Defect`. I don't think we should try to catch it here, and instead, let it bubble and crash the node.

codex/rest/api.nim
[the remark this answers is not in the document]
emizzle: Would a 507 be more of an OS-level error, and not a Codex-level (application level) error?

tests/integration/testsales.nim
emizzle: Good point Adam. My personal preference is to rather have the tests covered for now than wait for someone to do the mocking 😅  We will have parallel tests (soon ™️) and maybe these can be put them in some kind of extended suite as well. 

Additionally, I would either put these tests in `testrestapi` or create a new integration test, eg `testrestapi_validation`. The more modules, the more we'll benefit from parallelisation. 

These tests can also be sped up by spinning up only a single node in each test:
```diff
-salesConfig
+NodeConfigs(clients: CodexConfigs.init(nodes = 1).some)
```

2-towns: Okay, so I separated the REST API validation tests into a new file testrestapivalidation. This file starts only one node at the beginning of the suite and closes it at the end. Obviously, it speeds up the tests a lot, there are now 14 integration tests running in under 1 second.

codex/rest/api.nim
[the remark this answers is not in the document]
2-towns: Okay I changed the way of testing and it does not raise Defect anymore.

AuHau (approved):
LGTM - great 👍

tests/helpers.nim
```diff
@@ -35,3 +35,14 @@ proc allFuturesThrowing*[T, E]( # https://github.com/nim-lang/Nim/issues/23432
...
+template safeEventually*(expression: untyped, waitingTime = 250, timeout = 5000): bool =
+  proc safeEventually(): Future[bool] {.async.} =
+    let endTime = getTime() + initDuration(milliseconds = timeout)
+    while not expression:
+      if endTime < getTime():
+        return false
+      await sleepAsync(waitingTime)
+    return true
+
+  await safeEventually()
```
gmega, line 48: We probably want to have a PR against asynctest for this. I'll do it now.

PR #1106, opened 2025-02-10
upload OpenAPI artifacts and publish docs only when tagged

marcinczenko:
Currently the API docs are being published on every PR and every push to the master. This results in out of sync API spec, which should only be updated when there is a release.

In this PR, I am deploying the OpenAPI spec only when tagged with a tag matching `v*.*.*`.

> There is a linting problem on master that will prevent this PR from being merged. It is addressed in #1107 to make it visible (I still do not get how it is even possible). Thus once #1107 is merged, I will rebase this one.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -2,17 +2,17 @@ name: OpenAPI
 
 on:
   push:
-    branches:
-      - 'master'
+    tags:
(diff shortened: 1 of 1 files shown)


veaceslavdoina (approved):
Looks good, thanks!

benbierens (approved):

PR #1109, opened 2025-02-11
Fix/repostore deletes for non-orphan blocks

### LINKED ISSUE #1108: Block refcounts underflow when a non-orphan block is deleted, re-inserted, and then deleted again
The following test will cause Codex to crash with an `RangeDefect`:

```nim
  test "should not crash when we insert and then delete a block twice":
    # create a dataset
    let 
      repo = RepoStore.new(repoDs, metaDs, clock = mockClock, quotaMaxBytes = 1000'nb)
      dataset = await makeRandomBlocks(datasetSize = 512, blockSize = 256'nb)
      blk = dataset[0]
      (manifest, tree) = makeManifestAndTree(dataset).tryGet()
      treeCid = tree.rootCid.tryGet()
      proof = tree.getProof(1).tryGet()
    
    (await repo.putBlock(blk)).tryGet()

    # associate block to dataset
    (await repo.putCidAndProof(treeCid, 0, blk.cid, proof)).tryGet()
    
    # delete fails silently because refcount is > 0, but at the same time it
    # decrements the refcount.
    (await repo.delBlock(treeCid, 0.Natural)).tryGet()

    # reinsertion will not increase the refcount, causing the check below to fail.
    (await repo.putBlock(blk)).tryGet()

    # if you try to delete the block, you'll get an underflow
    (await repo.delBlock(treeCid, 0.Natural)).tryGet()
```

This happens because we decrement the ref counter but don't clear the leaf metadata. When the block gets reinserted, this causes us not to initialize the ref count for the reinserted block properly, causing the crash on the second delete attempt.

gmega:
Fixes #1108

This PR:
* fixes the broken deletes for non-orphan blocks (i.e., blocks that are part of a dataset);
* makes `deleteBlock` more strict by actually returning an error on illegal deletes instead of failing silently;
* adds additional tests to cover the peculiar aspects of deletes for non-orphan blocks.

### DIFF
--- a/codex/stores/repostore/store.nim
+++ b/codex/stores/repostore/store.nim
@@ -221,6 +221,10 @@ method delBlock*(
     else:
       return failure(err)
 
+  if err =? (await self.delLeafMetadata(treeCid, index)).errorOption:
+    error "Failed to delete leaf metadata, block will remain on disk.", err = err.msg
+    return failure(err)
+
   if err =?
       (await self.updateBlockMetadata(leafMd.blkCid, minusRefCount = 1)).errorOption:
     if not (err of BlockNotFoundError):
@@ -294,6 +298,18 @@ method listBlocks*(
 proc createBlockExpirationQuery(maxNumber: int, offset: int): ?!Query =
   let queryKey = ?createBlockExpirationMetadataQueryKey()
   success Query.init(queryKey, offset = offset, limit = maxNumber)
+
+proc blockRefCount*(self: RepoStore, cid: Cid): Future[?!Natural] {.async.} =
+  ## Returns the reference count for a block. If the count is zero;
+  ## this means the block is eligible for garbage collection.
+  ##
+  without key =? createBlockExpirationMetadataKey(cid), err:
+    return failure(err)
+
+  without md =? await get[BlockMetadata](self.metaDs, key), err:
+    return failure(err)
+
+  return success(md.refCount)
 
 method getBlockExpirations*(
     self: RepoStore, maxNumber: int, offset: int
--- a/codex/stores/repostore/store.nim
+++ b/codex/stores/repostore/store.nim
@@ -186,12 +186,12 @@ method putBlock*(
 
   return success()
 
-method delBlock*(self: RepoStore, cid: Cid): Future[?!void] {.async.} =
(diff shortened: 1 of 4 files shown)


codex/stores/repostore/store.nim
```diff
@@ -205,12 +205,28 @@ method delBlock*(self: RepoStore, cid: Cid): Future[?!void] {.async.} =
...
+  ##
+
+  logScope:
+    cid = cid
+
+  without outcome =? await self.delBlockInternal(cid), err:
+    return failure(err)
+
+  case outcome
+  of InUse:
```
dryajov, line 222: why not return the actual error here, instead of just a string?

gmega: Can I return anything other than a string with `Future[?!void]`?

gmega: Oh, if you mean having a custom exception type with the error code - I've considered that, but didn't seem worth the trouble. We're not handling these errors on the client side, and it felt like adding this was just increasing complexity for no good reason.

gmega: Not all of the outcomes are errors, and the method returns `Future[?!void]`, so I don't have a lot of choices. I could create a custom exception type to describe the single case in which this is a failure, but didn't seem worth it, particularly since we're not handling these errors on the client side other than logging the error message anyway.

dryajov (approved):

codex/stores/repostore/store.nim
dryajov: Ok, I think we can rework this later on.

gmega:
@munna0908 ping :-)

munna0908 (approved):

munna0908:
@gmega LGTM 🚀

tests/codex/examples.nim
```diff
@@ -37,8 +37,8 @@ proc example*(_: type SignedState): SignedState =
 proc example*(_: type Pricing): Pricing =
   Pricing(address: EthAddress.example, price: uint32.rand.u256)
 
-proc example*(_: type bt.Block): bt.Block =
-  let length = rand(4096)
+proc example*(_: type bt.Block, size: int = 4096): bt.Block =
+  let length = rand(size)
```
gmega, line 41: Oh shit... what had I smoked here....

PR #1113, opened 2025-02-13
fix: bumps CI timeouts to 3h

gmega:
We're often seeing cancelled jobs:

image

I suspect those are due to timeouts. I'm bumping this to 3h to see if it fixes the problem.


gmega:
Nevermind, I don't think this is needed.

PR #1115, opened 2025-02-14
fix(statemachine): do not raise from state.run

emizzle:
Prevents the state machine's run function from raising exceptions, ensuring there are no exceptions leaking out, and that each state can handle exceptions on an individual basis, instead of handling them all the same at the statemachine level.


codex/purchasing/states/submitted.nim
```diff
@@ -44,5 +47,10 @@ method run*(state: PurchaseSubmitted, machine: Machine): Future[?State] {.async.
     await wait().withTimeout()
   except Timeout:
     return some State(PurchaseCancelled())
+  except CancelledError as e:
+    trace "PurchaseStarted.run was cancelled", error = e.msgDetail
```
markspanbroek, line 51:
```suggestion
    trace "PurchaseSubmitted.run was cancelled", error = e.msgDetail
```

Same below

codex/sales/salesagent.nim
```diff
@@ -68,41 +69,48 @@ proc subscribeCancellation(agent: SalesAgent) {.async.} =
...
+      let market = agent.context.market
+      let expiry = await market.requestExpiresAt(data.requestId)
+
+      while true:
+        let deadline = max(clock.now, expiry) + 1
+        trace "Waiting for request to be cancelled", now = clock.now, expiry = deadline
+        await clock.waitUntil(deadline)
+
+        without state =? await agent.retrieveRequestState():
+          error "Uknown request", requestId = data.requestId
```
markspanbroek, line 86:
```suggestion
          error "Unknown request", requestId = data.requestId
```

codex/sales/states/proving.nim
```diff
@@ -129,27 +130,35 @@ method run*(state: SaleProving, machine: Machine): Future[?State] {.async.} =
...
+          try:
+            await state.loop.cancelAndWait()
+          except CatchableError as e:
+            error "Error during cancellation of proving loop", msg = e.msg
+
+        state.loop = nil
+
+    return some State(SalePayout())
+  except CancelledError as e:
+    trace "SalePreparing.run onCleanUp was cancelled", error = e.msgDetail
```
markspanbroek, line 161:
```suggestion
    trace "SaleProving.run onCleanup was cancelled", error = e.msgDetail
```

Same below

codex/sales/states/preparing.nim
```diff
@@ -37,62 +37,70 @@ method onSlotFilled*(
...
+        return some State(SaleIgnored(reprocessSlot: true))
+
+      return some State(SaleErrored(error: error))
+
+    trace "Reservation created successfully"
+
+    data.reservation = some reservation
+    return some State(SaleSlotReserving())
+  except CancelledError as e:
+    trace "SalePreparing.run onCleanUp was cancelled", error = e.msgDetail
```
markspanbroek, line 103:
```suggestion
    trace "SalePreparing.run was cancelled", error = e.msgDetail
```

codex/sales/states/slotreserving.nim
```diff
@@ -36,23 +38,29 @@ method run*(state: SaleSlotReserving, machine: Machine): Future[?State] {.async.
...
+
+      trace "Slot successfully reserved"
+      return some State(SaleDownloading())
+    else:
+      # do not re-add this slot to the queue, and return bytes from Reservation to
+      # the Availability
+      debug "Slot cannot be reserved, ignoring"
+      return some State(SaleIgnored(reprocessSlot: false, returnBytes: true))
+  except CancelledError as e:
+    trace "SaleSlotReserving.run onCleanUp was cancelled", error = e.msgDetail
```
markspanbroek, line 63:
```suggestion
    trace "SaleSlotReserving.run was cancelled", error = e.msgDetail
```

codex/sales/states/unknown.nim
```diff
@@ -26,34 +27,42 @@ method onCancelled*(state: SaleUnknown, request: StorageRequest): ?State =
...
+      return some State(SaleFailed())
+    of SlotState.Cancelled:
+      return some State(SaleCancelled())
+    of SlotState.Repair:
+      let error = newException(
+        SlotFreedError, "Slot was forcible freed and host was removed from its hosting"
+      )
+      return some State(SaleErrored(error: error))
+  except CancelledError as e:
+    trace "SaleUnknown.run onCleanUp was cancelled", error = e.msgDetail
```
markspanbroek, line 65:
```suggestion
    trace "SaleUnknown.run was cancelled", error = e.msgDetail
```

markspanbroek (approved):
Nice, thanks @emizzle! It is indeed much nicer to let states handle their own errors.

codex/purchasing/states/started.nim
```diff
@@ -28,15 +31,24 @@ method run*(state: PurchaseStarted, machine: Machine): Future[?State] {.async.}
...
+      ended.cancelSoon()
+      return some State(PurchaseFailed())
+    else:
+      failed.cancelSoon()
+      return some State(PurchaseFinished())
+  except CancelledError as e:
     ended.cancelSoon()
-    return some State(PurchaseFailed())
-  else:
     failed.cancelSoon()
```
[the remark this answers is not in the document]
emizzle: It's a good point and I honestly don't know the answer. My initial assumption was that once the reference to the run proc was GC'd (during an exception), that the futures (`ended` and `failed`) would be GC'd, but i spent a bit of time looking at the values in the heap (using `-d:nimTypeNames` and `dumpNumberOfInstances()`) and honestly I can't tell if those futures are GC'd or not. So I've asked Eugene: https://discord.com/channels/613988663034118151/616299964242460682/1341328004923658251

codex/utils/asyncstatemachine.nim
```diff
@@ -46,24 +47,14 @@ proc schedule*(machine: Machine, event: Event) =
   except AsyncQueueFullError:
     raiseAssert "unlimited queue is full?!"
 
-method run*(state: State, machine: Machine): Future[?State] {.base, async.} =
+method run*(
+    state: State, machine: Machine
+): Future[?State] {.base, async: (raises: []).} =
```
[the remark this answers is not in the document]
emizzle: It's there, but it's not obvious. On line 7 above, we do `{.push raises: [].}`
which applies an implicit `{. raises: [].}` pragma on each routine.

emizzle:
> Did you encounter some problem related to this? Not fan of the verbosity, but I guess overall is an improvement 😅

Yes, overall, I have come back to fixing this at least 3 times because I've spent so much time debugging it. When there's CancelledErrors leaking out, it because very difficult to reason about, especially when they're propagating the stack invisibly. Changing to `async: (raises: [])` provides a bit of debugging relief, and allows the log output to be much more easily followed.

> Just to double-check, I see lots of swallowing of CancelledErrors - that is because the states runs are asyncSpawned right?

Yes, that's correct. If they are not swallowed, the `asyncSpawn` causes a Defect

> An idea - maybe as to elevate some of the added verbosity, we could allow CancelledError in the raises and it could be left to propagate to the Machine.scheduler() where it is already now discarded anyhow?

Yes, that's a good idea, which as you said is how we have it now, but in my mind, it's easier to reason about `run` not raising anything at all. When `CancelledErrors` are raised, debugging becomes a pain in the ass (see above).

codex/sales/states/proving.nim
emizzle: Yikes, sorry mark, I missed this one. IMO, `onCleanup` should be removed entirely from that trace log. Can do in a future PR.

PR #1119, opened 2025-02-19
Add basic retry functionality

dryajov:


codex/blockexchange/engine/engine.nim
```diff
@@ -100,124 +93,147 @@ type
...
+
+      trace "Peers for block"
+      if peers.with.len > 0:
+        self.pendingBlocks.setInFlight(address, true)
+        await self.sendWantBlock(@[address], peers.with.randomPeer)
+      else:
+        self.pendingBlocks.setInFlight(address, false)
+        if peers.without.len > 0:
+          await self.sendWantHave(@[address], peers.without)
+        self.discovery.queueFindBlocksReq(@[address.cidOrTreeCid])
```
marcinczenko, line 190: So the important change here is that previously we were sending `WantHaves` even if we had a peer having the request block (`BlockAddress`). Now we are calling `sendWantHave` **only** if there are no peers with the requested address? It makes sense, but I would like to understand why previously it was decided otherwise...

Another thing I would like to confirm, is how the `inFlight` status is used. In this new version, the `inFlight` status of the given blocks seems to be used only to regulate when `WantHave` will be sent in `blockPresenceHandler`:

```nim
let ourWantCids = ourWantList.filter do(address: BlockAddress) -> bool:
  if address in peerHave and not self.pendingBlocks.retriesExhausted(address) and
      not self.pendingBlocks.isInFlight(address):
    self.pendingBlocks.setInFlight(address, true)
    self.pendingBlocks.decRetries(address)
    true
  else:
    false
  
if ourWantCids.len > 0:
  trace "Peer has blocks in our wantList", peer, wants = ourWantCids
  await self.sendWantBlock(ourWantCids, peerCtx)
```

With the new implementation we have in `downloadInternal`, while in the loop, we are continuously observing the peers context via `peers.with`. Thus, if we receive a presence message, wouldn't it be enough to just update the peer context in the `blockPresenceHandler`, and let the processing in `downloadInternal` to pick it up? And if we are afraid that we miss it because we are on the edge with the "retries" counter, we could just increase it here in the `blockPresenceHandler`. With all this, we could even get rid of the `inFlight` completely. Just a thought...

marcinczenko (commented):
I have started the review, and I will be working on by adding more comments today and over next days if possible (will try to look up a couple of things over the weekend as well). No need to wait for me, if someone is faster. My objective is to have some observations confirmed and some questions answered. Because I have documented the whole flow recently, I just want to take advantage of this more substantial change to improve the docs.

codex/blockexchange/engine/engine.nim
```diff
@@ -228,82 +244,99 @@ proc blockPresenceHandler*(
...
-    trace "Peer has blocks in our wantList", peer, wants = wantCids
-    await b.sendWantBlock(wantCids, peerCtx)
+  let ourWantCids = ourWantList.filter do(address: BlockAddress) -> bool:
+    if address in peerHave and not self.pendingBlocks.retriesExhausted(address) and
+        not self.pendingBlocks.isInFlight(address):
+      self.pendingBlocks.setInFlight(address, true)
+      self.pendingBlocks.decRetries(address)
+      true
+    else:
+      false
```
marcinczenko, line 259: Following the reasoning I have outlined in https://github.com/codex-storage/nim-codex/pull/1119#discussion_r1965496590, perhaps we could even remove that part?

codex/blockexchange/engine/engine.nim
```diff
@@ -228,82 +244,99 @@ proc blockPresenceHandler*(
...
 
-  # if none of the connected peers report our wants in their have list,
-  # fire up discovery
-  b.discovery.queueFindBlocksReq(
-    toSeq(b.pendingBlocks.wantListCids).filter do(cid: Cid) -> bool:
-      not b.peers.anyIt(cid in it.peerHaveCids)
-  )
+  if ourWantCids.len > 0:
+    trace "Peer has blocks in our wantList", peer, wants = ourWantCids
+    await self.sendWantBlock(ourWantCids, peerCtx)
```
marcinczenko, line 263: This also could be removed if what I am saying above and in https://github.com/codex-storage/nim-codex/pull/1119#discussion_r1965496590 makes sense...

codex/blockexchange/engine/engine.nim
```diff
@@ -365,18 +398,22 @@ proc blocksDeliveryHandler*(
...
   codex_block_exchange_blocks_received.inc(validatedBlocksDelivery.len.int64)
 
-  let peerCtx = b.peers.get(peer)
+  let peerCtx = self.peers.get(peer)
 
   if peerCtx != nil:
-    await b.payForBlocks(peerCtx, blocksDelivery)
+    await self.payForBlocks(peerCtx, blocksDelivery)
     ## shouldn't we remove them from the want-list instead of this:
     peerCtx.cleanPresence(blocksDelivery.mapIt(it.address))
```
marcinczenko, line 409: Just above the highlighted lines we are calling `resolveBlocks`. `resolveBlocks` will call `cancelBlocks` which should take care for removing the relevant presence entries.

codex/blockexchange/network/network.nim
```diff
@@ -21,17 +21,18 @@ import ../../blocktype as bt
...
-export network, payments
+export networkpeer, payments
 
 logScope:
   topics = "codex blockexcnetwork"
 
 const
   Codec* = "/codex/blockexc/1.0.0"
-  MaxInflight* = 100
+  DefaultMaxInflight* = 100
```
gmega, line 35: For reference, Deluge allows 1500 pending block requests **per peer** in its [most aggressive profiles](https://github.com/arvidn/libtorrent/blob/e5fe95f92273bd775961e2f1f3b4ec45f8012d9e/src/session.cpp#L172).

codex/blockexchange/engine/engine.nim
dryajov: > Now we are calling sendWantHave only if there are no peers with the requested address? It makes sense, but I would like to understand why previously it was decided otherwise...

One of the objective was to avoid sending too many `WantHave` messages, because this incurs in significant cost for the remote node, this is why we're now only sending this messages when `with == 0`.

> With the new implementation we have in downloadInternal, while in the loop, we are continuously observing the peers context via peers.with. Thus, if we receive a presence message, wouldn't it be enough to just update the peer context in the blockPresenceHandler, and let the processing in downloadInternal to pick it up? And if we are afraid that we miss it because we are on the edge with the "retries" counter, we could just increase it here in the blockPresenceHandler. With all this, we could even get rid of the inFlight completely. Just a thought...

The `InFlight` flag is used to indicate there is already a `WantBlock` request in progress. The reason we do not want to let this being handled by the retry loop, is because it increases the latency quite significantly as opposed to sending the request right away, as we're doing it in this case. This is mostly an optimization.

codex/blockexchange/network/network.nim
```diff
@@ -282,48 +290,61 @@ proc dialPeer*(b: BlockExcNetwork, peer: PeerRecord) {.async.} =
...
-method init*(b: BlockExcNetwork) =
+method init*(self: BlockExcNetwork) =
   ## Perform protocol initialization
   ##
 
   proc peerEventHandler(peerId: PeerId, event: PeerEvent) {.async.} =
     if event.kind == PeerEventKind.Joined:
-      b.setupPeer(peerId)
+      self.setupPeer(peerId)
     else:
```
gmega, line 313: Hm... I see there are three `PeerEventKind`s:

```nim
  PeerEventKind* {.pure.} = enum
    Left
    Joined
    Identified
```

as it is now, if we get `Indentified`, we'll drop the peer. I'm assuming you're aware of this but pointing it out anyhow.

codex/conf.nim
```diff
@@ -251,15 +251,15 @@ type
       desc:
         "Time interval in seconds - determines frequency of block " &
         "maintenance cycle: how often blocks are checked " & "for expiration and cleanup",
-      defaultValue: DefaultBlockMaintenanceInterval,
-      defaultValueDesc: $DefaultBlockMaintenanceInterval,
+      defaultValue: DefaultBlockInterval,
```
gmega, line 254: Not sure this is improving what this constant contains...

codex/node.nim
```diff
@@ -235,19 +244,29 @@ proc streamEntireDataset(
...
+      if err =? (
+        await self.fetchBatched(manifest, DefaultFetchBatch, fetchLocal = false)
+      ).errorOption:
+        error "Unable to fetch blocks", err = err.msg
+    except CatchableError as exc:
+      error "Error fetching blocks", exc = exc.msg
 
-    if err =? (await erasureJob()).errorOption:
-      return failure(err)
+  self.trackedFutures.track(prefetch())
```
gmega, line 269: OK I see why you're doing this, but now `streamEntireDataset` will effectively continue downloading even if the client stops streaming. Anyhow, I guess we're gonna drop streaming soon enough anyway.

codex/rng.nim
```diff
@@ -55,6 +55,15 @@ proc sample*[T](
...
     break
 
+proc sample*[T](
+    rng: Rng, sample: openArray[T], limit: int
+): seq[T] {.raises: [Defect, RngSampleError].} =
+  if limit > sample.len:
+    raise newException(RngSampleError, "Limit cannot be larger than sample!")
+
+  for _ in 0 ..< min(sample.len, limit):
+    result.add(rng.sample(sample, result))
```
gmega, line 65: These sample procedures aren't great, I'm going to do a PR on `rng.nim`.

codex/blockexchange/engine/discovery.nim
```diff
@@ -143,8 +143,6 @@ proc start*(b: DiscoveryEngine) {.async.} =
     asyncSpawn fut
 
   b.discoveryLoop = b.discoveryQueueLoop()
-  b.trackedFutures.track(b.discoveryLoop)
-  asyncSpawn b.discoveryLoop
```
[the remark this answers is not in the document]
dryajov: No, not really - this doesn't affect stopping since we call cancel the tracked futures on stop. The `asyncSpawn` call tries to do some logging and re-throw any future error as a Defect, but that was really a workaround because we lacked checked exceptions on async code. I do intend to make tracked futures only accept `Future.Raises([])` for that matter.

codex/blockexchange/engine/engine.nim
```diff
@@ -100,124 +93,147 @@ type
...
-    b: BlockExcEngine, handle: Future[Block], address: BlockAddress, peerId: PeerId
-) {.async.} =
-  try:
-    discard await handle
-  except CancelledError as exc:
-    trace "Block handle cancelled", address, peerId
-  except CatchableError as exc:
-    warn "Error block handle, disconnecting peer", address, exc = exc.msg, peerId
+proc randomPeer(peers: seq[BlockExcPeerCtx]): BlockExcPeerCtx =
+  Rng.instance.sample(peers)
```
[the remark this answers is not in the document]
dryajov: yeah, that was quite annoying when I realized what was going on.

codex/blockexchange/engine/engine.nim
```diff
@@ -100,124 +93,147 @@ type
...
+        peersWithout = peers.without.len
+
+      trace "Peers for block"
+      if peers.with.len > 0:
+        self.pendingBlocks.setInFlight(address, true)
+        await self.sendWantBlock(@[address], peers.with.randomPeer)
+      else:
+        self.pendingBlocks.setInFlight(address, false)
+        if peers.without.len > 0:
+          await self.sendWantHave(@[address], peers.without)
```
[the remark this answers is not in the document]
dryajov: Yes, and we will constrain this to a subset of peers, I actually wanted to randomize this, but it ended up affecting download perf more than I expected, so I left it as this for now, but I do intend to fine tune this.

codex/blockexchange/engine/engine.nim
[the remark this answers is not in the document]
dryajov: If a lookup is already in progress it will not schedule it, so it's already half way there. Otherwise I agree, but this is left for another pass.

codex/blockexchange/engine/engine.nim
```diff
@@ -100,124 +93,147 @@ type
...
+        self.discovery.queueFindBlocksReq(@[address.cidOrTreeCid])
 
-  if not b.pendingBlocks.isInFlight(address):
-    let peers = b.peers.getPeersForBlock(address)
+      await (handle or sleepAsync(self.pendingBlocks.retryInterval))
```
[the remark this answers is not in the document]
dryajov: The timeout is mostly to allow time for the requested block to arrive, otherwise we'll just be spamming nodes with `WantHave/WantBlock` requests (which will lead to duplicate blocks).

codex/blockexchange/engine/engine.nim
```diff
@@ -100,124 +93,147 @@ type
...
-  let blockFuture = b.pendingBlocks.getWantHandle(address, b.blockFetchTimeout)
+  let handle = self.pendingBlocks.getWantHandle(address)
+  trace "Downloading block"
+  try:
+    while address in self.pendingBlocks:
+      logScope:
+        retries = self.pendingBlocks.retries(address)
+        interval = self.pendingBlocks.retryInterval
+
+      if self.pendingBlocks.retriesExhausted(address):
```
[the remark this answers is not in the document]
dryajov: Yeah, I've been thinking about this as well. Not sure, but this makes things easier to test for now at least.

codex/blockexchange/engine/engine.nim
```diff
@@ -228,82 +244,99 @@ proc blockPresenceHandler*(
...
-  if payment =? engine.wallet.pay(peer, price):
+  if payment =? self.wallet.pay(peer, price):
     trace "Sending payment for blocks", price, len = blocksDelivery.len
     await sendPayment(peer.id, payment)
 
-proc validateBlockDelivery(b: BlockExcEngine, bd: BlockDelivery): ?!void =
-  if bd.address notin b.pendingBlocks:
+proc validateBlockDelivery(self: BlockExcEngine, bd: BlockDelivery): ?!void =
+  if bd.address notin self.pendingBlocks:
     return failure("Received block is not currently a pending block")
```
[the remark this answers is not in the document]
dryajov: Not sure, a node can spam you with unwanted blocks otherwise. I lean towards erroring now and adding smarter controls later on. For example, if a peer keeps sending unwanted blocks it should probably be throttled/disconnected.

codex/blockexchange/engine/engine.nim
```diff
@@ -343,11 +376,11 @@ proc blocksDeliveryHandler*(
       peer = peer
       address = bd.address
 
-    if err =? b.validateBlockDelivery(bd).errorOption:
+    if err =? self.validateBlockDelivery(bd).errorOption:
       warn "Block validation failed", msg = err.msg
```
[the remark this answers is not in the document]
dryajov: Really? I haven't seen this message in my runs?

codex/blockexchange/engine/engine.nim
```diff
@@ -413,7 +455,6 @@ proc wantListHandler*(b: BlockExcEngine, peer: PeerId, wantList: WantList) {.asy
                 address: e.address, `type`: BlockPresenceType.DontHave, price: price
               )
             )
-          peerCtx.peerWants.add(e)
```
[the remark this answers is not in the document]
dryajov: yep, this is what caused all the fuzz.

codex/blockexchange/engine/pendingblocks.nim
```diff
@@ -34,66 +32,76 @@ declareGauge(
...
+
+    proc cleanUpBlock(data: pointer) {.raises: [].} =
+      self.blocks.del(address)
+      self.updatePendingBlockGauge()
+
+    handle.addCallback(cleanUpBlock)
+    handle.cancelCallback = proc(data: pointer) {.raises: [].} =
+      if not handle.finished:
+        handle.removeCallback(cleanUpBlock)
+        cleanUpBlock(nil)
```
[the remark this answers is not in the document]
dryajov: ah, good catch - it should cleanup the block in either case.

codex/blockexchange/network/network.nim
dryajov: Good to know, we can definitely adjust.

codex/blockexchange/network/network.nim
dryajov: Oh, not really, thats a new event 🤔. Good catch!

codex/conf.nim
dryajov: The `SuperLongSelfDescriptingValueIdentifierSuperConstant` is quite annoying tho 😁. Matter of taste, but I can live with it if we find it useful.

codex/node.nim
dryajov: Not sure why you think that streaming is going to be dropped? Otherwise, good observation.

codex/blockexchange/engine/discovery.nim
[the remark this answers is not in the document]
gmega: Wait but you're not adding the discoveryLoop future to the trackedFutures.

codex/blockexchange/engine/engine.nim
marcinczenko: make sense, thanks

codex/blockexchange/engine/discovery.nim
```diff
@@ -143,8 +143,6 @@ proc start*(b: DiscoveryEngine) {.async.} =
...
 
 proc stop*(b: DiscoveryEngine) {.async.} =
```
marcinczenko, line 147: Since you are changing `b` => `self` all over the places, maybe you want to do it here as well?

codex/blockexchange/engine/engine.nim
```diff
@@ -100,124 +93,147 @@ type
...
+  self.blockexcRunning = false
 
   trace "NetworkStore stopped"
 
 proc sendWantHave(
-    b: BlockExcEngine, addresses: seq[BlockAddress], peers: seq[BlockExcPeerCtx]
+    self: BlockExcEngine, addresses: seq[BlockAddress], peers: seq[BlockExcPeerCtx]
 ): Future[void] {.async.} =
   for p in peers:
     let toAsk = addresses.filterIt(it notin p.peerHave)
```
marcinczenko, line 140: I know you did not touch that line, but it would be cleaner to have `sendWantHave` not to do any filtering (I mean, this should happen elsewhere, where it is really relevant). Especially that `sendWantBlock` is nice and clean without side-effected (the caller expects the request to be just sent)... Maybe a small separate PR...

codex/blockexchange/engine/engine.nim
```diff
@@ -100,124 +93,147 @@ type
...
+      handle.fail(exc)
+  finally:
+    self.pendingBlocks.setInFlight(address, false)
 
-    await b.sendWantHave(@[address], peers.without)
+proc requestBlock*(
+    self: BlockExcEngine, address: BlockAddress
+): Future[?!Block] {.async: (raises: [CancelledError]).} =
+  if address notin self.pendingBlocks:
+    self.trackedFutures.track(self.downloadInternal(address))
```
marcinczenko, line 213: So, what are we really tracking here in:

```nim
self.trackedFutures.track(self.downloadInternal(address))
```

 Which future will the `downloadInternal` return? If I read it well it will be the result of `await (handle or sleepAsync(self.pendingBlocks.retryInterval))`, which returns a new future which will complete once either `handle` or `sleepAsync` complete, which could well be `sleepAsync`. I think we always want to track `handle`.

codex/blockexchange/engine/engine.nim
```diff
@@ -228,82 +244,99 @@ proc blockPresenceHandler*(
...
-  let wantCids = wantList.filterIt(it in peerHave)
-
-  if wantCids.len > 0:
-    trace "Peer has blocks in our wantList", peer, wants = wantCids
-    await b.sendWantBlock(wantCids, peerCtx)
+  let ourWantCids = ourWantList.filter do(address: BlockAddress) -> bool:
+    if address in peerHave and not self.pendingBlocks.retriesExhausted(address) and
+        not self.pendingBlocks.isInFlight(address):
+      self.pendingBlocks.setInFlight(address, true)
+      self.pendingBlocks.decRetries(address)
```
marcinczenko, line 256: It is convenient to do it here, but I never like having side effects in `filter`... it also feels a bit out of context here when you read the code. We could add `setInFlight` that takes list of addresses. The same could be done for `self.pendingBlocks.decRetries`. We could also consider if  `self.pendingBlocks.setInFlight(address, true)` should be part of `sendWantBlock` (and then `self.self.pendingBlocks.setInFlight(address, false)` to be inside `sendWantHave` (but yeah, also said in other comment earlier that I like those two to be as focused as possible (still as long as they do not filter our blocks I am kind of fine with it).

Of course it is a bit faster to do it here, but , that could be a bit premature optimization....

codex/blockexchange/engine/engine.nim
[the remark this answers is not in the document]
marcinczenko: Shouldn't it rather be:

```nim
if peers.without.len > 0:
  await self.sendWantHave(@[address], peers.without)
else:
  self.discovery.queueFindBlocksReq(@[address.cidOrTreeCid])
```

and even then, is it good that we hit discovery on every iteration (at least before we discover some peers)?

codex/blockexchange/engine/engine.nim
```diff
@@ -228,82 +244,99 @@ proc blockPresenceHandler*(
...
-      b.peers.mapIt(
-        b.network.request.sendWantCancellations(peer = it.id, addresses = addrs)
+    if blocks.len > 0:
+      trace "Sending block request cancellations to peer", peer = peerCtx.id, blocks
+      await self.network.request.sendWantCancellations(
+        peer = peerCtx.id, addresses = blocks
       )
-    )
-  ).filterIt(it.failed)
+      peerCtx.cleanPresence(addrs)
```
marcinczenko, line 301: And not `peerCtx.cleanPresence(blocks)`?

codex/blockexchange/engine/engine.nim
```diff
@@ -395,9 +432,14 @@ proc wantListHandler*(b: BlockExcEngine, peer: PeerId, wantList: WantList) {.asy
 
     if idx < 0: # Adding new entry to peer wants
       let
-        have = await e.address in b.localStore
-        price = @(b.pricing.get(Pricing(price: 0.u256)).price.toBytesBE)
+        have = await e.address in self.localStore
+        price = @(self.pricing.get(Pricing(price: 0.u256)).price.toBytesBE)
+
+      if e.cancel:
+        trace "Received cancelation for untracked block, skipping", address = e.address
```
marcinczenko, line 439:
```suggestion
        trace "Received cancellation for untracked block, skipping", address = e.address
```

codex/blockexchange/engine/engine.nim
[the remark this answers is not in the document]
marcinczenko: It is indeed a bit confusing: `peerCtx.peerWants` is a list of blocks for which the remote peer explicitly asked, and not the blocks that remote peer would like to have.... Perhaps a better naming for `peerWants` would make it easier to understand.

codex/blockexchange/engine/engine.nim
```diff
@@ -100,124 +93,147 @@ type
...
+proc randomPeer(peers: seq[BlockExcPeerCtx]): BlockExcPeerCtx =
+  Rng.instance.sample(peers)
 
-    # TODO: really, this is just a quick and dirty way of
-    # preventing hitting the same "bad" peer every time, however,
-    # we might as well discover this on or next iteration, so
-    # it doesn't mean that we're never talking to this peer again.
-    # TODO: we need a lot more work around peer selection and
-    # prioritization
+proc downloadInternal(
```
marcinczenko, line 157: Just a though. I am wondering if `downloadInternal` does not better belong to `pendingblocks`. It would perhaps drag with it other things that should be there. For example `RetriesExhaustedError` is defined in `pendingblocks`, yet never used there but here.

codex/blockexchange/engine/engine.nim
marcinczenko: Also notice that having `let fut = await (fut1 or fut2)`, if `fut` is cancelled, `fut1` and `fut2` futures WILL NOT BE cancelled.

marcinczenko: well, in my reasoning, I ignored the fact that `downloadInternal` has a `while` loop. Thus, `downloadInternal` will return another Future I guess, which we are effectively tracking, but it still looks to me that `handle` will not be cancelled because of the "or" thing - I tried to test that with some examples, and it looks like indeed this is the case.

codex/blockexchange/engine/engine.nim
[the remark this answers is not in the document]
gmega: Fair enough.

codex/blockexchange/engine/engine.nim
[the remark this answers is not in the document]
marcinczenko: In BitTorrent peers use `have` (equivalent of our `presence`) to say they have a block. But then a peer seems to never say which blocks it wants without wanting to download them immediately (so, our `WantBlock`). On the very first message the downloaders send `bitfield` message to indicate which blocks they already have. Downloaders which don't have anything yet may skip the 'bitfield' message. And then they have `interested` `not interested` to indicate if they have interest in receiving blocks (any blocks, those messages do not have payload, so they do not tell which blocks you are or are no longer interested in). Interest state must be kept up to date at all times - whenever a downloader doesn't have something they currently would ask a peer for in `unchoked`, they must express lack of interest, despite being `choked`.

codex/blockexchange/engine/engine.nim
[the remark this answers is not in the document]
gmega: OK I see: we'll end up degenerating into a $k$-multicast where $k$ is the number of retries. I feel there might be a more efficient way do this if we allow a larger number of outstanding block requests instead of doing short retries per-block. 

If we have a large enough outstanding request set, we might even be able to get away with just having the actual block request timeout as the retry interval; i.e. instead of requesting the same block to several peers and getting duplicates, request many blocks to many different peers and get useful stuff.

codex/blockexchange/engine/engine.nim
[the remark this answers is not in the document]
gmega: OK maybe I'm recalling the the buggy version we have in master which generates duplicates like crazy...

benbierens:
Commit `18b5d6c` passes release tests.

I took some time to run a few reliability tests on this and compare it to the current master image.
This PR seems to significantly improve reliability, reducing the retries at the API level to zero. \o/
For performance I got some conflicting numbers. The download speed seems to be better than master for 'small' setups. (10 nodes or less) but slower than master for anything 20 nodes or larger.

It would still probably be good to have this included in the upcoming release.

gmega:
> Commit `18b5d6c` passes release tests.
> 
> I took some time to run a few reliability tests on this and compare it to the current master image. This PR seems to significantly improve reliability, reducing the retries at the API level to zero. \o/ For performance I got some conflicting numbers. The download speed seems to be better than master for 'small' setups. (10 nodes or less) but slower than master for anything 20 nodes or larger.
> 
> It would still probably be good to have this included in the upcoming release.

I saw some pretty dramatic changes (10x improvements) in my [preliminary experiments](https://rpubs.com/giuliano_mega/codex-exploratory) with a network of size 10, 4 seeders. Granted, I haven't even tried master for larger swarms cause I didn't think experiments would complete this century...

Can share which settings you used for your experiments?

codex/blockexchange/engine/engine.nim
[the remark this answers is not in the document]
marcinczenko: oh, that answered already "If a lookup is already in progress it will not schedule it, so it's already half way there. Otherwise I agree, but this is left for another pass."

codex/blockexchange/engine/engine.nim
```diff
@@ -228,82 +244,99 @@ proc blockPresenceHandler*(
...
+  if self.peers.len == 0:
+    return
+
   trace "Sending block request cancellations to peers",
-    addrs, peers = b.peers.mapIt($it.id)
+    addrs, peers = self.peers.peerIds
+
+  proc mapPeers(peerCtx: BlockExcPeerCtx): Future[BlockExcPeerCtx] {.async.} =
+    let blocks = addrs.filter do(a: BlockAddress) -> bool:
+      a in peerCtx.blocks
```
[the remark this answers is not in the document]
marcinczenko: Suggestion: we have `contains` defined for `BlockExcPeerCtx`. Perhaps better to just use `a in peerCtx` (`peerCtx.blocks` could be considered implementation detail and maybe should be private).

codex/blockexchange/engine/engine.nim
```diff
@@ -514,14 +558,14 @@ proc taskHandler*(b: BlockExcEngine, task: BlockExcPeerCtx) {.gcsafe, async.} =
```
marcinczenko, line 558: Above where we compute `wantBlocks` it is done like this:

```nim
var wantsBlocks =
    task.peerWants.filterIt(it.wantType == WantType.WantBlock and not it.inFlight)
```

I think you have just made sure that `peerWants` only keeps `WantBlock` entries (you removed `peerCtx.peerWants.add(e)` from `WantHave`) in `wantListHandler` above, so, we should not need to filter it here anymore on `WantType.WantBlock` - this would be misleading even I guess.

Like this:

```nim
var wantsBlocks =
    task.peerWants.filterIt(not it.inFlight)
```

benbierens:
Images used:
            "codexstorage/nim-codex:sha-f6aee4f-dist-tests", // master
            "codexstorage/nim-codex:sha-18b5d6c-dist-tests" // pr
Plan: 1 uploader node. N downloader nodes. All downloads start simultaneously.
Parameters: downloads=[10, 20], filesize=[1000mb]
(Basically the "fast subset" of what I did here: https://hackmd.io/rwPtPJ7KTw6cGjhN0zNYig )

For `18b5d6c`, I got these numbers:
10 nodes, 1000mb:
```
(3 mins, 14 secs)
(3 mins, 14 secs)
(3 mins, 14 secs)
(3 mins, 14 secs)
(3 mins, 14 secs)
(3 mins, 14 secs)
(3 mins, 14 secs)
(3 mins, 14 secs)
(3 mins, 14 secs)
(3 mins, 14 secs)
```

20 nodes, 1000mb:
```
(8 mins, 45 secs)
(8 mins, 44 secs)
(8 mins, 42 secs)
(8 mins, 46 secs)
(8 mins, 47 secs)
(8 mins, 47 secs)
(8 mins, 47 secs)
(8 mins, 39 secs)
(8 mins, 47 secs)
(8 mins, 47 secs)
(8 mins, 41 secs)
(8 mins, 40 secs)
(8 mins, 42 secs)
(8 mins, 47 secs)
(8 mins, 47 secs)
(8 mins, 47 secs)
(8 mins, 47 secs)
(8 mins, 44 secs)
(8 mins, 47 secs)
(8 mins, 41 secs)
```

These numbers are very consistent across the nodes. This is unusual when compared to master, where for example for the 20 nodes test there can easily be a difference of 7 minutes between the fastest and slowest nodes.

tests/codex/blockexchange/engine/testengine.nim
```diff
@@ -344,9 +349,168 @@ asyncchecksuite "NetworkStore engine handlers":
...
+    engine = BlockExcEngine.new(
+      localStore, wallet, network, discovery, advertiser, peerStore, pendingBlocks
+    )
+
+    peerCtx = BlockExcPeerCtx(id: peerId)
+    engine.peers.add(peerCtx)
+
+  test "Should exhaust retries":
+    var
+      retries = 2
```
marcinczenko, line 408: maybe use const for that (we use this number in two places), e.g. `const BlockRetires = 2`?

tests/codex/blockexchange/engine/testengine.nim
```diff
@@ -344,9 +349,168 @@ asyncchecksuite "NetworkStore engine handlers":
...
+    engine.network = BlockExcNetwork(
+      request: BlockExcRequest(
+        sendWantList: sendWantList, sendWantCancellations: NopSendWantCancellationsProc
+      )
+    )
+
+    let pending = engine.requestBlock(address)
+    await steps.wait()
+
+    # add blocks precense
```
marcinczenko, line 470: typo:

# add blocks presence

tests/codex/blockexchange/engine/testengine.nim
```diff
@@ -344,9 +349,168 @@ asyncchecksuite "NetworkStore engine handlers":
...
+      of WantHave:
+        check engine.pendingBlocks.isInFlight(address) == false
+        check engine.pendingBlocks.retriesExhausted(address) == false
+        steps.fire()
+      of WantBlock:
+        check engine.pendingBlocks.isInFlight(address) == true
+        check engine.pendingBlocks.retriesExhausted(address) == false
+        steps.fire()
+
+    engine.pendingBlocks.blockRetries = 10
```
marcinczenko, line 459: Hmm... I would expect `3` to be safe here...

tests/codex/blockexchange/engine/testengine.nim
```diff
@@ -344,9 +349,168 @@ asyncchecksuite "NetworkStore engine handlers":
...
+    let pending = engine.requestBlock(address)
+    await steps.wait()
+
+    # add blocks precense
+    peerCtx.blocks = blocks.mapIt(
+      (it.address, Presence(address: it.address, have: true, price: UInt256.example))
+    ).toTable
+
+    steps.clear()
+    await steps.wait()
```
marcinczenko, line 476: Comments?

\# waiting for first "WantBlock" being sent after receiving block presence message
\# `WantBlock` will trigger `blockDeliveryHandler` on the remote peer

tests/codex/blockexchange/engine/testengine.nim
```diff
@@ -344,9 +349,168 @@ asyncchecksuite "NetworkStore engine handlers":
...
+        addresses: seq[BlockAddress],
+        priority: int32 = 0,
+        cancel: bool = false,
+        wantType: WantType = WantType.WantHave,
+        full: bool = false,
+        sendDontHave: bool = false,
+    ) {.gcsafe, async.} =
+      done.complete()
+
+    engine.pendingBlocks.blockRetries = 10
```
marcinczenko, line 499: Will "1" not be enough here?

marcinczenko (approved):
I have finished. I've also reviewed the tests. I do not see anything braking - my comments are pretty much asking for clarifications or potential improvements (like this one <https://github.com/codex-storage/nim-codex/pull/1119/files#r1966579156>).

So, because I do not see anything braking I will approve.

If think it would be good that @dryajov looks quickly at my comments (should not take much time for him - it will improve my understanding at least).

codex/blockexchange/engine/discovery.nim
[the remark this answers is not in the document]
dryajov: Ah, good catch :)

codex/blockexchange/engine/discovery.nim
dryajov: kinda wanted to keep the noise down, it was probably a mistake to have done that in the engine, but I couldn't help myself :)

codex/blockexchange/engine/engine.nim
dryajov: I originally had this as a template in pendingBlocks, but I opted for moving it here, because the downloading logic should be handled in engine. `RetriesExhausted` is being used in pendingBlocks as part of the type declaration, so I don't see an issue.

codex/blockexchange/engine/engine.nim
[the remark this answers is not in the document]
dryajov: I don't think that increasing outstanding blocks amount will materially affect anything. We cannot trust the remote to ever send us anything and since this isn't a req/resp protocol, we cannot simply wait for the remote to disconnect either - it might simply never send us anything (until we hit some global timeout). 

This is a bit tricky overall. You want to keep a balance between decreasing the number of potential duplicates vs the latency incurred waiting for slow blocks. Right now I'm aiming for a balance between these two, with a very very naive algorithm.

Ideally, we would rate peers based on (among other things), how quickly they respond, and de-prioritize/disconnect peers that aren't fast enough. But this requires a mechanism to rate peers, which we don't have, so we're kinda stuck doing this type of simplistic fixes (for now).

tests/codex/blockexchange/engine/testengine.nim
marcinczenko: yes, I meant something different - wasn't clear enough I suppose. I meant this:

```nim
const BlockRetires = 2
var
  retries = BlockRetires
  address = BlockAddress.init(blocks[0].cid)

proc sendWantList(
    id: PeerId,
    addresses: seq[BlockAddress],
    priority: int32 = 0,
    cancel: bool = false,
    wantType: WantType = WantType.WantHave,
    full: bool = false,
    sendDontHave: bool = false,
) {.gcsafe, async.} =
  check wantType == WantHave
  check not engine.pendingBlocks.isInFlight(address)
  check engine.pendingBlocks.retries(address) == retries
  retries -= 1

engine.pendingBlocks.blockRetries = BlockRetires
```

This way it is more explicit that those two are related - not a big deal.

PR #1120, opened 2025-02-20
feat: Multi threading support for circom prover

munna0908:
This PR adds multi-threading support to the circompat backend to prevent blocking during proof generation and verification.

### DIFF
--- a/codex/slots/proofs/backends/circomcompat.nim
+++ b/codex/slots/proofs/backends/circomcompat.nim
@@ -9,7 +9,7 @@
 
 {.push raises: [].}
 
-import std/[sugar, atomics]
+import std/[sugar, atomics, locks]
 
 import pkg/chronos
 import pkg/taskpools
@@ -39,6 +39,7 @@ type
     backendCfg: ptr CircomBn254Cfg
     vkp*: ptr CircomKey
     taskpool: Taskpool
+    lock: ptr Lock
 
   NormalizedProofInputs*[H] {.borrow: `.`.} = distinct ProofInputs[H]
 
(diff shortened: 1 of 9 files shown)


codex/slots/proofs/backends/circomcompat.nim
```diff
@@ -143,58 +182,139 @@ proc prove[H](self: CircomCompat, input: NormalizedProofInputs[H]): ?!CircomProo
...
-    let res = verify_circuit(proofPtr, inputs.addr, self.vkp)
-    if res == ERR_OK:
-      success true
-    elif res == ERR_FAILED_TO_VERIFY_PROOF:
-      success false
-    else:
-      failure("Failed to verify proof - err code: " & $res)
-  finally:
-    inputs.releaseCircomInputs()
+    if error =? (await self.asyncVerify(proof, inputs, res)).errorOption:
```
dryajov, line 313: why not just return a bool here (or even void) and set to false on failure, true otherwise?

codex/slots/proofs/backends/converters.nim
```diff
@@ -52,3 +54,32 @@ func toG2*(g: CircomG2): G2Point =
 
 func toGroth16Proof*(proof: CircomProof): Groth16Proof =
   Groth16Proof(a: proof.a.toG1, b: proof.b.toG2, c: proof.c.toG1)
+
+proc newProof*(): ptr Proof =
```
dryajov, line 58: I think it would be better if we declare this as types and use the standard `new` conventino to construct them.

```nim
type
  ProofPtr = ptr Proof

proc new(_: ProofPtr): ProofPtr = 
 ...
```

Same for the verify types.

dryajov (changes requested):
I'm still reviwing this, but so far it's looking good!

codex/slots/proofs/backends/circomcompat.nim
```diff
@@ -143,58 +182,139 @@ proc prove[H](self: CircomCompat, input: NormalizedProofInputs[H]): ?!CircomProo
...
+    task[].success[].store(false)
+  else:
+    task[].success[].store(false)
+    error "Failed to verify proof", errorCode = res
+
+proc asyncVerify*[H](
+    self: CircomCompat,
+    proof: CircomProof,
+    inputs: ProofInputs[H],
+    success: ptr Atomic[bool],
```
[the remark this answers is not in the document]
dryajov: There is no advantage of using a pointer here at all, copying a pointer is equivalent to copying a bool value? I don't see material difference between this and prove procedure?

codex/slots/proofs/backends/circomcompat.nim
dryajov: The response is either passed or failed, right?

PR #1121, opened 2025-02-20
feat: enable chronos profiling

gmega:
Enables chronos profiling (chroprof). For now this is here just so you can see the changes and update the branch when needed.


gmega:
I think we have better ways of profiling chronos code these days.

PR #1123, opened 2025-02-21
fix: optimise erasure encode/decode 

munna0908:
This PR optimizes the asyncEncode and asyncDecode logic by eliminating unnecessary copying of block and parity data, enabling shared memory usage.


dryajov (approved):

gmega (approved):
Indeed looking nicer :-)

PR #1136, opened 2025-02-28
chore: create new httpClient per request

### LINKED ISSUE #1124: Use unique HttpClient in integration tests
`presto` does not support HTTP pipelining by default, and it returns a header `Connection: close`, meaning that the `HttpClient` should only be used once when making requests.

Keeping the same client _may_ work, but it is in a `faulty state`, which is why the error comes up only under certain conditions (CI on Windows + slow request for example).

Moreover, this is 100% reproducible on the [persistent availabilities branch](https://github.com/codex-storage/nim-codex/tree/refs/heads/feat/persistent-availabilities) when removing the line `host.restart()`. After a response 204 from the `PATCH` request, a new connection has to be created, otherwise, the error `Connection was closed before the full request has been made [ProtocolError]` will pop up.

Creating a new HTTP client for each request should fix the errors and improve stability.

Some references:
https://github.com/status-im/nim-presto/issues/77
https://github.com/nim-lang/Nim/issues/16572
https://github.com/nim-lang/Nim/issues/7413

2-towns:
Fix #1124

### DIFF
--- a/tests/integration/codexclient.nim
+++ b/tests/integration/codexclient.nim
@@ -15,16 +15,23 @@ export purchasing
 
 type CodexClient* = ref object
   baseurl: string
+  httpClients: seq[HttpClient]
 
 type CodexClientError* = object of CatchableError
 
 const HttpClientTimeoutMs = 60 * 1000
 
 proc new*(_: type CodexClient, baseurl: string): CodexClient =
-  CodexClient(baseurl: baseurl)
+  CodexClient(baseurl: baseurl, httpClients: newSeq[HttpClient]())
 
-proc http*(_: CodexClient): HttpClient =
-  return newHttpClient(timeout = HttpClientTimeoutMs)
+proc http*(client: CodexClient): HttpClient =
+  let httpClient = newHttpClient(timeout = HttpClientTimeoutMs)
+  client.httpClients.insert(httpClient)
+  return httpClient
+
+proc close*(client: CodexClient): void =
+  for httpClient in client.httpClients:
+    httpClient.close()
 
 proc info*(client: CodexClient): ?!JsonNode =
   let url = client.baseurl & "/debug/info"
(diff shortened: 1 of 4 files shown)


dryajov (approved):

tests/integration/codexprocess.nim
```diff
@@ -68,7 +68,6 @@ method stop*(node: CodexProcess) {.async.} =
 
   trace "stopping codex client"
   if client =? node.client:
-    client.close()
```
gmega, line 71: Hm... are we leaking anything by never closing any of the clients? Otherwise LGTM.

tests/integration/codexclient.nim
```diff
@@ -14,59 +14,57 @@ import pkg/codex/sales/reservations
...
 proc new*(_: type CodexClient, baseurl: string): CodexClient =
-  CodexClient(
-    http: newHttpClient(timeout = HttpClientTimeoutMs),
-    baseurl: baseurl,
-    session: HttpSessionRef.new({HttpClientFlag.Http11Pipeline}),
-  )
+  CodexClient(baseurl: baseurl)
+
+proc http*(_: CodexClient): HttpClient =
+  return newHttpClient(timeout = HttpClientTimeoutMs)
```
gmega, line 27: Do we need to keep track of the previous client instance and close it before doing this? Maybe have this implemented as `restart` previously was?

gmega (approved):
Some minor comments but otherwise LGTM.

tests/integration/codexprocess.nim
2-towns: That's a good point. Nim-presto sends the header "Connection: close," so the `httpclient` will close the connection itself (see [Nim HTTP Client](https://github.com/nim-lang/Nim/blob/f8294ce06e61752d15624e0422c5296292ef5467/lib/pure/httpclient.nim#L850-L853)).

However, it may be good practice to close it explicitly. We could close each client that we open, or we could keep track of each `httpclient` created and then close them at the end of the tests.

gmega (approved):
Looking good!

PR #1134, opened 2025-02-28
chore: add missing custom errors

2-towns:
Should fix #1133

PR #1139, opened 2025-03-03
fix(marketplace): catch Marketplace_SlotIsFree and continue the cancelled process

### LINKED ISSUE #1138: [BUG] calling market.freeSlot in SaleCancelled causes exception when slot not filled
When request expires, in `SaleCancelled` state we call `market.freeSlot`. But if the slot was never filled this will result in `Marketplace_SlotIsFree` as freeSlot is protected by `slotIsNotFree`:

```solidity
modifier slotIsNotFree(SlotId slotId) {
  if (_slots[slotId].state == SlotState.Free) revert Marketplace_SlotIsFree();
    _;
  }

function freeSlot(
    SlotId slotId,
    address rewardRecipient,
    address collateralRecipient
  ) public slotIsNotFree(slotId) {
// ...
}

function freeSlot(SlotId slotId) public slotIsNotFree(slotId) {
    return freeSlot(slotId, msg.sender, msg.sender);
  }

```

This in turn will cause the `returnedCollateral` to be never retuned to the availability as `onCleanUp` in `SaleCancelled` (calling subsequently `sales.cleanUp`) will not be reached. I encountered this failure while validating the local marketplace demo using locally build codex.

2-towns:
Fix #1138

### DIFF
--- a/codex/contracts/market.nim
+++ b/codex/contracts/market.nim
@@ -288,7 +288,9 @@ method freeSlot*(
 
       discard await freeSlot.confirm(1)
     except Marketplace_SlotIsFree:
-      raiseMarketError("Slot is free")
+      raise newException(
+        SlotStateMismatchError, "Failed to free slot, slot is already free", parent
+      )
 
 method withdrawFunds(
     market: OnChainMarket, requestId: RequestId
--- a/codex/contracts/market.nim
+++ b/codex/contracts/market.nim
@@ -232,7 +232,7 @@ method getHost(
 
 method currentCollateral*(
     market: OnChainMarket, slotId: SlotId
-): Future[UInt256] {.async.} =
+): Future[UInt256] {.async: (raises: [MarketError, CancelledError]).} =
   convertEthersError("Failed to get slot's current collateral"):
     return await market.contract.currentCollateral(slotId)
 
--- a/codex/contracts/market.nim
+++ b/codex/contracts/market.nim
@@ -287,7 +287,7 @@ method freeSlot*(
         freeSlot = market.contract.freeSlot(slotId)
 
       discard await freeSlot.confirm(1)
-    except Marketplace_SlotIsFree:
+    except Marketplace_SlotIsFree as parent:
       raise newException(
         SlotStateMismatchError, "Failed to free slot, slot is already free", parent
       )
--- a/codex/contracts/market.nim
+++ b/codex/contracts/market.nim
@@ -221,7 +221,7 @@ method requestExpiresAt*(
 
 method getHost(
     market: OnChainMarket, requestId: RequestId, slotIndex: uint64
-): Future[?Address] {.async.} =
+): Future[?Address] {.async: (raises: [CancelledError, MarketError]).} =
   convertEthersError("Failed to get slot's host"):
     let slotId = slotId(requestId, slotIndex)
     let address = await market.contract.getHost(slotId)
@@ -270,22 +270,25 @@ method freeSlot*(
     market: OnChainMarket, slotId: SlotId
 ) {.async: (raises: [CancelledError, MarketError]).} =
   convertEthersError("Failed to free slot"):
-    var freeSlot: Future[Confirmable]
-    if rewardRecipient =? market.rewardRecipient:
-      # If --reward-recipient specified, use it as the reward recipient, and use
-      # the SP's address as the collateral recipient
-      let collateralRecipient = await market.getSigner()
-      freeSlot = market.contract.freeSlot(
-        slotId,
-        rewardRecipient, # --reward-recipient
-        collateralRecipient,
-      ) # SP's address
-    else:
-      # Otherwise, use the SP's address as both the reward and collateral
-      # recipient (the contract will use msg.sender for both)
-      freeSlot = market.contract.freeSlot(slotId)
+    try:
+      var freeSlot: Future[Confirmable]
+      if rewardRecipient =? market.rewardRecipient:
+        # If --reward-recipient specified, use it as the reward recipient, and use
+        # the SP's address as the collateral recipient
+        let collateralRecipient = await market.getSigner()
+        freeSlot = market.contract.freeSlot(
+          slotId,
+          rewardRecipient, # --reward-recipient
+          collateralRecipient,
+        ) # SP's address
+      else:
+        # Otherwise, use the SP's address as both the reward and collateral
+        # recipient (the contract will use msg.sender for both)
+        freeSlot = market.contract.freeSlot(slotId)
 
-    discard await freeSlot.confirm(1)
+      discard await freeSlot.confirm(1)
+    except Marketplace_SlotIsFree:
+      raiseMarketError("Slot is free")
 
 method withdrawFunds(
     market: OnChainMarket, requestId: RequestId
--- a/codex/sales/states/cancelled.nim
+++ b/codex/sales/states/cancelled.nim
@@ -3,6 +3,7 @@ import ../../utils/exceptions
 import ../salesagent
 import ../statemachine
 import ./errored
+from ../../contracts/marketplace import Marketplace_SlotIsFree
 
 logScope:
   topics = "marketplace sales cancelled"
@@ -27,7 +28,11 @@ method run*(
     debug "Collecting collateral and partial payout",
       requestId = data.requestId, slotIndex = data.slotIndex
     let currentCollateral = await market.currentCollateral(slot.id)
-    await market.freeSlot(slot.id)
+
+    try:
+      await market.freeSlot(slot.id)
+    except Marketplace_SlotIsFree as e:
+      info "The slot cannot be freed because it is already free."
 
     if onClear =? agent.context.onClear and request =? data.request:
       onClear(request, data.slotIndex)
--- a/codex/sales/states/cancelled.nim
+++ b/codex/sales/states/cancelled.nim
@@ -32,7 +32,7 @@ method run*(
     try:
       await market.freeSlot(slot.id)
     except Marketplace_SlotIsFree as e:
-      info "The slot cannot be freed because it is already free."
+      debug "The slot cannot be freed because it is already free."
 
     if onClear =? agent.context.onClear and request =? data.request:
       onClear(request, data.slotIndex)
--- a/codex/sales/states/cancelled.nim
+++ b/codex/sales/states/cancelled.nim
@@ -42,8 +42,8 @@ method run*(
 
       try:
         await market.freeSlot(slot.id)
-      except MarketError as e:
-        trace "Error when trying to free the slot", error = e.msg
+      except SlotStateMismatchError as e:
+        warn "Failed to free slot because slot is already free", error = e.msg
 
       returnedCollateral = currentCollateral.some
 
--- a/codex/sales/states/cancelled.nim
+++ b/codex/sales/states/cancelled.nim
@@ -3,7 +3,6 @@ import ../../utils/exceptions
 import ../salesagent
 import ../statemachine
 import ./errored
-from ../../contracts/marketplace import Marketplace_SlotIsFree
 
 logScope:
   topics = "marketplace sales cancelled"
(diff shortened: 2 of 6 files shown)


codex/sales/states/cancelled.nim
```diff
@@ -27,7 +28,11 @@ method run*(
     debug "Collecting collateral and partial payout",
       requestId = data.requestId, slotIndex = data.slotIndex
     let currentCollateral = await market.currentCollateral(slot.id)
-    await market.freeSlot(slot.id)
+
+    try:
+      await market.freeSlot(slot.id)
+    except Marketplace_SlotIsFree as e:
+      info "The slot cannot be freed because it is already free."
```
emizzle, line 35: If this was a SP that had filled a slot, and the slot could not be freed, I would think this should be a warning. However, in other cases, this is normal behaviour. So I'd probably reduce this to a `debug` level because in either case, it's not essential that the user sees the log and we don't really need it to display in default Codex logging settings.

tests/codex/helpers/mockmarket.nim
```diff
@@ -381,8 +384,8 @@ method canReserveSlot*(
 func setCanReserveSlot*(market: MockMarket, canReserveSlot: bool) =
   market.canReserveSlot = canReserveSlot
 
-func setReserveSlotThrowError*(market: MockMarket, error: ?(ref MarketError)) =
-  market.reserveSlotThrowError = error
+func setSlotThrowError*(market: MockMarket, error: ?(ref CatchableError)) =
```
emizzle, line 387: Along the lines of [the comment in the other PR](https://github.com/codex-storage/nim-codex/pull/1134/files#r1986460502), i'd make this specific to `freeSlot`, ie `setErrorOnFreeSlot`

codex/sales/states/failed.nim
```diff
@@ -28,12 +29,17 @@ method run*(
...
       requestId = data.requestId, slotIndex = data.slotIndex
+
     await market.freeSlot(slot.id)
 
     let error = newException(SaleFailedError, "Sale failed")
     return some State(SaleErrored(error: error))
   except CancelledError as e:
     trace "SaleFailed.run was cancelled", error = e.msgDetail
+  except Marketplace_SlotIsFree as e:
+    info "The slot cannot be freed because it is already free.", error = e.msg
```
emizzle, line 40: Same as above, re: `info` -> `debug`

tests/codex/sales/states/testcancelled.nim
```diff
@@ -52,3 +58,20 @@ asyncchecksuite "sales state 'cancelled'":
...
+    market.fillSlot(
+      requestId = request.id,
+      slotIndex = slotIndex,
+      proof = Groth16Proof.default,
+      host = Address.example,
+      collateral = currentCollateral,
+    )
+
+    let error = newException(Marketplace_SlotIsFree, "")
+    market.setSlotThrowError(some cast[ref CatchableError](error))
```
emizzle, line 72: Same [comment in other PR](https://github.com/codex-storage/nim-codex/pull/1134/files#r1986456981) regarding use of generics to prevent a conversion/cast.

tests/codex/sales/states/testcancelled.nim
```diff
@@ -40,6 +41,11 @@ asyncchecksuite "sales state 'cancelled'":
     agent.onCleanUp = onCleanUp
     state = SaleCancelled.new()
 
+  teardown:
+    returnBytesWas = bool.none
+    reprocessSlotWas = bool.none
+    returnedCollateralValue = UInt256.none
```
emizzle, line 47: This should be in `setup`, no?

tests/codex/sales/states/testcancelled.nim
```diff
@@ -52,3 +58,20 @@ asyncchecksuite "sales state 'cancelled'":
...
+      collateral = currentCollateral,
+    )
+
+    let error = newException(Marketplace_SlotIsFree, "")
+    market.setSlotThrowError(some cast[ref CatchableError](error))
+
+    discard await state.run(agent)
+    check eventually returnBytesWas == some true
+    check eventually reprocessSlotWas == some false
+    check eventually returnedCollateralValue == some currentCollateral
```
emizzle, line 77: I think maybe we can improve the assumptions we're testing here because although we call `onCleanUp` directly in `run`, we might also go to `SaleErrored` in `run` which also calls `onCleanUp`, and I believe this test passes because the returned collateral in `onCleanUp` is not the default value of `none`, which makes this test a bit fragile. We should instead maybe split this between two tests:
1. Moves to `SaleErrored` if an unexpected error is raised
2. Remains in `SaleCancelled` if "slot not free" error is raised

The test here covers #2, but would just need to `check await state.run(agent) == none State` instead of checking the values of `onCleanUp` that were called.

2-towns: Done

tests/codex/sales/states/testcancelled.nim
2-towns: Yes done

codex/sales/states/failed.nim
2-towns: Done

tests/codex/helpers/mockmarket.nim
2-towns: Done

codex/sales/states/cancelled.nim
2-towns: Done

codex/contracts/market.nim
```diff
@@ -225,7 +225,7 @@ method getHost(
 
 method currentCollateral*(
     market: OnChainMarket, slotId: SlotId
-): Future[UInt256] {.async.} =
+): Future[UInt256] {.async: (raises: [MarketError, AsyncLockError, CancelledError]).} =
```
AuHau, line 228: Do we need the `AsyncLockError`? The `currentCollateral` is a view, so not sending transaction here and hence we should not need nonce-generation => "locking"...

codex/sales/states/cancelled.nim
```diff
@@ -3,6 +3,7 @@ import ../../utils/exceptions
 import ../salesagent
 import ../statemachine
 import ./errored
+from ../../contracts/marketplace import Marketplace_SlotIsFree
```
AuHau, line 6: 🔴 This is wrong. You are breaking the `market` abstraction here. This is coming from the contracts module directly, but you are operating with `market` abstraction. 

I can think of two approaches:

1) Bring these exceptions to Market level. It would be nice as we could use them with the async's `raises` on the market level, but on the other hand we might not need all of those as not everything is really relevant 
2) Convert or wrap them to some other error that operates on the Market abstraction and do assertion on them here.

@emizzle @markspanbroek WDYT?

AuHau (changes requested):
@2-towns you are breaking the `market` abstraction here. We should think of how to do this cleanly.

Also I am not so convinced about the bug report and hence this fix. How it can happen that the node would arrive to Cancelled state, did deposit the collateral **but did not fill the slot**? Because to my understanding of the bug report that is what happened? And that is what we are fixing here?

IMHO with this fix you could easily broke the collateral accounting. For example when the node is in Downloading state, the Request would expiry and the state machine would transition to Cancelled state, the `freeSlot` would get called, but the error would get swallowed, then the following `onCleanUp` would get actually called potentially returning non-existing collateral.

So, I think - yes - we should handle this case, but IMHO at least the `onCleanUp` should in such case not return the collateral. And we should double check what we are actually gonna be returning makes sense.

2-towns:
> Also I am not so convinced about the bug report and hence this fix. How it can happen that the node would arrive to Cancelled state, did deposit the collateral **but did not fill the slot**? Because to my understanding of the bug report that is what happened? And that is what we are fixing here?

The idea was to handle the `Marketplace_SlotIsFree` error raised by `freeSlot` and continue the cancelling state because it seemed natural to me. The state machine is trying to free the slot; the slot is already free, so let's move on. 
I didn't think about the scenario you mentioned, my bad. Next time, I will keep this in mind.

emizzle:
Good point @AuHau. We should probably add a condition to the cancelled state to handle the two different scenarios: the slot is filled or the slot is not filled. We can do that with the check (or abstract out this check):
```nim
proc slotIsFilledByMe(market: Market, requestId: RequestId, slotIndex: UInt256): bool =
  let host = await market.getHost(requestId, slotIndex)
  let me = await market.getSigner()
  host == me.some

if slotIsFilledByMe:
  # freeSlot, on "slot not free" error, show warning
  # onCleanUp(returnCollateral = some...)
else: 
  # onCleanUp(returnCollateral = none...)
```

I think we need to keep in mind that, although unlikely, `freeSlot` may have been called by the SP manually/externally before arriving here.

codex/sales/states/cancelled.nim
markspanbroek: Indeed, good catch Adam. I'd go with Eric's idea, and wrap it in a MarketError.

We should probably also be careful to try and catch specific errors from the contract. Going through the testnet logs, I've seen that from Geth we often only get an "execution reverted" error with no specifics. So trying to catch specific errors works nicely with hardhat, but is probably moot in testnet.

AuHau:
> I think we need to keep in mind that, although unlikely, freeSlot may have been called by the SP manually/externally

Yeah, I don't think that will happen (often), but I guess nothing hurts to have this in.

codex/sales/states/cancelled.nim
AuHau: Alright then, wrapping with `MarketError` it is then. My original follow-up question was if we should really handle this exact error but wrapped (eq. "slot is free"), but I guess thanks to Eric's suggestion below we don't really care why this call actually fails because the cleanup should be called anyhow and the decision on the collateral returning should be done with Eric's suggestion.

However, what I do not understand is how come that these Solidity errors are thrown in this unwrapped form?
Looking at the [`freeSlot`](https://github.com/codex-storage/nim-codex/blob/master/codex/contracts/market.nim#L256) implementation, it uses the `convertEthersError` template which should be catching `EthersError` and the `SolidityError` is inheriting from `EthersError` - shouldn't that be caught by this? Or does it not work like this in Nim? 😅 @emizzle any ideas?

codex/contracts/market.nim
2-towns: Yes we need `AsyncLockError`. It should be removed once https://github.com/codex-storage/nim-codex/pull/1146 is merged.

AuHau: Oh nice! Yeah just noticed that. Maybe lets merge this PR after that one is in so we can remove it?

2-towns: Yep

codex/sales/states/cancelled.nim
```diff
@@ -23,11 +31,23 @@ method run*(
...
+      debug "Collecting collateral and partial payout",
+        requestId = data.requestId, slotIndex = data.slotIndex
+
+      let slot = Slot(request: request, slotIndex: data.slotIndex)
+      let currentCollateral = await market.currentCollateral(slot.id)
+
+      try:
+        await market.freeSlot(slot.id)
+      except MarketError as e:
+        trace "Error when trying to free the slot", error = e.msg
```
emizzle, line 46: So I think we'll need to wrap the exception into its own type in the Market abstraction, so that we can disambiguate from other exception types and handle them differently here. Only when we have a "slot is not free" error, is it ok to log and continue. Other errors, we still need to go to SaleErrored (caught below).

```nim
# market.nim
except Marketplace_SlotIsFree as parent:
  raise newException (SlotStateMismatchError, "Failed to free slot, slot is already free", parent)

# cancelled.nim
try:
  await market.freeSlot(slot.id)
except SlotStatMismatchError as e:
  warn "Failed to free slot because slot is already free", error = e.msg
```

codex/sales/states/cancelled.nim
```diff
@@ -23,11 +31,23 @@ method run*(
...
+
+      returnedCollateral = currentCollateral.some
+    else:
+      debug "The slot is not filled by me"
```
emizzle, line 50: 🟡  I'd probably leave this out.

codex/sales/states/cancelled.nim
emizzle: > However, what I do not understand is how come that these Solidity errors are thrown in this unwrapped form?
Looking at the [freeSlot](https://github.com/codex-storage/nim-codex/blob/master/codex/contracts/market.nim#L256) implementation, it uses the convertEthersError template which should be catching EthersError and the SolidityError is inheriting from EthersError - shouldn't that be caught by this? Or does it not work like this in Nim? 😅 @emizzle any ideas?

I don't think the error would have propagated to the cancelled state. It would have been converted to a MarketError in `convertEthersError` as you mentioned. The compiler doesn't stop you from catching errors that don't actually propagate to the callsite 🤷

AuHau (approved):
LGTM, one suggestion but nothing blocking - thanks!

emizzle (approved):

PR #1140, opened 2025-03-03
fix(ci): handle coverage as a string to enable gcc 14 on linux

2-towns:
The CI take input as string not as boolean using `true` or `false`. This PR fixes this when enabling gcc 14 on linux.


veaceslavdoina (approved):
Looks good, thank you!

dryajov (approved):

PR #1144, opened 2025-03-06
chore: add request validations

2-towns:
Reflect the validation rules applied in the contracts side https://github.com/codex-storage/codex-contracts-eth/blob/1982e71d52b6f593cc6c20a3bfad2f70b751d081/contracts/Marketplace.sol#L131-L160

### DIFF
--- a/codex/rest/api.nim
+++ b/codex/rest/api.nim
@@ -652,10 +652,39 @@ proc initPurchasingApi(node: CodexNodeRef, router: var RestRouter) =
       without params =? StorageRequestParams.fromJson(body), error:
         return RestApiResponse.error(Http400, error.msg, headers = headers)
 
+      without expiry =? params.expiry:
+        return RestApiResponse.error(
+          Http422, "Expiry needs to be bigger than zero", headers = headers
+        )
+
+      if expiry <= 0 or expiry >= params.duration:
+        return RestApiResponse.error(
+          Http422,
+          "Expiry needs value bigger then zero and smaller then the request's duration",
+          headers = headers,
+        )
+
+      if params.proofProbability <= 0:
+        return RestApiResponse.error(
+          Http422, "Proof probability needs to be bigger than zero", headers = headers
+        )
+
+      if params.collateralPerByte <= 0:
+        return RestApiResponse.error(
+          Http422, "Collateral per byte needs to be bigger than zero", headers = headers
+        )
+
+      if params.pricePerBytePerSecond <= 0:
+        return RestApiResponse.error(
+          Http422,
+          "Price per byte per second needs to be bigger than zero",
+          headers = headers,
+        )
+
       let requestDurationLimit = await contracts.purchasing.market.requestDurationLimit
       if params.duration > requestDurationLimit:
         return RestApiResponse.error(
-          Http400,
+          Http422,
           "Duration exceeds limit of " & $requestDurationLimit & " seconds",
           headers = headers,
         )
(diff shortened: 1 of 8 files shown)


codex/rest/api.nim
```diff
@@ -637,10 +637,36 @@ proc initPurchasingApi(node: CodexNodeRef, router: var RestRouter) =
       without params =? StorageRequestParams.fromJson(body), error:
         return RestApiResponse.error(Http400, error.msg, headers = headers)
 
+      let expiry = params.expiry
+
+      if expiry <= 0 or expiry >= params.duration:
+        return RestApiResponse.error(
+          Http422,
+          "Expiry needs value bigger then zero and smaller then the request's duration",
```
emizzle, line 645: 🟡 
```suggestion
          "Expiry must be greater than zero and less than the request's duration",
```

codex/rest/api.nim
```diff
@@ -637,10 +637,36 @@ proc initPurchasingApi(node: CodexNodeRef, router: var RestRouter) =
...
+          headers = headers,
+        )
+
+      if params.proofProbability <= 0:
+        return RestApiResponse.error(
+          Http422, "Proof probability needs to be bigger than zero", headers = headers
```
emizzle, line 651: 🟡 
```suggestion
          Http422, "Proof probability must be greater than zero", headers = headers
```

codex/rest/api.nim
```diff
@@ -637,10 +637,36 @@ proc initPurchasingApi(node: CodexNodeRef, router: var RestRouter) =
...
+        )
+
+      if params.collateralPerByte <= 0:
+        return RestApiResponse.error(
+          Http422, "Collateral per byte needs to be bigger than zero", headers = headers
```
emizzle, line 656: 🟡 
```suggestion
          Http422, "Collateral per byte must be greater than zero", headers = headers
```

codex/rest/api.nim
```diff
@@ -637,10 +637,36 @@ proc initPurchasingApi(node: CodexNodeRef, router: var RestRouter) =
...
+        )
+
+      if params.pricePerBytePerSecond <= 0:
+        return RestApiResponse.error(
+          Http422,
+          "Price per byte per second needs to be bigger than zero",
```
emizzle, line 662: 🟡 
```suggestion
          "Price per byte per second must be greater than zero",
```

emizzle (approved):
Nicely done. Just some suggestions around wording, but no blockers so 👍

codex/rest/api.nim
2-towns: Done

codex/rest/api.nim
2-towns: Done

codex/rest/api.nim
2-towns: Done

AuHau (approved):
LGTM, but I would ask you to backfill one related issue - in this PR https://github.com/codex-storage/nim-codex/pull/1088/ you have added the error types for the added validations, but I forgot to ask you to also add them into proper places where they are thrown. They should be listed in the `errors` pragma at the right place.
But I will approve as it is non-blocking. If you want you can add it in follow up PR as well.

2-towns:
> LGTM, but I would ask you to backfill one related issue - in this PR #1088 you have added the error types for the added validations, but I forgot to ask you to also add them into proper places where they are thrown. They should be listed in the `errors` pragma at the right place.
> But I will approve as it is non-blocking. If you want you can add it in follow up PR as well.

Okay so I added the custom errors related to the request validations. 

`Periods_InvalidSecondsPerPeriod` can be raised in the marketplace constructor (I guess). 

`Proofs_InvalidProbability` is not used anymore

PR #1146, opened 2025-03-07
fix: ethers no longer leaks AsyncLockError

emizzle:
NOTE: Dependent upon https://github.com/codex-storage/nim-ethers/pull/109

### DIFF
--- a/codex/contracts/market.nim
+++ b/codex/contracts/market.nim
@@ -86,8 +86,7 @@ method loadConfig*(
       market.configuration = some fetchedConfig
 
     return success()
-  except AsyncLockError, EthersError:
-    let err = getCurrentException()
+  except EthersError as err:
     return failure newException(
       MarketError,
       "Failed to fetch the config from the Marketplace contract: " & err.msg,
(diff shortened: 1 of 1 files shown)


codex/contracts/market.nim
```diff
@@ -195,7 +194,7 @@ method slotState*(
     try:
       let overrides = CallOverrides(blockTag: some BlockTag.pending)
       return await market.contract.slotState(slotId, overrides)
-    except AsyncLockError as err:
+    except EthersError as err:
```
2-towns, line 197: Can't we remove the except here ? 

The EthersError should be caught by `convertEthersError`.

emizzle: I left it there because of the contextual error message. However, I've just pushed a commit that adds contextual error messages to `convertEthersError` and updated this block of code 👍

2-towns: Oh that's great ! Thanks.

2-towns (approved):

AuHau (approved):
Very nice 👍

PR #1148, opened 2025-03-07
chore(marketplace): notify sales when duration, minPricePerBytePerSecond or totalCollateral is updated

2-towns:
Currently, when an availability is updated, `sales` is notified only when the freeSize increases.

This PR changes this behavior to notify `sales` when freeSize, duration, minPricePerBytePerSecond, or totalCollateral is increased. All those parameters are taken into consideration when trying to find an availability matching the slot.

### DIFF
--- a/codex/sales/reservations.nim
+++ b/codex/sales/reservations.nim
@@ -300,7 +300,9 @@ proc updateAvailability(
 
   let res = await self.updateImpl(obj)
 
-  if oldAvailability.freeSize < obj.freeSize: # availability added
+  if oldAvailability.freeSize < obj.freeSize or oldAvailability.duration < obj.duration or
+      oldAvailability.minPricePerBytePerSecond < obj.minPricePerBytePerSecond or
+      oldAvailability.totalCollateral < obj.totalCollateral: # availability updated
     # inform subscribers that Availability has been modified (with increased
     # size)
     if onAvailabilityAdded =? self.onAvailabilityAdded:
--- a/codex/sales/reservations.nim
+++ b/codex/sales/reservations.nim
@@ -82,11 +82,11 @@ type
     availabilityLock: AsyncLock
       # Lock for protecting assertions of availability's sizes when searching for matching availability
     repo: RepoStore
-    OnAvailabilitySaved: ?OnAvailabilitySaved
+    OnAvailabilityUpserted: ?OnAvailabilityUpserted
 
   GetNext* = proc(): Future[?seq[byte]] {.upraises: [], gcsafe, closure.}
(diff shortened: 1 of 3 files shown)


codex/sales/reservations.nim
```diff
@@ -300,21 +300,23 @@ proc updateAvailability(
...
-        await onAvailabilityAdded(obj)
+        await OnAvailabilitySaved(obj)
       except CancelledError as e:
         raise e
       except CatchableError as e:
         # we don't have any insight into types of exceptions that
-        # `onAvailabilityAdded` can raise because it is caller-defined
-        warn "Unknown error during 'onAvailabilityAdded' callback", error = e.msg
+        # `OnAvailabilitySaved` can raise because it is caller-defined
+        warn "Unknown error during 'OnAvailabilitySaved' callback", error = e.msg
```
emizzle, line 319: Might be a good idea to annotate `OnAvailableSaved` with `{.async: (raises: [].}` and remove this try/except.

emizzle (commented):
Nice job 👏 

Wdyt about `OnAvailabilitySaved` -> `OnAvailabilityUpdated`? Just a preference/opinion, not 100% needed.

codex/sales/reservations.nim
2-towns: I did it in the Persistent availabilities PR : https://github.com/codex-storage/nim-codex/pull/1099/files#diff-2d8ab913b094a37ce60c67f82043dbc414d6010611f055d1b6e586b4407f95b7R292

emizzle:
> I named it `OnAvailabilitySaved` because this callback is called when the availability is created or updated, so I feel that `save` has those two meanings. But yes, it is a preference/opinion, if you prefer `OnAvailabilityUpdated` I don't mind.

`Upsert` might work here... `OnAvailabilityUpserted`. But yeah I don't mind that much as well.

codex/sales/reservations.nim
emizzle: That's right, I remember having seen that comment be removed but thought maybe it was duplicated elsewhere. Nice 👍

emizzle (approved):

2-towns:
> Uffff "upserted"? 😅 That is ugly word to me. Honestly I don't even know what does that mean 😂 I would rather prefer "saved", but I won't force you into another renaming pain :-D

I know `upsert` only because of sql/sqlite 😅. 
Okay no problem I can go back to "saved", it is not a huge effort.

gmega:
Not sure why you're so upsert by this @AuHau!

AuHau:
Thanks @2-towns 🙏

PR #1150, opened 2025-03-07
chore(restapi): add headers to support on progress when downloading

2-towns:
This PR adds 2 headers to support [download component with progress](https://github.com/codex-storage/codex-marketplace-ui/issues/98):

1. `Content-Length` to enable [lengthComputable](https://developer.mozilla.org/en-US/docs/Web/API/ProgressEvent/lengthComputable)  
2. `Access-Control-Expose-Headers` to expose the `Content-Disposition` to get the filename from JavaScript.

### DIFF
--- a/codex/rest/api.nim
+++ b/codex/rest/api.nim
@@ -103,6 +103,8 @@ proc retrieveCid(
     else:
       resp.setHeader("Content-Disposition", "attachment")
 
+    resp.setHeader("Content-Length", $manifest.datasetSize.uint64)
+
     await resp.prepareChunked()
 
     while not stream.atEof:
@@ -328,6 +330,7 @@ proc initDataApi(node: CodexNodeRef, repoStore: RepoStore, router: var RestRoute
       resp.setCorsHeaders("GET", corsOrigin)
       resp.setHeader("Access-Control-Headers", "X-Requested-With")
 
+    resp.setHeader("Access-Control-Expose-Headers", "Content-Disposition")
     await node.retrieveCid(cid.get(), local = false, resp = resp)
 
   router.api(MethodGet, "/api/codex/v1/data/{cid}/network/manifest") do(


gmega (approved):
Yep, good change. I was fetching the manifest separately to do this calculation, much better to have it in a standard header.

codex/rest/api.nim
```diff
@@ -103,6 +103,8 @@ proc retrieveCid(
     else:
       resp.setHeader("Content-Disposition", "attachment")
 
+    resp.setHeader("Content-Length", $manifest.datasetSize.uint64)
```
marcinczenko, line 106: Currently the `manifest.dataSize` is `NBytes` (which is `range[0..high(int)]`). perhaps `uint` is more "inline" here, but on the other hand in BitTorrent stuff I am currently using `uint64`, so maybe for the future it is actually better... So, just to reflect.

tests/integration/codexclient.nim
```diff
@@ -76,15 +76,15 @@ proc downloadNoStream*(client: CodexClient, cid: Cid): ?!string =
 proc downloadBytes*(
     client: CodexClient, cid: Cid, local = false
 ): Future[?!seq[byte]] {.async.} =
-  let uri =
-    parseUri(client.baseurl & "/data/" & $cid & (if local: "" else: "/network/stream"))
+  let uri = client.baseurl & "/data/" & $cid & (if local: "" else: "/network/stream")
 
-  let (status, bytes) = await client.session.fetch(uri)
+  let client = newHttpClient()
+  let response = client.get(uri)
```
marcinczenko, line 82: You may consider using a different name for the new `client` - maybe `httpClient`? Looks like you will be shadowing `client: CodexClient`, which has different type - not a bit deal, but it may create a glitch for the reader.

marcinczenko (approved):

codex/rest/api.nim
2-towns: Yes , I applied the change

tests/integration/codexclient.nim
2-towns: Done

PR #1155, opened 2025-03-11
chore(sales): log error when datasets size is different then expected

AuHau:
This is a safeguard for when the downloaded Slot's size does not match what was expected. I want to add this, because back when I was working on IPFS I discovered a security bug where you could forge the manifest in a way that it would report a different size than the actual data's size. You could have then stored an "infinite" amount of data with the pinning services because they relied on the manifest info and did not double-check actual data size. 

Originally part of https://github.com/codex-storage/nim-codex/pull/798/ but that PR got reworked by Arnoud at https://github.com/codex-storage/nim-codex/pull/1099/ so just pulling this part out of it.


emizzle (commented):
If the data is smaller than `StorageRequest.ask.slotSize` then unused capacity in a Reservation gets returned to the Availability on clean up. If the data is greater than the Reservation capacity, we should get a `BytesOutOfBoundsError` exception in the result. So from an accounting perspective, I believe we're already ok. However, from a data integrity perspective, wouldn't it be better to recreate the Cid from the dataset and see if it matches? Slower, yes, but more accurate.

Also, maybe `node.onStore` might be a better place to implement that.

dryajov:
> However, from a data integrity perspective, wouldn't it be better to recreate the Cid from the dataset and see if it matches? Slower, yes, but more accurate.

This is already the case, we always check the integrity of the downloaded data.

codex/sales/states/downloading.nim
```diff
@@ -75,6 +75,15 @@ method run*(
       return some State(SaleErrored(error: err, reprocessSlot: false))
 
     trace "Download complete"
+    without keyId =? reservation.key, error:
+      error "Couldn't get reservation key", id = $reservation.id, error = error.msg
+
+    if updatedReservation =? await reservations.get(keyId, Reservation):
+      if updatedReservation.size != 0:
```
dryajov, line 82: This should not fail the contract tho, if this is whats going on. We should either return that extra capacity back, or just ignore it, since the user is allegedly already paying for it.

PR #1159, opened 2025-03-12
feat(integration): use async client instead of standard Nim HTTP client

### LINKED ISSUE #1163: [BUG] Codex sets `Content-Length` and then uses chunked encoding to transfer data
That's for the `/api/codex/v1/data/{cid}/network/stream` endpoint. The [RFC7230](https://datatracker.ietf.org/doc/html/rfc7230#section-3.3.2) is pretty clear we shouldn't do that, and it's not working well with some client implementations, like the Chronos HTTP client.

gmega:
This PR replaces the standard Nim client with the chronos http client for all integration tests. The trigger for this was us realizing that the blocking client will choke the (chronos) output processor and cause tests to freeze because it starves it.

As an easter egg, and because the chronos client is a lot less lenient with header screw-ups, this PR also:

Fixes #1163

### DIFF
--- a/codex/rest/api.nim
+++ b/codex/rest/api.nim
@@ -116,7 +116,7 @@ proc retrieveCid(
 
     resp.setHeader("Content-Length", $manifest.datasetSize.int)
 
-    await resp.prepareChunked()
+    await resp.prepare(HttpResponseStreamType.Plain)
 
     while not stream.atEof:
       var
@@ -129,7 +129,7 @@ proc retrieveCid(
 
       bytes += buff.len
 
-      await resp.sendChunk(addr buff[0], buff.len)
+      await resp.send(addr buff[0], buff.len)
     await resp.finish()
     codex_api_downloads.inc()
   except CancelledError as exc:
--- a/codex/rest/api.nim
+++ b/codex/rest/api.nim
@@ -114,7 +114,12 @@ proc retrieveCid(
     else:
       resp.setHeader("Content-Disposition", "attachment")
 
-    resp.setHeader("Content-Length", $manifest.datasetSize.int)
+    # For erasure-coded datasets, we need to return the _original_ length; i.e.,
+    # the length of the non-erasure-coded dataset, as that's what we will be
+    # returning to the client.
+    let contentLength =
+      if manifest.protected: manifest.originalDatasetSize else: manifest.datasetSize
+    resp.setHeader("Content-Length", $(contentLength.int))
 
     await resp.prepare(HttpResponseStreamType.Plain)
 
--- a/tests/integration/codexclient.nim
+++ b/tests/integration/codexclient.nim
@@ -113,10 +113,17 @@ proc setLogLevel*(
     response = await client.post(url, headers = headers, body = "")
   assert response.status == 200
 
+proc uploadRaw*(
+    client: CodexClient, contents: string, headers: seq[HttpHeaderTuple] = @[]
+): Future[HttpClientResponseRef] {.
+    async: (raw: true, raises: [CancelledError, HttpError])
+.} =
+  return client.post(client.baseurl & "/data", body = contents, headers = headers)
+
 proc upload*(
     client: CodexClient, contents: string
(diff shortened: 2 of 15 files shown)


emizzle (commented):
I know this is a WIP, but just thought I'd comment after looking at the progress: 

I don't think we need to create new `HttpSessionRef` each time, but instead manage a single one on the `CodexClient` instance. That's because, when `HttpClientRequestRef.send` is called, it creates a new connection in the session, or reuses an idle connection from the session's connection pool: https://github.com/status-im/nim-chronos/blob/e4cb48088c49068caf07777a510ee3ed37701b98/chronos/apps/http/httpclient.nim#L1187-L1195

tests/integration/codexclient.nim
```diff
@@ -15,63 +15,106 @@ export purchasing
...
+      HttpSessionRef.new(), url, headers = headers, body = body
+    ).get
+    response = await request.send()
+  return (response.status, await response.getBodyBytes())
+
+proc delete(
+    t: typedesc[HttpClientRequestRef],
+    session: HttpSessionRef,
+    url: string,
+    version: httputils.HttpVersion = HttpVersion11,
```
emizzle, line 46: `HttpVersion11` is fine, but it won't reuse a connection, because of presto sending a `connection: close` header in the response, which [`send` will interpret](https://github.com/status-im/nim-chronos/blob/e4cb48088c49068caf07777a510ee3ed37701b98/chronos/apps/http/httpclient.nim#L937-L944) and then [no keep-alive header will be added](if connectionFlag and not(newConnectionAlways) and not(closeConnection) and
     httpPipeline:
    res.connection.flags.incl(HttpClientConnectionFlag.KeepAlive)).

2-towns:
@gmega Did you consider using `AsyncHttpClient` from Nim ?

gmega:
Ugh no, I didn't want to step onto async territory that's not chronos @2-towns.

tests/integration/codexclient.nim
gmega: OK, my reasoning is that if it was indeed closing connections automatically (which I hadn't yet confirmed) this would be easier as we wouldn't need any bookkeeping on our end, but OK this is not hard to do at all. I'll add it.

gmega: OK that should be addressed in my latest commit.

emizzle (approved):
LGTM. I don't think we need the `waitFor`'s, but that's not really blocking, so will approve now. Thanks very much for doing this!

codex/rest/api.nim
```diff
@@ -114,9 +114,14 @@ proc retrieveCid(
...
-    resp.setHeader("Content-Length", $manifest.datasetSize.int)
+    # For erasure-coded datasets, we need to return the _original_ length; i.e.,
+    # the length of the non-erasure-coded dataset, as that's what we will be
+    # returning to the client.
+    let contentLength =
+      if manifest.protected: manifest.originalDatasetSize else: manifest.datasetSize
+    resp.setHeader("Content-Length", $(contentLength.int))
 
-    await resp.prepareChunked()
+    await resp.prepare(HttpResponseStreamType.Plain)
```
2-towns, line 124: What is the impact in terms of performance? Do you expect streaming to be more efficient than chunking?

gmega: I'd expect the difference to be close to zero. You do get a tiny bit of efficiency as you don't need to send the chunk lengths, but in practical terms the two things are very similar.

2-towns (approved):

tests/testTaiko.nim
```diff
@@ -24,7 +24,7 @@ suite "Taiko L2 Integration Tests":
     )
     node1.waitUntilStarted()
 
-    let bootstrap = (!node1.client.info())["spr"].getStr()
+    let bootstrap = (!waitFor node1.client.info())["spr"].getStr()
```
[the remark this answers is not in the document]
gmega: Fixed

PR #1164, opened 2025-03-14
monitor background tasks on streaming dataset

dryajov:
re-add stream/task monitoring, thanks to @marcinczenko for catching this!

### DIFF
--- a/codex/node.nim
+++ b/codex/node.nim
@@ -271,6 +271,8 @@ proc streamEntireDataset(
   ##
   trace "Retrieving blocks from manifest", manifestCid
 
+  var jobs: seq[Future[void]]
+  let stream = LPStream(StoreStream.new(self.networkStore, manifest, pad = false))
   if manifest.protected:
     # Retrieve, decode and save to the local store all EС groups
     proc erasureJob(): Future[void] {.async: (raises: []).} =
@@ -284,14 +286,25 @@ proc streamEntireDataset(
       except CatchableError as exc:
         trace "Error erasure decoding manifest", manifestCid, exc = exc.msg
 
-    self.trackedFutures.track(erasureJob())
+    jobs.add(erasureJob())
 
-  self.trackedFutures.track(self.fetchDatasetAsync(manifest, fetchLocal = false))
-    # prefetch task should not fetch from local store
+  jobs.add(self.fetchDatasetAsync(manifest))
+
+  # Monitor stream completion and cancel background jobs when done
+  proc monitorStream() {.async: (raises: []).} =
+    try:
+      await stream.join()
+    except CatchableError as exc:
+      warn "Stream failed", exc = exc.msg
+    finally:
+      await noCancel allFutures(jobs.mapIt(it.cancelAndWait))
+
+  self.trackedFutures.track(monitorStream())
 
   # Retrieve all blocks of the dataset sequentially from the local store or network
   trace "Creating store stream for manifest", manifestCid
-  LPStream(StoreStream.new(self.networkStore, manifest, pad = false)).success
+
+  stream.success
 
 proc retrieve*(
     self: CodexNodeRef, cid: Cid, local: bool = true


codex/node.nim
```diff
@@ -284,14 +286,25 @@ proc streamEntireDataset(
       except CatchableError as exc:
         trace "Error erasure decoding manifest", manifestCid, exc = exc.msg
 
-    self.trackedFutures.track(erasureJob())
+    jobs.add(erasureJob())
 
-  self.trackedFutures.track(self.fetchDatasetAsync(manifest, fetchLocal = false))
-    # prefetch task should not fetch from local store
+  jobs.add(self.fetchDatasetAsync(manifest))
```
marcinczenko, line 291: I am wondering if we should not close the stream when prefetching fails. Before introducing `fetchDatasetAsync`, we could take advantage of having stream in the closure:

```nim
proc prefetch(): Future[void] {.async.} =
  try:
    if err =? (
      await self.fetchBatched(torrentManifest, codexManifest, onPieceReceived)
    ).errorOption:
      error "Unable to fetch blocks", err = err.msg
      await stream.close()  # <========================== HERE
  except CancelledError:
    trace "Prefetch job cancelled"
  except CatchableError as exc:
    await stream.close() # <========================== HERE
    error "Error fetching blocks", exc = exc.msg
``` 

I am doing it while aggregating pieces in torrent, because it does not make much sense to continue streaming if piece validation fails. I could also just use the `onBatch` callback where the piece validation is performed, but I thought that it makes sense to close the stream also when fetching blocks fails. Perhaps we should do the same for regular codex content (we would probably have to either bring the `prefetch` back so that we can access the stream, pass the stream to `fetchDatasetAsync`, or introduce an optional callback to `fetchDatasetAsync`)?

I was also wondering if we should not make sure that `fetchBatched` can only raise `CancelledError` since it already uses `Result` to communicate failure. It would allow us to get rid of `except CatchableError` above...

marcinczenko (approved):

codex/node.nim
dryajov: Not really, there isn't any stream for the background task, it just writes to disk, this is the intent of the task. There might be other cases were we do a similar thing, but I would consider handling it adhoc in place. What is your specific use case?

marcinczenko: `fetchDatasetAsync` is a background task - correct. And before it was only used in the context without streaming. But in #1130, where you refactored that proc and moved it to `node.nim`, you also decided to reuse it in `streamEntireDataset` in place of custom `prefetch` which I am mentioning in the comment above. So in the context of `streamEntireDataset`, streaming becomes very much relevant. With this PR we are bringing back the monitoring of the stream, so in case the stream has been closed, we will cancel the "prefetch" task. This is good.

But I was wondering about the reverse direction. If the prefetch task (which currently is done by `fetchDatasetAsync`) fails fetching a single batch, it will not even try fetching remaining batches. So basically, a failure to fetch a single block will effectively kill the prefetching.

So one could argue if we should not immediately close the stream in such a case.

If we don't then because we are creating stream with instance of `NetworkStore`, while streaming the block, in `StoreStream.readOnce`, we will call `networkStore.getBlock` again, which - not being able to find the block in local store (because the prefetch operation failed) - will request the block again from the network. If this one fails again trying to fetch the block from the network, then the stream will err, and will get closed. It may be getting those missing blocks in the end, but it will choke I guess, because it will be trying to get them one, by one, without any batching.

So maybe, it is all fine. We opportunistically try to prefetch as much as we can, and if that fails, we will try again while streaming, but less efficiently. Still I am wondering if it would not be better and cleaner to forcibly close the stream on error from pre-fetch (and to do so, we would have to either bring the `prefetch` proc inside of `streamEntireDataset` - so as it was before and then just add closing the stream on error, or change slightly `fetchDatasetAsync` so that it has a failure callback in which we can close the stream on error).

codex/node.nim
```diff
@@ -284,14 +286,25 @@ proc streamEntireDataset(
...
+
+  # Monitor stream completion and cancel background jobs when done
+  proc monitorStream() {.async: (raises: []).} =
+    try:
+      await stream.join()
+    except CatchableError as exc:
+      warn "Stream failed", exc = exc.msg
```
marcinczenko, line 298: `stream.join` has the following signature:

```nim
proc join*(
    s: LPStream
): Future[void] {.async: (raises: [CancelledError], raw: true), public.} =
```

Thus, it can only raise `CancelledError`. Thus, wouldn't it be better to be more specific catching only the `CancelledError` and having a more precise message (like `Stream cancelled`)?

dryajov: Thats fine, this workers shouldn't allow anything to leak, so having a catch all exception is fine IMO.

codex/node.nim
dryajov: > But I was wondering about the reverse direction. If the prefetch task (which currently is done by fetchDatasetAsync) fails fetching a single batch, it will not even try fetching remaining batches. So basically, a failure to fetch a single block will effectively kill the prefetching.

Yeah, the prefetch right now is very naive and I would not attempt to address it in this PR.

> So maybe, it is all fine. We opportunistically try to prefetch as much as we can,

Yep, thats the idea, a bigger upgrade to the retry and prefetch is coming, but this isn't what this PR is for.

gmega: OK so I guess for now we're treating the prefetch as an optimization and don't want to close the stream if it fails. I guess I'm fine with that for now, but I'd really like to decouple streaming to the client from the download strategy so we can actually a better level of control on what's going on.

gmega (approved):
LGTM

PR #1166, opened 2025-03-17
chore(marketplace): switch to websocket

2-towns:
Linea is supporting only Websocket. 

So it would be better to use it in our testing. 

Depends on https://github.com/codex-storage/nim-ethers/pull/112.

### DIFF
--- a/tests/contracts/testClock.nim
+++ b/tests/contracts/testClock.nim
@@ -23,6 +23,10 @@ ethersuite "On-Chain Clock":
     let future = (getTime() + 42.years).toUnix
     discard await ethProvider.send("evm_setNextBlockTimestamp", @[%future])
     discard await ethProvider.send("evm_mine")
+
+    # Ensure that state updates are sync with WS
+    discard await ethProvider.getBlock(BlockTag.latest)
+
     check eventually clock.now() == future
 
   test "can wait until a certain time is reached by the chain":
--- a/tests/ethertest.nim
+++ b/tests/ethertest.nim
@@ -11,19 +11,30 @@ import ./checktest
 ## changes to the blockchain do not persist.
 template ethersuite*(name, body) =
   asyncchecksuite name:
-    var ethProvider {.inject, used.} = JsonRpcProvider.new("ws://localhost:8545")
-    var accounts {.inject, used.} = waitFor ethProvider.listAccounts()
+    var ethProvider {.inject, used.}: JsonRpcProvider
+    var accounts {.inject, used.}: seq[Address]
     var snapshot: JsonNode
+    var fut: Future[void]
+
+    proc resubscribeOnTimeout() {.async.} =
+      while true:
+        await sleepAsync(5.int64.minutes)
+        #await ethProvider.resubscribeAll()
 
     setup:
+      ethProvider = JsonRpcProvider.new("ws://localhost:8545")
       snapshot = await send(ethProvider, "evm_snapshot")
+      accounts = await ethProvider.listAccounts()
 
+      fut = resubscribeOnTimeout()
     teardown:
+      if not fut.isNil:
+        await fut.cancelAndWait()
+
       discard await send(ethProvider, "evm_revert", @[snapshot])
 
+      await ethProvider.close()
     body
-
-    waitFor ethProvider.close()
 
 export asynctest
 export ethers except `%`
--- a/tests/ethertest.nim
+++ b/tests/ethertest.nim
@@ -11,22 +11,19 @@ import ./checktest
 ## changes to the blockchain do not persist.
 template ethersuite*(name, body) =
   asyncchecksuite name:
-    var ethProvider {.inject, used.}: JsonRpcProvider
-    var accounts {.inject, used.}: seq[Address]
+    var ethProvider {.inject, used.} = JsonRpcProvider.new("ws://localhost:8545")
+    var accounts {.inject, used.} = waitFor ethProvider.listAccounts()
     var snapshot: JsonNode
 
     setup:
-      ethProvider = JsonRpcProvider.new(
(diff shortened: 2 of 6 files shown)


emizzle (commented):
Nice job. What about handling websocket resubscriptions after 5 minutes?

2-towns:
> Nice job. What about handling websocket resubscriptions after 5 minutes?

I am still exploring the issue, I am not sure how to reproduce it.

tests/integration/multinodes.nim
```diff
@@ -92,6 +92,11 @@ template multinodesuite*(name: string, body: untyped) =
     var accounts {.inject, used.}: seq[Address]
     var snapshot: JsonNode
 
+    if nodeConfigs.hardhat.isNone:
+      ethProvider = JsonRpcProvider.new(
+        jsonRpcProviderUrl, pollingInterval = chronos.milliseconds(100)
+      )
+
```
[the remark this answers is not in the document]
2-towns: > 🟡 I think I understand that your motivation for making this change here was so that ethProvider would not need to be reinstantiated during each test setup when Hardhat is run "globally". Is that correct?

Yes exactly. 

So I did that first because I was facing an error with the WebSocket connection. Moreover, I feel it makes more sense to open one WebSocket connection for all the suite tests and then close it at the end. What is the need for opening a new connection for each test? 
But after taking another look today, I realized the real problem was that `ethProvider.close()` was called before sending `evm_revert`. 
So I can revert the changes, I don't have a strong option either.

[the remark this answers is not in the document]
2-towns: > We don't need to specify `pollingInterval` for websockets. We can use the default value.

Yeah I forgot to remove it after copy / paste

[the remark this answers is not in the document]
marcinczenko: On the other hand it is nice to start each test fresh with the same initial state, so if the connections are cheap maybe it is better to create a new one for each test?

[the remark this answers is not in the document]
2-towns: Not super convinced but yes okay, I reverted the changes

tests/ethertest.nim
```diff
@@ -14,18 +14,26 @@ template ethersuite*(name, body) =
     var ethProvider {.inject, used.}: JsonRpcProvider
     var accounts {.inject, used.}: seq[Address]
     var snapshot: JsonNode
+    var fut: Future[void]
+
+    proc resubscribeOnTimeout() {.async.} =
+      while true:
+        await sleepAsync(5.int64.minutes)
+        #await ethProvider.resubscribeAll()
```
2-towns, line 22: @emizzle What do you think about this approach ? I would add a `resubscribeAll` to `JsonRpcProvider` in `nim-ethers`.

emizzle: I don't mind the idea, though I'm not convinced we should put this functionality in ethers, as it is a client-specific (hardhat) workaround. However, i'm not entirely sure we'll have a choice. 

Also, we are going to run into issues when we cancel all subscriptions and then resubscribe: there will have been blocks mined in between that potentially have events in the logs and we may not have subscribed in time to call the respective subscription callbacks. I believe we discussed this in the past and we reasoned that we might need to inspect the logs for the blocks that had been mined between unsubscribing and re-subscribing. Maybe we could get away with simply re-subscribing only (not unsubscribing), then deduplicating subscription callbacks somehow?

2-towns: I found another approach to keep the resubscribe in the client, but it requires one change in ethers: move the `logEvents` (currently in `PollingSubscriptions`) to `JsonRpcSubscriptions`. Thanks to that, we can keep track of the events per subscription in `WebSocketSubscriptions` as well (see [here](https://github.com/codex-storage/nim-ethers/pull/112)).

Then, in the client, I access the private fields of the `JsonRpcSubscriptions` and register first a new subscription and then remove the old one so we don't lose any block mined event.

tests/helpers.nim
```diff
@@ -4,6 +4,12 @@ import helpers/templeveldb
 
 import std/sequtils, chronos
 
+from std/importutils import privateAccess
+# from pkg/json_rpc/rpcclient import RpcClient
+# from pkg/ethers import JsonRpcProvider, JsonRpcSubscriptions
```
AuHau, line 9: Remove commented out imports please ;-)

AuHau (approved):
Well the Hardhat bug is just unfortunate :-( It is hacky approach, but I can't think of better one, so LGTM.

AuHau:
Actually, I have (I think 😅) a better idea - what about adding this functionality to the nim-ethers **under compile flag** that would do the resubscription automatically in nim-ethers? No hacky private access needed. It will be cleanly turned on only when the nim-ether's user needs... IMHO better then this.

emizzle (approved):

tests/contracts/testClock.nim
```diff
@@ -23,6 +23,10 @@ ethersuite "On-Chain Clock":
     let future = (getTime() + 42.years).toUnix
     discard await ethProvider.send("evm_setNextBlockTimestamp", @[%future])
     discard await ethProvider.send("evm_mine")
+
+    # Ensure that state updates are sync with WS
+    discard await ethProvider.getBlock(BlockTag.latest)
```
emizzle, line 28: How does fetching the latest block ensure state updates are synchronous?

2-towns: I am not sure I was gonna to ask for a review. 

But my guess is that with Websocket, since it is asynchronous and does not wait for the state update, the client might not immediately reflect the new block's state after evm_mine, so we need to get the latest block to ensure the client is synced with the blockchain.

2-towns: @emizzle are these changes okay for you ?

PR #1167, opened 2025-03-20
deps: upgrade libp2p & constantine

munna0908:
This PR upgrades libp2p and constantine packages


dryajov (approved):

codex/blockexchange/protobuf/message.nim
```diff
@@ -254,16 +254,14 @@ proc decode*(
 proc protobufDecode*(_: type Message, msg: seq[byte]): ProtoResult[Message] =
   var
     value = Message()
-    pb = initProtoBuffer(msg, maxSize = MaxMessageSize)
+    pb = initProtoBuffer(msg)
```
dryajov, line 257: Just curious why we're removing the limit here?

PR #1169, opened 2025-03-25
make sure we do not call "get" on unverified Result while fetching in batches

marcinczenko:
While fetching blocks in batches we were doing the following:

```nim
while not iter.finished:
  let blocks = collect:
    for i in 0 ..< batchSize:
      if not iter.finished:
        let address = BlockAddress.init(cid, iter.next())
        if not (await address in self.networkStore) or fetchLocal:
          self.networkStore.getBlock(address)

  let res = await allFinishedFailed(blocks)
  if res.failure.len > 0:
    trace "Some blocks failed to fetch", len = res.failure.len
    return failure("Some blocks failed to fetch (" & $res.failure.len & " )")

  if not onBatch.isNil and
      batchErr =? (await onBatch(blocks.mapIt(it.read.get))).errorOption:
    return failure(batchErr)
```

Above in:

```nim
let res = await allFinishedFailed(blocks)
if res.failure.len > 0:
  trace "Some blocks failed to fetch", len = res.failure.len
  return failure("Some blocks failed to fetch (" & $res.failure.len & " )")
```

we are only testing the futures, but not the underlying result. `getBlock` returns `Future[?!Block]` which means that even for completed futures the underlying result can still be a failure.

Then in `onBatch(blocks.mapIt(it.read.get))` we are calling `get` on potentially failed `Result`, which will endup in:

```
Unhandled defect: Trying to access value with err Result [ResultDefect]
```

This PR should fix that.


dryajov (approved):
Yeah, we shouldn't be doing `get()` anywhere except tests.

munna0908 (approved):

PR #1178, opened 2025-03-28
chore(openapi): add required parameters

2-towns:
While working on [generating the typescript types from the openapi yaml file](https://github.com/codex-storage/codex-js/blob/4750e0b6fe9430cef3d85b15bbddd937f454cbb8/src/openapi.ts#L348), I noticed that the required parameter was missing for several definitions. 

This PR add those required parameters and fix few minor mistakes.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -27,10 +27,6 @@ components:
       maxLength: 66
       example: 0x...
 
-    BigInt:
-      type: string
-      description: Integer represented as decimal string
-
     Cid:
       type: string
       description: Content Identifier as specified at https://github.com/multiformats/cid
@@ -55,17 +51,18 @@ components:
       description: The amount of tokens paid per byte per second per slot to hosts the client is willing to pay
 
     Duration:
-      type: string
-      description: The duration of the request in seconds as decimal string
+      type: integer
+      format: int64
+      description: The duration of the request in seconds
 
     ProofProbability:
       type: string
       description: How often storage proofs are required as decimal string
 
     Expiry:
-      type: string
+      type: integer
+      format: int64
       description: A timestamp as seconds since unix epoch at which this request expires if the Request does not find requested amount of nodes to host the data.
-      default: 10 minutes
 
     SPR:
       type: string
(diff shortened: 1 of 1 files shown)


2-towns:
For better context: https://swagger.io/docs/specification/v3_0/data-models/data-types/#required-properties 

> By default, all object properties are optional. You can specify the required properties in the required list

openapi.yaml
```diff
@@ -283,7 +326,12 @@ components:
     StorageAsk:
       type: object
       required:
+        - slots
+        - slotSize
+        - duration
+        - proofProbability
         - pricePerBytePerSecond
+        - maxSlotLoss
```
marcinczenko, line 334: Does it make sense to also add `collateralPerByte`? It is correct returned by the rest interface (e.g. from `/storage/purchases/{id}`) but for some reason we do not have here...

openapi.yaml
```diff
@@ -870,7 +945,7 @@ paths:
         "200":
           description: Node's SPR
           content:
-            plain/text:
+            text/plain:
```
marcinczenko, line 948: Nice catch!

marcinczenko (approved):
Just small comment from me on perhaps including `collateralPerByte` in storage ask.

AuHau (approved):
LGTM 👍

openapi.yaml
```diff
@@ -173,25 +197,35 @@ components:
...
               description: Unused size of availability's storage in bytes as decimal string
+              readOnly: true
+            totalRemainingCollateral:
+              type: string
+              description: Total collateral effective (in amount of tokens) that can be used for matching requests
+              readOnly: true
 
     SalesAvailabilityCREATE:
       allOf:
         - $ref: "#/components/schemas/SalesAvailability"
```
AuHau, line 221: I think we don't need this one then no?

openapi.yaml
2-towns: Oh yes thanks ! I spent a lot of time to review everything but hard to catch everything.

openapi.yaml
2-towns: Yes I wasn't sure. We can delete it because it is not bringing a real value or we can keep it to have the same pattern as SalesAvailabilityREAD. 

But I guess we can just remove it.

PR #1179, opened 2025-03-29
checked exceptions in stores

marcinczenko:
While working on BitTorrent stuff and trying to keep exceptions checked, I continuously hit the wall of set of "stores" which mostly do not use anything but `.async.`.

This PR adds checked exceptions to all stores and dependent abstractions.

To keep the PR still maintainable it focuses on the "stores", thus it has pretty much a minimal set of changes to get them compile.

Most of the changes are easy and syntactical, but there were a few exceptions where it was a big pain to move forward.
It all boils down to `AsynIter`, most probably created before `async: (raises: [CancelledError])` era, that makes checking exceptions much harder where it is used.

The only way to get it clean was for me to create a version of AsyncIter (called `SafeAsyncIter`) that uses `Results` and only ever allows for `async: (raises: [CancelledError])`. I applied it to `listBlocks` and `getBlockExpirations` and doing this I could remove problematic situations (especially in block maintenance). `SafeAsynIter` has the same API as `AsynIter` and requires very little work (pretty much syntactical) to use it in place of `AsyncIter`, which is still used in some other places, which can be dealt with in a separate PR to keep this one small.

To get some insights about challenges that you may encounter when working with checked (raising) exception, you may like to check the following note: https://publish.obsidian.md/bittorrent/10+Notes/Async+checked+(raising)+exceptions.

### DIFF
--- a/codex/errors.nim
+++ b/codex/errors.nim
@@ -7,6 +7,8 @@
 ## This file may not be copied, modified, or distributed except according to
 ## those terms.
 
+{.push raises: [].}
+
 import std/options
 import std/sugar
 import std/sequtils
--- a/codex/errors.nim
+++ b/codex/errors.nim
@@ -65,7 +65,7 @@ proc allFinishedFailed*[T](
   return res
 
 proc allFinishedValues*[T](
-    futs: seq[Future[T]]
+    futs: auto
 ): Future[?!seq[T]] {.async: (raises: [CancelledError]).} =
   ## If all futures have finished, return corresponding values,
   ## otherwise return failure
--- a/codex/errors.nim
+++ b/codex/errors.nim
@@ -47,7 +47,7 @@ func toFailure*[T](exp: Option[T]): Result[T, ref CatchableError] {.inline.} =
     T.failure("Option is None")
 
 proc allFinishedFailed*[T](
-    futs: seq[Future[T]]
+    futs: auto
 ): Future[FinishedFailed[T]] {.async: (raises: [CancelledError]).} =
   ## Check if all futures have finished or failed
   ##
--- a/codex/node.nim
+++ b/codex/node.nim
@@ -485,7 +485,7 @@ proc iterateManifests*(self: CodexNodeRef, onManifest: OnManifest) {.async.} =
     return
 
   for c in cidsIter:
-    if cid =? await c:
+    if cid =? (await cast[Future[?!Cid].Raising([CancelledError])](c)):
       without blk =? await self.networkStore.getBlock(cid):
         warn "Failed to get manifest block by cid", cid
         return
--- a/codex/node.nim
+++ b/codex/node.nim
@@ -187,7 +187,7 @@ proc fetchBatched*(
   #   )
 
   while not iter.finished:
-    let blockFutures: seq[Future[?!blocktype.Block].Raising([CancelledError])] = collect:
+    let blockFutures = collect:
       for i in 0 ..< batchSize:
         if not iter.finished:
           let address = BlockAddress.init(cid, iter.next())
@@ -197,8 +197,7 @@ proc fetchBatched*(
     if blockFutures.len == 0:
       continue
 
-    without blockResults =?
-      await allFinishedValues(cast[seq[Future[?!blocktype.Block]]](blockFutures)), err:
+    without blockResults =? await allFinishedValues[?!bt.Block](blockFutures), err:
       trace "Some blocks failed to fetch", err = err.msg
       return failure(err)
 
@@ -485,7 +484,7 @@ proc iterateManifests*(self: CodexNodeRef, onManifest: OnManifest) {.async.} =
     return
 
   for c in cidsIter:
-    if cid =? (await cast[Future[?!Cid].Raising([CancelledError])](c)):
+    if cid =? await c:
       without blk =? await self.networkStore.getBlock(cid):
         warn "Failed to get manifest block by cid", cid
         return
--- a/codex/node.nim
+++ b/codex/node.nim
@@ -78,7 +78,9 @@ type
   CodexNodeRef* = ref CodexNode
 
   OnManifest* = proc(cid: Cid, manifest: Manifest): void {.gcsafe, raises: [].}
-  BatchProc* = proc(blocks: seq[bt.Block]): Future[?!void] {.gcsafe, raises: [].}
+  BatchProc* = proc(blocks: seq[bt.Block]): Future[?!void] {.
+    gcsafe, async: (raises: [CancelledError])
+  .}
 
 func switch*(self: CodexNodeRef): Switch =
   return self.switch
@@ -109,7 +111,9 @@ proc storeManifest*(
 
   success blk
 
-proc fetchManifest*(self: CodexNodeRef, cid: Cid): Future[?!Manifest] {.async.} =
+proc fetchManifest*(
+    self: CodexNodeRef, cid: Cid
+): Future[?!Manifest] {.async: (raises: [CancelledError]).} =
   ## Fetch and decode a manifest block
   ##
 
@@ -144,7 +148,7 @@ proc connect*(
 
 proc updateExpiry*(
     self: CodexNodeRef, manifestCid: Cid, expiry: SecondsSince1970
-): Future[?!void] {.async.} =
+): Future[?!void] {.async: (raises: [CancelledError]).} =
   without manifest =? await self.fetchManifest(manifestCid), error:
     trace "Unable to fetch manifest for cid", manifestCid
     return failure(error)
@@ -154,7 +158,7 @@ proc updateExpiry*(
         self.networkStore.localStore.ensureExpiry(manifest.treeCid, it, expiry)
       )
 
-    let res = await allFinishedFailed(ensuringFutures)
+    let res = await allFinishedFailed(cast[seq[Future[?!void]]](ensuringFutures))
     if res.failure.len > 0:
       trace "Some blocks failed to update expiry", len = res.failure.len
       return failure("Some blocks failed to update expiry (" & $res.failure.len & " )")
@@ -172,7 +176,7 @@ proc fetchBatched*(
     batchSize = DefaultFetchBatch,
     onBatch: BatchProc = nil,
     fetchLocal = true,
-): Future[?!void] {.async, gcsafe.} =
+): Future[?!void] {.async: (raises: [CancelledError]), gcsafe.} =
   ## Fetch blocks in batches of `batchSize`
   ##
 
@@ -183,14 +187,18 @@ proc fetchBatched*(
   #   )
 
   while not iter.finished:
-    let blockFutures = collect:
+    let blockFutures: seq[Future[?!blocktype.Block].Raising([CancelledError])] = collect:
       for i in 0 ..< batchSize:
         if not iter.finished:
           let address = BlockAddress.init(cid, iter.next())
           if not (await address in self.networkStore) or fetchLocal:
             self.networkStore.getBlock(address)
 
-    without blockResults =? await allFinishedValues(blockFutures), err:
+    if blockFutures.len == 0:
+      continue
+
+    without blockResults =?
+      await allFinishedValues(cast[seq[Future[?!blocktype.Block]]](blockFutures)), err:
       trace "Some blocks failed to fetch", err = err.msg
       return failure(err)
 
@@ -215,7 +223,7 @@ proc fetchBatched*(
     batchSize = DefaultFetchBatch,
     onBatch: BatchProc = nil,
     fetchLocal = true,
-): Future[?!void] =
+): Future[?!void] {.async: (raw: true, raises: [CancelledError]).} =
   ## Fetch manifest in batches of `batchSize`
   ##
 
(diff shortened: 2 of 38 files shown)


codex/errors.nim
```diff
@@ -63,7 +65,7 @@ proc allFinishedFailed*[T](
   return res
 
 proc allFinishedValues*[T](
-    futs: seq[Future[T]]
+    futs: auto
```
dryajov, line 68: Hmm, why `auto` here?

codex/stores/cachestore.nim
```diff
@@ -164,14 +174,7 @@ method listBlocks*(
...
   )
-
-  return success(
-    map[Cid, ?Cid](
-      iter,
-      proc(cid: Cid): Future[?Cid] {.async.} =
-        some(cid),
-    )
-  )
+  success(iter)
```
dryajov, line 177: Why don't we need the map anymore?

codex/utils/safeasynciter.nim
```diff
@@ -0,0 +1,233 @@
...
+  SafeAsyncIter[U].new(
+    genNext = () => iter.next().flatMap(fn),
+    isFinished = () => iter.finished,
+    finishOnErr = finishOnErr,
+  )
+
+proc mapFilter*[T, U](
+    iter: SafeAsyncIter[T],
+    mapPredicate: SafeFunction[?!T, Option[?!U]],
+    finishOnErr: bool = true,
```
dryajov, line 173: Why do we need this flag now?

dryajov (approved):
Overall this looks great, thanks for puting in the work! 

Ok, I think it is safe to replace `AsyncIter` with `SafeAsyncIter` as it stands. This might happen in a follow up PR, to keep things moving.

Other than that and a few questions I have, this looks good.

codex/stores/cachestore.nim
marcinczenko: The `map` was just converting the yielded value to an optional because advertiser expects an `Option[Cid]` to be yielded. `SafeAsyncIter` yields a `Result` and so it can be directly used by the advertiser, and so mapping is not needed anymore. Also, in the original version, mapping was happening after filtering, and if there were something unexpected happening while iterating over `self.cache.keys` it would result in an exception. With `SafeAsyncIter`, if there is any exception while iterating it will result in a failure, which would then be detected by the advertiser. Because by default `SafeAsyncIter` stops on failure, we maintain the same semantics as before where an exception would interrupt the flow.

codex/utils/safeasynciter.nim
marcinczenko: Do you mean `finishOnErr`? If so, this is because, previously an exception would be thrown if yielding next item raises. In `SafeAsyncIter`, `next` is not allowed to throw anything but `CancelledError` and the error is communicated by returning `failure`. This flag allows the user to choose what to do with that failure - should the failure be just yielded and allowing iterator to continue, or shall it stop (the default).

codex/errors.nim
dryajov: I guess this is fine, but 🤔

dryajov:
I think this is good to go right now.

codex/utils/safeasynciter.nim
```diff
@@ -0,0 +1,233 @@
...
+): SafeAsyncIter[U] =
+  ## Creates new Iter from a slice
+  ##
+
+  let iter = Iter[U].new(slice)
+  mapAsync[U, U](
+    iter,
+    proc(i: U): Future[?!U] {.async: (raises: [CancelledError]).} =
+      success[U](i),
+  )
```
markspanbroek, line 120: Shouldn't `mapAsync` have parameter `finishOnErr = finishOnErr`, like in the proc below?

tests/codex/utils/testsafeasynciter.nim
```diff
@@ -0,0 +1,389 @@
...
+
+    expect CancelledError:
+      for fut in iter2:
+        if i =? (await fut):
+          collected.add(i)
+        else:
+          fail()
+
+    check:
+      collected == @["0", "1"]
```
markspanbroek, line 388: I find this test hard to understand. From the condition `(i < 3)` from the `mapFilter` invocation above I would expect the iterator to return values `"0", "1", "2"` .

markspanbroek (approved):
I like this a lot, thanks @marcinczenko! Just a few comments below.

codex/node.nim
```diff
@@ -154,7 +158,7 @@ proc updateExpiry*(
         self.networkStore.localStore.ensureExpiry(manifest.treeCid, it, expiry)
       )
 
-    let res = await allFinishedFailed(ensuringFutures)
+    let res = await allFinishedFailed(cast[seq[Future[?!void]]](ensuringFutures))
```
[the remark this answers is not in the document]
marcinczenko: Good point. I left it because I wanted to be visible why I had to use `auto` in `allFinishedValues` in `codex/errors.nim`. The same basically applies to `allFinishedFailed`, which has the following signature:

```nim
proc allFinishedFailed*[T](
    futs: seq[Future[T]]
): Future[FinishedFailed[T]] {.async: (raises: [CancelledError]).} =
```

Having `seq[Future[T]]` is what requires the cast and we cannot use the actual explicit type here (which would be `Future[T].Raising([CancelledError])`) because of the reasons I explain in the [presentation](https://drive.proton.me/urls/Q562MNN8VG#6iy1sTyjagcW) at the end of Section "Generic Types". The correct solution here is to use `auto` (the same as we did in the above mentioned `allFinishedValues`).

Thus, I will resolve it by following the same practice in `allFinishedFailed` as I already do in `allFinishedValues`: it will use `auto` as the type of the argument.

codex/node.nim
```diff
@@ -648,13 +657,15 @@ proc onStore(
...
+  proc updateExpiry(
+      blocks: seq[bt.Block]
+  ): Future[?!void] {.async: (raises: [CancelledError]).} =
     trace "Updating expiry for blocks", blocks = blocks.len
 
     let ensureExpiryFutures =
       blocks.mapIt(self.networkStore.ensureExpiry(it.cid, expiry.toSecondsSince1970))
 
-    let res = await allFinishedFailed(ensureExpiryFutures)
+    let res = await allFinishedFailed(cast[seq[Future[?!void]]](ensureExpiryFutures))
```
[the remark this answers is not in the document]
marcinczenko: Addressed in https://github.com/codex-storage/nim-codex/pull/1179#discussion_r2100286753

tests/codex/utils/testsafeasynciter.nim
marcinczenko: Yes, this test is a bit confusing. I added the following comment to the test that explains this:

```nim
# When `mapFilter` is called, it already performs first iteration
# step: this is necessary, so that if there is nothing there left
# after filtering, the iterator state should be market as "finished"
# before event trying to call `next()` for the very first time (a standard
# practice is for the called to check if the iterator is finished before
# attempting to call `next()`). Thus, internally, the value that is to be
# returned for the first iteration is already resolved and ready to be returned.
# And this follows in the same for the next iterations. On calling `next()`
# the iterator first makes a temporary copy of the value already captured in
# the precious step, awaits for the next value (and if there is no more values
# to be returned it marks the iterator as finished), and then returns the
# local copy of the previously captured value.
# Now, to make sure that this mechanism works, and to document its
# cancellation semantics, this test shows that when the async predicate
# function is cancelled, this cancellation has immediate effect, which means
# that `next()` (or more precisely `getNext()` in `mapFilter` function), is 
# interrupted immediately. If this is the case, the the iterator be interrupted
# before `next()` returns this locally captured value from the previous
# iteration and this is exactly the reason why at the end of the test
# we expect only values "0" and "1" to be collected while value "2" - although
# already resolved and ready to be returned, is not returned because of the
# cancellation of the async predicate function.
```

codex/utils/safeasynciter.nim
marcinczenko: Yes, it should. Good catch! Fixed.

codex/stores/cachestore.nim
```diff
@@ -145,12 +148,19 @@ method listBlocks*(
   proc isFinished(): bool =
     return finished(cids)
 
-  proc genNext(): Future[Cid] {.async.} =
-    cids()
+  proc genNext(): Future[?!Cid] {.async: (raises: [CancelledError]).} =
+    try:
+      let cid = cids()
+      success(cid)
+    except Exception as err:
```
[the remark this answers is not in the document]
marcinczenko: Very right observation. This comes from the `lrucache` which `raises: [Defect]` for `keys` iterator. LRUCache perhaps needs to be updated in the near future if that's possible.

codex/stores/queryiterhelper.nim
```diff
@@ -63,3 +100,29 @@ proc filterSuccess*[T](
...
+): Future[SafeAsyncIter[tuple[key: Key, value: T]]] {.
+    async: (raises: [CancelledError])
+.} =
+  ## Filters out any items that are not success
+
+  proc mapping(
+      resOrErr: ?!QueryResponse[T]
+  ): Future[Option[?!KeyVal[T]]] {.async: (raises: [CancelledError]).} =
+    without res =? resOrErr, error:
+      error "Error occurred when getting QueryResponse", msg = error.msg
```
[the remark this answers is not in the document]
marcinczenko: Yes, I would agree that this should rather be `debug`. Here we are filtering out unsuccessful query responses, meaning, that unsuccessful responses is something we do expect as sort-of "normal" or "possible" thing to happen. `error` indeed indicates that this is more "catastrophic".

The same strategy is being applied in other places related to query responses, e.g. in the `getBlockExpirations` which calls the `filterSuccess`, which we are commenting here. In my filling changing the logging strategy falls outside of this PR, so I would suggest we get back to it when we review the logging practices on another occasion. Would you agree?

codex/stores/cachestore.nim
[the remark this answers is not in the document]
marcinczenko: It looks like it is a Defect though:
image

tests/codex/utils/testsafeasynciter.nim
```diff
@@ -0,0 +1,389 @@
...
+      iter1,
+      proc(i: ?!int): Future[Option[?!string]] {.async: (raises: [CancelledError]).} =
+        if i =? i:
+          if (i < 3):
+            return some(success($i))
+        return await fut,
+    )
+
+    proc cancelFut(): Future[void] {.async.} =
+      await sleepAsync(100.millis)
```
dryajov, line 398: Hmm, not sure if we want to make this non-deterministic with a `sleep`, perhaps it's better to use another future for signaling, which makes things less prone to intermittent failures?

codex/stores/cachestore.nim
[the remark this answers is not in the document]
marcinczenko: It is here:

```nim
{.push raises: [Defect].}

# ...

iterator keys*[K,T](cache: LruCache[K,T]): lent K =
  var it = cache.list.head
  while not isNil(it):
    yield it.value.key
    it = it.next
```

dryajov:
I think we can merge this, but I would address the non-deterministic sleep in https://github.com/codex-storage/nim-codex/pull/1179#discussion_r2100831743 (could be a follow up PR).

codex/stores/cachestore.nim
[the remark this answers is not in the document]
marcinczenko: But strangely enough, if I change the `{.push raises: [Defect].}` in the lrucache then I still get the compiler error:

```bash
 Error: Exception can raise an unlisted exception: Exception
```

and I have that now:

```nim
proc genNext(): Future[?!Cid] {.async: (raises: [CancelledError]).} =
    let cid = cids()
    success(cid)
    # try:
    #   let cid = cids()
    #   success(cid)
    # except Exception as err:
    #   failure(err.msg)
```

So I am a bit puzzled right now...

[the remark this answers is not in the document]
marcinczenko: Ok, I think I got this. So we have:

```nim
{.push raises: [].}
```

on top of `codex/stores/cachestore.nim`.

But it does not apply to a return type that happens to be a iterator type:

```nim
func cids(self: CacheStore): (iterator (): Cid {.gcsafe.}) =
  return
    iterator (): Cid =
      for cid in self.cache.keys:
        yield cid
```

If I annotate it explicitly with `raises: []` like this:

```nim
func cids(self: CacheStore): (iterator (): Cid {.gcsafe, raises: [].}) =
  return
    iterator (): Cid =
      for cid in self.cache.keys:
        yield cid
```

Then I do not have that exception any more even though LRUChache still raises `Defect`.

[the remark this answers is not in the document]
marcinczenko: I have created an example to replicate this in an isolated context:

```nim
import pkg/chronos
import std/tables

# without `raises: []` it will fail with "Error: Exception can raise an unlisted exception: Exception" in genNext
# func cids(): (iterator (): string {.gcsafe.}) = 
func cids(): (iterator (): string {.gcsafe, raises: [].}) =
  let cache = {
    "a": 1,
    "b": 2,
    "c": 3
  }.toTable
  return
    iterator (): string =
      for cid in cache.keys:
        yield cid

proc f1() {.raises: [].} =
  let cids = cids()

  proc genNext(): Future[?!string] {.async: (raises: [CancelledError]).} =
    success(cids())
```

What is the right explanation of this?

tests/codex/utils/testsafeasynciter.nim
marcinczenko: Yes, perhaps sleep is a bit too risky in that test. There will be follow-up on removing remaining references to the old async iter, and then I get back to it and use something more deterministic. I will than merge this one.

codex/stores/cachestore.nim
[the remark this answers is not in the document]
marcinczenko: Same here = let's reiterate on this in the follow-up PR. It is now pretty much clear and eventual changes here will not have any functional impact (I might be dealing with some compiler "quirks" here).

PR #1094, opened 2025-01-31
perf: contract storage optimizations

AuHau:

PR #1182, opened 2025-03-31
chore: adding perf optimization tweaks to openapi

AuHau:
The contract's [perf optimization PR](https://github.com/codex-storage/nim-codex/pull/1094) changed some API types from string to integer without documenting them, so I am mainly adding these.


2-towns (approved):
Thanks a lot @AuHau . It will save me a lot of `parseInt` in the codex sdk.

PR #1183, opened 2025-04-01
fix: Support for mapping multiple listener address 

### LINKED ISSUE #1158: [BUG] Client hanging on exit
I've lately noticed that attempting to exit Codex will print:

```text
➜  nim-codex git:(master) ✗ ./build/codex --data-dir=./codex-data
...
NTC 2025-03-12 11:13:43.390-03:00 Shutting down after having received SIGINT topics="codex" tid=48666
NTC 2025-03-12 11:13:43.390-03:00 Stopping codex node                        topics="codex node" tid=48666
NTC 2025-03-12 11:13:43.393-03:00 Stopping Codex                             topics="codex" tid=48666
INF 2025-03-12 11:13:45.775-03:00 Exiting blockexc task runner               topics="codex blockexcengine" tid=48666
INF 2025-03-12 11:13:45.775-03:00 Exiting blockexc task runner               topics="codex blockexcengine" tid=48666
INF 2025-03-12 11:13:45.775-03:00 Exiting blockexc task runner               topics="codex blockexcengine" tid=48666
INF 2025-03-12 11:13:45.775-03:00 Exiting blockexc task runner               topics="codex blockexcengine" tid=48666
INF 2025-03-12 11:13:45.775-03:00 Exiting blockexc task runner               topics="codex blockexcengine" tid=48666
INF 2025-03-12 11:13:45.775-03:00 Exiting blockexc task runner               topics="codex blockexcengine" tid=48666
INF 2025-03-12 11:13:45.775-03:00 Exiting blockexc task runner               topics="codex blockexcengine" tid=48666
INF 2025-03-12 11:13:45.775-03:00 Exiting blockexc task runner               topics="codex blockexcengine" tid=48666
INF 2025-03-12 11:13:45.775-03:00 Exiting blockexc task runner               topics="codex blockexcengine" tid=48666
INF 2025-03-12 11:13:45.775-03:00 Exiting blockexc task runner               topics="codex blockexcengine" tid=48666
NTC 2025-03-12 11:13:45.775-03:00 REST service stopped                       tid=48666 address=127.0.0.1:8080
INF 2025-03-12 11:13:45.776-03:00 Exiting advertise task loop                topics="codex discoveryengine advertiser" tid=48666
NTC 2025-03-12 11:13:45.776-03:00 Exited codex                               topics="codex" tid=48666
```
and then hang. Insisting on the SIGTERM will print:

```text
^CNTC 2025-03-12 11:15:04.702-03:00 Shutting down after having received SIGINT topics="codex" tid=48666
NTC 2025-03-12 11:15:04.702-03:00 Stopping codex node                        topics="codex node" tid=48666
NTC 2025-03-12 11:15:04.702-03:00 REST service stopped                       tid=48666 address=127.0.0.1:8080
WRN 2025-03-12 11:15:04.702-03:00 service is already stopped                 topics="libp2p switch" tid=48666
WRN 2025-03-12 11:15:04.702-03:00 TCP transport already stopped              topics="libp2p tcptransport" tid=48666
WRN 2025-03-12 11:15:04.703-03:00 Failed to stop NAT port mapping renewal thread topics="nat" tid=48666 exc="/home/giuliano/Work/Status/nim-codex/vendor/nim-taskpools/taskpools/instrumentation/contracts.nim(91, 15) `\nworkerContext.currentTask.task.isRootTask()` \n    Contract violated for pre-condition at taskpools.nim:394\n        workerContext.currentTask.task.isRootTask()\n    The following values are contrary to expectations:\n        workerContext.currentTask.task.isRootTask()  [Worker 0 on taskpool 0x0000616a041f6ac0]\n"
```

If I run codex with `codex --nat:none`, then the problem doesn't manifest itself. That taken together with the log above suggests this is related to NAT.

munna0908:
This PR will fix #1158

### DIFF
--- a/codex/codex.nim
+++ b/codex/codex.nim
@@ -163,8 +163,8 @@ proc start*(s: CodexServer) {.async.} =
 
   await s.codexNode.switch.start()
 
-  let (announceAddrs, discoveryAddrs) = nattedAddress(
-    s.config.nat, s.codexNode.switch.peerInfo.addrs, s.config.discoveryPort
+  let (announceAddrs, discoveryAddrs) = s.codexNode.nat.nattedAddress(
+    s.codexNode.switch.peerInfo.addrs, s.config.discoveryPort
   )
 
   s.codexNode.discovery.updateAnnounceRecord(announceAddrs)
@@ -287,7 +287,7 @@ proc new*(
       interval = config.blockMaintenanceInterval,
       numberOfBlocksPerInterval = config.blockMaintenanceNumberOfBlocks,
     )
-
+    natManager = NatManager.new(config.nat)
     peerStore = PeerCtxStore.new()
     pendingBlocks = PendingBlocksManager.new()
     advertiser = Advertiser.new(repoStore, discovery)
@@ -312,6 +312,7 @@ proc new*(
       discovery = discovery,
       prover = prover,
       taskPool = taskpool,
+      nat = natManager,
     )
 
     restServer = RestServerRef
--- a/codex/nat.nim
+++ b/codex/nat.nim
@@ -28,14 +28,27 @@ const
   PORT_MAPPING_INTERVAL = 20 * 60 # seconds
   NATPMP_LIFETIME = 60 * 60 # in seconds, must be longer than PORT_MAPPING_INTERVAL
 
-var
-  upnp {.threadvar.}: Miniupnp
-  npmp {.threadvar.}: NatPmp
-  strategy = NatStrategy.NatNone
-  internalTcpPort: Port
-  externalTcpPort: Port
-  internalUdpPort: Port
-  externalUdpPort: Port
+type NatConfig* = object
+  case hasExtIp*: bool
+  of true: extIp*: IpAddress
+  of false: nat*: NatStrategy
+
+type PortMapping* = object
+  externalTcpPort*: Port
+  internalTcpPort*: Port
+  externalUdpPort*: Port
+  internalUdpPort*: Port
+  description*: string
+
+type NatManager* = ref object
+  portMappings: seq[PortMapping]
+  thread: Thread[ptr NatManager]
+  config: NatConfig
+  upnp: Miniupnp
+  npmp: NatPmp
+  strategy: NatStrategy
+  threadStarted: bool
+  natCloseChan: Channel[bool]
 
 logScope:
   topics = "nat"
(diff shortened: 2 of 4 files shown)


codex/codex.nim
```diff
@@ -287,7 +287,7 @@ proc new*(
       interval = config.blockMaintenanceInterval,
       numberOfBlocksPerInterval = config.blockMaintenanceNumberOfBlocks,
     )
-
+    natManager = NatManager.new(config.nat)
```
gmega, line 290: :yellow_circle: Hm... not sure I'm very comfortable scoping the object that gets its reference passed onto another thread here. What you had before were a bunch of module-private globals so I think that was OK, but here you find yourself with unwanted accesses to fields in this object (in the future) as it's read from another thread.

codex/nat.nim
```diff
@@ -28,14 +28,29 @@ const
...
+  internalTcpPort*: Port
+  externalUdpPort*: Port
+  internalUdpPort*: Port
+  description*: string
+
+type NatManager* = ref object
+  portMappings: seq[PortMapping]
+  thread: Thread[ptr NatManager]
+  config: NatConfig
+  upnp: Miniupnp
```
gmega, line 47: :red_circle: Those were threadvars before and initialized per-thread. Why are we not doing that anymore, and is it safe?

gmega (changes requested):
Overall looks OK, but I'm not convinced about the safety of some of those things. Need to have them clarified first.

codex/codex.nim
gmega: Here you're also initializing a ref object that contains a sequence and then passing it to another thread to access. What am I missing? How's that more GC-safe than having it as a module-private global?

codex/nat.nim
gmega: I wouldn't know, I don't know why they were done as threadvars in the first place, but it would seem to indicate that you can't initialize them in one thread then use them in another. Previously we had them as module-level globals, which are also stable references? Am I missing anything?

codex/codex.nim
gmega: Btw I realize we're not currently accessing NatManager from multiple threads, what I meant to say is that we are leaving a reference to it here which someone in the future might misuse.

dryajov: Good point. But as long as access to `NatManager` is synchronized internally it should be OK. So, `NatManager` should be thread safe for the sake of not messing things up.

dryajov:
I went through the code a bit more today and although it is true that the background thread doesn't mutate anything (as far as I can tell), the code is quite convoluted and fragile overall (I know that this is a consequence of us "borrowing" this from somewhere else, but we should still clean this up sometime soon). 

Nothing should be touching mutating the manager after the initial setup in the main thread, but just in case I added locking around `upnp` and `pmp` objects.

@munna0908 can you please test this a bit more throughly with my current changes to make sure it's all working, then we can merge this as soon as that is done.

dryajov (approved):
I went through the code a bit more today and although it is true that the background thread doesn't mutate anything (as far as I can tell), the code is quite convoluted and fragile overall (I know that this is a consequence of us "borrowing" this from somewhere else, but we should still clean this up sometime soon). 

Nothing should be touching mutating the manager after the initial setup in the main thread, but just in case I added locking around `upnp` and `pmp` objects.

@munna0908 can you please test this a bit more throughly with my current changes to make sure it's all working, then we can merge this as soon as that is done.

PR #1184, opened 2025-04-01
ci: switch out from ubuntu 20.04

### LINKED ISSUE #1141: Switch out from Ubuntu 20.04 in workflows
Recently, during [Chore/update nim version #1052](https://github.com/codex-storage/nim-codex/pull/1052), we've switched to `Ubuntu 24.04` for some of our workflows, but some still are using `Ubuntu 20.04`.

`Ubuntu 20.04` will be deprecated very soon(**2025-04-15**) and we need to solve that - [The Ubuntu 20.04 Actions runner image will begin deprecation on 2025-02-01 and will be fully unsupported by 2025-04-15 #11101](https://github.com/actions/runner-images/issues/11101).

 - [x] [CI](https://github.com/codex-storage/nim-codex/blob/master/.github/workflows/ci.yml) (coverage) workflow
 - [x] ~[Nim matrix](https://github.com/codex-storage/nim-codex/blob/master/.github/workflows/nim-matrix.yml) workflow~ #1140
 - [x] [Release](https://github.com/codex-storage/nim-codex/blob/master/.github/workflows/release.yml) workflow

For Release workflow, it would be good to switch to `Ubuntu 22.04` to have a better binaries compatibility (#939). But we also should take care because of the existing self-hosted runners with a similar labels (#1102).

For CI (coverage) we have a "lcov was updated to 2.x version in ubuntu-latest" compatibility issue - [Having trouble upgrading LCOV to v2.0 in gitHub CI/CD pipeline #238](https://github.com/linux-test-project/lcov/issues/238). And comment from the mentioned PR
 - https://github.com/codex-storage/nim-codex/commit/b5848da18821a1679524b3efc798946ffcc26c79#r151129750
 - https://github.com/codex-storage/nim-codex/pull/1052#discussion_r1908904378

veaceslavdoina:
This PR closes #1141
- We've switched to `ubuntu-latest` (`24.04`) for coverage because 20.04 will be fully unsupported by 2025-04-15
- To fix the issue with the `lcov` v2 (which is a default on ubuntu-24.04) it was required to pass [`--keep-going`](https://man.archlinux.org/man/lcov.1.en) and use default GCC 13 (GCC 14 compatibility issue need to be solved, but let's do it later)
    > Do not stop if error occurs: attempt to generate a result, however flawed.
- Release workflow was switched to `ubuntu-22.04`
- GCC-14 is not easy available on Ubuntu < 24.04 - so, install and use it only on the new versions

And to fix CI, it was required to update submodule [nim-leveldbstatic to 0.2.1](https://github.com/codex-storage/nim-leveldb/pull/4).

### DIFF
--- a/.github/actions/nimbus-build-system/action.yml
+++ b/.github/actions/nimbus-build-system/action.yml
@@ -92,10 +92,16 @@ runs:
       if : ${{ inputs.os == 'linux' && inputs.coverage != 'true' }}
       shell: ${{ inputs.shell }} {0}
       run: |
-        # Add GCC-14 to alternatives
-        sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-14 14
-        # Set GCC-14 as the default
-        sudo update-alternatives --set gcc /usr/bin/gcc-14
+        # Skip for older Ubuntu versions
+        if [[ $(lsb_release -r | awk -F '[^0-9]+' '{print $2}') -ge 24 ]]; then
+          # Install GCC-14
+          sudo apt-get update -qq
+          sudo apt-get install -yq gcc-14
+          # Add GCC-14 to alternatives
+          sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-14 14
+          # Set GCC-14 as the default
+          sudo update-alternatives --set gcc /usr/bin/gcc-14
+        fi
 
     - name: Install ccache on Linux/Mac
       if: inputs.os == 'linux' || inputs.os == 'macos'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,11 +61,7 @@ jobs:
           suggest: true
 
   coverage:
-    # Force to stick to ubuntu 20.04 for coverage because
-    # lcov was updated to 2.x version in ubuntu-latest
-    # and cause a lot of issues.
-    # See https://github.com/linux-test-project/lcov/issues/238
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     steps:
       - name: Checkout sources
         uses: actions/checkout@v4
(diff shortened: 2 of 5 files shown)


2-towns:
`--keep-going` works with GCC 14 @veaceslavdoina  ?

veaceslavdoina:
> `--keep-going` works with GCC 14 @veaceslavdoina ?

No, and this is why we are still using a condition
```yaml
    - name: Install gcc 14 on Linux
      # We don't want to install gcc 14 for coverage (Ubuntu 20.04)
      if : ${{ inputs.os == 'linux' && inputs.coverage != 'true' }}
```

2-towns:
> > `--keep-going` works with GCC 14 @veaceslavdoina ?
> 
> No, and this is why we are still using a condition and rely on the preinstalled GCC 13
> 
> https://github.com/codex-storage/nim-codex/blob/0627d018ed4b7974d879ec52d01ad21ed75dde4e/.github/actions/nimbus-build-system/action.yml#L90-L92

Okay so lcov will run with Ubuntu 22, which will not use GCC 14, so ``--keep-going` works for now. Then we will need to update to Ubuntu 24, we will need to find a solution. Am I right ?

veaceslavdoina:
> Okay so lcov will run with Ubuntu 22, which will not use GCC 14, so `--keep-going` works for now. Then we will need to update to Ubuntu 24, we will need to find a solution. Am I right ?

Almost yes
- lcov is running on `ubuntu-24.04` but uses default available GCC 13, we do not install GCC 14 for codecov
- `ubuntu-22.04` is used only for a Release workflow

And as you mentioned, there is still an issue with lcov v2 and GCC 14 which needs to be solved. That PR is focused on the  `ubuntu-20.04` deprecation and we can check for a solution later.

2-towns (approved):
LGTM

markspanbroek (approved):

PR #1189, opened 2025-04-07
feat(marketplace) - add command line arg for maxPriorityFeePerGas

2-towns:
This PR adds a new command line argument to configure the maxPriorityFeePerGas available now with [EIP-1559 implementation] (https://github.com/codex-storage/nim-ethers/pull/113)

### DIFF
--- a/codex/conf.nim
+++ b/codex/conf.nim
@@ -44,7 +44,7 @@ import ./utils
 import ./nat
 import ./utils/natutils
 
-from ./contracts/config import DefaultRequestCacheSize
+from ./contracts/config import DefaultRequestCacheSize, DefaultMaxPriorityFeePerGas
 from ./validationconfig import MaxSlots, ValidationGroups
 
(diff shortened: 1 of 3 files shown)


codex/conf.nim
```diff
@@ -370,6 +370,15 @@ type
         hidden
       .}: uint16
 
+      maxPriorityFeePerGas* {.
+        desc:
+          "Sets the maximum priority fee per gas for Ethereum EIP-1559 transactions in wei.",
```
emizzle, line 375: Maybe something like this is a bit more clear?
```suggestion
          "Sets the default maximum priority fee per gas for Ethereum EIP-1559 transactions, in wei, when not provided by the network.",
```

emizzle (approved):

AuHau:
Btw. is EIP 1559 available on our PoA network? It is running old version of Geth, but I think it is the last version before The Merge, so it should be available as EIP 1559 was introduced before The Merge if I remember correctly?

But anyhow, nim-ethers will work correctly if EIP 1559 is not available right?

PR #1193, opened 2025-04-15
chore: add marketplace topic to reservations

AuHau:
We use `marketplace` everywhere related to Marketplace, but it is missing in the `reservations` module.


markspanbroek (approved):
Good find :+1:

emizzle (approved):

PR #1195, opened 2025-04-15
chore(marketplace): use hardhat ignition

2-towns:
This PR update the contracts module in order to use hardhat ignition for deployments and replace the [deleted increaseAllowance](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4585) by allowance / approve functions.


markspanbroek:
I ran into the same problem with #1196 and solved it with a lock (https://github.com/codex-storage/nim-codex/pull/1196/commits/94236ca99b05cd07aef86cd9aafae802167f0103) to ensure that we don't get any race conditions while updating the allowance.

tests/integration/hardhatprocess.nim
```diff
@@ -63,7 +63,28 @@ method start*(node: HardhatProcess) {.async.} =
...
+      stdoutHandle = AsyncProcess.Pipe,
+    )
+
+    discard await startProcess(
+      node.executable,
+      node.workingDir,
+      @[
+        "node", "ignition", "deploy", "ignition/modules/marketplace.js", "--network",
+        "localhost",
+      ].concat(node.arguments),
```
emizzle, line 87: Instead of starting a (long running) process, it would be better to execute a command using `execCommandEx`, eg
```nim
let execResult = await execCommandEx("node run scripts/mine.js --network localhost")
```

emizzle (changes requested):
Quick suggestion...

tests/integration/multinodes.nim
```diff
@@ -146,6 +146,14 @@ template multinodesuite*(name: string, body: untyped) =
         raiseMultiNodeSuiteError "hardhat node not started: " & e.msg
 
       trace "hardhat node started"
+
+      try:
+        await node.postStart()
+      except NodeProcessError as e:
+        raiseMultiNodeSuiteError "cannot run the post start scripts: " & e.msg
```
emizzle, line 153: I think it would be better to run `postStart` from `HardhatProcess.start` since it would be not great if someone ran a hardhat process but the MP contract was not deployed. If we need a situation where these scripts are NOT run, we could pass a parameter to `start`.

AuHau: Yeah, I would also go with that route of encapsulating everything in the `start` method.

2-towns:
> Seems like the tests are failing on the test that you reduced the polling for? Maybe try to revert it?

Yeah actually I would prefer to have #1268 merged first and rebase this PR on master.

PR #1196, opened 2025-04-15
Marketplace vault

markspanbroek:
Uses the latest version of the Marketplace contract that utilize the Vault contract.

This includes the following changes:
- separate reward and collateral addresses are no longer supported
- request cancelled event is no longer emitted
- slot state 'paid' no longer exists
- ERC20 token contract no longer supports increaseAllowance
- repair reward is added to slot queue, and used to calculate profitability
- sales agent keeps track of slot queue item
- all marketplace configuration getters are now synchronous
- separate types for timestamps, duration, tokens, and tokens per second

And fixes the following bug:
- slotqueue and repostore misinterpreted expiry duration for expiry timestamp

Depends on the fix in #1201 
Depends on https://github.com/codex-storage/codex-contracts-eth/pull/220
Depends on https://github.com/codex-storage/codex-contracts-eth/pull/223


markspanbroek:
Note to self:

https://github.com/codex-storage/nim-codex/blob/307d85a7001e86f37bb07bd333b2bbb17f32ec3f/codex/sales/states/proving.nim#L84

This is no longer true because I removed the cancelled callback, we need to exit the proving loop here.

AuHau (approved):
Generally LGTM, but honestly this is the type of PR that is really hard to review.

For next time, please really break it up. From my POV, the "hard limit" of a "too big PR" is when the browser has a problem rendering the UI for the PR ... and my Brave was already having problems 😅

PR #1197, opened 2025-04-16
chore: bump contracts to master

emizzle:
Bump contracts to master branch.

There was a change that allowed hardhat to have multiple blocks with the same timestamp, so this needed to be reflected in two tests.

PR #1198, opened 2025-04-16
chore: bumps contracts to allow slot reservations when in repair

emizzle:
Allows slots to be reserved when they are in the Repair state.

Note: based on #1197


markspanbroek (approved):
:+1:

PR #1201, opened 2025-04-17
fix(integration): fix api validation test

markspanbroek:
Fixes an intermittent problem with other tests because the api validation test kept a codex node alive after it finishes.

Also includes a fix for an unrelated flaky market test, to make the CI pass.


gmega:
Ugh, I think the reason @2-towns did this was to make the API tests faster.

emizzle:
There's a lot of integration tests improvements lingering in the parallel integration tests branch, particularly regarding tearing down tests

2-towns (approved):
LGTM thanks !

PR #1205, opened 2025-04-18
Experiment with simple swarms

gmega:
Not a real PR, just to keep the thing visible if someone's interested.

### DIFF
--- a/codex/blockexchange/swarm.nim
+++ b/codex/blockexchange/swarm.nim
@@ -63,6 +63,7 @@ type
     # Network interface
     network*: BlockExcNetwork
     discovery*: Discovery
+    peerEventHandler: PeerEventHandler
 
     completionHandle: Future[?!void]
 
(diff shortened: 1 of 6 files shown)


codex/blockexchange/swarm.nim
```diff
@@ -0,0 +1,612 @@
...
+  try:
+    while true:
+      trace "Advertising CID", cid = self.cid
+      await self.discovery.provide(self.cid)
+      trace "Avertiser going to sleep"
+      await sleepAsync(DefaultAdvertisementInterval)
+  except CancelledError:
+    trace "Advertisement loop cancelled. Exiting."
+
+proc loadBlockKnowledge*(self: Swarm): Future[void] {.async: (raises: []).} =
```
dryajov, line 197: I'm not sure this is needed, why duplicate this if it is already in the blockstore? If you're worried about hitting the repostore too oftne, then we can always put a cache of some sort in front?

codex/blockexchange/swarm.nim
```diff
@@ -0,0 +1,612 @@
...
+    try:
+      if not self.peers[peer].pendingBlockKnowledgeRequest:
+        trace "Illegal state for pending block knowledge request (already cleared)",
+          peer = peer
+      self.peers[peer].pendingBlockKnowledgeRequest = false
+    except KeyError:
+      trace "Error updating pending block knowledge request state (peer dropped)",
+        peer = peer
+
+proc handleBlockRequest(
```
dryajov, line 408: I know this is an experiment, but we should be lazy with handling requests, so I would still put this behind some throttling mechanism, such as the asyncqueue.

dryajov (commented):
Overall, I'm liking this direction very much. I think this might be close to being ussable. One thing we need to keep in mind, is that we still want the ability to stream requests, as this is a really imoprtant use case. I don't think anything specifically prevents it in this implementation, but just making sure it's know :)

PR #1207, opened 2025-04-22
Codex-contracts hash in version information.

benbierens:
Following a discussion on how to track compatibility of codex-contracts with the main application.

This PR adds the contracts revision hash to both `--version` and the debug/info version object:
```
	"codex": {
		"version": "v0.2.0",
		"revision": "22f5150d",
		"contracts": "0bf1385"
	}
```

This enabled the automatic checking of compatibility in tests. (Including situations where docker commands are not available to check the docker-image labels added in an earlier PR.)

### DIFF
--- a/codex/conf.nim
+++ b/codex/conf.nim
@@ -477,14 +477,9 @@ proc getCodexRevision(): string =
   var res = strip(staticExec("git rev-parse --short HEAD"))
   return res
 
-proc getCodexContractsRevision(): string =
-  let
-    res = strip(
-        staticExec("git ls-tree -z -d --abbrev=7 HEAD -- ../vendor/codex-contracts-eth")
-      )
-      .replace('\t', ' ')
-    tokens = res.split(' ')
-  return tokens[2]
+proc getCodexContractsRevision() : string =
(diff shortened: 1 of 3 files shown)


veaceslavdoina (approved):
Thanks, a highly desirable improvement 😄

codex/conf.nim
```diff
@@ -477,17 +477,22 @@ proc getCodexRevision(): string =
   var res = strip(staticExec("git rev-parse --short HEAD"))
   return res
 
+proc getCodexContractsRevision() : string =
+  let res = strip(staticExec("git rev-parse --short HEAD:vendor/codex-contracts-eth"))
```
gmega, line 481: This makes me super uneasy (what happens if you try to compile Codex from a source tarball instead of a git repo?) but I suppose you're not the one introducing this pattern in our codebase (see `getCodexRevision`) so I'll go with it.

gmega (approved):
I'd rather not have this inside of `conf.nim` but I suppose it's no worse than what's already in there (e.g. `getCodexRevision`), and it is pretty simple to revert if we decide this is not acceptable down the line.

PR #1211, opened 2025-04-23
docs(openapi): freeSize non-optional

AuHau:


2-towns (approved):

2-towns:
Just a note: I think we could remove `freeSize` from `RestAvailability`.  

When creating an availability, the `totaliSize` is actually the `freeSize`. 
When updating an availability, it is not allowed to edit the `freeSize`.

emizzle:
> Just a note: I think we could remove `freeSize` from `RestAvailability`.
> 
> When creating an availability, the `totaliSize` is actually the `freeSize`. When updating an availability, it is not allowed to edit the `freeSize`.

@2-towns, I created an issue for this: https://github.com/codex-storage/nim-codex/issues/1225

PR #1212, opened 2025-04-23
fix(api): availability creation validation

AuHau:
When testing the new release I noticed that our "required" parameters are not really required...


2-towns (approved):

markspanbroek (approved):
Good catch :+1:

PR #1217, opened 2025-05-06
chore: fix custom error handling when simulating invalid proofs

2-towns:


codex/contracts/market.nim
```diff
@@ -349,7 +349,12 @@ method submitProof*(
     market: OnChainMarket, id: SlotId, proof: Groth16Proof
 ) {.async: (raises: [CancelledError, MarketError]).} =
   convertEthersError("Failed to submit proof"):
-    discard await market.contract.submitProof(id, proof).confirm(1)
+    try:
+      discard await market.contract.submitProof(id, proof).confirm(1)
+    except Proofs_InvalidProof as parent:
+      raise newException(
+        ProofInvalidError, "Failed to fill slot because the slot is not free", parent
```
markspanbroek, line 356: This particular error is only raised by the contracts when the [Groth16 verifier says that the ZK proof is invalid](https://github.com/codex-storage/codex-contracts-eth/blob/470a4df4152b53a8065c2984d55efcd7350b549a/contracts/Proofs.sol#L208), so this error message incorrectly states that this is due to the slot not being free.

emizzle: That's correct, but this error is discarded in the `provingsimulated` state (and not in the `proving` state).

2-towns: > This particular error is only raised by the contracts when the [Groth16 verifier says that the ZK proof is invalid](https://github.com/codex-storage/codex-contracts-eth/blob/470a4df4152b53a8065c2984d55efcd7350b549a/contracts/Proofs.sol#L208), so this error message incorrectly states that this is due to the slot not being free.

Okay I updated the error message

markspanbroek (approved):

PR #1218, opened 2025-05-06
[feat] Add support for content ranges/seeking

Zorlin:
Very naive implementation, but in local testing it works! Use VLC and stream a video and you can repeatedly seek within it.

### DIFF
--- a/codex/node.nim
+++ b/codex/node.nim
@@ -14,6 +14,7 @@ import std/sequtils
 import std/strformat
 import std/sugar
 import times
+import std/streams
 
 import pkg/taskpools
 import pkg/questionable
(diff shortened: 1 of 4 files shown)


Zorlin:
Currently this implementation relies on retrieving the entire file (in the case of a node that doesn't have the full content) before it streams from a position, which is pretty inefficient. Looking into fixes.

vpavlin:
How is this looking @Zorlin ?:) I could really use it for one of my hacks:D

Zorlin:
> How is this looking @Zorlin ?:) I could really use it for one of my hacks:D

Will be doing a rewrite based on the latest code, and get back to you soon :pray:

Issue #1210, opened 2025-04-23
Storage request ignored


benbierens:
Testnet v0.2.1
RequestId:
`7e8b80db3a44dde9b860db68e19efaaabe688c142000dcf5609d60164abd82be`

Kibana search: https://kibana.testnet.codex.storage/app/r/s/WMyWR

Storage nodes mention request in log message `pushing item to queue`
and then never again until the request times out. Then `removing item from queue`.

Slot-reservations-full events never received for this request.

Are all the workers full?
There are 10 storage nodes.
Default: 3 workers per node.
Total: 30 workers.

Testnet metrics shows a maximum of 8 open storage requests concurrently.
Each request has 5 slots.
Total slots: 40.

How can we detect a full slot queue, or no-workers-available?

benbierens:
It seems this issue got progressively worse during the testnet run of release v0.2.1. And it seems to have brought the net to a halt after nearly 1 hour.

AuHau:
The slot queue will never be "full" as it will replace items. 
Regarding workers depletion - I don't think we have any way currently to detect that unfortunatelly.

benbierens:
I think I was wrong to suspect a full queue or zero workers. Even when there are very few requests available, their slots are still not being reserved.

2-towns:
We are facing this issue in the Chrys slot repair branch. Here is the scenario:

- 3 SPs are created  
- SP-0 creates an availability for 2 slots  
- SP-1 creates an availability for 1 slot  
- The contract starts  
- SP-2 creates an availability for 1 slot  
- SP-1 frees the slots  
- SP-2 is expected to take the freed slot  

However, instead of picking the freed slot, SP-2 gets stuck on `pushing item to queue`, as mentioned in the issue description.

Diving deeper into the slot queue, I found that in our case, SP-2 ignores the 3 slots at the beginning because no availability is created yet, so the sale moves to the `SaleIgnored` state.

But the issue is that the following line is hanging the 3 workers, so no one is available to pick up the new slot:

```nim
await onProcessSlot(item)
```

Digging into sales, I found that `completed.fire()` is never called in the `onCleanUp` function.

Further, in the state machine, the stop function is hanging at:

```nim
await machine.trackedFutures.cancelTracked()
```

I am still exploring the issue to try to find an explanation.

2-towns:
I think we are in a catch-22 situation.

Let's say a sale moves from `SalePreparing` to `SaleIgnored`.

[...]

PR #1224, opened 2025-05-14
fix(slotqueue): simplify slot queue workers

markspanbroek:
- worker is now just an async running loop
- instead of passing a "done" Future, use an AsyncEvent to signal completion

Attempts to fix, or at least understand better #1210

### DIFF
--- a/codex/sales.nim
+++ b/codex/sales.nim
@@ -111,11 +111,7 @@ proc remove(sales: Sales, agent: SalesAgent) {.async.} =
     sales.agents.keepItIf(it != agent)
 
 proc cleanUp(
-    sales: Sales,
-    agent: SalesAgent,
-    reprocessSlot: bool,
-    returnedCollateral: ?UInt256,
-    processing: Future[void],
+    sales: Sales, agent: SalesAgent, reprocessSlot: bool, returnedCollateral: ?UInt256
 ) {.async.} =
   let data = agent.data
 
@@ -181,36 +177,38 @@ proc cleanUp(
 
   await sales.remove(agent)
 
-  # signal back to the slot queue to cycle a worker
-  if not processing.isNil and not processing.finished():
-    processing.complete()
-
-proc filled(
-    sales: Sales, request: StorageRequest, slotIndex: uint64, processing: Future[void]
-) =
+proc filled(sales: Sales, request: StorageRequest, slotIndex: uint64) =
   if onSale =? sales.context.onSale:
     onSale(request, slotIndex)
 
-  # signal back to the slot queue to cycle a worker
-  if not processing.isNil and not processing.finished():
-    processing.complete()
-
-proc processSlot(sales: Sales, item: SlotQueueItem, done: Future[void]) =
+proc processSlot(
+    sales: Sales, item: SlotQueueItem
+) {.async: (raises: [CancelledError]).} =
   debug "Processing slot from queue", requestId = item.requestId, slot = item.slotIndex
 
   let agent =
     newSalesAgent(sales.context, item.requestId, item.slotIndex, none StorageRequest)
 
+  let completed = newAsyncEvent()
+
   agent.onCleanUp = proc(
       reprocessSlot = false, returnedCollateral = UInt256.none
   ) {.async.} =
-    await sales.cleanUp(agent, reprocessSlot, returnedCollateral, done)
+    trace "slot cleanup"
+    await sales.cleanUp(agent, reprocessSlot, returnedCollateral)
+    completed.fire()
 
   agent.onFilled = some proc(request: StorageRequest, slotIndex: uint64) =
-    sales.filled(request, slotIndex, done)
+    trace "slot filled"
+    sales.filled(request, slotIndex)
+    completed.fire()
 
   agent.start(SalePreparing())
   sales.agents.add agent
+
+  trace "waiting for slot processing to complete"
+  await completed.wait()
+  trace "slot processing completed"
 
 proc deleteInactiveReservations(sales: Sales, activeSlots: seq[Slot]) {.async.} =
   let reservations = sales.context.reservations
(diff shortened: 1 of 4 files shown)


codex/sales/slotqueue.nim
```diff
@@ -386,25 +309,24 @@ proc clearSeenFlags*(self: SlotQueue) =
...
+  trace "slot queue worker loop started"
+  while queue.running:
     try:
-      if self.paused:
+      if queue.paused:
         trace "Queue is paused, waiting for new slots or availabilities to be modified/added"
 
       # block until unpaused is true/fired, ie wait for queue to be unpaused
-      await self.unpaused.wait()
+      await queue.unpaused.wait()
```
dryajov, line 320: Why is the pause/unpause mechanism needed?

codex/sales.nim
```diff
@@ -181,37 +177,39 @@ proc cleanUp(
...
-proc processSlot(sales: Sales, item: SlotQueueItem, done: Future[void]) =
+proc processSlot(
+    sales: Sales, item: SlotQueueItem
+) {.async: (raises: [CancelledError]).} =
   debug "Processing slot from queue", requestId = item.requestId, slot = item.slotIndex
 
   let agent =
     newSalesAgent(sales.context, item.requestId, item.slotIndex, none StorageRequest)
 
+  let completed = newAsyncEvent()
```
dryajov, line 192: If this is single use (which it looks like), perhaps a simple `Future` is simpler here, `AsyncEvent` semantincs might get confusing at times. But just a suggestion :)

codex/sales.nim
```diff
@@ -181,37 +177,39 @@ proc cleanUp(
...
+
   agent.onCleanUp = proc(
       reprocessSlot = false, returnedCollateral = UInt256.none
   ) {.async.} =
-    await sales.cleanUp(agent, reprocessSlot, returnedCollateral, done)
+    trace "slot cleanup"
+    await sales.cleanUp(agent, reprocessSlot, returnedCollateral)
+    completed.fire()
```
dryajov, line 199: Could `onCleanUp` and `onFilled` both fire, or is this guaranteed to be "either or"?

dryajov (approved):
Nice simplification, I just have a few clarifying questions, but other than that :thumbsup

dryajov:
So, I have a couple of questions about the overall flow.

- First, why are we pausing and unpausing the queue on seen items?
- Why are we handling seen items as a special case?
  - Shouldn't reprocessing just follow the same flow at the queue level (it might require special handling elsewhere, but not at the queue level)?

codex/sales/slotqueue.nim
```diff
@@ -386,25 +309,24 @@ proc clearSeenFlags*(self: SlotQueue) =
...
 
-      let worker =
-        await self.workers.popFirst() # if workers saturated, wait here for new workers
-      let item = await self.queue.pop() # if queue empty, wait here for new items
+      let item = await queue.queue.pop() # if queue empty, wait here for new items
```
emizzle, line 322: Precisely why I preferred `self` over `queue` 😉 `queue.queue` is a bit cumbersome to me 🤷‍♂️ But also, it doesn't really matter at all.

codex/sales/slotqueue.nim
```diff
@@ -413,29 +335,26 @@ proc run(self: SlotQueue) {.async: (raises: []).} =
...
 
       trace "processing item"
-
-      let fut = self.dispatch(worker, item)
-      self.trackedFutures.track(fut)
-
-      await sleepAsync(1.millis) # poll
+      without onProcessSlot =? queue.onProcessSlot:
+        error "slot queue onProcessSlot not set?"
+        continue
```
emizzle, line 350: Should we raise a `Defect` here instead? Seems like a pretty unrecoverable error to me. This how we handle callbacks not set elsewhere in the sales state machine.

tests/codex/sales/testslotqueue.nim
```diff
@@ -50,19 +50,14 @@ suite "Slot queue start/stop":
 suite "Slot queue workers":
   var queue: SlotQueue
 
-  proc onProcessSlot(
-      item: SlotQueueItem, doneProcessing: Future[void]
-  ) {.async: (raises: []).} =
+  proc onProcessSlot(item: SlotQueueItem) {.async: (raises: []).} =
     # this is not illustrative of the realistic scenario as the
     # `doneProcessing` future would be passed to another context before being
     # completed and therefore is not as simple as making the callback async
```
emizzle, line 56: Probably no longer needed?

emizzle (commented):
Nice work, Mark. Left a few comments/suggestions.

codex/sales/slotqueue.nim
markspanbroek: I did this refactor in a couple of steps. At first, I had `run(self: Worker, queue: Queue)`, but then it turned out that the `Worker` object was no longer needed, so I removed the `self` parameter. Now we can indeed rename `queue` to `self` again.

codex/sales.nim
emizzle: I think that dance is warranted since we want to prevent it from being fired twice.

codex/sales.nim
markspanbroek: When we cannot fill a slot, then only `onCleanup` is called.
When we do fill a slot, then first `onFilled` is called, and much later `onCleanup` is called (e.g. when the request is finished)

codex/sales.nim
markspanbroek: > I think that dance is warranted since we want to prevent it from being fired twice.

The AsyncEvent ensures that we don't go through the `await completed.wait()` call more than once, even if you call `completed.fire()` more than once.

emizzle (approved):
LGTM, nice job!

codex/sales/slotqueue.nim
emizzle: Futhermore, removing the queue pausing mechanism in the new sales design also means that updating an availability will not attempt to fill historically added slots. This is ok since availabilities in the new sales design represent future sales only.

PR #1229, opened 2025-05-19
fix(availabilities): use totalRemainingCollateral instead of totalCollateral for comparaison 

2-towns:
This PR uses `totalRemainingCollateral` instead of `totalCollateral` when trying to detect availability changes for notifying the sales module during the availability update. 

 The  `totalRemainingCollateral`  has to be used because this is the value updated when a reservation is created or deleted. 
The `totalCollateral` is the initial value used when creating / updating the availability from the JSON api.


emizzle (approved):
Good catch!

markspanbroek (approved):

PR #1233, opened 2025-05-20
fix(sales): handle cancellation of slot cleanup

markspanbroek:
Possible fix for #1210 as suggested by @emizzle: https://github.com/codex-storage/nim-codex/issues/1210#issuecomment-2894016920

Ensures that processing slots from the slot queue continues even when cleanup of a slot is cancelled.
Also adds `{.async: (raises:[]}.}` to `onCleanup` to make sure that we didn't miss any other surprises here.

### DIFF
--- a/codex/sales.nim
+++ b/codex/sales.nim
@@ -105,14 +105,14 @@ proc new*(
     subscriptions: @[],
   )
 
-proc remove(sales: Sales, agent: SalesAgent) {.async.} =
+proc remove(sales: Sales, agent: SalesAgent) {.async: (raises: [CancelledError]).} =
   await agent.stop()
   if sales.running:
     sales.agents.keepItIf(it != agent)
 
 proc cleanUp(
     sales: Sales, agent: SalesAgent, reprocessSlot: bool, returnedCollateral: ?UInt256
-) {.async.} =
+) {.async: (raises: [CancelledError]).} =
   let data = agent.data
 
   logScope:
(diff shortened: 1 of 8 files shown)


emizzle (approved):
LGTM, thanks for doing this, Mark. 

For completeness in the reservations module, maybe we could also add `async: (raises: [CancelledError])` to `reservations.createReservation` and `reservations.returnBytesToAvailability`?

codex/sales.nim
```diff
@@ -105,14 +105,14 @@ proc new*(
...
-proc remove(sales: Sales, agent: SalesAgent) {.async.} =
+proc remove(sales: Sales, agent: SalesAgent) {.async: (raises: [CancelledError]).} =
   await agent.stop()
   if sales.running:
     sales.agents.keepItIf(it != agent)
 
 proc cleanUp(
     sales: Sales, agent: SalesAgent, reprocessSlot: bool, returnedCollateral: ?UInt256
-) {.async.} =
+) {.async: (raises: [CancelledError]).} =
```
emizzle, line 115: I think we should avoid allowing cleanup routines to be cancelled, and thus should raise `[]` here. For each await in this routine, we can prefix with `noCancel`, eg `await noCancel sales.context.reservations.returnBytesToAvailability` which will prevent those routines from being cancelled and raising `CancelledError` (so that exception doesn't need to be handled).

emizzle (changes requested):

codex/sales.nim
markspanbroek: Yes, I agree. I'd suggest to solve this in #1234, which is based on this PR.

2-towns (approved):
LGTM, thanks.

PR #1234, opened 2025-05-22
fix: sales cleanup cancellation

2-towns:
When an SP was processing a slot and the sale was ignored, the cleanup function was blocked, preventing workers from picking up new items from the slot queue.

This PR should fix that and maybe #1210.

### DIFF
--- a/codex/sales.nim
+++ b/codex/sales.nim
@@ -116,7 +116,7 @@ proc remove(sales: Sales, agent: SalesAgent) {.async: (raises: []).} =
 
 proc cleanUp(
     sales: Sales, agent: SalesAgent, reprocessSlot: bool, returnedCollateral: ?UInt256
-) {.async: (raises: [CancelledError]).} =
+) {.async: (raises: []).} =
   let data = agent.data
 
   logScope:
@@ -133,7 +133,7 @@ proc cleanUp(
   # there are not really any bytes to be returned
   if request =? data.request and reservation =? data.reservation:
     if returnErr =? (
-      await sales.context.reservations.returnBytesToAvailability(
+      await noCancel sales.context.reservations.returnBytesToAvailability(
         reservation.availabilityId, reservation.id, request.ask.slotSize
       )
     ).errorOption:
@@ -143,7 +143,7 @@ proc cleanUp(
     # delete reservation and return reservation bytes back to the availability
     if reservation =? data.reservation and
         deleteErr =? (
-          await sales.context.reservations.deleteReservation(
+          await noCancel sales.context.reservations.deleteReservation(
             reservation.id, reservation.availabilityId, returnedCollateral
           )
         ).errorOption:
@@ -152,7 +152,8 @@ proc cleanUp(
   # Re-add items back into the queue to prevent small availabilities from
   # draining the queue. Seen items will be ordered last.
   if data.slotIndex <= uint16.high.uint64 and reprocessSlot and request =? data.request:
-    let res = await sales.context.market.slotCollateral(data.requestId, data.slotIndex)
+    let res =
+      await noCancel sales.context.market.slotCollateral(data.requestId, data.slotIndex)
     if res.isErr:
       error "Failed to re-add item back to the slot queue: unable to calculate collateral",
         error = res.error.msg
@@ -192,10 +193,7 @@ proc processSlot(
       reprocessSlot = false, returnedCollateral = UInt256.none
   ) {.async: (raises: []).} =
     trace "slot cleanup"
-    try:
-      await sales.cleanUp(agent, reprocessSlot, returnedCollateral)
-    except CancelledError as e:
-      trace "slot cleanup was cancelled", error = e.msgDetail
+    await sales.cleanUp(agent, reprocessSlot, returnedCollateral)
     completed.fire()
 
   agent.onFilled = some proc(request: StorageRequest, slotIndex: uint64) =
@@ -270,10 +268,7 @@ proc load*(sales: Sales) {.async.} =
     agent.onCleanUp = proc(
         reprocessSlot = false, returnedCollateral = UInt256.none
     ) {.async: (raises: []).} =
-      try:
-        await sales.cleanUp(agent, reprocessSlot, returnedCollateral)
-      except CancelledError as e:
-        trace "slot cleanup was cancelled", error = e.msgDetail
+      await sales.cleanUp(agent, reprocessSlot, returnedCollateral)
 
     # There is no need to assign agent.onFilled as slots loaded from `mySlots`
     # are inherently already filled and so assigning agent.onFilled would be
--- a/codex/sales.nim
+++ b/codex/sales.nim
@@ -128,30 +128,36 @@ proc cleanUp(
   # that the cleanUp was called before the sales process really started, so
   # there are not really any bytes to be returned
   if request =? data.request and reservation =? data.reservation:
-    if returnErr =? (
-      await sales.context.reservations.returnBytesToAvailability(
-        reservation.availabilityId, reservation.id, request.ask.slotSize
-      )
-    ).errorOption:
-      error "failure returning bytes",
-        error = returnErr.msg, bytes = request.ask.slotSize
-
-  # delete reservation and return reservation bytes back to the availability
-  if reservation =? data.reservation and
-      deleteErr =? (
-        await sales.context.reservations.deleteReservation(
-          reservation.id, reservation.availabilityId, returnedCollateral
+    try:
+      if returnErr =? (
+        await sales.context.reservations.returnBytesToAvailability(
+          reservation.availabilityId, reservation.id, request.ask.slotSize
         )
       ).errorOption:
-    error "failure deleting reservation", error = deleteErr.msg
+        error "failure returning bytes",
+          error = returnErr.msg, bytes = request.ask.slotSize
+    except CancelledError:
+      debug "returning bytes was cancelled"
+    except CatchableError as e:
+      error "failure returning bytes", error = e.msg
 
-  if data.slotIndex > uint16.high.uint64:
-    error "Cannot cast slot index to uint16", slotIndex = data.slotIndex
-    return
+  try:
+    # delete reservation and return reservation bytes back to the availability
+    if reservation =? data.reservation and
+        deleteErr =? (
+          await sales.context.reservations.deleteReservation(
+            reservation.id, reservation.availabilityId, returnedCollateral
+          )
+        ).errorOption:
+      error "failure deleting reservation", error = deleteErr.msg
+  except CancelledError:
+    debug "deleting reservation was cancelled"
+  except CatchableError as e:
+    error "failure deleting reservation", error = e.msg
 
   # Re-add items back into the queue to prevent small availabilities from
   # draining the queue. Seen items will be ordered last.
-  if reprocessSlot and request =? data.request:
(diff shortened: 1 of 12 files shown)


codex/sales.nim
```diff
@@ -195,7 +206,8 @@ proc processSlot(
       reprocessSlot = false, returnedCollateral = UInt256.none
   ) {.async.} =
     trace "slot cleanup"
-    await sales.cleanUp(agent, reprocessSlot, returnedCollateral)
+    let fut = sales.cleanUp(agent, reprocessSlot, returnedCollateral)
+    sales.trackedFutures.track(fut)
```
markspanbroek, line 210: I don't think that this is the correct strategy. We're now signaling completion of the worker before we've done cleanup. That means that the worker can pick up the next queue item while the reserved bytes haven't been returned the availability yet.

Maybe it's better to call the `cleanup` function with `noCancel`?

codex/sales.nim
```diff
@@ -128,30 +128,36 @@ proc cleanUp(
...
+        await sales.context.reservations.returnBytesToAvailability(
+          reservation.availabilityId, reservation.id, request.ask.slotSize
         )
       ).errorOption:
-    error "failure deleting reservation", error = deleteErr.msg
+        error "failure returning bytes",
+          error = returnErr.msg, bytes = request.ask.slotSize
+    except CancelledError:
+      debug "returning bytes was cancelled"
+    except CatchableError as e:
```
markspanbroek, line 141: It might be nicer to base this on top of #1233, so that the reservations function no longer raise `CatchableError`

codex/sales.nim
```diff
@@ -171,11 +177,16 @@ proc cleanUp(
       trace "pushing ignored item to queue, marked as seen"
       if err =? queue.push(seenItem).errorOption:
         error "failed to readd slot to queue", errorType = $(type err), error = err.msg
-    except MarketError as e:
+    except CatchableError as e:
       error "Failed to re-add item back to the slot queue.", error = e.msg
       return
```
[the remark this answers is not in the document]
markspanbroek: Me neither, seems like a bug

markspanbroek (commented):
Nice work on the test @2-towns ! I think there are a few issues with the solution still...

codex/sales.nim
2-towns: Oh great I will do it

codex/sales.nim
```diff
@@ -128,30 +128,36 @@ proc cleanUp(
...
+      error "failure deleting reservation", error = deleteErr.msg
+  except CancelledError:
+    debug "deleting reservation was cancelled"
+  except CatchableError as e:
+    error "failure deleting reservation", error = e.msg
 
   # Re-add items back into the queue to prevent small availabilities from
   # draining the queue. Seen items will be ordered last.
-  if reprocessSlot and request =? data.request:
+  if data.slotIndex <= uint16.high.uint64 and reprocessSlot and request =? data.request:
```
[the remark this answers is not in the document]
2-towns: The check was already there in the lines above but I preferred to include it here to avoid to `return`, because I think we should not return from the cleanup function.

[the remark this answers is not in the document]
markspanbroek: Ah right, I missed that :+1:

codex/sales.nim
2-towns: Okay so I rebased my PR on yours. Waiting that you move it out of draft to do it as well.

codex/sales.nim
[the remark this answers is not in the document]
2-towns: As discussed in the weekly meeting, adding `cleanup` function into `trackedFutures` is not a good idea because bytes and collateral are returned in this function, so `sales.remove(agent)`  is added to  `trackedFutures` instead to not block the `cleanup` function.

codex/sales.nim
```diff
@@ -105,14 +105,18 @@ proc new*(
...
+  except CancelledError:
+    trace "Agent stop was cancelled"
+
   if sales.running:
     sales.agents.keepItIf(it != agent)
 
 proc cleanUp(
     sales: Sales, agent: SalesAgent, reprocessSlot: bool, returnedCollateral: ?UInt256
-) {.async.} =
+) {.async: (raises: [CancelledError]).} =
```
emizzle, line 119: I think we should avoid raising any exceptions here, including CancelledError. Cleanup routines should happen no matter what, and any awaited calls inside should be prefixed with `noCancel`.

codex/sales.nim
```diff
@@ -136,29 +140,24 @@ proc cleanUp(
...
-      ).errorOption:
-    error "failure deleting reservation", error = deleteErr.msg
-
-  if data.slotIndex > uint16.high.uint64:
-    error "Cannot cast slot index to uint16", slotIndex = data.slotIndex
-    return
+    # delete reservation and return reservation bytes back to the availability
+    if reservation =? data.reservation and
+        deleteErr =? (
+          await sales.context.reservations.deleteReservation(
```
emizzle, line 146:
```suggestion
          await noCancel sales.context.reservations.deleteReservation(
```

codex/sales.nim
```diff
@@ -193,9 +190,12 @@ proc processSlot(
...
   agent.onCleanUp = proc(
       reprocessSlot = false, returnedCollateral = UInt256.none
-  ) {.async.} =
+  ) {.async: (raises: []).} =
     trace "slot cleanup"
-    await sales.cleanUp(agent, reprocessSlot, returnedCollateral)
+    try:
+      await sales.cleanUp(agent, reprocessSlot, returnedCollateral)
+    except CancelledError as e:
+      trace "slot cleanup was cancelled", error = e.msgDetail
```
emizzle, line 198: This won't be needed once `cleanUp` no longer raises. We don't want `cleanUp` to be cancelled in any circumstance, no matter who calls it.

codex/sales.nim
```diff
@@ -269,8 +269,11 @@ proc load*(sales: Sales) {.async.} =
 
     agent.onCleanUp = proc(
         reprocessSlot = false, returnedCollateral = UInt256.none
-    ) {.async.} =
-      await sales.cleanUp(agent, reprocessSlot, returnedCollateral)
+    ) {.async: (raises: []).} =
+      try:
+        await sales.cleanUp(agent, reprocessSlot, returnedCollateral)
+      except CancelledError as e:
+        trace "slot cleanup was cancelled", error = e.msgDetail
```
emizzle, line 276: Same as above.

codex/sales/salesagent.nim
```diff
@@ -132,7 +132,7 @@ proc subscribe*(agent: SalesAgent) {.async.} =
   await agent.subscribeCancellation()
   agent.subscribed = true
 
-proc unsubscribe*(agent: SalesAgent) {.async.} =
+proc unsubscribe*(agent: SalesAgent) {.async: (raises: [CancelledError]).} =
```
emizzle, line 135: This routine as well should not be cancellable imo, since it invokes teardown mechanisms.
```suggestion
proc unsubscribe*(agent: SalesAgent) {.async: (raises: []).} =
```

codex/sales/salesagent.nim
```diff
@@ -143,6 +143,6 @@ proc unsubscribe*(agent: SalesAgent) {.async.} =
 
   agent.subscribed = false
 
-proc stop*(agent: SalesAgent) {.async.} =
+proc stop*(agent: SalesAgent) {.async: (raises: [CancelledError]).} =
```
emizzle, line 146: This routine as well should not be cancellable imo, since it invokes teardown mechanisms.
```suggestion
proc stop*(agent: SalesAgent) {.async: (raises: []).} =
```

tests/integration/testmarketplace.nim
```diff
@@ -144,16 +225,20 @@ marketplacesuite "Marketplace payouts":
...
+      clients: CodexConfigs
+        .init(nodes = 1)
+        #  .debug() # uncomment to enable console log output.debug()
+        .withLogFile()
+        # uncomment to output log file to tests/integration/logs/<start_datetime> <suite_name>/<test_name>/<node_role>_<node_idx>.log
+        .withLogTopics("node", "erasure").some,
+      providers: CodexConfigs
+        .init(nodes = 1)
+        #  .debug() # uncomment to enable console log output
+        .withLogFile()
```
emizzle, line 237: Probably better to avoid committing with `withLogFile` enabled.
```suggestion
        # uncomment to output log file to tests/integration/logs/<start_datetime> <suite_name>/<test_name>/<node_role>_<node_idx>.log
        .withLogTopics("node", "erasure").some,
      providers: CodexConfigs
        .init(nodes = 1)
        #  .debug() # uncomment to enable console log output
```

emizzle (approved):
This LGTM, nice!! I really think this should finally fix the starved workers testnet issue.

codex/sales.nim
2-towns: Yep done

codex/sales/reservations.nim
```diff
@@ -248,7 +248,7 @@ proc exists*(
   let exists = await self.repo.metaDs.ds.contains(key)
   return exists
 
-iterator items(self: StorableIter): Future[?seq[byte]] =
+iterator items(self: StorableIter): auto =
```
emizzle, line 251: Out of curiosity, was there a need to change this to auto?

codex/sales/reservations.nim
```diff
@@ -470,109 +476,116 @@ method createReservation*(
     slotIndex: uint64,
     collateralPerByte: UInt256,
     validUntil: SecondsSince1970,
-): Future[?!Reservation] {.async, base.} =
-  withLock(self.availabilityLock):
-    without availabilityKey =? availabilityId.key, error:
-      return failure(error)
+): Future[?!Reservation] {.async: (raises: [CancelledError]), base.} =
```
emizzle, line 479: This made me realise that because createReservation performs multiple async ops, it should not be cancelled when called from the state machine, otherwise the Availability/Reservation may end up in an inconsistent state depending how far along createReservation was at the time of cancellation. So maybe as part of this PR we should also change (in `sales/states/preparing`):
```diff
without reservation =?
-      await reservations.createReservation(
+      await noCancel reservations.createReservation(
        availability.id, request.ask.slotSize, request.id, data.slotIndex,
        request.ask.collateralPerByte, requestEnd,
      ), error:
```

codex/sales.nim
```diff
@@ -105,14 +105,18 @@ proc new*(
     subscriptions: @[],
   )
 
-proc remove(sales: Sales, agent: SalesAgent) {.async.} =
-  await agent.stop()
+proc remove(sales: Sales, agent: SalesAgent) {.async: (raises: []).} =
+  try:
+    await agent.stop()
+  except CancelledError:
+    trace "Agent stop was cancelled"
```
emizzle, line 112: `agent.stop` does not raise in this PR, so the try/except should not be needed. 
```suggestion
proc remove(sales: Sales, agent: SalesAgent) {.async: (raises: []).} =
  await agent.stop()
```

emizzle (commented):
Sorry, just a couple more comments/suggestions that I didn't catch in the last round.

codex/sales/reservations.nim
markspanbroek: A return type of `Future[?seq[byte]]` can raise a `CatchableException` when `await`ed. The return type of `self.next()` captures that it can only raise `CancelledError`, but it's not a simple `Future` type but generated by the `{.async.}` macro.

codex/sales.nim
2-towns: Done

PR #1235, opened 2025-05-22
Add support for slot reconstruction on unavailable slot detection

cnanakos:


2-towns:
The integration tests can be found [on this branch](https://github.com/codex-storage/nim-codex/blob/chore/slot-repair-tests/tests/integration/testslotrepair.nim).

3 PRs need to be merged to fix MP issues: 

https://github.com/codex-storage/nim-codex/pull/1238
https://github.com/codex-storage/nim-codex/pull/1234
https://github.com/codex-storage/nim-codex/pull/1216

codex/blockexchange/engine/advertiser.nim
```diff
@@ -124,6 +124,10 @@ proc start*(b: Advertiser) {.async: (raises: []).} =
 
   trace "Advertiser start"
 
+  if b.advertiserRunning:
+    warn "Starting advertiser twice"
```
benbierens, line 128: I think there is never a situation in which the advertiser should be started twice. If that's correct, I propose we put a raiseAssert here. This way the node will crash and we'll have to fix it.

codex/erasure/erasure.nim
```diff
@@ -659,6 +669,44 @@ proc decode*(self: Erasure, encoded: Manifest): Future[?!Manifest] {.async.} =
 
   return decoded.success
 
+proc repair*(self: Erasure, encoded: Manifest): Future[?!void] {.async.} =
+  ## Repair a protected manifest slot
```
benbierens, line 673: This comment suggests we're repairing 1 slot in the protected manifest. But I think the code reconstructs the entire dataset. Otherwise, I would have expected a slotIndex to be a required argument.

codex/node.nim
```diff
@@ -676,32 +672,45 @@ proc onStore(
...
-  without blksIter =? indexer.getIndicies(slotIdx.int).catch, err:
-    trace "Unable to get indicies from strategy", err = err.msg
-    return failure(err)
+  if isRepairing:
+    trace "start repairing slot", slotIdx
+    try:
+      let erasure = Erasure.new(
+        self.networkStore, leoEncoderProvider, leoDecoderProvider, self.taskpool
+      )
+      if err =? (await erasure.repair(manifest)).errorOption:
```
benbierens, line 685: Yeah, because slotIdx is not used to perform repair, this node will end up with the entire dataset stored locally. That's not a problem. (In fact, for the network that's kind of nice!) But it's not what a storage-provider signed up for. They agreed to store a slot. Not the whole thing.

It would be ideal if we could somehow fetch *only those block* we need to repair the slot we are trying to store, and nothing more. I understand that's a complicated puzzle. And for our first try to tackle repair, I entirely support the way you have implemented it here. However, everyone should be aware that it is being implemented this way.

codex/blockexchange/engine/advertiser.nim
benbierens: I think that's a good plan. I would like to see more raiseAsserts in general for situations that should not happen. I think this is a prime example.

codex/utils/iter.nim
```diff
@@ -89,6 +90,36 @@ proc new*[T](_: type Iter[T], items: seq[T]): Iter[T] =
 
   Iter[int].new(0 ..< items.len).map((i: int) => items[i])
 
+proc new*[T](_: type Iter[T], iter: Iterable[T]): Iter[T] =
+  ## Creates a new Iter from an iterator
+  ##
+  var nextOrErr: Option[Result[T, ref CatchableError]]
```
dryajov, line 96: Perhaps we can just use `?!` instead of `Result` here. `?!` is already defined as `Result[T, ref CatchableError]`

codex/utils/iter.nim
```diff
@@ -89,6 +90,36 @@ proc new*[T](_: type Iter[T], items: seq[T]): Iter[T] =
...
+  proc tryNext(): void =
+    nextOrErr = Result[T, ref CatchableError].none
+    while not iter.finished:
+      try:
+        let t: T = iter()
+        if not iter.finished:
+          nextOrErr = some(success(t))
```
dryajov, line 103: Also, why do we need to wrap the result in an optional?

codex/indexingstrategy.nim
```diff
@@ -44,39 +48,47 @@ func getIter(first, last, step: int): Iter[int] =
...
+    self.getLinearIndices(iteration)
   of StrategyType.SteppedStrategy:
-    self.getSteppedIndicies(iteration)
+    self.getSteppedIndices(iteration)
+
+func getIndices*(
+    self: IndexingStrategy, iteration: int
+): Iter[int] {.raises: [IndexingError].} =
+  self.checkIteration(iteration)
+  {.cast(noSideEffect).}:
```
dryajov, line 76: Any idea where the sideEffect is coming from?

tests/codex/node/testslotrepair.nim
```diff
@@ -0,0 +1,312 @@
...
+
+    await nodes[6].switch.stop() # slot 2 missing now
+
+    # repair missing slot from repaired slots
+    (await nodes[9].onStore(request, 2.uint64, nil, isRepairing = true)).tryGet()
+
+    let
+      stream = (await nodes[10].retrieve(verifiableBlock.cid, local = false)).tryGet()
+      expectedData = await fetchStreamData(stream, datasetSize)
+    assert expectedData.len == data.len
```
dryajov, line 243: use `check/check:` in tests

codex/utils/iter.nim
```diff
@@ -89,6 +90,36 @@ proc new*[T](_: type Iter[T], items: seq[T]): Iter[T] =
...
+        break
+      except CatchableError as err:
+        nextOrErr = some(T.failure(err))
+      except Exception:
+        assert(false)
```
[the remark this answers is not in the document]
dryajov: Why do we have to catch `Exception`?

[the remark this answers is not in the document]
gmega: Probably some ancient API that leaks it...

codex/utils/iter.nim
```diff
@@ -89,6 +90,36 @@ proc new*[T](_: type Iter[T], items: seq[T]): Iter[T] =
...
+
+  proc genNext(): T {.raises: [CatchableError].} =
+    without u =? nextOrErr.unsafeGet, err:
```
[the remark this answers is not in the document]
gmega: I think raising `Defect` could be fine if you want to crash the node on this error. `unsafeGet` is nasty because, from the docs:

```nim
template unsafeGet*[T: not void, E](self: Result[T, E]): T =
  self.unsafeValue()

template unsafeValue*[T: not void, E](self: Result[T, E]): T =
  ## Fetch value of result if set, undefined behavior if unset
  ## See also: `unsafeError`
  self.vResultPrivate
```

So Defect/CatchableError can both be valid approaches, undefined behavior isn't. :-)

gmega (approved):
OK I'm approving this - though I see a compilation error. Both that and the tests must be running smooth here before merging.

codex/erasure/erasure.nim
```diff
@@ -259,6 +266,11 @@ proc prepareDecodingData(
 
     resolved.inc
 
+  pendingBlockFutures.apply(
```
dryajov, line 269: you can use `pendingBlockFutures.applyIt(it.cancel())` from `sequtils`.

codex/node.nim
```diff
@@ -675,32 +671,45 @@ proc onStore(
 
     return success()
 
-  without indexer =?
-    manifest.verifiableStrategy.init(0, manifest.blocksCount - 1, manifest.numSlots).catch,
-    err:
-    trace "Unable to create indexing strategy from protected manifest", err = err.msg
-    return failure(err)
-
   if slotIdx > int.high.uint64:
```
dryajov, line 674: Could this ever happen - I don't think it can? It would die during conversion since you can't simply fit a larger number than what the type could handle.

tests/codex/node/testslotrepair.nim
```diff
@@ -0,0 +1,312 @@
...
+    for i in 0 ..< protected.numSlots.uint64:
+      (await nodes[i + 1].onStore(request, i, nil, isRepairing = false)).tryGet()
+
+    await nodes[0].switch.stop() # acts as client
+    await nodes[1].switch.stop() # slot 0 missing now
+
+    # repair missing slot
+    (await nodes[4].onStore(request, 0.uint64, nil, isRepairing = true)).tryGet()
+
+    await nodes[2].switch.stop() # slot 1 missing now
```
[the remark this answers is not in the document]
dryajov: Huh, I guess we can check that it actually did stop, but might be cumbersome - usually stopping the node is quite reliable, otherwise everything is broken :)

dryajov (changes requested):
This looks very good and I think it's ready to merge, with the exception of the comment I left in the `onStore` logic, I think when`isRepairing` is set, after the repari, we should continue with the slot setup flow, instead of having it as an if/else.

codex/node.nim
dryajov: ah, we're converting down to an int.

codex/node.nim
```diff
@@ -675,32 +671,45 @@ proc onStore(
...
     error "Cannot cast slot index to int", slotIndex = slotIdx
     return
 
-  without blksIter =? indexer.getIndicies(slotIdx.int).catch, err:
-    trace "Unable to get indicies from strategy", err = err.msg
-    return failure(err)
+  if isRepairing:
```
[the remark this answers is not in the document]
dryajov: Right, I see it now, we're either repairing or fetching, ok then that makes sense.

dryajov (approved):

PR #1236, opened 2025-05-26
fix: Support for mapping multiple listener address

munna0908:
This PR will fix https://github.com/codex-storage/nim-codex/issues/1158.

### DIFF
--- a/tests/contracts/testMarket.nim
+++ b/tests/contracts/testMarket.nim
@@ -541,14 +541,14 @@ ethersuite "On-Chain Market":
       (await market.queryPastStorageRequestedEvents(blocksAgo = 2))
     )
 
-  test "pays rewards and collateral to host":
+  test "pays rewards and returns collateral to host":
     await market.requestStorage(request)
 
     let address = await host.getAddress()
     switchAccount(host)
     await market.reserveSlot(request.id, 0.uint64)
     await market.fillSlot(request.id, 0.uint64, proof, request.ask.collateralPerSlot)
-    let filledAt = (await ethProvider.currentTime())
+    let filledAt = await ethProvider.blockTime(BlockTag.latest)
 
     for slotIndex in 1 ..< request.ask.slots:
       await market.reserveSlot(request.id, slotIndex.uint64)
@@ -629,7 +629,7 @@ ethersuite "On-Chain Market":
     check collateral ==
       request.ask.collateralPerSlot - (request.ask.collateralPerSlot * 10).div(100.u256)
 
-  test "the request is added in cache after the fist access":
+  test "the request is added to cache after the first access":
     await market.requestStorage(request)
 
     check market.requestCache.contains($request.id) == false
--- a/tests/integration/codexclient.nim
+++ b/tests/integration/codexclient.nim
@@ -1,4 +1,3 @@
-import std/httpclient
 import std/strutils
 
 from pkg/libp2p import Cid, `$`, init
--- a/tests/integration/testrestapivalidation.nim
+++ b/tests/integration/testrestapivalidation.nim
@@ -1,45 +1,23 @@
-import std/httpclient
 import std/times
 import pkg/ethers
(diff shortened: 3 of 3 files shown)


codex/nat.nim
```diff
@@ -233,30 +243,23 @@ proc repeatPortMapping(args: PortMappingArgs) {.thread, raises: [ValueError].} =
...
-      if dataAvailable:
-        return
-      else:
-        let currTime = now()
-        if currTime >= (lastUpdate + interval):
-          discard doPortMapping(tcpPort, udpPort, description)
-          lastUpdate = currTime
+    while natClosed.load() == false:
+      let
+        # we're being silly here with this channel polling because we can't
```
dryajov, line 248: this comment is probably not needed anymore?

codex/nat.nim
```diff
@@ -268,54 +271,68 @@ proc stopNatThread() {.noconv.} =
...
-    (externalTcpPort, externalUdpPort) = result.get()
+    let (externalTcpPort, externalUdpPort) = result.get()
     # needed by NAT-PMP on port mapping deletion
-    internalTcpPort = tcpPort
-    internalUdpPort = udpPort
     # Port mapping works. Let's launch a thread that repeats it, in case the
     # NAT-PMP lease expires or the router is rebooted and forgets all about
     # these mappings.
-    natCloseChan.open()
+    activeMappings.add(
```
dryajov, line 318: Since this is accessed in other threads concurrently, I would protect it with a lock. I think this is the only thing that would need locking. Here is some useful info on locks - https://nim-lang.org/docs/manual.html#guards-and-locks

dryajov (commented):
OK, this is much better. Lets make that one change with the locking.

codex/nat.nim
gmega: I don't think this is accessed by multiple threads?

codex/nat.nim
```diff
@@ -268,54 +271,68 @@ proc stopNatThread() {.noconv.} =
...
+        externalTcpPort: externalTcpPort,
+        internalUdpPort: udpPort,
+        externalUdpPort: externalUdpPort,
+        description: description,
+      )
+    )
     try:
-      natThread.createThread(
-        repeatPortMapping, (externalTcpPort, externalUdpPort, description)
+      natThreads.add(Thread[PortMappingArgs]())
```
gmega, line 328: I'm guessing we're doing a thread per mapping cause that's the easiest way.

codex/nat.nim
```diff
@@ -233,30 +243,23 @@ proc repeatPortMapping(args: PortMappingArgs) {.thread, raises: [ValueError].} =
...
+        # we're being silly here with this channel polling because we can't
+        # select on Nim channels like on Go ones
+        currTime = now()
+      if currTime >= (lastUpdate + interval):
+        discard doPortMapping(strategy, tcpPort, udpPort, description)
+        lastUpdate = currTime
+
         sleep(sleepDuration)
 
 proc stopNatThread() {.noconv.} =
```
gmega, line 257:
```suggestion
proc stopNatThreads() {.noconv.} =
```

codex/nat.nim
```diff
@@ -268,54 +271,68 @@ proc stopNatThread() {.noconv.} =
...
-        repeatPortMapping, (externalTcpPort, externalUdpPort, description)
+      natThreads.add(Thread[PortMappingArgs]())
+      natThreads[^1].createThread(
+        repeatPortMapping, (strategy, externalTcpPort, externalUdpPort, description)
       )
       # atexit() in disguise
-      addQuitProc(stopNatThread)
+      if natThreads.len == 1:
+        # we should register the thread termination function only once
+        addQuitProc(stopNatThread)
```
gmega, line 335:
```suggestion
        addQuitProc(stopNatThreads)
```

gmega (approved):
LGTM

PR #1249, opened 2025-06-02
fix(purchasing): fix crash completing future more than once

### LINKED ISSUE #962: [BUG] Crash: attempt to complete future more than once.
Application crashed with exception in async-proc. Seen in v0.1.6 - Autoclient run.

Unhandled exception in async proc, aborting topics="codex" tid=3841 msg="

An attempt was made to complete a Future more than once. 
Details:
  Future ID: 132808371808528
  Creation location:
    submitted.nim(29)    [unspecified]
  First completion location:
    submitted.nim(31)    [unspecified]
  Second completion location:
    submitted.nim(31)    [unspecified]

"

markspanbroek:
Fixes #962

### DIFF
--- a/codex/purchasing/states/submitted.nim
+++ b/codex/purchasing/states/submitted.nim
@@ -30,12 +30,12 @@ method run*(
     requestId = purchase.requestId
 
   proc wait() {.async.} =
-    let done = newFuture[void]()
+    let done = newAsyncEvent()
     proc callback(_: RequestId) =
-      done.complete()
+      done.fire()
 
     let subscription = await market.subscribeFulfillment(request.id, callback)
-    await done
+    await done.wait()
     await subscription.unsubscribe()
 
   proc withTimeout(future: Future[void]) {.async.} =


codex/purchasing/states/submitted.nim
```diff
@@ -30,12 +30,12 @@ method run*(
     requestId = purchase.requestId
 
   proc wait() {.async.} =
-    let done = newFuture[void]()
+    let done = newAsyncEvent()
     proc callback(_: RequestId) =
-      done.complete()
+      done.fire()
```
dryajov, line 35: Is the callback supposed to be called more than once, if not this is probably an issue somewhere else?

emizzle (approved):
Thanks Mark. As with #1248, this is a fix for preventing the node from crashing, but it doesn't quite explain why this subscription callback is being called twice. Therefore, I'd be hesitant to close #947 when merging this.

codex/purchasing/states/submitted.nim
emizzle: Agreed. This smells like a subscription callback or ethers error, or maybe the event is indeed being fired more than once in the contract.

The root issue will be masked by this change. I don't think it's a big deal personally, but it could potentially hide other issues.

markspanbroek: Agreed that it should not be called more than once. However it might be called more than once if the JSON-RPC provider decides to invoke the callback more than once for whatever reason (chain reorg?), so we'd better not crash if it does :smile:

markspanbroek:
I took a look into `market.nim` the `nim-ethers` subscription handling, and a possible reason why it could be invoked more than once could be because of the resubscribing to events when we encounter an error in the JSON-RPC response when polling. A good fix for that would be to implement https://github.com/codex-storage/nim-ethers/issues/96.

Issue #1253, opened 2025-06-04
Improve release process - Phase 1 / Conventional Commits


veaceslavdoina:
During the last offsite, @AuHau proposed to improve an existing release process.

It was discussed to split implementation in two phases
 1. PR title adheres to the [Conventional Commits](https://www.conventionalcommits.org) specification
 2. Automate a release to build changelog based on PR titles

### Phase 1 - PR title adheres to the Conventional Commits specification

 1. Repository rulesets permits to set [metadata restrictions](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-rulesets/creating-rulesets-for-a-repository#adding-metadata-restrictions) just [Using regular expressions for commit metadata](https://docs.github.com/en/enterprise-cloud@latest/organizations/managing-organization-settings/creating-rulesets-for-repositories-in-your-organization#using-regular-expressions-for-commit-metadata) and **not for PR**.
 2. However, we could use [ytanikin/pr-conventional-commits](https://github.com/ytanikin/pr-conventional-commits) action to follow Conventional Commits in PR titles.

### Phase 1 - Actions to perform

 During the first phase we are planning to perform
 1. Add GitHub actions to the CI workflow to check PR titles - #1254
 2. **Required:** PR titles must follow Conventional Commits
 3. Optional: Commits messages should follow Conventional Commits

### Examples

```shell
# CI
ci: use macos arm runners (#1174)

# Docs with scope
docs(openapi): freeSize non-optional (#1211)

# Breaking
network!: support for uPnP and PMP nat traversal #1050

# Chore
chore: bump Nim to 2.2.4 (#1242)
```

**Template**
```shell
<type>[optional scope]: <description>
```

**Types**
```shell
fix, feat, chore, ci, test, docs
build, style, refactor, perf, revert
```

**Scopes**
```shell
cli
configuration
api
openapi
network
smart contracts
p2p
```

### Known issues

1. When PR was created with a wrong title, to pass title check will require a new push or a force-push

PR #1254, opened 2025-06-04
ci: validate pr title to adhere conventional commits

veaceslavdoina:
PR intends to implement a PR title check for CI workflow for #1253.

It works in the following way
1. User create a PR
2. All workflow jobs are run including `pr-title`, which was added in this PR
3. Job `pr-title` is mandatory to pass branch status check
4. When PR title is not align wit the rules, we should edit it and make a force-push to re-run `pr-title` job

> [!NOTE]
> We will add `pr-title` job to the branch protection rules only after this PR will be merged.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,6 +16,15 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  pr-title:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


AuHau (approved):
LGTM!

I agree with taking the second approach. People will learn (hopefully fast) and worst case is force push 👍

2-towns (approved):

PR #1258, opened 2025-06-05
fix(sales): fix marketplace block expiry

markspanbroek:
Previously we used request.expiry as the block expiry, but the former is a duration in seconds, whereas the latter is a timestamp.

- Slot queue items can now either contain an expiry (when they represent a newly created request) or not contain an expiry (when they represent a recently freed slot that need repairing)
- Slot queue items are now added to the sales agent state, so that they can be re-added easily, instead of having to recreate the slot queue item from on-chain state.

Extracted the fix from #1196

### DIFF
--- a/tests/codex/helpers/mockmarket.nim
+++ b/tests/codex/helpers/mockmarket.nim
@@ -14,6 +14,7 @@ from pkg/ethers import BlockTag
 import codex/clock
 
 import ../examples
+import ./mockclock
 
 export market
 export tables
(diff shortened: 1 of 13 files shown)


benbierens:
Commit `sha-6e55e37` passes the release tests.

tests/codex/helpers/mockmarket.nim
```diff
@@ -541,15 +547,23 @@ method queryPastStorageRequestedEvents*(
 ): Future[seq[StorageRequested]] {.async.} =
   return market.requested.map(
     request =>
-      StorageRequested(requestId: request.id, ask: request.ask, expiry: request.expiry)
+      StorageRequested(
+        requestId: request.id,
+        ask: request.ask,
+        expiry: market.requestExpiry[request.id].uint64,
```
2-towns, line 553: Here the expiry should be the duration no ?

2-towns: Oh no sorry, it is ok

PR #1259, opened 2025-06-05
feat(cirdl): auto-discovery of marketplace contract

### LINKED ISSUE #987: Add markerplace contract autodiscovery for Circuit download tool
**Is your feature request related to a problem? Please describe.**
Simplify `cirld` usage and Codex storage node upgrade.

**Describe the solution you'd like**
Circuit downloader (`cirdl`) should support marketplace contract address autodiscovery, like it is done for Codex now.

**Describe alternatives you've considered**
Now, we are passing marketplace address manually and it is not so convenient, especially if we would like to automate Codex upgrade.

**Additional context**
`cirdl` binary should support the following CLI arguments
- `cirdl --help` - show help about how to use the app
- `cirdl --version` - show app version, to identify compatible Codex version

We also, should be able to run `cirdl` by passing just `[circuitPath] [rpcEndpoint]`
```shell
cirdl data/circuits http://rpc.url
```
and optionally pass a custom marketplace address if we would like to use a different address
```
cirdl data/circuits http://rpc.url 0x5Bd66fA15Eb0E546cd26808248867a572cFF5706
```

In that way we can simplify Codex upgrade process and skip marketplace address passing. It would be very usefully, especially for Storage providers, because for now latest contract can be found in the documentation or [nim-codex source code](https://github.com/codex-storage/nim-codex/blob/master/codex/contracts/deployment.nim).

AuHau:
Closes #987 

I know it is bit hacky having the `CodexConf` object here, but I did not really want to refactor the `Deployment` module to move this part out so went about it like this.

### DIFF
--- a/codex/codex.nim
+++ b/codex/codex.nim
@@ -105,7 +105,7 @@ proc bootstrapInteractions(s: CodexServer): Future[void] {.async.} =
         quit QuitFailure
       signer = wallet
 
-    let deploy = Deployment.new(provider, config)
+    let deploy = Deployment.new(provider, config.marketplaceAddress)
     without marketplaceAddress =? await deploy.address(Marketplace):
       error "No Marketplace address was specified or there is no known address for the current network"
       quit QuitFailure
--- a/codex/contracts/deployment.nim
+++ b/codex/contracts/deployment.nim
@@ -33,7 +33,9 @@ proc getKnownAddress(T: type, chainId: UInt256): ?Address =
   return knownAddresses[id].getOrDefault($T, Address.none)
 
 proc new*(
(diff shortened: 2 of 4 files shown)


AuHau:
The failing test does not really seems like related to my changes - @benbierens have you seen something like this before?

AuHau:
> Where this address (0x322813fd9a801c5507c9de605d63cea4f2ce6c44) is coming from?

Well, you have specified it as CLI param ;-)  `--marketplace-address=0x322813fd9a801c5507c9de605d63cea4f2ce6c44`

> The failing test does not really seems like related to my changes

Hmm, it seems like it was a flaked run? Second run passed.

benbierens (approved):
This should do it!

PR #1260, opened 2025-06-06
Simple optimizations to block protocol

gmega:


gmega:
This got in as part of #1324

PR #1262, opened 2025-06-09
feat(docker): add codex network support for docker-entrypoint

veaceslavdoina:
PR updates Docker entrypoint and add additional variables to simplify Codex run using Docker
- `BOOTSTRAP_NODE_FROM_URL` - we can pass an URL with the list of bootstrap nodes (https://spr.codex.storage/testnet).
- `MARKETPLACE_ADDRESS_FROM_URL` - we can pass an URL with the marketplace address (https://marketplace.codex.storage/codex-testnet/latest).
    Even if we already have marketplace autodiscovery for `codex` and `cirdl`(#1259), it is limited to a single contract for a chain. Let's consider that variable as a temporary workaround.
- `NETWORK` - we can simplify `BOOTSTRAP_NODE_FROM_URL` variable configuration by passing just a network name like `testnet`/`devnet`.
- Show variables and arguments passed at run.

**Examples**
```shell
docker run --rm \
  -e NETWORK=testnet \
  codexstorage/nim-codex:sha-3c7d50b-dist-tests

docker run --rm \
  -e BOOTSTRAP_NODE_FROM_URL=https://spr.codex.storage/testnet \
  codexstorage/nim-codex:sha-3c7d50b-dist-tests

docker run --rm \
  -e BOOTSTRAP_NODE_FROM_URL=https://spr.codex.storage/testnet \
  -e MARKETPLACE_ADDRESS_FROM_URL=https://marketplace.codex.storage/codex-testnet/latest \
  codexstorage/nim-codex:sha-3c7d50b-dist-tests
```
Screenshot 2025-06-09 at 19 28 55

### DIFF
--- a/codex/erasure/erasure.nim
+++ b/codex/erasure/erasure.nim
@@ -419,8 +419,8 @@ proc encodeData(
 
         trace "Adding parity block", cid = blk.cid, idx
         cids[idx] = blk.cid
-        if isErr (await self.store.putBlock(blk)):
-          trace "Unable to store block!", cid = blk.cid
+        if error =? (await self.store.putBlock(blk)).errorOption:
+          warn "Unable to store block!", cid = blk.cid, msg = error.msg
           return failure("Unable to store block!")
         idx.inc(params.steps)
 
@@ -619,8 +619,8 @@ proc decode*(self: Erasure, encoded: Manifest): Future[?!Manifest] {.async.} =
             return failure(error)
 
           trace "Recovered block", cid = blk.cid, index = i
-          if isErr (await self.store.putBlock(blk)):
-            trace "Unable to store block!", cid = blk.cid
+          if error =? (await self.store.putBlock(blk)).errorOption:
+            warn "Unable to store block!", cid = blk.cid, msg = error.msg
             return failure("Unable to store block!")
 
           cids[idx] = blk.cid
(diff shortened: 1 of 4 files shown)


docker/docker-entrypoint.sh
```diff
@@ -9,7 +9,12 @@ if [[ -n "${ENV_PATH}" ]]; then
   set +a
 fi
 
-# Bootstrap node from URL
+# Codex Network
+if [[ -n "${NETWORK}" ]]; then
+  export BOOTSTRAP_NODE_FROM_URL="${BOOTSTRAP_NODE_FROM_URL:-https://spr.codex.storage/${NETWORK}}"
```
AuHau, line 14: Shouldn't there also be a setting of `MARKETPLACE_ADDRESS_FROM_URL` based on the `$NETWORK`? Or are you already depending on the auto-discovery?

AuHau: > So, we could set codex-testnet/latest value from the linea-testnet/latest and later update a default value in Docker entrypoint and end users only have to restart their nodes to fetch an updated address.

Honestly, not a good idea. Given that specific smart contract deployments need specific support for it in `nim-codex`, it could easily happen that a user would "simply restart" their node and accidentally start to use non-compatible smart contract version.

AuHau (approved):

Issue #1263, opened 2025-06-10
[BUG] Disk full of orphaned blocks


benbierens:
Seen in the v0.2.3 testnet.

Cloud node 03 and 04 are logging:
`LevelDbDatastore.put exception: IO error: /data/meta/089621.ldb: No space left on device`

The cloud nodes each have 100GB of space allocated to them.

The `/data` endpoint shows 3 datasets present. Their total size: 987430912 => 0.9GB
The `/space` endpoint shows:
```json
{
	"totalBlocks": 1530468,
	"quotaMaxBytes": 105226698752,
	"quotaUsedBytes": 100272881146,
	"quotaReservedBytes": 4928552960
}
```

"quotaUsedBytes": 100272881146 => 93.4 GB

The disk is full of blocks that do not belong to a dataset: Orphaned blocks.

Maintenance module is in charge of cleaning up these blocks. It appears to be running, but unable to delete the blocks. Logs reveal many occurances of:
```
TRC 2025-06-09 09:30:32.428+00:00 Attempting to delete a block               topics="codex repostore" tid=1 cid=zDx*6GeRLL count=30577438
TRC 2025-06-09 09:30:32.428+00:00 Unable to delete block from repoStore      tid=1 count=30577439
```

To fix:
Maintenance module code does not show the error message. So we don't know why it can't delete the blocks.
Maintenance logging doesn't use a log topic.
Timer util used by maintenance module catches CancelledError, but (maybe?) silently swallows any other errors.

benbierens:
Oh shit. It's a deliverate design choice by LevelDB:
```
      if (sync_error) {
        // The state of the log file is indeterminate: the log record we
        // just added may or may not show up when the DB is re-opened.
        // So we force the DB into a mode where all future writes fail.
        RecordBackgroundError(status);
      }
```
Once it hits this state, it knows it can't recover:

LevelDB never recovers from a "no space left on device" (or similar fatal write error) because it records the error as a permanent background error and refuses all future write operations. This is by design, to avoid data corruption or inconsistent state.

PR #1264, opened 2025-06-10
Improves logging in maintenance module and erasure.

benbierens:
This will help us figure out what's wrong with orphaned blocks filling up the disk. ( https://github.com/codex-storage/nim-codex/issues/1263 )

### DIFF
--- a/codex/utils/timer.nim
+++ b/codex/utils/timer.nim
@@ -24,7 +24,7 @@ type
     name: string
     loopFuture: Future[void]
 
-proc new*(T: type Timer, timerName = "Unnamed Timer"): Timer =
+proc new*(T: type Timer, timerName: string): Timer =
   ## Create a new Timer intance with the given name
   Timer(name: timerName)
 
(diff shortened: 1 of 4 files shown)


codex/utils/timer.nim
```diff
@@ -35,6 +35,9 @@ proc timerLoop(timer: Timer) {.async: (raises: []).} =
       await sleepAsync(timer.interval)
   except CancelledError:
     discard # do not propagate as timerLoop is asyncSpawned
+  except CatchableError as err:
+    error "CatchableError in timer loop", name = timer.name, msg = err.msg
```
emizzle, line 39: How did this compile before if there were no leaking `CatchableErrors` (due to `{.async: (raises: []).}`? If it is now not compiling, where were new exceptions introduced?

codex/utils/timer.nim
```diff
@@ -35,6 +35,9 @@ proc timerLoop(timer: Timer) {.async: (raises: []).} =
...
+  info "Timer loop has stopped", name = timer.name
```
emizzle, line 40: Is this really `info` worthy?

emizzle (approved):
Nice! This should definitely help understand what's happening in the testnet. I left a couple of suggestions, but nothing blocking, so will approve now.

codex/utils/timer.nim
benbierens: Normally, the info only shows up when a timer is stopped. That's not a common thing, but not necessarily an error. However, if there are issues it's good to be able to see whether the timer was stopped.

PR #1265, opened 2025-06-10
ci: docker stable tag

AuHau:
This PR enables building Docker images for custom branch or commit and also adds mutable tag `stable` that will point to the latest released version.

### DIFF
--- a/codex/erasure/erasure.nim
+++ b/codex/erasure/erasure.nim
@@ -419,8 +419,8 @@ proc encodeData(
 
         trace "Adding parity block", cid = blk.cid, idx
         cids[idx] = blk.cid
-        if isErr (await self.store.putBlock(blk)):
-          trace "Unable to store block!", cid = blk.cid
+        if error =? (await self.store.putBlock(blk)).errorOption:
+          warn "Unable to store block!", cid = blk.cid, msg = error.msg
           return failure("Unable to store block!")
         idx.inc(params.steps)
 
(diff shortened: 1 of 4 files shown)


.github/workflows/docker.yml
```diff
@@ -40,5 +45,7 @@ jobs:
     needs: get-contracts-hash
     with:
       tag_latest: ${{ github.ref_name == github.event.repository.default_branch || startsWith(github.ref, 'refs/tags/') }}
+      tag_stable: ${{ startsWith(github.ref, 'refs/tags/') }}
```
emizzle, line 48: What exactly does this do? It checks the ref of the workflow branch for starting with /ref/tags? And somehow this resolves to the greatest version tag??

emizzle (approved):

.github/workflows/docker.yml
AuHau: No, this checks if the build was triggered on `tag` event, eq. when there is release.

AuHau:
Testing the run here: https://github.com/codex-storage/nim-codex/actions/runs/15561890311

.github/workflows/docker.yml
AuHau: No worries ;-) Happy to explain.

> So does this build a stable release?

Well, yes and no 😅 It builds Docker images for the commit/tag that triggered it and for releases (eq. for builds triggered by tags), it will also add Docker tag `stable` to the images. Docker tags are mutable, so it will replace the old `stable`. It is similar to `latest`, but `latest` track latest `master` commits and I wanted something that will track "latest proper release". You can then host Codex with this tag and for updates you just need to redeploy, or there are automatized tools (like [watchtower](https://github.com/containrrr/watchtower)) that will bump your Docker containers when there is new release, but for that I need the `stable` (as I don't want to update for every master commit with `latest`)

AuHau:
> How it is intended to work?

@veaceslavdoina, well if you run the Docker action now, then `actions/checkout` checks out the default branch (if you don't specify the `ref`). So with this PR and the `git-ref` I am enabling building Docker images for commits/branches that have not been merged yet.

AuHau:
Well I just tested it and you are right, it works like that 🙈 Alright, I will scratch my `git-ref` work then and leave only the `stable` tag.

PR #1268, opened 2025-06-11
chore: improve marketplace integration tests

2-towns:
This PR replaces the usage of `check eventually` with an event subscription mechanism in order to reduce the number of requests sent in our tests and avoid connection timeouts due to overload.


tests/integration/marketplacesuite.nim
```diff
@@ -18,10 +18,22 @@ template marketplacesuite*(name: string, body: untyped) =
     var period: uint64
     var periodicity: Periodicity
     var token {.inject, used.}: Erc20Token
+    var requestStartedFut = Future[void].Raising([CancelledError]).init()
+    var requestStartedSubscription: Subscription
+
+    proc onRequestStarted(eventResult: ?!RequestFulfilled) {.raises: [].} =
+      requestStartedFut.complete()
```
markspanbroek, line 25: This will crash if the future is already completed. I would suggest to use an AsyncEvent instead of a raw Future, which can handle this case nicely.

tests/integration/testmarketplace.nim
```diff
@@ -124,12 +123,16 @@ marketplacesuite "Marketplace":
...
+    check eventually(
+      (await token.balanceOf(hostAccount)) - startBalanceHost >=
+        (duration - 5 * 60).u256 * pricePerSlotPerSecond * ecNodes.u256,
+      timeout = 5000,
+      pollInterval = 1000,
+    )
     # Checking that client node receives some funds back that were not used for the host nodes
     check eventually(
       (await token.balanceOf(clientAccount)) - clientBalanceBeforeFinished > 0,
+      pollInterval = 1000,
```
markspanbroek, line 135: This should be the default in asynctest nowadays. Are we using the latest asynctest with this fix: https://github.com/codex-storage/asynctest/commit/73c08f77afc5cc2a5628d00f915b97bf72f70c9b

tests/integration/testmarketplace.nim
```diff
@@ -61,9 +62,8 @@ marketplacesuite "Marketplace":
       tolerance = ecTolerance,
     )
 
-    check eventually(
-      await client.purchaseStateIs(id, "started"), timeout = 10 * 60 * 1000
-    )
+    discard await waitForRequestToStart()
```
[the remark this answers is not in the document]
2-towns: Here is the code of `waitForRequestToStart`: 

```nim
    proc waitForRequestToStart(): Future[Period] {.
        async: (raises: [CancelledError, AsyncTimeoutError])
    .} =
      await requestStartedFut.wait(timeout = chronos.seconds((5 * 60) + 10))
      # Recreate a new future if we need to wait for another request
      requestStartedFut = Future[void].Raising([CancelledError]).init()
```

Do I still to to apply `wait` for `waitForRequestToStart` ?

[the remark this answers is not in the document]
markspanbroek: Ah, you're right, I missed that. Then it's good :+1:

tests/integration/marketplacesuite.nim
markspanbroek: You can do the following: `await event.wait().wait(timeout)`

`AsyncEvent.wait()` gives you a future, and you can call `.wait(timeout)` on any future.

2-towns: Oh thanks !

dryajov: Why are we doing this? Isn't the feature being completed an indication that something is broken?

markspanbroek: > Why are we doing this? Isn't the feature being completed twice an indication that something is broken?

No, there could be multiple requests being started, then this callback is called for each of these requests.

tests/integration/testproofs.nim
```diff
@@ -78,277 +76,286 @@ marketplacesuite "Hosts submit regular proofs":
...
-      # .withLogFile() # uncomment to output log file to tests/integration/logs/<start_datetime> <suite_name>/<test_name>/<node_role>_<node_idx>.log
-      # .withLogTopics("node", "marketplace", "clock")
-      .some,
-      providers: CodexConfigs
-        .init(nodes = 1)
-        .withSimulateProofFailures(idx = 0, failEveryNProofs = 1)
+marketplacesuite(
+  name = "Simulate invalid proofs",
+  stopOnRequestFail = false,
+  body = block:
```
markspanbroek, line 82: I think you can also write this as:
`marketplacesuite("Simulate invalid proofs", stopOnRequestFail=false):`

Then you don't need to change the indentation below.

tests/integration/testvalidator.nim
```diff
@@ -15,214 +15,168 @@ export logutils
...
-        .withSimulateProofFailures(idx = 0, failEveryNProofs = 1)
-        # .debug() # uncomment to enable console log output
-        # .withLogFile() # uncomment to output log file to tests/integration/logs/<start_datetime> <suite_name>/<test_name>/<node_role>_<node_idx>.log
-        # .withLogTopics("sales", "onchain")
-        .some,
-      validators: CodexConfigs
+marketplacesuite(
+  name = "Validation",
+  stopOnRequestFail = false,
+  body = block:
```
markspanbroek, line 21: Same here:
`marketplacesuite("Validation", stopOnRequestFail = false):`

markspanbroek (approved):

tests/integration/testproofs.nim
2-towns: Yes, I already tried but got this error: 

```
Expected one of (first mismatch at [position]):
[3] template marketplacesuite(name: string; body: untyped; stopOnRequestFail = true)
  positional param was already given as named param
```

markspanbroek: Yes, the body would need to be the last parameter of `marketplacesuite` for it to work with the `:` syntax.

benbierens (approved):
I think we need this, ASAP.

PR #1277, opened 2025-06-17
Ensures same verifier artifacts are used for release tests as testnet

benbierens:
Verifier artifacts in nim-codex release v0.2.4 were broken. This was not detected by the containerized testing setup, because, it has its own verifier/networks folder and artifacts.

This PR deletes those artifacts and uses a symlink to ensure that the tests are running on the same artifacts that would be used for the testnet. This will help us detect verifier-related issues in the dist/release tests.


benbierens:
Merging this PR will not be necessary. This fix will be included in:
https://github.com/codex-storage/nim-codex/pull/1278

markspanbroek:
Can we close this now @benbierens?

PR #1280, opened 2025-06-19
fix: Disable erasure early exit during streaming to prevent cancellation conflicts

cnanakos:
When streamEntireDataset() runs background erasure decoding while client streams blocks, erasure early exit was cancelling downloads that the streaming client still needed, causing CancelledError failures.

Add allowEarlyExit parameter to erasure.decode() and disable it during streaming. This makes the background erasure job act as a prefetching helper rather than competing with client reads for the same blocks.

The early exit optimization is less valuable during streaming since the client typically needs most blocks anyway, and cooperative prefetching provides better user experience than download cancellation conflicts.

### DIFF
--- a/codex/erasure/erasure.nim
+++ b/codex/erasure/erasure.nim
@@ -211,6 +211,7 @@ proc prepareDecodingData(
     parityData: ref seq[seq[byte]],
     cids: ref seq[Cid],
     emptyBlock: seq[byte],
+    allowEarlyExit = true,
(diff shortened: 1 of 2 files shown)


gmega (approved):
Right now as things are implemented, I think this is a pretty reasonable thing to do. Later on, as we do better dispersal of data/parity blocks and improve download scheduling, I'm not sure I'd agree with this reasoning; i.e., that it's useless to decode when you're streaming.

PR #1281, opened 2025-06-19
fix(tests): auto import all tests files and fix forgotten tests

markspanbroek:
Imports all files that are named `test*.nim` automatically, so we cannot forget to add them.
Includes compilation fixes for tests that were forgotten, and were therefore not run in the CI before.


AuHau (approved):
Nice!

dryajov (approved):

PR #1287, opened 2025-06-23
refactor(ci): run integration tests in parallel by spinning up more runners

### LINKED ISSUE #1177: Group integration tests so they can be run on separate runners
## Problem
Integration tests take too long to run on CI. Due to the resource requirements of the parallel integration tests, Windows runners were pushed past their limits and were not able to provide proofs in time. Thus, period duration had to be increased, which caused the integration tests to run slower.

## Solution
As discussed with @veaceslavdoina, we can group integration tests such that each group can be run on a separate runner. This, in effect, parallelises the runners. The integration tests being run on each runner can also take advantage of the parallel integration tests. This would also allow us to revert the increased period duration.

```shell
tests
├── integration
│   └── tests
│       ├── api
│       │   ├── testcli.nim
│       │   ├── testecbug.nim
│       │   ├── testrestapi.nim
│       │   └── testrestapivalidation.nim
│       ├── client
│       │   ├── testblockexpiration.nim
│       │   └── testupdownload.nim
│       ├── marketplace
│       │   ├── testmarketplace.nim
│       │   ├── testpurchasing.nim
│       │   └── testsales.nim
│       ├── proving
│       │   └── testproofs.nim
│       └── validation
│           └── testvalidator.nim
├── testIntegration.nim
```
Each group could be run by passing the group name to the GROUP parameter in make, eg:
```bash
make -j12 PARALLEL=1 GROUP=all testIntegration
make -j12 PARALLEL=1 GROUP=api testIntegration
make -j12 PARALLEL=1 GROUP=client testIntegration
make -j12 PARALLEL=1 GROUP=marketplace testIntegration
make -j12 PARALLEL=1 GROUP=proving testIntegration
make -j12 PARALLEL=1 GROUP=validation testIntegration
```
Inside of `testIntegration.nim`, `GROUP` could be used to run all tests in the group directory

markspanbroek:
Spins up a large number of runners that run the integration tests in small batches. Each batch of integration tests does not run for more than 30 minutes.

The goal is to reduce the waiting time for CI to complete.

Consists of a number of changes:
- integration tests are moved into folders that indicate their maximum run length:
  - `tests/integration/1_minute`
  - `tests/integration/5_minutes`
  - `tests/integration/30_minutes`
- a new environment variable `CODEX_INTEGRATION_TEST_INCLUDES` indicates which tests to run when `make testIntegration` is called
- `tools/scripts/ci-job-matrix.sh` creates the github actions jobs that each run a batch of integration tests

Depends on #1288 
Fixes #1177

### DIFF
--- a/tools/scripts/ci-job-matrix.sh
+++ b/tools/scripts/ci-job-matrix.sh
@@ -108,6 +108,19 @@ integration_test () {
   for tests in $(find_tests tests/integration/1_minute | batch 30 ","); do
     integration_test_job $tests
   done
+
+  # fail when there are integration tests with an unknown duration
+  local filter='1_minute\|5_minutes\|30_minutes'
+  local unknown=$(find_tests tests/integration | grep -v "$filter")
+  if [ "$unknown" != "" ]; then
+    echo "Error: Integration tests need to be in either the 1_minute," >&2
+    echo "       5_minutes, or 30_minutes directory, based on the maximum" >&2
+    echo "       running time of the test. This is used to group the" >&2
+    echo "       integration tests into batches to speed up the the" >&2
+    echo "       continuous integration." >&2
+    echo "       Offending tests: $unknown" >&2
+    exit 1
+  fi
 }
 
 # outputs jobs for all test types


dryajov (approved):

emizzle (approved):
This looks great Mark! Once this is in, enabling the parallel integration tests (per runner) will further reduce the time required in CI. 

My only comment is when creating new integration tests, how will a developer know how long it will take to run on CI, and therefore which directory to put it into?

markspanbroek:
> My only comment is when creating a new integration test, how will a developer know how long it will take to run on CI, and therefore which directory to put it into?

Yes, I was thinking about that. I wanted to put in a timeout for the tests, so that if a test takes longer than allowed it would fail in the CI. But that wasn't that straightforward to do, so I didn't implement it. Let's see if that really becomes a problem, then I can invest some more time into this.

Issue #923, opened 2024-09-26
[BUG] Change codex default datadir from compile-time to run-time


veaceslavdoina:
**Describe the bug**
Change codex default datadir from compile-time to run-time

**To Reproduce**
1. Install latest release
   <details>
   <summary>code</summary>

   ```shell
   # Get
   version=v0.1.4
   platform=$(echo `uname -s` | awk '{print tolower($0)}')
   architecture=$([[ `uname -m` == 'x86_64' ]] && echo amd64 || echo arm64)

   curl -LO https://github.com/codex-storage/nim-codex/releases/download/${version}/codex-${version}-${platform}-${architecture}.tar.gz
   curl -LO https://github.com/codex-storage/nim-codex/releases/download/${version}/codex-${version}-${platform}-${architecture}.tar.gz.sha256


   # Check
   shasum -a 256 -c codex-${version}-${platform}-${architecture}.tar.gz.sha256

   # Unpack
   tar -zxvf codex-${version}-${platform}-${architecture}.tar.gz

   # Install
   sudo install codex-${version}-${platform}-${architecture} /usr/local/bin/codex

   # Cleanup
   rm -f \
     codex-${version}-${platform}-${architecture} \
     codex-${version}-${platform}-${architecture}.tar.gz \
     codex-${version}-${platform}-${architecture}.tar.gz.sha256
   ```
2. Run codex
   ```shell
   codex
   ```
3. See error
   ```shell
   WRN 2024-09-26 20:10:02.753+03:00 `--nat` is set to loopback, your node wont properly announce over the DHT topics="codex" tid=5306318
   FAT 2024-09-26 20:10:02.753+03:00 Could not create data directory            tid=5306318 data_dir="/Users/runner/Library/Application Support/Codex" errorMsg="Permission denied" errorCode=13
   ```
   <img width="1344" alt="Screenshot 2024-09-26 at 20 16 39" src="https://github.com/user-attachments/assets/9afe0336-259f-42d6-bbb3-8098427a0108">

**Expected behavior**
Codex should use directories it get on run-time instead of the compile-time one which is `runner` in case of the GitHub releases.

**Screenshots**
<img width="681" alt="Screenshot 2024-09-26 at 20 10 58" src="https://github.com/user-attachments/assets/8768071d-24b9-4a2e-b5bb-7427dddecb2c">

**Environment:**
 - OS: `macOS`, `Linux`, `Windows`
 - Version: `all versions`
 - Tag/Commit: `master`
[...]

PR #1292, opened 2025-06-25
make default dirs runtime, not compile time.

dryajov:


emizzle (approved):

dryajov:
> So, it fixes #923?

yes

Issue #1220, opened 2025-05-06
Implement a Codex library so that clients can use Codex directly from their apps


jrainville:
As a client of Codex, I would like to have access to its APIs and functionalities as part of a library instead of having to spin it in a different process manually.

This also aligns with the new goals of the IFT of having all the subprojects available as libraries that can be used by anyone.

The end goal is that as part of another app, we can start and use Codex features by calling them directly from our code instead of relying on a REST API. This can be done using C Bindings.

PR #1322, opened 2025-09-09
feat: add c binding 

2-towns:
Closes: #1220

### DIFF
--- a/library/codex_thread_requests/requests/node_lifecycle_request.nim
+++ b/library/codex_thread_requests/requests/node_lifecycle_request.nim
@@ -56,11 +56,7 @@ proc readValue*(r: var JsonReader, val: var SignedPeerRecord) =
   val = res.get()
 
 proc readValue*(r: var JsonReader, val: var ThreadCount) =
-  let res = ThreadCount.parse(r.readValue(string))
-  if res.isErr:
-    raise
-      newException(SerializationError, "Cannot parse the thread count: " & res.error())
-  val = res.get()
+  val = ThreadCount(r.readValue(int))
 
 proc readValue*(r: var JsonReader, val: var NBytes) =
   val = NBytes(r.readValue(int))


library/codex_context.nim
```diff
@@ -0,0 +1,198 @@
...
+        "Exception " & eventName & " when calling 'eventCallBack': " &
+        getCurrentExceptionMsg()
+      cast[CodexCallback](ctx[].eventCallback)(
+        RET_ERR, unsafeAddr msg[0], cast[csize_t](len(msg)), ctx[].eventUserData
+      )
+
+proc sendRequestToCodexThread*(
+    ctx: ptr CodexContext,
+    reqType: RequestType,
+    reqContent: pointer,
```
gmega, line 79: Hm... how will this work when we have to upload data to Codex? I suppose reqContent is meant to be a string or byte array, so we will probably have to constrain loading to happen from files and pass a local path here? Also wondering what'll be the best approach for a download API...

library/codex_context.nim
```diff
@@ -0,0 +1,198 @@
...
+from ../codex/codex import CodexServer
+
+type CodexContext* = object
+  thread: Thread[(ptr CodexContext)]
+
+  # This lock is only necessary while we use a SP Channel and while the signalling
+  # between threads assumes that there aren't concurrent requests.
+  # Rearchitecting the signaling + migrating to a MP Channel will allow us to receive
+  # requests concurrently and spare us the need of locks
+  lock: Lock
```
gmega, line 28: Not sure this is a big deal as requests run async and the actual sending of the request to the Codex thread should be relatively fast, plus this shouldn't be a very "hot" API anyway?

library/codex_thread_requests/codex_thread_request.nim
```diff
@@ -0,0 +1,81 @@
...
+    T: type CodexThreadRequest, request: ptr CodexThreadRequest, codex: ptr CodexServer
+) {.async.} =
+  ## Processes the request in the Codex thread.
+  ## Dispatch to the appropriate request handler based on reqType.
+  let retFut =
+    case request[].reqType
+    of LIFECYCLE:
+      cast[ptr NodeLifecycleRequest](request[].reqContent).process(codex)
+
+  handleRes(await retFut, request)
```
gmega, line 78: Hmmm... are we calling back the C code directly from the Chronos thread?

library/codex_thread_requests/requests/node_lifecycle_request.nim
```diff
@@ -0,0 +1,165 @@
...
+
+proc readValue*[T: InputFile | InputDir | OutPath | OutDir | OutFile](
+    r: var JsonReader, val: var T
+) =
+  val = T(r.readValue(string))
+
+proc readValue*(r: var JsonReader, val: var MultiAddress) =
+  val = MultiAddress.init(r.readValue(string)).get()
+
+proc readValue*(r: var JsonReader, val: var NatConfig) =
```
emizzle, line 36: It feels like these deserialisations are actually quite generic and should maybe live elsewhere. In ethers, we redirect all de/serializations to  nim-serde instead:
```nim
import pkg/serde/json

## Generic conversions to use nim-serde instead of nim-json-serialization for
## json rpc serialization purposes
##  writeValue => `%`
##  readValue  => fromJson

proc writeValue*[T: not JsonNode](
  writer: var JsonWriter[JrpcConv],
  value: T) {.raises:[IOError].} =

  writer.writeValue(%value)

proc readValue*[T: not JsonNode](
  r: var JsonReader[JrpcConv],
  result: var T) {.raises: [SerializationError, IOError].} =

  var json = r.readValue(JsonNode)
  result = T.fromJson(json).getOrRaise(SerializationError)
```

emizzle (commented):
Definitely looks like it's going in the right direction! I tried it out locally and it worked well! 

Just a few suggestions/small comments even though I know this is a draft. I can imagine you were saving exception handling for "clean up". Best for last right 😂

codex/conf.nim
```diff
@@ -517,62 +518,86 @@ proc parseCmdArg*(
     quit QuitFailure
   ma
 
-proc parseCmdArg*(T: type ThreadCount, input: string): T {.upraises: [ValueError].} =
-  let count = parseInt(input)
-  if count != 0 and count < 2:
-    warn "Invalid number of threads", input = input
+proc parse*(T: type ThreadCount, p: string): Result[ThreadCount, string] =
```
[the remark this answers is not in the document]
2-towns: you will still need to import `CodexConf` from `conf`. But yeah I can import only `CodexConf` and move the parsing method into a parser.nim as you suggested.

library/codex_thread_requests/codex_thread_request.nim
```diff
@@ -0,0 +1,81 @@
...
+proc handleRes[T: string | void](
+    res: Result[T, string], request: ptr CodexThreadRequest
+) =
+  ## Handles the Result responses, which can either be Result[string, string] or
+  ## Result[void, string].
+  defer:
+    deallocShared(request)
+
+  if res.isErr():
+    foreignThreadGc:
```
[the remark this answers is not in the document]
2-towns: Well, I’m not sure that’s always the case, especially when a Go routine is involved.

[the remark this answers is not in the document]
2-towns: I mean from what I understood, Go routines will run threads that are considered as `foreign` by Nim so that my be an issue with the GC.

library/libcodex.nim
```diff
@@ -0,0 +1,367 @@
...
+  init(ctx, callback, userData)
+
+  let res = codex_context.destroyCodexContext(ctx)
+  if res.isErr:
+    return callback.error(res.error, userData)
+
+  return callback.success("", userData)
+
+proc codex_upload_init(
+    ctx: ptr CodexContext, filepath: cstring, callback: CodexCallback, userData: pointer
```
emizzle, line 250: What happens with the filepath if you're uploading via chunks? Is it not used?

library/codex_context.nim
```diff
@@ -0,0 +1,208 @@
...
+    # Pop a request from the channel
+    let recvOk = ctx.reqChannel.tryRecv(request)
+    if not recvOk:
+      error "codex thread could not receive a request"
+      continue
+
+    # yield immediately to the event loop
+    # with asyncSpawn only, the code will be executed
+    # synchronously until the first await
+    asyncSpawn (
```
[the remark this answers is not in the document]
2-towns: > I don't think it's really necessary to yield here because there's not a lot of io between here and the first await in `CodexThreadRequest.process`

It is actually needed, I made a try with `start` method. Without this, the start will run and block the thread instead of returning `RET_OK`.

```go
	if C.cGoCodexStart(self.ctx, bridge.resp) != C.RET_OK {
		return bridge.CallError("cGoCodexStart")
	}

	log.Println("Codex is starting... Press Ctrl+C to stop.")

	_, err := bridge.wait()
```

```
2025/09/22 15:44:44 Starting Codex...
DBG 2025-09-22 15:44:44.332+02:00 starting switch for peer                   topics="libp2p switch" tid=462122 
################ Other Codex logs
NTC 2025-09-22 15:44:46.557+02:00 REST service started                       tid=462122 address=127.0.0.1:8080
2025/09/22 15:44:46 Codex is starting... Press Ctrl+C to stop.

```

library/libcodex.nim
2-towns: Well, the filepath can be obviously the absolute file path when you upload a file. 
But when you upload with chunks, it is used as `filename`. Then the `node.store` will use it for the filename and the mimetype.

build.nims
```diff
@@ -121,3 +145,23 @@ task showCoverage, "open coverage html":
...
+  var params = ""
+  when compiles(commandLineParams):
+    for param in commandLineParams():
+      if param.len > 0 and param.startsWith("-"):
+        params.add " " & param
+
+  let name = "libcodex"
+  buildLibrary name, "library/", params, "dynamic"
+
+task libcodextatic, "Generate bindings":
```
nipsysdev: `libcodextatic` seems to be a typo. It is referenced as `libcodexStatic` in the Makefile:
https://github.com/codex-storage/nim-codex/pull/1322/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R263

emizzle (approved):

PR #1325, opened 2025-09-30
feat: Block exchange optimizations

cnanakos:


codex/blockexchange/engine/engine.nim
```diff
@@ -63,30 +65,58 @@ declareCounter(codex_block_exchange_blocks_sent, "codex blockexchange blocks sen
...
+  codex_block_exchange_requests_failed_total,
+  "Total number of block requests that failed after exhausting retries",
+)
 
 const
-  DefaultMaxPeersPerRequest* = 10
+  # The default max message length of nim-libp2p is 100 megabytes, meaning we can
+  # in principle fit up to 1600 64k blocks per message, so 20 is well under
+  # that number.
+  DefaultMaxBlocksPerMessage = 20
```
gmega: I'm surprised the "optimal" number is this low but, then again, the intuition on how to tune this under multiple peers and single-threaded Chronos is not simple.

codex/blockexchange/engine/engine.nim
```diff
@@ -63,30 +65,58 @@ declareCounter(codex_block_exchange_blocks_sent, "codex blockexchange blocks sen
...
+    proc(peers: seq[BlockExcPeerCtx]): BlockExcPeerCtx {.gcsafe, raises: [].}
 
   BlockExcEngine* = ref object of RootObj
     localStore*: BlockStore # Local block store for this instance
-    network*: BlockExcNetwork # Petwork interface
+    network*: BlockExcNetwork # Network interface
     peers*: PeerCtxStore # Peers we're currently actively exchanging with
     taskQueue*: AsyncHeapQueue[BlockExcPeerCtx]
-      # Peers we're currently processing tasks for
+    selectPeer*: PeerSelector # Peers we're currently processing tasks for
```
gmega:
```suggestion
    taskQueue*: AsyncHeapQueue[BlockExcPeerCtx] # Peers we're currently processing tasks for
    selectPeer*: PeerSelector  # Algorithm for selecting a peer from which to request a block
```

The comment refers to `taskQueue`, I believe. I've inlined them but nph will probably reformat. 🙂

codex/blockexchange/engine/engine.nim
```diff
@@ -154,8 +183,133 @@ proc sendWantBlock(
...
+        peer.wantsUpdated
+    else:
+      trace "No changes in want list, skipping send", peer = peer.id
+      peer.lastSentWants = toAsk
+  else:
+    trace "Sending full want list", peer = peer.id, length = toAsk.len
+
+    let toAskSeq = toAsk.toSeq
+    var offset = 0
+    while offset < toAskSeq.len:
```
gmega: Seems like you could extract the batched send of the want lists in a small internal proc? This looks very similar to line 215.

codex/blockexchange/engine/engine.nim
```diff
@@ -154,8 +183,133 @@ proc sendWantBlock(
...
+    # small updates.
+    #
+
+    # In dynamic swarms, staleness will dominate latency.
+    let
+      hasNewBlocks = peer.lastRefresh < self.pendingBlocks.lastInclusion
+      isKnowledgeStale = peer.isKnowledgeStale
+
+    if isKnowledgeStale or hasNewBlocks:
+      if not peer.refreshInProgress:
```
gmega: I'm also a bit afraid of this... if the peer doesn't reply for some reason, you'll never request it again. At the same time retries should probably not be handled at this layer, and we'd likely want to have a consistent policy in which we timeout peers that fail to respond in a timely fashion.

The way this is implemented now, however, we have these "untracked" requests in which a reply is expected but then we don't track it and don't do anything about it.

In the future, we probably want to have a "reply expected" abstraction in which we set a timeout and penalize the peer for failing to be timely somehow (either by decreasing its score or disconnecting it). None of that stuff is structured clearly in the current implementation.

codex/blockexchange/engine/engine.nim
```diff
@@ -154,8 +183,133 @@ proc sendWantBlock(
...
+        peer.refreshRequested()
+        await self.refreshBlockKnowledge(
+          peer, skipDelta = isKnowledgeStale, resetBackoff = hasNewBlocks
+        )
+    else:
+      trace "Not refreshing: peer is up to date", peer = peer.id
+
+    if (Moment.now() - lastIdle) >= runtimeQuota:
```
gmega: I guess this is to avoid blocking the event loop for the entire update cycle? You could get a more fine-grained schedule if you embedded this into the batch sends... not sure it's worth the trouble though given that your numbers are already good.

codex/blockexchange/peers/peercontext.nim
```diff
@@ -25,28 +25,77 @@ import ../../logutils
...
-proc peerHave*(self: BlockExcPeerCtx): seq[BlockAddress] =
-  toSeq(self.blocks.keys)
+proc refreshReplied*(self: BlockExcPeerCtx) =
+  self.refreshInProgress = false
+  self.lastRefresh = Moment.now()
+  self.refreshBackoff = min(self.refreshBackoff * 2, MaxRefreshBackoff)
 
-proc peerHaveCids*(self: BlockExcPeerCtx): HashSet[Cid] =
-  self.blocks.keys.toSeq.mapIt(it.cidOrTreeCid).toHashSet
+proc havesUpdated(self: BlockExcPeerCtx) =
```
gmega: Bleh I added this but can't fully recall the rationale...

codex/node.nim
```diff
@@ -186,33 +189,61 @@ proc fetchBatched*(
...
-    without blockResults =? await allFinishedValues[?!bt.Block](blockFutures), err:
-      trace "Some blocks failed to fetch", err = err.msg
-      return failure(err)
+    inc(successfulBlocks)
+    inc(completedInWindow)
 
-    let blocks = blockResults.filterIt(it.isSuccess()).mapIt(it.value)
+    if not onBatch.isNil:
+      blockData.add(blk)
+      if blockData.len >= maxCallbackBlocks:
```
gmega: This is OK, but I'm surprised you had to change this - is it actually doing anything in your case? `fetchBatched` is called with a null callback on the regular download path, so you should not run into memory issues here with benchmarks we had.

If you're running marketplace, on the other hand, then you get a non-nil callback from `onStore`, and what you're doing here should make more of a difference.

gmega (commented):
Code-wise this looks fine (added a few comments but see no showstoppers), but the tests are not passing? 🙂 

I'm gonna approve this anyway as we have to expedite it and I don't block you.

codex/blockexchange/engine/engine.nim
```diff
@@ -154,8 +183,133 @@ proc sendWantBlock(
   ) # we want this remote to send us a block
   codex_block_exchange_want_block_lists_sent.inc()
 
-proc randomPeer(peers: seq[BlockExcPeerCtx]): BlockExcPeerCtx =
-  Rng.instance.sample(peers)
+proc refreshBlockKnowledge(
+    self: BlockExcEngine, peer: BlockExcPeerCtx, skipDelta = false, resetBackoff = false
+) {.async: (raises: [CancelledError]).} =
+  if self.pendingBlocks.wantListLen == 0:
+    if peer.lastSentWants.len > 0:
```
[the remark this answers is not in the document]
cnanakos: Yes you are right, this is a memory leak. Stale entries are never cleaned up when blocks got resolved/cancelled. 

This can be fixed by adding incremental cleanup. On each peer refresh we remove a batch of stale entries with proper yielding to avoid blocking the event loop.  Cleanup can happen gradually across refresh cycles, keeping memory bounded  without a huge performance impact.

codex/blockexchange/engine/engine.nim
cnanakos: Good catch! Pushed a commit with the fix.

PR #1326, opened 2025-10-03
chore: skip marketplace integration tests 

2-towns:
This PR skips the Marketplace integration tests (located in the 30-minutes folder).

The tests take around 30 minutes to run and slow to run locally. For now, we are not using Marketplace integrations, so it makes sense to disable them.


emizzle:
Maybe we should remove testsales and testpurchasing from the 1- and 5-minute folders?

PR #1339, opened 2025-10-31
fix(ci): Windows SIGILL in CI 

emizzle:
Prevents the SIGILL in windows ci, which indicates that nim was being built with an illegal instruction set. By specifying --march=x86_64, it tells gcc to enable a more compatible (generic) set of cpu instructions. See https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html#index-march-15 for more info.

Also reverts "excludes" from test import as it is no longer needed.


2-towns (approved):
Thanks !

cnanakos (approved):
LGTM.

PR #1342, opened 2025-11-05
fix(ci): introduce a number of integration test fixes

emizzle:
A number of bugs were encountered in windows CI during development of the parallel integration tests. Fixes for these bugs were ported to this PR without enabling parallel integration tests to see if this alleviates intermittent test failures in windows ci.

Changes include:
- refactor of integration tests to include manually force killing windows processes, preventing hangs
- more graceful shutdown, allowing for successful restarts 
- adding unique cache key constraints per workflow and run in an attempt to avoid "illegal instruction" failures in CI.


2-towns (approved):
Okay I let few comments. 

Thanks for this huge work ! 

And sorry to didn't pay attention to this before.

PR #1343, opened 2025-11-11
fix(block exchange): decrements block retries count when downloading with no peers

emizzle:
There is a limitation in `downloadInternal` that infinitely loops for peers when attempting to download a block if the node does not have peers. Currently, the block retry counter is only decremented if there exists peers and a download attempt is made, otherwise the loop will continue until peers are found, ad infinium. This fix decrements the retry counter on each loop that no peers are found. Additionally, it waits for the `DiscoveryRateLimit` interval before retrying, as otherwise looking for peers would not produce new results.

### DIFF
--- a/codex/blockexchange/engine/engine.nim
+++ b/codex/blockexchange/engine/engine.nim
@@ -972,7 +972,7 @@ proc blockexcTaskRunner(self: BlockExcEngine) {.async: (raises: []).} =
       let peerCtx = await self.taskQueue.pop()
       await self.taskHandler(peerCtx)
   except CancelledError:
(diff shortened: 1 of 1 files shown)


codex/blockexchange/engine/engine.nim
```diff
@@ -371,6 +371,12 @@ proc downloadInternal(
           break
         # If we still don't have the block, we'll go for another cycle.
         trace "No peers for block, will retry shortly"
+        # Without decremeting the retries count, this would infinitely loop
```
cnanakos: decremeting -> decrementing

cnanakos (approved):
LGTM!

PR #1344, opened 2025-11-11
fix: parse config using int for NBytes instead of string

2-towns:
This issue was discovered first [here](https://github.com/codex-storage/codex-go-bindings/issues/12) when trying to use `StoringQuota`. The value expected in the json should be a int and not a string.


emizzle:
> Actually I think `BlockTtl` should be a string in the Go config object because it makes sense to have a string, the duration can be parsed like "10H".

Ah, that's really nice! But in that case, `Config.BlockTtl` should be an `string` instead of an `int`. Also, would "10" be valid for seconds? I ask this because the Codex help text makes me think a single digit, even as a string, should work: 
> Default block timeout in seconds - 0 disables the ttl

PR #1350, opened 2025-12-11
chore: assorted bumps

arnetheduck:
* fix branch names in gitmodules that were pointing to non-existant branches (allows running `git submodule update --remote` to find out-of-date submodules)
* make sure commits from canonical branches are used for submodules
* "small" library bumps that deal with small correctness issues, memory leaks, threading fixes and other general performance/bugfixing
* metrics: make it work better with multithreading, which makes debugging easier
* stew: more compile-time support and an assortment of other fixes
* taskpools: block invalid use of GC types at compile time
* nimcrypto: accelerated sha2
* upraises: remove submodule

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -37,22 +37,17 @@
 	path = vendor/nim-nitro
 	url = https://github.com/status-im/nim-nitro.git
 	ignore = untracked
-	branch = master
+	branch = main
 [submodule "vendor/questionable"]
 	path = vendor/questionable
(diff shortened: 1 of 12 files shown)


arnetheduck:
#1351 compiles where this one doesn't - the difference is nimcrypto, though I suspect that this is a red herring since this PR also worked on one CI host (https://github.com/logos-storage/logos-storage-nim/actions/runs/20133800291 )

PR #1360, opened 2025-12-17
feat: async tree building v2

arnetheduck:
This PR updates the async tree building, making it compatible with refc and fixing several efficiency problems in the original implementation.

The changes are split in to 3 parts described more in detail in the individual commits:

* [minimal threading fixes](https://github.com/logos-storage/logos-storage-nim/commit/99a78a41ee0ba5b156a51852812f4df342ac6647) - this commit addresses the main correctness issues in the original proposal - in particular, garbage-collected types must not cross threads when using `refc` since they are allocated from a shared heap
* [perf fixes](https://github.com/logos-storage/logos-storage-nim/commit/db8f81cd63f7250c9da0337556b1351977a2a9a3) - many of the threading and performance issues can actually be traced back to choices made in the original implementation - refactoring it helps both with performance and code simplicity
* [taskpools bump](https://github.com/logos-storage/logos-storage-nim/commit/ac5e3c733ba3b02ae1919b3879cdea842c795932) - which further simplifies the code

### DIFF
--- a/tests/codex/merkletree/testcodextree.nim
+++ b/tests/codex/merkletree/testcodextree.nim
@@ -115,8 +115,8 @@ suite "Test CodexTree":
 
     # Single-leaf trees have their root separately computed
     let
-      atree1 = (await CodexTree.init(tp, leaves = expectedLeaves[0..0]))
-      stree1 = CodexTree.init(leaves = expectedLeaves[0..0])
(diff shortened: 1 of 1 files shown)


arnetheduck:
> I've [studied those changes](https://hackmd.io/VRn7qfUIQ1mAXRjtiGUdnw), and it looks ready to merge to me.

added a few comments

PR #1363, opened 2025-12-24
chore: update nix config

2-towns:
This PR updates the Nix configuration to build Logos Storage.

It updates the sha256 and fixes a regex used for revision version matching.

### DIFF
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -83,9 +83,14 @@ in pkgs.gcc13Stdenv.mkDerivation rec {
   '';
 
   installPhase = ''
-    mkdir -p $out/lib $out/include
-    cp build/*storage* $out/lib/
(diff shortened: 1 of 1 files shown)


nix/tools.nix
```diff
@@ -9,7 +9,12 @@ in {
     let
       linesFrom = file: splitString "\n" (fileContents file);
       matching = regex: lines: map (line: match regex line) lines;
-      extractMatch = matches: last (flatten (remove null matches));
+      extractMatch = matches:
+        let xs = flatten (remove null matches);
+        in if xs == [] then
+          throw "findKeyValue: no match for regex '${regex}' in ${toString sourceFile}"
```
markoburcul: Good catch, thanks :)

PR #1376, opened 2026-01-16
fix: downsizing fixes to release workflow and docker entrypoint

gmega:
We were still trying to build cirdl at every merge; then I ended up finding a few more things to remove.

Plus: 
 - prevent segmentation fault when the discovery store is already closed.
 - remove the discovery `stop` method in the server `stop` method because it is already stopped in the node `stop` method


2-towns:
@gmega can we merge this ?

PR #1398, opened 2026-02-02
fix(ci): release process producing wrong artifacts

### LINKED ISSUE #1397: Release process producing wrong artifacts
Each release should produce:
1. logos storage binary for each os/arch
2. libstorage library for each os/arch
3. additional dlls for windows

emizzle:
Fixes #1397.

Release was missing quite a few artifacts as assets on the release. This PR rectifies that.

It also: 
- prevents double zipping that was occurring
- prevents unnecessary artifacts from being added to release (eg libstorage.dll)
- renames many workflow steps for clarity
- renames and normalises env vars for readability
- normalises run type conditional (release vs branch)

Example output for a released version: https://github.com/logos-storage/logos-storage-nim/releases/tag/v0.3.1
Example output for a branch merge (eg on PR merged to master): https://github.com/logos-storage/logos-storage-nim/actions/runs/21585089878

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -96,18 +96,18 @@ jobs:
         if: matrix.os == 'linux'
         run: |
           sudo apt-get update && sudo apt-get install -y zip
-          cd ./build && zip -j "${{env.storage_binary}}.zip" ./*
+          zip -j "${{ env.build_dir }}/${{env.storage_binary}}.zip" ./${{ env.build_dir }}/*
           
       - name: Package ${{ env.storage_binary_base }} MacOS (compress and preserve perms)
         if: matrix.os == 'macos'
         run: |
-          cd ./build && zip -j "${{env.storage_binary}}.zip" ./*
+          zip -j "${{ env.build_dir }}/${{env.storage_binary}}.zip" ./${{ env.build_dir }}/*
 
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -59,109 +60,123 @@ jobs:
...
+          7z a -tzip "${{ env.build_dir }}/${{ env.storage_binary }}-dlls.zip" ./${{ env.build_dir }}/dlls/*.dll
+
+      - name: Upload Windows dlls to workflow artifacts
         if: matrix.os == 'windows'
         uses: actions/upload-artifact@v4
         with:
-          name: ${{ env.storage_binary }}-dlls
-          path: ${{ env.build_dir }}/*.dll
+          name: ${{ env.storage_binary }}-dlls.zip
+          path: ${{ env.build_dir }}/${{ env.storage_binary }}-dlls.zip
```
gmega: Hmm so we were uploading those without zipping, now we're zipping them? Oh well.

emizzle: Faster uploads, counts for less storage quota, and can be parallelised. I didn't think there was much of a downside unless there were automations like docker downloading the uncompressed archive directly. Anyway, the filename has changed, so if these processes exist, they will need to be updated regardless.

PR #1400, opened 2026-02-04
fix: propagate callback cancellation

gmega:
The callback in the block maintainer was swallowing the cancellation exception, causing `stop` to sometimes get stuck on node shutdown. This is also a nasty one to test: you'd need to guarantee that you'd invoke stop right as the timer is running the callback, which is difficult to do without some refactoring.


emizzle:
Nice find! 
> need to guarantee that you'd invoke stop right as the timer is running the callback, which is difficult to do without some refactoring.

I think this can be done by trying the following:

In the test, you could start a new instance of `BlockMaintainer` with a very short interval so `onTimer` would be triggered quickly. Inside `onTimer` do a `sleepAsync` for 1 second. Then immediately `stop` the `BlockMaintainer`. That should set the situation up where the awaited sleepAsync yields to stop the blockmaintainer/timer.

PR #1409, opened 2026-02-16
feat(cli)!: change listen addr to listen port

emizzle:
Allows specifying the port only for the listening address, the address that is announced on the DHT for peer connections. It makes the developer experience less confusing, and easier to change for things like port forwarding.

### DIFF
--- a/library/storage_thread_requests/requests/node_debug_request.nim
+++ b/library/storage_thread_requests/requests/node_debug_request.nim
@@ -52,15 +52,14 @@ proc getDebug(
   let node = storage[].node
   let table = RestRoutingTable.init(node.discovery.protocol.routingTable)
 
-  let json =
-    %*{
-      "id": $node.switch.peerInfo.peerId,
-      "addrs": node.switch.peerInfo.addrs.mapIt($it),
-      "spr":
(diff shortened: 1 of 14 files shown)


2-towns (approved):
Thanks Eric ! Exactly what I need.

codex/conf.nim
```diff
@@ -135,19 +135,29 @@ type
...
+      defaultValueDesc: "Listens on all addresses.",
       abbr: "i",
-      name: "listen-addrs"
-    .}: seq[MultiAddress]
+      name: "listen-ip"
+    .}: IpAddress
+
+    listenPort* {.
+      desc:
+        "Port to listen on for remote peer connections. Announced in the DHT as a multiaddress, eg /ip4/<listen-ip>/tcp/<listen-port>. Port must be in the range 0-65535.",
```
gmega:
```suggestion
        "TCP port to listen on for remote peer connections. Selects a random port if none is specified.",
```

Again, I don't think we need to mention the DHT; also 0-65535 is the valid port range so I don't think it needs to be here either.

gmega (approved):
I think this is a good idea, I'm providing some suggestions to the CLI parameter descriptions.

PR #1411, opened 2026-02-20
chore: Block exchange protocol rewrite

cnanakos:


library/ffi_types.nim
```diff
@@ -19,7 +19,7 @@ const RET_PROGRESS*: cint = 3
 ## Returns RET_OK as acknowledgment and call the callback
 ## with RET_OK code and the provided message.
 proc success*(callback: StorageCallback, msg: string, userData: pointer): cint =
-  callback(RET_OK, cast[ptr cchar](msg), cast[csize_t](len(msg)), userData)
+  callback(RET_OK, unsafeAddr msg[0], cast[csize_t](len(msg)), userData)
```
arnetheduck: not sure if it can happen here, but as a rule it's better to use `baseAddr` from `stew` which avoids a panic when `msg.len == 0`

cnanakos: Good catch! Switched to `baseAddr` from `stew/ptrops`.

storage/blockexchange/engine/downloadmanager.nim
```diff
@@ -0,0 +1,182 @@
...
+logScope:
+  topics = "storage downloadmanager"
+
+const
+  DefaultBlockRetries* = 200
+  DefaultRetryInterval* = 2.seconds
+
+type DownloadManager* = ref object of RootObj
+  nextDownloadId*: uint64 = 1 # 0 is invalid
+  blockRetries*: int
```
gmega:
```suggestion
  maxBlockRetries*: int
```

storage/manifest/coders.nim
```diff
@@ -37,31 +37,32 @@ proc encode*(manifest: Manifest): ?!seq[byte] =
...
   #
-  # var treeRootVBuf = initVBuffer()
   var header = initProtoBuffer()
-  header.write(1, manifest.treeCid.data.buffer)
-  header.write(2, manifest.blockSize.uint32)
-  header.write(3, manifest.datasetSize.uint64)
-  header.write(4, manifest.codec.uint32)
-  header.write(5, manifest.hcodec.uint32)
-  header.write(6, manifest.version.uint32)
+  header.write(1, manifest.manifestVersion)
```
arnetheduck, line 53: fyi https://github.com/status-im/nim-protobuf-serialization is getting some attention by @nitely right now, might be a good moment to play with it and find bugs

storage/blockexchange/engine/downloadmanager.nim
```diff
@@ -0,0 +1,183 @@
...
+
+proc cancelDownload*(self: DownloadManager, cid: Cid) =
+  self.downloads.withValue(cid, innerTable):
+    var toCancel: seq[ActiveDownload] = @[]
+    for _, download in innerTable[]:
+      toCancel.add(download)
+    for download in toCancel:
+      self.cancelDownload(download)
+
+proc releaseDownload*(self: DownloadManager, cid: Cid) =
```
gmega: Why those aliases? Can't we make do with just `cancelDownload`?

storage/stores/networkstore.nim
```diff
@@ -33,52 +33,31 @@ type NetworkStore* = ref object of BlockStore
...
+    let handle = downloadOpt.get().getWantHandle(address)
+    without blk =? (await self.localStore.getBlock(address)), err:
+      if not (err of BlockNotFoundError):
+        handle.cancel()
+        return failure err
+      return await handle
+    discard downloadOpt.get().completeWantHandle(address, some(blk))
+    return success blk
+
+  return await self.localStore.getBlock(address)
```
gmega, line 53: This looks very awkward. Why can't we do:

```nim
method getBlock*(
    self: NetworkStore, address: BlockAddress
): Future[?!Block] {.async: (raises: [CancelledError]).} =
   # Local lookup
    without blk =? (await self.localStore.getBlock(address)), err:
      if not (err of BlockNotFoundError):
        return failure err
    return blk # block found, just return

  # If block not available locally, return completion handle. The download process
  # will eventually get to it.
  let downloadOpt = self.engine.downloadManager.getDownload(address.treeCid)
  if downloadOpt.isSome:
    return await downloadOpt.get().getWantHandle(address)

  # If we got thus far, it means the block is not available locally and there's no
  # active download.
  failure (ref BlockNotFoundError)("Block not found"); # or something like this
```

i.e., I don't get the whole `getWantHandle` followed by immediate `completeWantHandle`.

storage/blockexchange/engine/engine.nim
```diff
@@ -83,68 +63,86 @@ declareCounter(
...
+    activeDownloads*: HashSet[uint64] # Track running download workers by download ID
+
+  DownloadHandleGeneric*[T] = object
+    cid*: Cid
+    downloadId*: uint64
+    iter*: SafeAsyncIter[T]
+    completionFuture*: Future[?!void].Raising([CancelledError])
+
+  DownloadHandle* = DownloadHandleGeneric[Block]
+  DownloadHandleOpaque* = DownloadHandleGeneric[void]
```
gmega, line 93: Do we ever iterate over `DownloadHandleOpaque`? Does it make sense to have `iter` in it?

storage/blockexchange/engine/engine.nim
```diff
@@ -83,68 +63,86 @@ declareCounter(
...
+
+proc waitForComplete*[T](
+    h: DownloadHandleGeneric[T]
+): Future[?!void] {.async: (raises: [CancelledError]).} =
+  return await h.completionFuture
+
+proc nextBlock*(
+    h: DownloadHandle
+): Future[?!Block] {.async: (raw: true, raises: [CancelledError]).} =
+  h.iter.next()
```
gmega:
```suggestion
```
This doesn't seem to be used anywhere, not even test code.

storage/blockexchange/engine/engine.nim
```diff
@@ -83,68 +63,86 @@ declareCounter(
...
+  DownloadHandle* = DownloadHandleGeneric[Block]
+  DownloadHandleOpaque* = DownloadHandleGeneric[void]
+
+proc finished*[T](h: DownloadHandleGeneric[T]): bool =
+  h.iter.finished
+
+proc next*[T](
+    h: DownloadHandleGeneric[T]
+): Future[?!T] {.async: (raw: true, raises: [CancelledError]).} =
+  h.iter.next()
```
gmega: I really wonder if we should maintain this. The only place where we seem to iterate over a `DownloadHandle` (of any kind) is in test code (`downloadDataset`), and even there we're not looking at the blocks themselves, just their CIDs, and even that code could be replaced by something else (e.g. some form of callback). Plus, if we drop support for this, we can remove the whole logic for the asynciterators (I suppose).

storage/blockexchange/engine/engine.nim
```diff
@@ -161,673 +159,948 @@ proc stop*(self: BlockExcEngine) {.async: (raises: []).} =
...
+    if connectedPeers.len > 0:
+      trace "Initial presence window broadcast",
+        cid = cid,
+        windowStart = windowStart,
+        windowCount = windowCount,
+        totalBlocks = download.ctx.totalBlocks
+
+      await self.broadcastWantHave(
+        download, cid, windowStart, windowCount, connectedPeers
+      )
```
gmega: Are we broadcasting this to **every** connected peer? What if there are peers which we are connected to, but are not sharing the file that the `ActiveDownload` refers to (e.g. when we're downloading several files at once?)?

storage/blockexchange/engine/engine.nim
```diff
@@ -161,673 +159,959 @@ proc stop*(self: BlockExcEngine) {.async: (raises: []).} =
...
 
-  let handle = self.pendingBlocks.getWantHandle(address)
-  trace "Downloading block"
   try:
-    while address in self.pendingBlocks:
-      logScope:
-        retries = self.pendingBlocks.retries(address)
-        interval = self.pendingBlocks.retryInterval
+    let
+      (windowStart, windowCount) = download.ctx.currentPresenceWindow()
```
gmega, line 485:
```suggestion
      (windowStart, windowLength) = download.ctx.currentPresenceWindow()
```

I was genuinely confused by the use of the word "count" here. Looking at `currentPresenceWindow` this seems to be window's length:

```nim
proc currentWindow(p: WindowCursor): tuple[start: uint64, count: uint64] =
  (start: p.windowStart, count: min(p.windowSize, p.totalBlocks - p.windowStart))
```

i.e. it's always `windowSize` except for the last window which gets truncated to `totalBlocks - windowStart`. So I'd suggest naming this "windowLength" (or "windowSize" as you did above) across the board.

storage/blockexchange/engine/engine.nim
```diff
@@ -161,673 +159,959 @@ proc stop*(self: BlockExcEngine) {.async: (raises: []).} =
...
+        let (newStart, newCount) = ctx.advancePresenceWindow()
+
+        ctx.trimPresenceBeforeWatermark()
+
+        # Broadcast want-have for the new window to swarm peers only
+        var swarmPeers: seq[PeerContext] = @[]
+        for peerId in ctx.swarm.connectedPeers():
+          let peerCtx = self.peers.get(peerId)
+          if not peerCtx.isNil:
+            swarmPeers.add(peerCtx)
```
gmega, line 519: OK I see that you've scoped those broadcasts to be in-swarm - I think this is what makes the most sense.

storage/blockexchange/engine/engine.nim
```diff
@@ -161,673 +159,959 @@ proc stop*(self: BlockExcEngine) {.async: (raises: []).} =
...
+          cid = cid,
+          newWindowStart = newStart,
+          newWindowCount = newCount,
+          watermark = ctx.scheduler.completedWatermark(),
+          swarmPeers = swarmPeers.len
+
+        await self.broadcastWantHave(download, cid, newStart, newCount, swarmPeers)
+
+      # Broadcast availability to peers
+      if not download.fetchLocal and ctx.shouldBroadcastAvailability():
```
gmega, line 531: I suppose those are the push presence updates.

storage/blockexchange/engine/downloadcontext.nim
```diff
@@ -0,0 +1,318 @@
...
+    PresenceBroadcastIntervalMin.milliseconds.int ..
+      PresenceBroadcastIntervalMax.milliseconds.int
+  ).milliseconds
+
+proc shouldBroadcast(t: BroadcastAvailabilityTracker, watermark: uint64): bool =
+  case t.policy
+  of spRandomWindow:
+    t.pendingRanges.len > 0
+  of spSequential:
+    let newBlocks = watermark - t.lastBroadcastedWatermark
```
gmega: Why only watermark differences? You might have completed several batches but the watermark hasn't yet advanced.

storage/blockexchange/engine/downloadcontext.nim
```diff
@@ -0,0 +1,318 @@
...
+  )
+
+proc currentPresenceWindow*(ctx: DownloadContext): tuple[start: uint64, count: uint64] =
+  ctx.scheduler.currentPresenceWindow()
+
+proc needsNextPresenceWindow*(ctx: DownloadContext): bool =
+  ctx.scheduler.needsNextPresenceWindow()
+
+proc advancePresenceWindow*(ctx: DownloadContext): tuple[start: uint64, count: uint64] =
+  ctx.availabilityTracker.addPendingRange(ctx.scheduler.currentPresenceWindow())
```
gmega, line 168: Does this mean we ONLY broadcast the availability of the blocks for a window when we have all the blocks in that window?

storage/blockexchange/engine/engine.nim
```diff
@@ -161,673 +159,959 @@ proc stop*(self: BlockExcEngine) {.async: (raises: []).} =
...
+        var lastIdle = Moment.now()
+        let runtimeQuota = 100.milliseconds
+
+        for i in start ..< start + count:
+          if (Moment.now() - lastIdle) >= runtimeQuota:
+            await idleAsync()
+            lastIdle = Moment.now()
+
+          let address = download.makeBlockAddress(i)
+          if download.isBlockExhausted(address):
```
gmega, line 591: Why are we testing this here? Shouldn't we test retries when we're about to retry something? Why would this stop us from doing a local lookup for a block?

Plus, if retries are exhausted and we definitely cannot retrieve a block, the download has failed. We should probably not even continue doing any work?

Another aspect that bothers me a bit now that I think about it is that we retry batches, but our retry counters are per block. We might as well track retries per batch as once one block fails this batch is never going to succeed.

storage/blockexchange/engine/engine.nim
```diff
@@ -161,673 +159,959 @@ proc stop*(self: BlockExcEngine) {.async: (raises: []).} =
...
+            break localCheck
+          let exists =
+            try:
+              await address in self.localStore
+            except CatchableError:
```
gmega: This is too broad - we should probably use the proper error type here, as otherwise we'll interpret arbitrary errors as some block not existing.

storage/blockexchange/engine/engine.nim
```diff
@@ -161,673 +159,959 @@ proc stop*(self: BlockExcEngine) {.async: (raises: []).} =
...
+              false
+          if not exists:
+            if download.fetchLocal:
+              download.failLocalMissing(address)
+              break localCheck
+            break localCheck
+
+        for i in start ..< start + count:
```
gmega: We're doing two local block lookups per block, this can probably be condensed in one and merge the two loops?

storage/blockexchange/engine/engine.nim
```diff
@@ -161,673 +159,959 @@ proc stop*(self: BlockExcEngine) {.async: (raises: []).} =
...
-): Future[?!Block] {.async: (raw: true, raises: [CancelledError]).} =
-  self.requestBlock(BlockAddress.init(cid))
+          await self.broadcastWantHave(download, cid, start, count, connectedPeers)
+          # Give peers a short time to respond with presence
+          await sleepAsync(50.milliseconds)
+        else:
+          trace "No connected peers, searching for new peers"
+          self.searchForNewPeers(cid)
+          await self.handleBatchRetry(download, cid, start, count, retryInterval)
+          continue
```
gmega, line 647: This reads a bit like a Hail Mary. We already have push updates from peers to keep us informed, and we already send our want lists on window switches, does it make sense to to this again here? What's the assumption?

storage/blockexchange/engine/engine.nim
```diff
@@ -161,673 +159,959 @@ proc stop*(self: BlockExcEngine) {.async: (raises: []).} =
...
-proc completeBlock*(self: BlockExcEngine, address: BlockAddress, blk: Block) =
-  if address in self.pendingBlocks.blocks:
-    self.pendingBlocks.completeWantHandle(address, blk)
-  else:
-    warn "Attempted to complete non-pending block", address
+      if self.peers.len == 0:
+        trace "No connected peers available for batch, searching"
+        self.searchForNewPeers(cid)
+        await self.handleBatchRetry(download, cid, start, count, 100.milliseconds)
+        continue
```
gmega, line 651: You can probably move this before the the test for shouldBroadcast and then remove the duplication from lines 712-715.

storage/blockexchange/engine/swarm.nim
```diff
@@ -0,0 +1,340 @@
...
+  return false
+
+proc recordPeerTimeout*(swarm: Swarm, peerId: PeerId): bool =
+  ## return true if peer should be removed
+  swarm.peers.withValue(peerId, peer):
+    peer[].recordTimeout()
+    return peer[].timeoutCount >= swarm.config.maxPeerTimeouts
+  return false
+
+proc recordPeerSuccess*(swarm: Swarm, peerId: PeerId) =
```
gmega: This doesn't seem to be called from "live" code.

storage/blockexchange/engine/engine.nim
```diff
@@ -161,673 +159,959 @@ proc stop*(self: BlockExcEngine) {.async: (raises: []).} =
...
-proc blockPresenceHandler*(
-    self: BlockExcEngine, peer: PeerId, blocks: seq[BlockPresence]
-) {.async: (raises: []).} =
-  trace "Received block presence from peer", peer, len = blocks.len
-  let
-    peerCtx = self.peers.get(peer)
-    ourWantList = toHashSet(self.pendingBlocks.wantList.toSeq)
+      let staleUnknown = swarm.staleUnknownPeers()
+      if staleUnknown.len > 0:
+        let rangeAddress = download.makeBlockAddress(start)
```
gmega, line 655: Why are we singling these out for re-broadcast? The regular broadcasts we do already covers these peers, and if they fail to respond we're already timing them out. We should also be getting push updates from those, so technically there's no reason to re-query them. What's the assumption here? What case are you trying to cover?

storage/blockexchange/engine/engine.nim
```diff
@@ -161,673 +159,959 @@ proc stop*(self: BlockExcEngine) {.async: (raises: []).} =
...
+                downloadId = download.id,
+              )
+              .wait(DefaultWantHaveSendTimeout)
+          except AsyncTimeoutError:
+            trace "Re-query stale unknown peer send timed out", peer = peerId
+          except CatchableError as err:
+            trace "Failed to re-query stale unknown peer",
+              peer = peerId, error = err.msg
+
+        await sleepAsync(50.milliseconds)
```
gmega, line 684: And why sleep 50 milliseconds here?

storage/blockexchange/types.nim
```diff
@@ -0,0 +1,206 @@
...
+
+  case avail.kind
+  of bakUnknown:
+    false
+  of bakComplete:
+    true
+  of bakRanges:
+    let reqEnd = start + count
+    for (rangeStart, rangeCount) in avail.ranges:
+      if rangeCount > high(uint64) - rangeStart:
```
gmega, line 90: I suppose we shouldn't be allowing overflowing ranges to enter this list in the first place...

storage/blockexchange/engine/swarm.nim
```diff
@@ -0,0 +1,340 @@
...
+  else:
+    swarm.config.deltaTarget - active
+
+proc connectedPeers*(swarm: Swarm): seq[PeerId] =
+  for peerId in swarm.peers.keys:
+    result.add(peerId)
+
+proc peersWithRange*(swarm: Swarm, start: uint64, count: uint64): seq[PeerId] =
+  for peerId, peer in swarm.peers:
+    if not peer.isStale and peer.availability.hasRange(start, count):
```
gmega, line 197: Why test this here? This way we're embedding decisions on whether peers should be considered for selection or not inside of a function that supposedly checks if peers have a range of blocks. If we want to disregard stale peers, then we should probably do this in `selectPeerForBatch`, not here.

storage/blockexchange/engine/engine.nim
```diff
@@ -161,673 +159,914 @@ proc stop*(self: BlockExcEngine) {.async: (raises: []).} =
...
+    peer = peer.id
+
   let
-    peerHave = peer.peerHave
-    toAsk = toHashSet(self.pendingBlocks.wantList.toSeq.filterIt(it notin peerHave))
+    requestStartTime = Moment.now()
+    requestResult = await self.requestWantBlocks(
+      peer.id, BlockRange(treeCid: treeCid, ranges: ranges)
+    )
+    rttMicros = (Moment.now() - requestStartTime).microseconds.uint64
```
gmega, line 261: This is not really RTT, it's more like batch completion time.

storage/blockexchange/network/networkpeer.nim
```diff
@@ -47,24 +63,82 @@ proc readLoop*(self: NetworkPeer, conn: Connection) {.async: (raises: []).} =
...
+        let respResult = await readWantBlocksResponse(conn, dataLen)
+        if respResult.isErr:
+          warn "Failed to read WantBlocks response",
+            peer = self.id, error = respResult.error.msg
+          return
+
+        let response = respResult.get
+        self.pendingWantBlocksRequests.withValue(response.requestId, fut):
+          if not fut[].finished:
+            fut[].complete(WantBlocksResult[WantBlocksResponse].ok(response))
```
gmega, line 123: OK this is good - we complete the request as soon as we're done reading. No callbacks. This means I can actually trust the timing measurements we're doing for the batch requests, which in turn means I can probably trust our estimates for transfer rates.

storage/blockexchange/engine/swarm.nim
```diff
@@ -0,0 +1,335 @@
...
+        untriedPeers.add(peer)
+
+  if untriedPeers.len > 0:
+    var
+      bestPeer = untriedPeers[0]
+      bestLoad = tracker.count(bestPeer.id)
+    for i in 1 ..< untriedPeers.len:
+      let load = tracker.count(untriedPeers[i].id)
+      if load < bestLoad:
+        bestLoad = load
```
gmega, line 238: Hm... is this really the best proxy for load? Wouldn't load be pipeline utilization, not just size; i.e, optimalPipelineDepth - batchesInFlight / optimalPipelineDepth?

storage/blockexchange/engine/swarm.nim
```diff
@@ -0,0 +1,335 @@
...
+    var peersWithCapacity: seq[PeerContext]
+    for peer in peers:
+      let
+        pipelineDepth = peer.optimalPipelineDepth(batchBytes)
+        currentLoad = tracker.count(peer.id)
+      if currentLoad < pipelineDepth:
+        peersWithCapacity.add(peer)
+
+    if peersWithCapacity.len > 0:
+      let idx = rand(peersWithCapacity.len - 1)
```
gmega, line 253: Maybe exclude the best peers before doing exploration?

storage/blockexchange/peers/peercontext.nim
```diff
@@ -7,127 +7,87 @@
...
+proc evalBDPScore*(
+    self: PeerContext, batchBytes: uint64, currentLoad: int, penalty: float
+): float =
+  let
+    pipelineDepth = self.optimalPipelineDepth(batchBytes)
+    capacityScore =
+      if currentLoad >= pipelineDepth:
+        100.0
+      else:
+        (currentLoad.float / pipelineDepth.float) * 10.0
```
gmega: I guess this 100.0 works as a $+\infty$ of sorts?

storage/blockexchange/peers/peercontext.nim
```diff
@@ -7,127 +7,87 @@
...
+
+    throughputScore =
+      if self.stats.throughputBps().isSome:
+        let bps = self.stats.throughputBps().get().float
+        if bps > 0:
+          ThroughputScoreBaseline / bps
+        else:
+          50.0
+      else:
+        25.0 # normalization fallback
```
gmega: These look very much like magic numbers. Would be good to have some justification for them.

storage/blockexchange/peers/peercontext.nim
```diff
@@ -7,127 +7,87 @@
...
+      if currentLoad >= pipelineDepth:
+        100.0
+      else:
+        (currentLoad.float / pipelineDepth.float) * 10.0
+
+    throughputScore =
+      if self.stats.throughputBps().isSome:
+        let bps = self.stats.throughputBps().get().float
+        if bps > 0:
+          ThroughputScoreBaseline / bps
```
gmega: `ThroughputScoreBaseline` is such a large number that'll simply dominate over everything if `bps` is small. We should probably normalize those and reweigh them before summing.

storage/blockexchange/peers/peercontext.nim
```diff
@@ -7,127 +7,87 @@
...
+      else:
+        25.0 # normalization fallback
+
+    rttScore =
+      if self.stats.avgRttMicros().isSome:
+        self.stats.avgRttMicros().get().float / 10000.0
+      else:
+        5.0 # normalization fallback
+
+  return capacityScore + throughputScore + rttScore + penalty
```
gmega: This seems very much prone to dimension mismatches, maybe we should standardize the quantities (e.g. do z-scores) and then apply weights to them.

storage/blockexchange/engine/engine.nim
```diff
@@ -161,673 +173,905 @@ proc stop*(self: BlockExcEngine) {.async: (raises: []).} =
...
+      let
+        batchBytes = download.ctx.batchBytes
+        selection = swarm.selectPeerForBatch(
+          self.peers, start, count, batchBytes, self.downloadManager.peerTracker
+        )
 
-  let blocksDelivered = toHashSet(addrs)
-  var scheduledCancellations: Table[PeerId, HashSet[BlockAddress]]
+      if selection.kind == pskNoPeers:
+        trace "No peer with range, searching for new peers"
```
gmega, line 693: Techincally you're not searching for new peers here, it's your swarm maintenance snippet above that should be doing that.

storage/blockexchange/engine/engine.nim
```diff
@@ -161,673 +173,905 @@ proc stop*(self: BlockExcEngine) {.async: (raises: []).} =
...
+          waitTime = if hasActivePeers: retryInterval else: 100.milliseconds
+        await download.handleBatchRetry(start, count, waitTime)
+        continue
 
-  if self.peers.len == 0:
-    return
+      if selection.kind == pskAtCapacity:
+        download.requeueBatch(start, count, front = false)
+        await sleepAsync(10.milliseconds)
+        continue
```
gmega, line 703: Why do we have separate retry paths for empty swarms and peers at capacity? Why `pskAtCapacity` doesn't cause the block retry counts to be decreased?

gmega: Also, I'm assuming we don't do an explicit peer lookup here is because we're trusting that the swarm maintenance snippet above (line 644) will handle it.

storage/blockexchange/engine/engine.nim
```diff
@@ -161,673 +173,905 @@ proc stop*(self: BlockExcEngine) {.async: (raises: []).} =
...
+  if download.cancelled:
     return
 
-  # We send only blocks that the peer hasn't already told us that they already have.
+  let treeCid = download.treeCid
+
+  var ranges: seq[tuple[start: uint64, count: uint64]] = @[]
+  if missingIndices.len > 0:
+    var
+      rangeStart = missingIndices[0]
```
gmega, line 236: Looks like `missingIndices` needs to be sorted for this work, yet this doesn't seem to be documented or enforced anywhere.

storage/blockexchange/engine/engine.nim
```diff
@@ -161,673 +173,905 @@ proc stop*(self: BlockExcEngine) {.async: (raises: []).} =
...
+    requestResult = await self.requestWantBlocks(
+      peer.id, BlockRange(treeCid: treeCid, ranges: ranges)
+    )
+    rttMicros = (Moment.now() - requestStartTime).microseconds.uint64
 
-  if toAsk.len == 0:
-    if peer.lastSentWants.len > 0:
-      trace "Clearing want list tracking, peer has all blocks", peer = peer.id
-      peer.lastSentWants.clear()
+  if download.cancelled:
```
gmega, line 263: Why do you spot check this here, and then never again? At every call to `await` you could have the state of `download` to be cancelled.

storage/blockexchange/engine/engine.nim
```diff
@@ -161,673 +173,905 @@ proc stop*(self: BlockExcEngine) {.async: (raises: []).} =
...
+      if swarm.removePeer(peer.id).isNone:
+        trace "Peer was not in swarm", peer = peer.id
 
-      await self.sendBatchedWantList(peer, newWants.toSeq, full = false)
+      download.handlePeerFailure(peer.id)
 
-      if resetBackoff:
-        peer.wantsUpdated
+      if swarm.needsPeers():
+        self.searchForNewPeers(download.manifestCid)
```
gmega: This to me feels in the wrong place. We're putting logic that tests swarm invariants and triggers peer search within some snippet hidden in `sendWantBlocksRequest`. In fact, swarm maintenance logic seems to be a bit all over: part here, part in dropAndBanPeer, part in downloadWorker, and yet we don't trigger maintenance as part of `peerDepartHandler`.

I personally believe it'd be easier to keep this logic focused elsewhere (e.g. swarm.nim) and then feed network reconfiguration events into it (all of them) instead of having slightly different flavours of such logic in different places, and missing from some others.

storage/blockexchange/engine/engine.nim
```diff
@@ -161,673 +173,905 @@ proc stop*(self: BlockExcEngine) {.async: (raises: []).} =
...
     else:
-      trace "No changes in want list, skipping send", peer = peer.id
-      peer.lastSentWants = toAsk
-  else:
-    trace "Sending full want list", peer = peer.id, length = toAsk.len
+      # we can requeue immediately (cancels timeout), no benefit waiting for timeout.
+      download.requeueBatch(start, count, front = true)
```
gmega, line 287: `handlePeerFailure` seems to be already requeueing the batches, sounds like we're doing it twice.

storage/blockexchange/engine/engine.nim
```diff
@@ -161,673 +173,905 @@ proc stop*(self: BlockExcEngine) {.async: (raises: []).} =
...
+    totalBytes: uint64 = 0
+    validCount: int = 0
+    receivedIndices: HashSet[uint64]
+
+  for view in allBlockViews:
+    if not bexutils.isIndexInRanges(
+      view.address.index.uint64, ranges, sortedRanges = true
+    ):
+      warn "Received unrequested block", index = view.address.index, ranges = ranges.len
+      continue
```
gmega, line 314: Why are we so nice to peers that send unrequested blocks, but immediately ban peers that send inconsistent data? Btw these blocks might well be corrupt/bogus data, but we don't know cause in this case we're not checking them.

storage/blockexchange/engine/engine.nim
```diff
@@ -161,673 +173,905 @@ proc stop*(self: BlockExcEngine) {.async: (raises: []).} =
...
+        if peerAvail.isSome and peerAvail.get().availability.hasRange(idx, 1):
+          penaltyAddresses.add(download.makeBlockAddress(idx))
+
+      let exhausted = download.decrementBlockRetries(penaltyAddresses)
+      if exhausted.len > 0:
+        warn "Blocks exhausted retries after partial delivery",
+          treeCid = treeCid, exhaustedCount = exhausted.len
+        download.failExhaustedBlocks(exhausted)
+        let exhaustedIndices = exhausted.mapIt(it.index.uint64).toHashSet
+        stillMissing = stillMissing.filterIt(it notin exhaustedIndices)
```
gmega, line 379: I'm not sure I understand the logic in this: we only decrement the retry count if we didn't get the block back BUT the peer says it had it? Otherwise we don't decrement at all? Why?

gmega (approved):
I won't be able to fully review this. I did review some of the key parts, and I'm hoping we will refine it as we go. Since the testing seems to show this works and is indeed much faster, I'd merge it anyway.

gmega:
* [Document which attempts to formalize the bandwidth-delay product calculations](https://hackmd.io/2brFqqXzS9GQ4Ei5xe0gLw#) we do;
* [Raw notes in which I try to reason through how the code works](https://hackmd.io/YKEKMFGgSAaHjVxKcz8n2w).

storage/blockexchange/engine/downloadmanager.nim
cnanakos: Yes it's cleaner, made the change.

storage/stores/networkstore.nim
cnanakos: The pattern defends against a race between the local check and the handle registration. Registering the handle before the local lookup ensures we catch any arrival during the yield. The `completeWantHandle` on the local-hit path resolves the handle for any other waiters.

storage/blockexchange/engine/engine.nim
cnanakos: Would prefer keeping it uniform rather than splitting. The cost is small and the `iter` al already wired up if we ever want per-block progress events on the opaque path.

storage/blockexchange/engine/engine.nim
cnanakos: Removed `next/nextBlock/finished` - no production callers. Kept `iter` for now since its `genNext` still wires up per-block handles used internally.

storage/blockexchange/engine/engine.nim
cnanakos: Outdated, we broadcast to swarm only.

storage/blockexchange/engine/downloadcontext.nim
cnanakos: Good catch! Let's add the OOO batches.

storage/blockexchange/engine/downloadcontext.nim
cnanakos: Yes, once the window is done (for `spRandomWindow` only).

storage/blockexchange/engine/swarm.nim
cnanakos: This is just a filter that restricts peers to loa [0,1], so utilization and count produce the same ordering. OTOH `evalBDPScore` already uses the ration.

storage/blockexchange/engine/swarm.nim
cnanakos: Nice catch, removed.

storage/blockexchange/engine/engine.nim
cnanakos: Removed redundant discovery triggers.

storage/blockexchange/engine/engine.nim
cnanakos: By design. If the peer sent an extra block that's legal for this dataset but not in our current request, we skip it without cost — just noise. If the peer is sending garbage outside our dataset or an invalid proof, we ban.

storage/blockexchange/engine/engine.nim
cnanakos: We are decrementing only for the subset the peer claimed to have but didn't deliver. Without this filter, asking a peer with partial range coverage would burn retry budget for blocks they never said they had.

storage/blockexchange/engine/engine.nim
cnanakos: Great catch!

storage/blockexchange/engine/engine.nim
cnanakos: It's just a short settling window for presence responses to arrive before the peer selection. Trying to factor the updated availability into this iteration rather than the next.

storage/blockexchange/engine/swarm.nim
cnanakos: Staleness is a reachability signal which means the peer hasn't been heard from in a while so they probably won't answer a request. So based on this both callers need reachable peers: the broadcast decision only cares if a reachable peer claims the range and selection can't pick a possible unreachable peer.

storage/blockexchange/engine/engine.nim
cnanakos: This worths a follow-up. Right now the checks sit at high-value exit points. Before and right after the big network wait where exiting will save a lot of work. The awaits after are short local store I/O so the cost of continuing is small.

storage/blockexchange/engine/downloadmanager.nim
cnanakos: They have different semantics. Multiple downloads can share a treeCid. `releaseDownload(downloadId, treeCid)` releases one specific handle - other downloads for the same CID keep running. When the last holder releases, the CID's download group is cleaned up. `cancelDownload(treeCid)` kills everything for that CID regardless of holders.

storage/blockexchange/engine/engine.nim
cnanakos: Added a comment.

storage/blockexchange/peers/peercontext.nim
cnanakos: Replaced the weighted unbounded sum with normalized components and explicit weights. Removes the 100.0 sentinel mentioned above, replaces magic fallbacks with named middle-of-range constants, and eliminates the dimension mismatch.

PR #1412, opened 2026-02-25
fix(repostore): prevent underflow when deleting datasets

emizzle:
Storing a manifest does not increase the refCount. So, when deleting a manifest, the refCount starts at zero, and was always being decremented, for both manifests and blocks. This PR adds a conditional check to only decrement the refCount when it's a block.

It also adds `max(0, ...)` calculations to `Natural` assignments to ensure no underflows happen.

### DIFF
--- a/storage/conf.nim
+++ b/storage/conf.nim
@@ -136,8 +136,7 @@ type
     .}: OutDir
 
     listenIp* {.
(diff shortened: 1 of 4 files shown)


cnanakos (approved):
LGTM

PR #1413, opened 2026-03-30
fix: cleanup chronicles writer in close function

2-towns:
This PR exposes the `logFile` to the `StorageServer` in order to close it during the cleanup phase.

### DIFF
--- a/storage/conf.nim
+++ b/storage/conf.nim
@@ -54,10 +54,6 @@ export
 
 type ThreadCount* = distinct Natural
 
-# Declare the log file globally in order to be collected properly
-# by the GC.
-var logFile: ?IoHandle
-
 proc `==`*(a, b: ThreadCount): bool {.borrow.}
 
 proc defaultDataDir*(): string =
@@ -289,6 +285,10 @@ type
     logFile* {.
       desc: "Logs to file", defaultValue: string.none, name: "log-file", hidden
     .}: Option[string]
+
+    # Declare the log file in StorageConf in order to be collected properly
+    # by the GC.
+    logFileIo: ?IoHandle
 
 func defaultAddress*(conf: StorageConf): IpAddress =
   result = static parseIpAddress("127.0.0.1")
@@ -567,7 +567,7 @@ proc setupLogging*(conf: StorageConf) =
       writeAndFlush(stdout, stripAnsi(msg))
 
     proc fileFlush(logLevel: LogLevel, msg: LogOutputStr) =
-      if file =? logFile:
+      if file =? conf.logFileIo:
         if error =? file.writeFile(stripAnsi(msg).toBytes).errorOption:
           error "failed to write to log file", errorCode = $error
 
@@ -579,7 +579,7 @@ proc setupLogging*(conf: StorageConf) =
         error "failed to open log file",
           path = logFilePath, errorCode = $logFileHandle.error
       else:
-        logFile = logFileHandle.option
+        conf.logFileIo = logFileHandle.option
         defaultChroniclesStream.outputs[2].writer = fileFlush
 
     defaultChroniclesStream.outputs[1].writer = noOutput
--- a/storage/conf.nim
+++ b/storage/conf.nim
@@ -90,13 +90,6 @@ type
     repoSQLite = "sqlite"
     repoLevelDb = "leveldb"
 
-  # LogFile is a ref object so that the `fileFlush` closure captures a reference
-  # to it rather than a copy. This ensures the Nim GC keeps the closure environment
-  # alive as long as the owning StorageServer holds a reference to it.
-  LogFile* = ref object
-    handle*: Option[IoHandle]
-    writer*: proc(logLevel: LogLevel, msg: LogOutputStr) {.gcsafe, raises: [].}
-
   StorageConf* = object
     configFile* {.
       desc: "Loads the configuration from a TOML file",
@@ -560,13 +553,12 @@ proc openLogFile(conf: StorageConf): Option[IoHandle] =
       return logFileHandle.option
   return IoHandle.none
 
-proc setupLogging*(conf: StorageConf): LogFile =
+proc setupLogging*(conf: StorageConf): Option[IoHandle] =
   let ioHandle =
     if conf.logFile.isSome:
       conf.openLogFile()
     else:
       IoHandle.none
-  let logFile = LogFile(handle: ioHandle)
 
   when defaultChroniclesStream.outputs.type.arity != 3:
     warn "Logging configuration options not enabled in the current build"
@@ -595,7 +587,6 @@ proc setupLogging*(conf: StorageConf): LogFile =
     defaultChroniclesStream.outputs[2].writer = noOutput
     if ioHandle.isSome:
       defaultChroniclesStream.outputs[2].writer = fileFlush
-      logFile.writer = fileFlush
 
     defaultChroniclesStream.outputs[1].writer = noOutput
 
@@ -624,7 +615,7 @@ proc setupLogging*(conf: StorageConf): LogFile =
     else:
       defaultChroniclesStream.outputs[0].writer = writer
 
-    return logFile
+    return ioHandle
 
 proc setupMetrics*(config: StorageConf) =
   if config.metricsEnabled:
--- a/storage/conf.nim
+++ b/storage/conf.nim
@@ -54,6 +54,10 @@ export
 
 type ThreadCount* = distinct Natural
 
+# Declare the log file globally in order to be collected properly
+# by the GC.
+var logFile : ?IoHandle
+
 proc `==`*(a, b: ThreadCount): bool {.borrow.}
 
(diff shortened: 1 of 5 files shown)


storage/conf.nim
```diff
@@ -54,6 +54,10 @@ export
 
 type ThreadCount* = distinct Natural
 
+# Declare the log file globally in order to be collected properly
+# by the GC.
+var logFile: ?IoHandle
```
emizzle: I think it would be better to track this against the `StorageConf` object or the `StorageServer` object instead of keeping in the global space.

2-towns: > I think it would be better to track this against the `StorageConf` object or the `StorageServer` object instead of keeping in the global space.

Okay it cannot be done in the `StorageConf` because conf-util expects serialisabled fields. So I added it in the `StorageServer`. I had to store the `fileFlush` ref as well.

2-towns:
Elaborating more on the PR: 

In our last engineering, we said that the function / variable **should** not be GC. 

Reminder: the problem appears when running multiple tests with multiple start / stop.

So I added some logs in the `fileFlush` function to check what's going on : 

```nim
proc fileFlush(logLevel: LogLevel, msg: LogOutputStr) =
      echo "ARNAUD: fileFlush called, hasFile=", logFile.isSome        
```

Surprisingly, the `logFile` become None during the start process: 

```
1: QDEBUG : TestStorageModule::test_updateLogLevel() StorageModulePlugin::init called with cfg: "{\"data-dir\": \"/build/logos-storage-test.bUddHm\", \"log-level\": \"DEBUG\", \"log-file\":>
1: QDEBUG : TestStorageModule::test_updateLogLevel() StorageModulePlugin::waitForSignal: Waiting for signal with timeout 1000 ms
1: ARNAUD: setupLogging
1: INF 2026-04-02 07:35:16.244+00:00 Creating a private key and saving it       tid=193
1: ARNAUD: fileFlush called, hasFile=true
1: DBG 2026-04-02 07:35:16.244+00:00 Rest API is disabled!                      topics="libstorage libstoragelifecycle" tid=193
1: ARNAUD: fileFlush called, hasFile=true
1: INF 2026-04-02 07:35:16.245+00:00 Threadpool started                         topics="libstorage libstoragelifecycle" tid=193 numThreads=12
1: ARNAUD: fileFlush called, hasFile=true
1: INF 2026-04-02 07:35:16.294+00:00 Updating announce record                   topics="storage discovery" tid=193 addrs=@[/ip4/0.0.0.0/tcp/0]
1: ARNAUD: fileFlush called, hasFile=false
```

I tried to add a finaliser in order to detect GC action: 

```
  logFile = logFileHandle.option
  var logFileRef: ref IoHandle
  new(logFileRef, proc(h: ref IoHandle) =
    echo "ARNAUD: IoHandle ref GC collected"
  )
```

I can see the finaliser is called: 

```
QDEBUG : TestStorageModule::test_updateLogLevel() StorageModulePlugin::waitForSignal: Waiting for signal with timeout 1000 ms
ARNAUD: setupLogging
INF 2026-04-02 12:05:53.433+04:00 Creating a private key and saving it       tid=627455
ARNAUD: fileFlush called, hasFile=true
DBG 2026-04-02 12:05:53.433+04:00 Rest API is disabled!                      topics="libstorage libstoragelifecycle" tid=627455
ARNAUD: fileFlush called, hasFile=true
ARNAUD: IoHandle ref GC collected
```

But another surprising thing: adding this called keep alive the `logFile` and `hasFile` is true until the end of the test, which makes the test passes. 

Another general note is, if I juste move the `logFile` on the top of the `conf.nim` file, the problem is fixed as well.

emizzle:
1. is `logFile` in config empty here?
2. is qtest parallelising the tests at all?
3. Can you echo `writeStackTrace` inside of the finaliser? 
4. I wonder if this has something to do with calling `initializeLibrary` on each call to `libstorage`, because this function sets up foreign thread gc. The doc on `setupForeignThreadGc` says it has a thread local guard, but it could be a bug of some sort... 
5. The above logs happen even with `logFile` as a property of `StorageServer`? Or is only when `logFile` is captured in the `setupLogging` closure?

2-towns:
> 1. is `logFile` in config empty here?
> 
>     2. is qtest parallelising the tests at all?
> 
>     3. Can you echo `writeStackTrace` inside of the finaliser?
> 
>     4. I wonder if this has something to do with calling `initializeLibrary` on each call to `libstorage`, because this function sets up foreign thread gc. The doc on `setupForeignThreadGc` says it has a thread local guard, but it could be a bug of some sort...
> 
>     5. The above logs happen even with `logFile` as a property of `StorageServer`? Or is only when `logFile` is captured in the `setupLogging` closure?

1. No it is not empty. `hasFile` is true in the logs, so the `logFile` is not `None`. Here is the config I pass to libstorage: 

```json
{
"data-dir": "/tmp/logos-storage-test.ZHdWGT", "log-level": "DEBUG", "log-file": "/tmp/logos-storage-test.ZHdWGT/storage.log"}
```

2. No 
3. It seems to not give much info: 

```
ARNAUD: IoHandle ref GC collected
Traceback (most recent call last, using override)
```

And the logs continue after that: 

```
INF 2026-04-02 13:46:19.711+04:00 Threadpool started                         topics="libstorage libstoragelifecycle" tid=648805 numThreads=12
```

4. I have no idea, will try to explore.
5. When `logFile` is captured in the setupLogging closure. I reverted back the changes to `StorageServer` locally because the test was passing.

emizzle:
4. I think you could maybe spend a bit of time trying to isolate the issue by reducing down the problem to a simple library implementation and see if you can figure out the issue from there. It's also good to know that there's a workaround available in 5 (`StorageServer.logFile`), so if we can't figure out the issue, it's not the end of the world.

2-towns:
> 4. I think you could maybe spend a bit of time trying to isolate the issue by reducing down the problem to a simple library implementation and see if you can figure out the issue from there. It's also good to know that there's a workaround available in 5 (`StorageServer.logFile`), so if we can't figure out the issue, it's not the end of the world.

WOW I think I found it ! 

I followed your advice and tried to isolate the issue by writing Nim test and C test: nothing. 

Then after hours of debugging I found that were [set chronicles ouput writer](https://github.com/logos-storage/logos-storage-nim/blob/4ea8350612b516d54aad11534f0c75429a6084ec/storage/conf.nim#L580) in `setupLogging`, but we never clean that !  

So what I think is happening is the following: 

1. The first test creates a storage context. It starts the node and runs the test successfully.
2. The node is stopped and the context is destroyed. However, the Chronicles output writer is not cleaned up.
3. The second test creates a new context. A new Chronicles output writer is created and assigned to the global array, replacing the old one.
4. Because the old writer is finally replaced in step 3, the GC cleans it up. Unfortunately, this GC cleanup process happens after the new writer is assigned. It ends up freeing the newly assigned writer. As a result, the new writer silently captures a `None` value, meaning no logs will be written for the second test.

I tried to clean up the Chronicles output writer and it does work ! 

So I am gonna to create another PR and put the reference here.

#1414

emizzle:
That's awesome that you found a way to get it to work! 

I still don't think this explains it. The first node's closure environment is completely separate from the second node's closure environment. Once node 2's `logFile` replaces node 1's, node 1's `logFile` can be GC'd. This does not mean that node 2's `logFile` would be `none`; it should not be impacted at all.

Regardless of GC, re-assigning the writer to a non-file output likely does reset some condition and allow it to work properly, whether that condition is in logos-storage-nim or chronicles. This makes me think that instead of assigning the writer to `noOutput`, could you try closing the file in `Storage.close` instead? The file is [opened in `setupLogging`](https://github.com/logos-storage/logos-storage-nim/blob/03ba5a38869c41b728f4ea105f6810212584bed8/storage/conf.nim#L575-L576), but it is never closed (as noted in [the original PR](https://github.com/logos-storage/logos-storage-nim/pull/558)).

storage/storage.nim
```diff
@@ -127,6 +128,16 @@ proc close*(s: StorageServer) {.async.} =
       error "Failed to stop the taskpool", failures = res.failure.len
       raiseAssert("Failure in taskpool shutdown:" & exc.msg)
 
+  when defaultChroniclesStream.outputs.type.arity >= 3:
+    proc noOutput(logLevel: LogLevel, msg: LogOutputStr) =
+      discard
+
+    defaultChroniclesStream.outputs[2].writer = noOutput
```
emizzle, line 132: Does it work with this removed? I thought this was only needed if we didn't hang `logFile` off of `StorageServer` to prevent it from being collected.

2-towns: No it does not work, I have this error : 

`logos-storage_module-tests> ERR 2026-04-07 13:02:16.822+00:00 failed to write to log file                tid=311 errorCode=9`

emizzle (approved):
I still don't think we got to the bottom of the issue, but we have spent quite a lot of time debugging it, and the fix/workaround implemented seems to work quite well, so will approve 👍

Issue #782, opened 2024-04-19
NAT / port forwarding / holepunching fixes


emizzle:

veaceslavdoina:
Is this one is related #763?

emizzle:
> Is this one is related #763?

Yes, it should be 👍

emizzle:
@gmega to estimate difficulty/duration.

PR #1417, opened 2026-04-10
feat: nat traversal relay

2-towns:
# NAT traversal

This PR introduces NAT traversal, following two design documents: [NAT Strategy implementation](https://hackmd.io/3DR2Of1qS9iwXl_qdGdkKg) and [Notes on NAT Traversal](https://hackmd.io/5C0PgLS3TAqmbk4jUXXTKw).

`Autonat` here refers to `AutonatV2` in libp2p.

## Breaking change

The SPR now exposes `localRecord`, which carries both TCP and UDP addresses, instead of the DHT record. Bootstrap nodes deployed before this PR are therefore incompatible and must be redeployed.

## AutoNAT integration

Autonat starts for nodes without `--nat=extip:<IP>` (extip nodes are assumed `Reachable`) and for `--autonat-server` nodes.
The Autonat workflow is the following:

1. Starts in client mode
2. Connects to Autonat servers (bootstrap nodes)
3. Autonat servers dial back the candidate addresses (derived via `guessDialableAddr`) to test reachability
4. `handleNatStatus` receives the Autonat response
   - Reachable => `clientMode` becomes false + DHT announcements
   - NotReachable => `clientMode` becomes true + try port mapping, fallback to Relay

After a port mapping is created, we need to wait for the next Autonat iteration to trigger reachability.

`AutonatV2Service` behaviour is tuned through four CLI flags:

1. `--nat-schedule-interval`
2. `--nat-num-peers-to-ask`
3. `--nat-max-queue-size`
4. `--nat-min-confidence`

Two more flags were added:

1. `--relay-server`
2. `--autonat-server`

They both require `--nat=extip:<IP>` because Relay and Autonat servers have to be `Reachable`.
For now they are meant for bootstrap nodes, but later we can expose them to any reachable node so it can contribute to the network.

## Hole punching

Because we want to control AutoNAT's result and try a port mapping before starting the relay, we cannot use libp2p's `HPService` (it depends on AutoNAT v1 and starts the relay immediately on `NotReachable`). The hole-punching logic is therefore adapted from libp2p.

## Address announcement

The addresses a node announces to the DHT depend on how it is reachable, and they are updated from three places:

1. A node started with `--nat=extip` is assumed reachable, so it announces its public address directly at startup, without waiting for AutoNAT.
2. `handleNatStatus`, on every reachability change.
   On `Reachable`, the node announces the dial-back address.
   On `NotReachable`, if the relay is not yet running, the previously announced direct addresses are removed.
3. `onReservation` (Relay's callback): the public circuit addresses are announced; no UDP address is derived from them

## SPR fix

The behaviour on master contains a bug: `updateAnnounceRecord` and `updateDhtRecord` in `discovery.nim` both call `d.protocol.updateRecord()`, which overwrites `localNode.record` with the new SPR, so the last call wins.

- [`updateAnnounceRecord`: discovery.nim:178](https://github.com/logos-storage/logos-storage-nim/blob/10408a72de8aa90c53db32df0401cc174ef8ecdd/storage/discovery.nim#L178)
- [`updateDhtRecord`: discovery.nim:192](https://github.com/logos-storage/logos-storage-nim/blob/10408a72de8aa90c53db32df0401cc174ef8ecdd/storage/discovery.nim#L192)
- [`localNode.record`: protocol.nim:287](https://github.com/logos-storage/logos-storage-nim-dht/blob/99884b5971759a0da437db3d2e834b92a058527d/codexdht/private/eth/p2p/discoveryv5/protocol.nim#L287)

On top of that, the SPR exposed by the REST API is the [dhtRecord](https://github.com/logos-storage/logos-storage-nim/blob/ebb1ae2599c0a9273018229ac968477c1b37def3/storage/rest/api.nim#L487-L490). This is confusing and breaks AutoNAT: evaluating reachability requires prior TCP connections to bootstrap nodes, which is impossible without TCP addresses.

This PR resolves by adding a method `getSpr` that returns an SPR containing TCP and UDP addresses. 
Moreover it ensures that `providerRecord` contains only TCP addresses and `localRecord` contains only UDP addresses.   

## Port mapping

The UPnP code was rewritten to use [nim-libplum](https://github.com/logos-storage/nim-libplum) instead of nim-nat-traversal, for three reasons: it adds PCP support, it has a simpler API, and it manages its threads internally.

## REST API and libstorage

NAT fields were added to debug (REST API and libstorage):

1. `reachability`: `NotReachable`, `Reachable`, `Unknown`
2. `clientMode`: true if the node runs in client mode
3. `relayRunning`: true if the node started a relay
4. `portMapping`: `none`, `upnp`, `pcp`, `pmp`, `direct`
5. `announceAddresses`: announced TCP/circuit addresses
6. `dhtAddresses`: announced UDP addresses for DHT routing

### Unit tests

- `testnatdetection`: reachability detection over a simulated NAT
- `testnatreaction`: the `handleNatStatus` state machine
- `testnatsimulation`: the NAT filtering modes

The simulated NAT is a test-only `NatTransport` that wraps `TcpTransport` and implements the filtering modes: `EndpointIndependent`, `AddressDependent`, `AddressAndPortDependent` and `DoubleNat`.

### Integration tests

This [hackmd](https://hackmd.io/Nn3fav3cTNCvJc5anoIdTA) was used as base for the tests.

Several NAT scenarios were created using containers. `podman` is preferred because it is rootless but `docker` can be used as well:

- `reachable`: node is `Reachable` using a manual port-forward
- `upnp` / `pcp`: node is `Reachable` using a real miniupnpd mapping
- `not-reachable`: node is `NotReachable` when there is no mapping and falls back to the relay
- `not-downloadable`: a NATed node with no relay is `NotReachable`, and a reachable peer cannot download from it (useful to validate the opposite of `relay-download`)
- `relay-download`: a reachable node downloads from a NATed node through the relay
- `hole-punching`: a relayed connection is upgraded to a direct one. 

## Dependencies

This PR adds `nim-libplum`.
This PR needs libp2p update to 2.1.0: https://github.com/logos-storage/logos-storage-nim/pull/1449

## Limitation

If PCP create a port mapping on a port which is the the listen port, it is not supported. The node will fallback into relay.


storage/nat.nim
```diff
@@ -42,14 +45,24 @@ type NatConfig* = object
...
+    natThreads: seq[Thread[PortMappingArgs]] = @[]
+else:
+  var
+    upnp {.threadvar.}: Miniupnp
+    npmp {.threadvar.}: NatPmp
+    strategy = NatStrategy.NatNone
+    natClosed: Atomic[bool]
+    extIp: Option[IpAddress]
+    activeMappings: seq[PortMappings]
+    natThreads: seq[Thread[PortMappingArgs]] = @[]
```
emizzle: The fact that these are globals is already a big no no. We saw the issues it creates in chronicles and multiple instances of libstorage. I would move these into a type, eg `StorageNat`, and then in the tests you can use `privateAccess(StorageNat)` to access the private fields.

storage/nat.nim
```diff
@@ -9,10 +9,13 @@
...
-  nat_traversal/[miniupnpc, natpmp],
-  json_serialization/std/net,
-  results
+  std/[options, os, strutils, times, net, atomics], json_serialization/std/net, results
+
+when defined(testing):
+  import ../tests/storage/helpers/mockminiupnpc
+  import ../tests/storage/helpers/mocknatpmp
+else:
+  import nat_traversal/[miniupnpc, natpmp]
```
emizzle: Generally speaking, I'd avoid this if possible. I think the best way to approach this is to inject Upnp and Nat pmp as type fields. That way you can inject a mock for testing...

```nim
# nat.nim

type StorageNat
  upnp: Miniupnp
  npmp: NatPmp
  strategy: NatStrategy
  natClosed: Atomic[bool]
  extIp: Option[IpAddress]
  activeMappings: seq[PortMappings]
  natThreads: seq[Thread[PortMappingArgs]]

func new*(self: StorageNat, strategy: NatStrategy): StorageNat =
  StorageNat(strategy: strategy) # don't need to instantiate upnp/npmp yet, will do in the thread
```
The big question is about how to handle the threadvar. So I think we could try making the entire StorageNat object a threadvar, which means it's available, but not instantiated, in each thread. So when launching a new thread, it needs to be instantiated again...

```nim
# storage/storage.nim

var storageNat {.threadvar.}: StorageNat
proc start*(s: StorageServer) {.async.} =
  # ...
  storageNat = StorageNat.new(strategy)
```

Then back in `nat.nim`:

```nim
proc getExternalIP*(self: StorageNat, natStrategy: NatStrategy, quiet = false): Option[IpAddress] =
  if self.upnp == nil:
      self.upnp = newMiniupnp()
  # ...
  if self.npmp == nil:
      self.npmp = newNatPmp()

proc repeatPortMapping(self: StorageNat, args: PortMappingArgs) {.thread, raises: [ValueError].} =
  # self is a copy with upnp/npmp non-initialised
  let ipres = getExternalIP(self, strategy, quiet = true)

proc redirectPorts*(
    self: StorageNat, strategy: NatStrategy, tcpPort, udpPort: Port, description: string
): Option[(Port, Port)] =
  # ...
  natThreads.add(Thread[PortMappingArgs]())
      natThreads[^1].createThread(
        repeatPortMapping, self, (strategy, externalTcpPort, externalUdpPort, description)
      )
```
You'll need to pass the `StorageNat` object into most of the functions in `nat.nim` and set all the fields.
(Side note: I still think this entire module is really poorly written, and should be rewritten...)

So then in the tests:
```nim
suite "getExternalIP":
  var nat: StorageNat
  setup:
    nat = StorageNat.new(NatStrategy.None)

  test "getExternalIp returns None when UPnP fails to discover devices":
    nat.upnp = mockupnp.newMiniupnp()
    nat.upnp.discoverResult = Result[int, cstring].err(cstring("no devices found"))
    #....
    check nat.getExternalIP(NatStrategy.NatUpnp) == none(IpAddress)
```

2-towns: That 's nice approach. 

But I cannot assign the upnp using my mock there are 2 differents types + the UPnP module defines proc and not func. 

I think I could achieve some kind of what you described by using a wrapper. StorageNat
 would be a wrapper that encapsulates upnp et then you can define functions that uses the upnp module that you can overwrite in the tests.

2-towns: But now I am going against my initial thought : don't touch it, because it works now as it is...

emizzle: > But now I am going against my initial thought : don't touch it, because it works now as it is...

Yeah it works, but it's very fragile and hard to read/maintain, so it's very much not ideal. Instead of making it more complicated and more fragile, better to rewrite it IMO.

2-towns: I agree. The question is: should we make it part of this PR ?

2-towns: Maybe this is something we can discuss tomorrow btw

storage/nat.nim
```diff
@@ -8,421 +8,361 @@
...
+    m: NatPortMapper
+): Future[Option[(Port, Port, MappingProtocol)]] {.
+    async: (raises: [CancelledError]), base, gcsafe
+.} =
+  if m.closed or m.natConfig.hasExtIp:
+    return none((Port, Port, MappingProtocol))
+
+  # If both mappings are still active, return the stored ports without recreating.
+  if m.activeTcpPort.isSome and m.activeUdpPort.isSome and m.activeMappingProtocol.isSome and
+      m.tcpMappingId.isSome and m.hasLiveMapping(m.tcpMappingId.get) and
```
emizzle: Imo, all of these should be fields of an object, eg PortMapping:
```nim
type PortMapping = object
  tcpMappingId: cint
  udpMappingId: cint
  activeMappingProtocol*: MappingProtocol
  activeTcpPort*: Port
  activeUdpPort*: Port

type NatPortMapper* = ref object of RootObj
  natConfig*: NatConfig
  tcpPort*: Port
  discoveryPort*: Port
  discoverTimeout*: int
  mappingTimeout*: int
  recheckPeriod*: int
  portMapping*: Option[PortMapping]
  plumInitialized: bool
  closed: bool
```

There's a lot of noise checking `isSome` and `isNone` on individual properties, and there's no branch where some of these values are set but others are not (that I have seen).

The idea could be to use the underlying libplum to get a port mapping and assign to `NatPortMapper.portMapping`, which can be used throughout this module. Then use libplum to check for liveness of that mapping if needed.

This allows the API to become a bit more readable and consistent:

```nim

# change from `hasMappingIds*(m: NatPortMapper)`
method hasPortMapping*(m: NatPortMapper): bool {.base, gcsafe.} =
  # Only checks that mappings were created, not that they are still live
  # (use hasReachablePortMapping() for liveness check).
  m.portMapping.isSome

# change from `isPortMapped*(m: NatPortMapper, port: Port)`
proc hasPortMapping*(m: NatPortMapper, port: Port): bool =
  if m.portMapping.isNone:
    return false
  
  return m.portMapping.get.activeTcpPort == port

# change from `hasLiveMapping*(m: NatPortMapper, id: cint)`
method hasLivePortMapping*(m: NatPortMapper): bool {.base, gcsafe.} =
  # Checks that mappings were created and uses libplum.hasMapping to confirm liveness of the port mapping
  if m.portMapping.isNone:
    return false
  
  let mappingId = m.portMapping.get.tcpMappingId
  return mappingId.hasMapping()
```

storage/nat.nim
```diff
@@ -8,421 +8,361 @@
...
-      return (none(IpAddress), NoRoutingInfo)
-    elif ip.get().isGlobalUnicast():
-      return (some(ip.get()), PrefSrcIsPublic)
-    else:
-      return (none(IpAddress), PrefSrcIsPrivate)
-  elif bindAddress.isGlobalUnicast():
-    return (some(bindIp), BindAddressIsPublic)
+        # In case of failure, close the port mapping in order to rerun discover
+        # on the next iteration
+        m.close()
```
emizzle, line 227: I'm confused as to what the purpose of closing the `NatPortMapper` is for. In the next AutoNAT check, `handleNatStatus` will be called and it will early return because `NatPortMapper` is closed. 

Even without the early return, `mapNatPorts` would always return None due to the close check at the start of the function.

storage/nat.nim
```diff
@@ -8,421 +8,361 @@
...
+
+      if maybePorts.isSome:
+        let (tcpPort, udpPort, protocol) = maybePorts.get()
+
+        info "Port mapping created successfully", tcpPort, udpPort, protocol
+
+        # The address mapper uses the mapped port to build the candidate address
+        # for AutoNAT; the announce happens once AutoNAT confirms Reachable.
+
+        hasPortMapping = true
```
emizzle: This whole block is a bit hard to follow. Possibly might be more readable with the api updates i suggested, but i think the multi-branching is really what could be finessed if possible.

On this line, I think an early return would be better than a single-use flag (`hasPortMapping`) here.

storage/storage.nim
```diff
@@ -76,24 +88,68 @@ proc start*(s: StorageServer) {.async.} =
...
+      let addrs = spr.data.addresses.mapIt(it.address)
+      await s.storageNode.switch.connect(spr.data.peerId, addrs)
+    except CancelledError as exc:
+      raise exc
+    except CatchableError as e:
+      warn "Cannot connect to bootstrap node", error = e.msg
 
-  s.storageNode.discovery.updateAnnounceRecord(announceAddrs)
-  s.storageNode.discovery.updateDhtRecord(discoveryAddrs)
+  await allFutures(findReachableNodes(s.bootstrapNodes).mapIt(connectBootstrapNode(it)))
```
emizzle: I'm not really sure if we need it, but using `noCancel` here prevents issues where `start` is cancelled but the `connectBootstrapNode` futures are not (because cancelling an `allFutures` future doesn't cancel the passed in futures).

emizzle (approved):
I didn't have nearly enough time to thoroughly review this PR, but from what i've seen it looks good. There are some suggestions on api design, and a potential issue with closing the nat mapper early that I flagged.

I also ran this through Claude who gave the following suggestions, but I didn't have time to run through and verify them:

## Important (fix before merge)

**1. Dead port mapping is pinned forever → permanent loss of direct-path recovery.** `storage/nat.nim`
- `hasMappingIds()` (149-152) checks only `isSome`, not liveness; the NotReachable keep-branch (220-222) gates on it. Once *any* mapping has been created and the node stays NotReachable, this branch is always taken — `mapNatPorts()`/`resetMappings()` never run again and `activeTcpPort` never clears.
- `setupMappedAddrMapper` (163-176) injects the mapped address at index 0 whenever `activeTcpPort.isSome` — **without** a `hasLiveMapping` check. A dead mapping keeps being offered to AutoNAT first.
- Net effect: if the router mapping expires while the node is still NotReachable, the stale mapped addr stays pinned, AutoNAT keeps testing it, and the node can never recover a direct path (relay still works, so not Critical).
- **Fix:** gate both the keep-decision and the mapper injection on `hasLiveMapping(...)`; on a dead mapping call `resetMappings()` and fall through to recreate/relay. The `{.base.}` method seams make this unit-testable without real I/O.

**2. Shutdown race: libplum `cleanup()` can run while an AutoNAT probe is mid-`mapNatPorts`.** `storage.nim:166-167`, `nat.nim:95,115-124,191`
- `stop()` calls `natMapper.stop()` synchronously → `close()` → global `cleanup()` immediately, but the AutoNAT service is stopped later via awaited futures, and `AutonatV2Service.stop` uses `cancelSoon()` (doesn't await the in-flight probe). The `if m.closed: return` guard is only checked at entry, so a probe parked at `await m.createMappingFor(...)` can resume and call into freed libplum → crash/leak on exit.
- **Fix:** re-check `m.closed` after each `await` in `mapNatPorts` (between the TCP/UDP creates and before storing results), bailing without creating/storing if closed.

**3. AutoNAT parameter defaults contradict the design doc — flagged independently by 3 reviewers.** `conf.nim:56,323,329,335`
- Code ships `numPeersToAsk=3, maxQueueSize=3, minConfidence=0.7, scheduleInterval=2min`; Doc 1's logos-storage table specifies `5, 5, 0.6, 5min` (and explicitly criticizes the `3/0.7` logos-delivery values the code adopted). This is not cosmetic: `3/0.7` requires a **unanimous 3/3** to confirm reachability (2/3≈0.67 < 0.7), whereas `5/0.6` confirms on a **3/5 majority** — a material behavior change. Values are now CLI-tunable, so it's a justified-but-undocumented deviation. **Fix:** pick the authoritative source, align defaults, and cite the doc in a `conf.nim` comment.

**4. `validateAutonatConfig` has asymmetric bounds.** `conf.nim:389-417`
- It bounds `maxQueueSize`, `numPeersToAsk`, `observedAddrMinCount`, `minConfidence`, but leaves `natScheduleInterval` (zero/negative Duration breaks the scheduler), `natMaxRelays` (0 silently disables relay), and the three port-mapping timeouts (negative → libplum `cint`) unchecked. **Fix:** add `> 0` lower bounds, mirroring the existing checks. (Also: `validateAutonatConfig` itself has no unit test despite being a pure `?!void`.)

**5. A disabled test reports GREEN, not skipped.** `tests/storage/testnatdetection.nim:260`
- The "mapped address is AutoNAT's first dial candidate" test was neutered with a bare `return` (commit `8263f9ea` "Skip test temporary"), so CI reports it **passing** while coverage of the mapping→AutoNAT-candidate handoff is silently gone. **Fix:** use a real skip marker + tracking issue; ideally stabilize, since this is exactly the path that Important issue #1 breaks.

storage/nat.nim
2-towns: This is because nim-libplum runs `discover` when on initialisation. So we if don't close it, we will run `discover` once. This could be improved be exporting a `discover` function from nim-libplum.

storage/nat.nim
2-towns: Yes the `PortMapping*` is something that I wanted to bring from nim-libplum actually but didn't have the time to do it.

2-towns:
> I didn't have nearly enough time to thoroughly review this PR, but from what i've seen it looks good. There are some suggestions on api design, and a potential issue with closing the nat mapper early that I flagged.
> 
> I also ran this through Claude who gave the following suggestions, but I didn't have time to run through and verify them:
> 
> ## Important (fix before merge)
> **1. Dead port mapping is pinned forever → permanent loss of direct-path recovery.** `storage/nat.nim`
> 
> * `hasMappingIds()` (149-152) checks only `isSome`, not liveness; the NotReachable keep-branch (220-222) gates on it. Once _any_ mapping has been created and the node stays NotReachable, this branch is always taken — `mapNatPorts()`/`resetMappings()` never run again and `activeTcpPort` never clears.
> * `setupMappedAddrMapper` (163-176) injects the mapped address at index 0 whenever `activeTcpPort.isSome` — **without** a `hasLiveMapping` check. A dead mapping keeps being offered to AutoNAT first.
> * Net effect: if the router mapping expires while the node is still NotReachable, the stale mapped addr stays pinned, AutoNAT keeps testing it, and the node can never recover a direct path (relay still works, so not Critical).
> * **Fix:** gate both the keep-decision and the mapper injection on `hasLiveMapping(...)`; on a dead mapping call `resetMappings()` and fall through to recreate/relay. The `{.base.}` method seams make this unit-testable without real I/O.
> 
> **2. Shutdown race: libplum `cleanup()` can run while an AutoNAT probe is mid-`mapNatPorts`.** `storage.nim:166-167`, `nat.nim:95,115-124,191`
> 
> * `stop()` calls `natMapper.stop()` synchronously → `close()` → global `cleanup()` immediately, but the AutoNAT service is stopped later via awaited futures, and `AutonatV2Service.stop` uses `cancelSoon()` (doesn't await the in-flight probe). The `if m.closed: return` guard is only checked at entry, so a probe parked at `await m.createMappingFor(...)` can resume and call into freed libplum → crash/leak on exit.
> * **Fix:** re-check `m.closed` after each `await` in `mapNatPorts` (between the TCP/UDP creates and before storing results), bailing without creating/storing if closed.
> 
> **3. AutoNAT parameter defaults contradict the design doc — flagged independently by 3 reviewers.** `conf.nim:56,323,329,335`
> 
> * Code ships `numPeersToAsk=3, maxQueueSize=3, minConfidence=0.7, scheduleInterval=2min`; Doc 1's logos-storage table specifies `5, 5, 0.6, 5min` (and explicitly criticizes the `3/0.7` logos-delivery values the code adopted). This is not cosmetic: `3/0.7` requires a **unanimous 3/3** to confirm reachability (2/3≈0.67 < 0.7), whereas `5/0.6` confirms on a **3/5 majority** — a material behavior change. Values are now CLI-tunable, so it's a justified-but-undocumented deviation. **Fix:** pick the authoritative source, align defaults, and cite the doc in a `conf.nim` comment.
> 
> **4. `validateAutonatConfig` has asymmetric bounds.** `conf.nim:389-417`
> 
> * It bounds `maxQueueSize`, `numPeersToAsk`, `observedAddrMinCount`, `minConfidence`, but leaves `natScheduleInterval` (zero/negative Duration breaks the scheduler), `natMaxRelays` (0 silently disables relay), and the three port-mapping timeouts (negative → libplum `cint`) unchecked. **Fix:** add `> 0` lower bounds, mirroring the existing checks. (Also: `validateAutonatConfig` itself has no unit test despite being a pure `?!void`.)
> 
> **5. A disabled test reports GREEN, not skipped.** `tests/storage/testnatdetection.nim:260`
> 
> * The "mapped address is AutoNAT's first dial candidate" test was neutered with a bare `return` (commit `8263f9ea` "Skip test temporary"), so CI reports it **passing** while coverage of the mapping→AutoNAT-candidate handoff is silently gone. **Fix:** use a real skip marker + tracking issue; ideally stabilize, since this is exactly the path that Important issue #1 breaks.

3. I kept 3 but reduce `minConfidence`. Will double check with Giuliano what should be the defaults. 
5. Not relevant anymore, the test was deleted because it is a known limitation for now.

I fixed the other points.

storage/nat.nim
2-towns: Done

storage/nat.nim
emizzle: But what about the next call to `handleNatStatus` (on the next iteration of AutoNAT)?

2-towns: Then it will re-initialise libplum on the next iteration, run discover internally and try port mapping again. `close` is a shortcut of stop in that case.

I think the boolean `closed` is really misleading here, it should be `stopped` I will update it. Plus I found that I was never setting this boolean to true again, so after it was stopped, that was forever. 

I fix that + tests in my last commit.

2-towns:
FIX #1417

2-towns:
FIX #782

storage/rest/json.nim
```diff
@@ -52,22 +55,36 @@ type
...
+    clientMode* {.serialize.}: bool
+    relayRunning* {.serialize.}: bool
+    portMapping* {.serialize.}: string
+
   DebugInfo* = object # Peer's ID
     id* {.serialize.}: PeerId
     # peer addresses known by the libp2p switch
     addrs* {.serialize.}: seq[MultiAddress]
+    # local data directory
+    repo* {.serialize.}: string
```
gmega: I think I dropped this from master...

storage/utils/addrutils.nim
```diff
@@ -40,22 +49,35 @@ func remapAddr*(
...
+  let relayWireAddr = MultiAddress.init(relayWireStr).valueOr:
+    return false
+  relayWireAddr.isPublicMA()
+
+proc dialableAddressPolicy*(ma: MultiAddress): bool {.gcsafe, raises: [].} =
+  # Use with switchBuilder.withAddressPolicy.
+  # Filter the peerInfo.addrs updated by libp2p without
+  # declaring another address mapper.
+  if ma.isCircuitRelayMA():
+    ma.hasPublicRelayTransport()
```
gmega, line 74: OK... I guess we never have private relays, but I suppose it doesn't hurt to check.

storage/conf.nim
```diff
@@ -343,11 +344,130 @@ type
...
 
+    natScheduleInterval* {.
+      desc: "Interval between AutoNAT reachability checks",
+      defaultValue: DefaultNatScheduleInterval,
+      defaultValueDesc: $DefaultNatScheduleInterval,
+      name: "nat-schedule-interval"
+    .}: Duration
+
+    natNumPeersToAsk* {.
+      desc: "Number of peers to ask per AutoNAT round",
```
gmega:
```suggestion
      desc: "Number of peers to contact per AutoNAT round",
```

storage/conf.nim
```diff
@@ -343,11 +344,130 @@ type
...
+      # peer.
+      desc: "Minimum confidence threshold to confirm reachability",
+      defaultValue: 0.6,
+      name: "nat-min-confidence"
+    .}: float
+
+    natObservedAddrMinCount* {.
+      desc:
+        "Number of identify observations of the same external address required " &
+        "before it is used as the node's dialable address",
```
gmega, line 378:
```suggestion
      desc:
        "Minimum number of times that an address must show up in identify replies" &
        "before it is used as the node's dialable address",
```

storage/discovery.nim
```diff
@@ -203,33 +204,46 @@ method removeProvider*(
...
-
-proc updateDhtRecord*(d: Discovery, addrs: openArray[MultiAddress]) =
-  ## Update providers record
-  ##
-
-  info "Updating Dht record", addrs = addrs
-  d.dhtRecord = SignedPeerRecord
-    .init(d.key, PeerRecord.init(d.peerId, @addrs))
+    let spr = SignedPeerRecord
+      .init(d.key, PeerRecord.init(d.peerId, tcpAddrs & udpAddrs))
```
gmega: As per the discussion we were having, [I don't think it makes sense to have the TCP addresses here](https://hackmd.io/eyjSeKI8QVKdRIDMzVx9eQ).

storage/discovery.nim
```diff
@@ -203,33 +204,46 @@ method removeProvider*(
...
+  ## Unlike announceDirectAddrs, no UDP address is derived so dhtAddrs is left untouched.
+  d.announceAddrs = @addrs
+  info "Updating announce record", addrs = d.announceAddrs
+  d.providerRecord = SignedPeerRecord
+    .init(d.key, PeerRecord.init(d.peerId, d.announceAddrs))
     .expect("Should construct signed record").some
 
   if not d.protocol.isNil:
-    d.protocol.updateRecord(d.dhtRecord).expect("Should update SPR")
+    d.protocol.updateRecord(d.providerRecord).expect("Should update SPR")
```
gmega: Same here - if you're under a relay, you basically have no usable discovery address, so I think this should remain undefined (assuming it's allowed!).

storage/nat.nim
```diff
@@ -8,427 +8,348 @@
...
-        )
-      if pmres.isErr:
-        error "NAT-PMP port mapping", msg = pmres.error, port
+
+      if maybePorts.isSome:
+        let (tcpPort, udpPort, protocol) = maybePorts.get()
+
+        info "Port mapping created successfully", tcpPort, udpPort, protocol
+
+        # The announce happens once AutoNAT confirms Reachable.
```
gmega, line 221: OK so basically you map the ports, then wait for the next cycle, is that it?

storage/nat.nim
```diff
@@ -8,427 +8,348 @@
...
+
         return
```
gmega, line 223: OK yes, this returns. In case there's a port mapping already in place, it'll fall through to the relay. It can't happen that you're trying this concurrently with an ongoing AutoNAT check, can it? Cause then it might happen that you get an unreachable signal after you open the port map, but from an AutoNAT probe that was done before you actually mapped the ports.

I suppose what protects you is the AutoNAT interval (i.e. if you're processing this in the reachability callback it'll be at least 2 minutes before the next run), but this could get screwed up if the chronos thread gets blocked doing something for too long...

storage/nat.nim
```diff
@@ -8,427 +8,348 @@
...
+): Future[bool] {.async: (raises: [CancelledError]).} =
+  proc tryConnect(
+      address: MultiAddress
+  ): Future[bool] {.async: (raises: [DialFailedError, CancelledError]).} =
+    debug "Trying to create direct connection", peerId, address
+    await switch.connect(peerId, @[address], true, false)
+    debug "Direct connection created."
+    return true
+
+  await sleepAsync(500.milliseconds) # wait for AddressBook to be populated
```
gmega, line 298: Ugh...

storage/nat.nim
```diff
@@ -8,427 +8,348 @@
...
+      addresses
+    return
+  info "Relay reservation updated", addresses = publicAddrs
+  # relay addresses are for download traffic only, not DHT routing
+  discovery.announceRelayAddrs(publicAddrs)
+
+# Hole punching logic below is adapted from libp2p's HPService
+# (libp2p/services/hpservice.nim). HPService cannot be used directly because it
+# depends on AutoNAT v1 and starts the relay immediately on NotReachable,
+# bypassing the UPnP step.
```
gmega, line 285: I suppose we could modify this behavior in libp2p eventually...

storage/storage.nim
```diff
@@ -129,24 +148,68 @@ proc start*(s: StorageServer) {.async.} =
...
-  for announceAddr in announceAddrs:
-    let (maybeIp, _) = getAddressAndPort(announceAddr)
-    if maybeIp.isSome and maybeIp.get.isGlobalUnicast():
-      hasPublicAddr = true
-      break
+  if s.natMapper.isSome:
+    s.natMapper.get.start()
+
+  # When listenPort is 0 the OS assigns a random port. For UDP, the port
+  # doesn't change so there is no need to update it.
```
gmega, line 155: Not sure I understand what you mean by "for UDP the port doesn't change"...

storage/storage.nim
```diff
@@ -129,24 +148,68 @@ proc start*(s: StorageServer) {.async.} =
...
-  if not hasPublicAddr:
-    warn "Unable to determine a public IP address. This node will only be reachable on a private network."
+  await s.storageNode.start()
 
-  s.storageNode.discovery.updateAnnounceRecord(announceAddrs)
-  s.storageNode.discovery.updateDhtRecord(discoveryAddrs)
+  # Connect to the bootstrap nodes in order to have connected peers
+  # for Autonat. The dials are run concurrently in case of
+  # a dead bootstrap node that could timeout.
+  proc connectBootstrapNode(
```
gmega: Technically you're connecting to known public AutoNAT/Relay peers, which may or may not be bootstrap nodes. :-)

gmega (approved):
OK I reviewed MOST of it, with the exception of some tests. :-) The only part I'm unsure is with the way you're handling the SPRs with TCP and UDP addresses, the rest looks solid.

storage/rest/json.nim
2-towns: Yep sorry, it was re introduced during the merge, will remove it back

storage/utils/addrutils.nim
2-towns: Yes just a double layer of protection, I think a private relay is okay for libp2p so I wanted to make sure that we don't have private address when using relay.

storage/nat.nim
2-towns: Yes exactly.

storage/nat.nim
2-towns: Yes exactly.  

From `nim-libplum` side, it has 2 timeouts of 10 seconds each by default for `discover` and the mapping it self. So we have have a lot of buffer with Autonat interval to 2 minutes. 

But yes if the chronos thread gets blocked for some reason this is an issue. Then we could increase the interval if needed.

storage/nat.nim
2-towns: Yes :sweat_smile: So I just copy pasted the code from libp2p because I think it can be easier to change the code (in storage nim or libp2p) later on, if we keep the code synced for now.

storage/nat.nim
2-towns: Yes exactly. We will need to discuss with libp2p team. I don't know why `HPService` is available with Autonat version 1. 

I didn't spend time to dig into it because the changes in Autonat V2 were taking already some time to be integrated. But this is something I can explore if we can spend more time on it.

storage/storage.nim
2-towns: I mean it stays equal to the config value.

For the `listenPort` we allow 0, so the OS picks a free port then we have to read it back and update the mapper. 
But for the discovery port, it is fixed (defaults to 8090), so there's nothing to update.

storage/storage.nim
2-towns: Yes it should be `connectAutonatServer`. Even `findReachableNodes` is not great because it is not only about Reachability, we want Autonat servers. I will update both.

storage/discovery.nim
2-towns: Okay addresses in this commit: https://github.com/logos-storage/logos-storage-nim/pull/1417/changes/5ee32683e858776952d2df644a691d47332ba177

storage/discovery.nim
2-towns: Yes but in your document, you mentioned that, `providerRecord` and `localRecord` (dht record) seem to be designed to contain TCP in the first one and UDP in the latter. 

So at this point, no need to update the local record. Or we could even update it with an empty array to make sure it does not contain anything, because under a relay, your `localRecord` should be empty.

storage/discovery.nim
```diff
@@ -203,33 +206,55 @@ method removeProvider*(
...
-
-  d.announceAddrs = @addrs
-
+proc getSpr*(d: Discovery): SignedPeerRecord =
+  ## Expose the discovery addresses (UDP) and provider addresses (TCP).
+  ## TCP addresses are needed because Autonat requires to connect to
+  ## Autonat servers and we are currently using the provider record for that.
+  ## We might get ride of that if we add a new configuration option that gives
+  ## the Autonat servers the TCP addresses directly, something like the Mix options
+  ## that we have.
```
gmega:
```suggestion
  ## We might get rid of that if we add a new configuration option that gives
  ## the Autonat servers the TCP addresses directly (as Mix options do), or
  ## once we get proper service discovery/migrate to the libp2p Kad DHT.
```

storage/discovery.nim
gmega: Yeah, would probably make sense to clear it.

gmega (approved):
Double approve! 😃

PR #1420, opened 2026-04-15
ci: add release tests workflow with ephemeral k8s cluster

emizzle:
This branch adds an end-to-end release test stage to the existing release workflow. A new `release-tests` job is introduced that, on tag pushes or manual dispatch, first builds a dedicated `dist-tests`-tagged Docker image, then provisions an ephemeral DigitalOcean Kubernetes cluster via Terraform (configuration added under `.github/release/clusters/codex-dist-tests-do-ams3/`), deploys a test job using a Kubernetes Job manifest and a kubeconfig template, streams logs until completion, and checks the job's success condition before tearing down the cluster. 
- A static in-cluster kubeconfig is created via a service account token and injected as a Kubernetes secret so the test runner can interact with the cluster. 
- The `release` job is updated to depend on both `build` and `release-tests` succeeding. 
- A `workflow_dispatch` input was also added to allow manually specifying the `cs-dist-tests` branch to run tests from, for example using `build/docker-cleanup` from https://github.com/logos-storage/logos-storage-nim-cs-dist-tests/pull/124.


gmega:
Is this working @emizzle? Can I merge it?

emizzle:
This is still a valid PR, however, #1439 supersedes it, as it moves to GCP with logging streamed to Cloud Logging. So we can close this in favour of #1439.

PR #1422, opened 2026-04-16
chore: update chronicles

2-towns:
This PR updates `chronicles`  to the last version.

Needs: https://github.com/logos-storage/nim-serde/pull/35

### DIFF
--- a/storage/logutils.nim
+++ b/storage/logutils.nim
@@ -92,6 +92,7 @@ import std/sugar
 import std/typetraits
 
 import pkg/chronicles except toJson, `%`
+import json_serialization/writer as json_serialization_writer
 from pkg/chronos import TransportAddress
 from pkg/libp2p import Cid, MultiAddress, `$`
 import pkg/questionable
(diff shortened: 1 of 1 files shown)


emizzle (approved):
Sucks that there's no tests for the `formatIt` usage inside Logos Storage, so it's hard to know if these changes work or not.

2-towns:
> Sucks that there's no tests for the `formatIt` usage inside Logos Storage, so it's hard to know if these changes work or not.

We do have tests no ? https://github.com/logos-storage/logos-storage-nim/blob/master/tests/storage/testlogutils.nim

emizzle:
> > Sucks that there's no tests for the `formatIt` usage inside Logos Storage, so it's hard to know if these changes work or not.
> 
> We do have tests no ? https://github.com/logos-storage/logos-storage-nim/blob/master/tests/storage/testlogutils.nim

Oh nice! I forgot I wrote those 😅

PR #1423, opened 2026-04-17
feat: add a warning when private address only

2-towns:
This PR introduces a warning when only private addresses are detected.

Two additional fixes are included:                                                                                                                                                          
                             
1. Fix wrong log message for `CLOSE_NODE` operation in libstorage.
2. Replace `raiseAssert` with `raise newException(StorageError, ...)` in `stop` and `close` so errors propagate correctly to libstorage instead of crashing the process.

### DIFF
--- a/library/storage_thread_requests/requests/node_lifecycle_request.nim
+++ b/library/storage_thread_requests/requests/node_lifecycle_request.nim
@@ -147,10 +147,6 @@ proc createStorage(
   else:
     debug "Rest API is enabled!"
 
-  if conf.allowPrivateAddress.isNone:
-    # Default to true so the node starts out of the box.
-    conf.allowPrivateAddress = true.some
-
   let server =
     try:
       StorageServer.new(conf, pk, logFile)
--- a/library/storage_thread_requests/requests/node_lifecycle_request.nim
+++ b/library/storage_thread_requests/requests/node_lifecycle_request.nim
@@ -91,6 +91,7 @@ proc createStorage(
     configJson: cstring
 ): Future[Result[StorageServer, string]] {.async: (raises: []).} =
   var conf: StorageConf
+
   try:
     conf = StorageConf.load(
       version = storageFullVersion,
@@ -146,6 +147,10 @@ proc createStorage(
   else:
     debug "Rest API is enabled!"
 
+  if conf.allowPrivateAddress.isNone:
+    # Default to true so the node starts out of the box.
+    conf.allowPrivateAddress = true.some
+
   let server =
     try:
       StorageServer.new(conf, pk, logFile)
@@ -185,6 +190,6 @@ proc process*(
     try:
       await storage[].close()
     except Exception as e:
-      error "Failed to STOP_NODE.", error = e.msg
+      error "Failed to CLOSE_NODE.", error = e.msg
       return err(e.msg)
   return ok("")
--- a/storage/storage.nim
+++ b/storage/storage.nim
@@ -80,8 +80,8 @@ proc start*(s: StorageServer) {.async.} =
 
   if not s.config.allowPrivateAddress.get(false):
     var hasPublicAddr = false
-    for ma in announceAddrs:
-      let (maybeIp, _) = getAddressAndPort(ma)
+    for announceAddr in announceAddrs:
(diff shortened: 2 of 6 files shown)


library/storage_thread_requests/requests/node_lifecycle_request.nim
```diff
@@ -146,6 +147,10 @@ proc createStorage(
   else:
     debug "Rest API is enabled!"
 
+  if conf.allowPrivateAddress.isNone:
+    # Default to true so the node starts out of the box.
```
emizzle: Could you please elaborate why this is default only for libstorage but not in the config? Why is operating a node via libstorage meant to be different?

storage/storage.nim
```diff
@@ -110,7 +122,7 @@ proc stop*(s: StorageServer) {.async.} =
 
   if res.failure.len > 0:
     error "Failed to stop Storage node", failures = res.failure.len
-    raiseAssert "Failed to stop Storage node"
+    raise newException(StorageError, "Failed to stop Storage node")
```
emizzle: Should probably output the failures here, eg `"Failed to stop Storage node: " & res.failure.join(", ")`

storage/storage.nim
```diff
@@ -137,7 +149,7 @@ proc close*(s: StorageServer) {.async.} =
 
   if res.failure.len > 0:
     error "Failed to close Storage node", failures = res.failure.len
-    raiseAssert "Failed to close Storage node"
+    raise newException(StorageError, "Failed to close Storage node")
```
emizzle: Same here

storage/storage.nim
```diff
@@ -77,6 +78,17 @@ proc start*(s: StorageServer) {.async.} =
     s.config.nat, s.storageNode.switch.peerInfo.addrs, s.config.discoveryPort
   )
 
+  if not s.config.allowPrivateAddress.get(false):
```
emizzle: A suggestion: 
If `allowPrivateAddress` is true, there should still be a warning. Something like 
```
Unable to determine a public IP address. This node will only be reachable on a private network. Use --allow-private-address=false to restrict usage to publicly reachable nodes.
```

emizzle (commented):
In my opinion, I'm not sure if this is worth a config flag. I think it would be better to show a warning. Even when a node is not reachable because of a private address on a public network, they can still download from the network, so setting --allow-private-address=false means they would not be able to do anything with the node.

library/storage_thread_requests/requests/node_lifecycle_request.nim
2-towns: The idea is that the message is quite obvious when you start your node as standalone and see it closing with an error message.

For libstorage, it might not be obvious, and since we integrated libstorage into some places (Status, Logos Storage Module), it could "break" the current behavior.

2-towns:
> In my opinion, I'm not sure if this is worth a config flag. I think it would be better to show a warning. Even when a node is not reachable because of a private address on a public network, they can still download from the network, so setting --allow-private-address=false means they would not be able to do anything with the node.

I don't have a strong opinion on this, it seems fine to me. Any thoughts, @gmega?

library/storage_thread_requests/requests/node_lifecycle_request.nim
emizzle: My opinion would be to make the right decision for Logos Storage, and inform downstream parties about the change, instead of making decisions on the basis of avoiding downstream changes.

In this case, I think the config should have a default value of `true`, and there would be no special case for libstorage.

Although, in general, I still think that a config flag for this is a bit of overkill.

2-towns: So you would remove the flag and just add a warning  right ?

2-towns: Okay I removed the config flag

emizzle (approved):
Nice Arnaud!

2-towns:
Okay so after re-thinking to it, I think the config argument is useful to avoid to announce private address on a public network. I am gonna to re-introduce it in the NAT Traversal PR.

PR #1425, opened 2026-04-24
chore: update nim 2.2.10

2-towns:
This PR updates Nim to 2.2.10 and performs another round of cleanup. 

It updates DHT to 0.6.2 to fix a crash issue when logging.


cnanakos (approved):
LGTM

PR #1433, opened 2026-05-01
chore: update nimbus build system

2-towns:
For the NAT Traversal PR, it is required to update nimbus build system which involves some changes in Nix configuration.


emizzle (approved):
LGTM but jenkins is failing

PR #1434, opened 2026-05-06
fix: validate WantBlocks request ranges to prevent DoS

cnanakos:


storage/blockexchange/engine/engine.nim
```diff
@@ -1141,6 +1141,21 @@ proc new*(
   proc wantBlocksRequestHandler(
       peer: PeerId, req: WantBlocksRequest
   ): Future[seq[BlockDelivery]] {.async: (raises: [CancelledError]).} =
+    let maxIndex = high(Natural).uint64
+    var totalCount: uint64 = 0
+    for r in req.ranges:
+      if r.count == 0 or r.count > MaxBlocksPerBatch or r.start > maxIndex or
```
gmega: Do we need to check that `r.count > MaxBlocksPerBatch`? Seems somewhat redundant since you're checking this below again (in totalCount).

gmega (approved):
LGTM

storage/blockexchange/engine/engine.nim
cnanakos: Right, was meant as fail-fast, but the aggregate check covers it. Dropped.

PR #1435, opened 2026-05-07
chore: remove unused cache and make discovery IP limits configurable

2-towns:
- Remove unused CacheStore instance and cacheSize CLI option                                                          
- Add hidden CLI options `--discovery-table-ip-limit` (default: 10) and `--discovery-bucket-ip-limit` (default: 2) to configure discv5 routing table IP limits, replacing the hardcoded `high(uint)` that disabled all protection

### DIFF
--- a/storage/conf.nim
+++ b/storage/conf.nim
@@ -272,15 +272,22 @@ type
       name: "block-retries"
     .}: int
 
-    cacheSize* {.
+    discoveryTableIpLimit* {.
+      desc: "Maximum number of nodes with the same IP in the discovery routing table",
+      defaultValue: 10'u,
+      defaultValueDesc: "10",
+      name: "discovery-table-ip-limit",
+      hidden
+    .}: uint
+
+    discoveryBucketIpLimit* {.
       desc:
-        "The size of the block cache, 0 disables the cache - " &
-        "might help on slow hardrives",
-      defaultValue: 0,
-      defaultValueDesc: "0",
-      name: "cache-size",
-      abbr: "c"
-    .}: NBytes
+        "Maximum number of nodes with the same IP per bucket in the discovery routing table",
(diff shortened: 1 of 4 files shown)


storage/conf.nim
```diff
@@ -272,15 +272,22 @@ type
...
-      defaultValueDesc: "0",
-      name: "cache-size",
-      abbr: "c"
-    .}: NBytes
+        "Maximum number of nodes with the same IP per bucket in the discovery routing table",
+      defaultValue: 2'u,
+      defaultValueDesc: "2",
+      name: "discovery-bucket-ip-limit",
+      hidden
+    .}: uint
```
cnanakos, line 290: There is no bounds validation for any of the values above. For example I might be able to pass 0 which will reject every same IP node or pass `uint.high` and disable protection again. 

Suggestion:
* `bucketIpLimit: range[1'u..BUCKET_SIZE'u]`
* `tableIpLimit: range[1'u..1024'u]` or similar reflecting any sensible production value.

2-towns: > There is no bounds validation for any of the values above. For example I might be able to pass 0 which will reject every same IP node or pass `uint.high` and disable protection again.
> 
> Suggestion:
> 
> * `bucketIpLimit: range[1'u..BUCKET_SIZE'u]`
> * `tableIpLimit: range[1'u..1024'u]` or similar reflecting any sensible production value.

This parameter is hidden. It is not something we should expose in documentation if we have good defaults. As a reminder, it currently sets the limit to high(uint) which can introduce vulnerabilities. 
If someone wants to change this value and breaks their node, I would say it is their responsibility, but I think it is fair to assume most people will not do that.

cnanakos (commented):
Fair enough. LGTM.

emizzle (approved):

PR #1437, opened 2026-05-21
feat: network presets

gmega:
Fixes https://github.com/logos-co/logos-storage-module/issues/31

This PR introduces network presets -- named sets of bootstrap nodes -- that define different Logos storage networks. By default, the node will currently bootstrap onto the Logos devnet (`logos.dev`).

Users maintain the option of overriding preset bootstrap nodes by manually specifying `--bootstrap-nodes`.

### DIFF
--- a/storage/conf.nim
+++ b/storage/conf.nim
@@ -415,6 +415,13 @@ proc parseCmdArg*(T: type Duration, val: string): T =
     quit QuitFailure
   dur
 
+proc parseCmdArg*(T: type NetworkPreset, p: string): NetworkPreset =
+  let res = NetworkPresets.find(p)
+  if res.isErr:
+    fatal "Cannot parse the network preset.", error = res.error(), input = p
+    quit QuitFailure
+  return res.get()
+
 proc readValue*(r: var TomlReader, val: var SignedPeerRecord) =
   without uri =? r.readValue(string).catch, err:
     error "invalid SignedPeerRecord configuration value", error = err.msg
@@ -478,6 +485,17 @@ proc readValue*(
     except CatchableError as err:
       raise newException(SerializationError, err.msg)
 
+proc readValue*(
+    r: var TomlReader, val: var NetworkPreset
+) {.raises: [SerializationError, IOError].} =
+  let
+    str = r.readValue(string)
+    preset = NetworkPresets.find(str)
+  if preset.isErr():
+    raise newException(SerializationError, preset.error())
+
+  val = preset.get()
+
 # no idea why confutils needs this:
 proc completeCmdArg*(T: type NBytes, val: string): seq[string] =
   discard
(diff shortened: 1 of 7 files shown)


storage/conf.nim
```diff
@@ -179,11 +181,18 @@ type
     bootstrapNodes* {.
       desc:
         "Specifies one or more bootstrap nodes to use when " &
-        "connecting to the network",
+        "connecting to the network. When specified, overrides " &
+        "the network preset option.",
       abbr: "b",
       name: "bootstrap-node"
     .}: seq[SignedPeerRecord]
```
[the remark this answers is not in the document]
gmega: I think this doesn't need to be addressed.

storage/presets.nim
```diff
@@ -0,0 +1,106 @@
...
+
+func `$`*(preset: NetworkPreset): string =
+  "[" & preset.name & "]: " & preset.description
+
+func `$`*[N](presets: array[N, NetworkPreset]): string =
+  result = ""
+  for preset in presets:
+    result &= $preset & "; "
+
+proc parse*(T: type SignedPeerRecord, p: string): Result[SignedPeerRecord, string] =
```
2-towns, line 39: Not sure why you moved here

storage/presets.nim
```diff
@@ -0,0 +1,106 @@
...
+    let e = getCurrentException()
+    return err(e.msg)
+
+proc `bootstrapNodes`*(self: NetworkPreset): seq[SignedPeerRecord] =
+  for record in self.unparsedRecords:
+    # Having an invalid SPR in a hardcoded config is a bug, a+
+    # it should crash the node.
+    result.add(parse(SignedPeerRecord, record).tryGet())
+
+const NetworkPresets* = [
```
2-towns, line 55: We could have a map as well I suppose so the find would be easier

storage/storage.nim
```diff
@@ -216,6 +216,15 @@ proc new*(
     error "Failed to initialize discovery datastore",
       path = providersPath, err = discoveryStoreRes.error.msg
 
+  let bootstrapNodes =
+    if config.bootstrapNodes.len > 0:
+      warn "Overriding network preset using custom bootstrap nodes",
```
2-towns: `warn` seems a bit too much, I would use `info`

2-towns (approved):
LGTM

storage/storage.nim
gmega: Yeah I guess the here I was trying to convey here is like "you should probably not be doing this", which is true in most cases. But OK I'll use info.

storage/presets.nim
gmega: The set of presets is small and you do a single lookup so it didn't seem worth the trouble.

storage/presets.nim
gmega: I need to parse those in the presets, and I had three options:

1. leave them in `conf.nim`, but then end up with a circular import;
2. move it to a separate module and then import it both under `conf.nim` and `presets.nim`; that looked like overkill;
3. just move it here, which is what I ended up deciding to do.

gmega: I suppose I could've also duplicated the logic here, but that didn't even cross my mind :slightly_smiling_face:

2-towns: Make sense :+1:

PR #1439, opened 2026-05-28
feat: end-to-end release test pipeline on GKE with structured logging

emizzle:
Establishes a production-grade end-to-end release test pipeline that provisions a GKE cluster, runs the dist-test suite, streams structured logs to GCP Cloud Logging, and tears everything down — all from a single GitHub Actions release workflow. This workflow is executed automatically for every new release.

- moves cluster creation scripts into logos-storage-nim (from logos-storage-nim-cs-dist-tests)
- cluster is setup and torn down in each release workflow run
- workflow can be triggered manually via Actions → Release → Run workflow
- test run summary is posted to the workflow summary, with per-fixture log links, deep linking to structured logs in GCP Cloud Logging, with a 30-day expiry window

## Note
This PR relies on https://github.com/logos-storage/logos-storage-nim-cs-dist-tests/pull/126 to be merged first for workflows to run automatically during the release workflow. However, the workflow can be run manually, by specifying the `fix/remove-system-critical-priority` branch in the workflow input parameters.

image


.github/release/clusters/logos-storage-rel-tests-gcp-europe-west4/main.tf
```diff
@@ -0,0 +1,29 @@
...
+  region  = var.region
+  zone    = var.zone
+
+  node_pool_name         = "runners-ci-e2-standard-2"
+  node_pool_machine_type = "e2-standard-2"
+  node_pool_min          = 1
+  node_pool_max          = 5
+  node_pool_labels = {
+    default-pool  = "true"
+    scaling-type  = "auto"
```
gmega: Does it make sense to have autoscaling for a transient cluster? It'll just make everything slower. Just spin it up with the number of nodes you need.

.github/release/clusters/logos-storage-rel-tests-gcp-europe-west4/main.tf
```diff
@@ -0,0 +1,29 @@
...
+    workload-type = "tests-runners-ci"
+  }
+
+  tests_pool_name         = "tests-e2-medium"
+  tests_pool_machine_type = "e2-medium"
+  tests_pool_max          = 5
+  tests_pool_zones        = ["europe-west4-a", "europe-west4-b", "europe-west4-c"]
+  tests_pool_labels = {
+    default-pool  = "false"
+    scaling-type  = "auto"
```
gmega: Same here... I'd just make a fixed-sized pool with no autoscaling. Will probably shave off a few minutes from the tests...

.github/workflows/release.yml
```diff
@@ -179,11 +185,265 @@ jobs:
...
+
+      - name: Wait for runners-ci node to be Ready
+        run: |
+          kubectl wait \
+            --for=condition=Ready \
+            node \
+            -l workload-type=tests-runners-ci \
+            --timeout=300s
+
+      - name: Create in-cluster app kubeconfig secret
```
gmega: Why are you doing this here? Shouldn't you be using terraform for this?

.github/release/clusters/modules/gke/variables.tf
```diff
@@ -0,0 +1,78 @@
...
+variable "tests_pool_labels" {
+  type        = map(string)
+  description = "Kubernetes labels to apply to nodes in the tests pool."
+  default     = {}
+}
+
+variable "tests_pool_zones" {
+  type        = list(string)
+  description = "Zones for the tests node pool. Spanning multiple zones increases spot instance availability."
+  default     = []
```
gmega: It also makes you pay for bandwidth: https://cloud.google.com/vpc/network-pricing. Also, I wouldn't use spot instances for testing: those should run quickly, and it'll just be another confounding factor when GCP decides to kill your instance halfway through a test run.

.github/release/clusters/logos-storage-rel-tests-gcp-europe-west4/main.tf
```diff
@@ -0,0 +1,29 @@
+# Both node pools are inline in the module so GCP provisions them in parallel.
+module "gke" {
+  source = "../modules/gke"
+
+  name    = "logos-storage-rel-tests"
+  project = var.project
+  region  = var.region
+  zone    = var.zone
```
gmega, line 8: You're hardcoding the zones for the node pools, not sure it makes much sense to have those read as variables.

.github/release/clusters/logos-storage-rel-tests-gcp-europe-west4/main.tf
```diff
@@ -0,0 +1,29 @@
...
+  node_pool_labels = {
+    default-pool  = "true"
+    scaling-type  = "auto"
+    workload-type = "tests-runners-ci"
+  }
+
+  tests_pool_name         = "tests-e2-medium"
+  tests_pool_machine_type = "e2-medium"
+  tests_pool_max          = 5
+  tests_pool_zones        = ["europe-west4-a", "europe-west4-b", "europe-west4-c"]
```
gmega: Why multiple zones?

.github/workflows/release.yml
```diff
@@ -179,11 +185,265 @@ jobs:
...
+    if: github.ref_type == 'tag' || github.event_name == 'workflow_dispatch'
+    # needs: build-docker-dist-tests
+    timeout-minutes: 60
+    permissions:
+      id-token: write
+      contents: read
+    env:
+      TF_VAR_project: ${{ vars.RELEASE_TESTS_GCP_PROJECT }}
+      TF_VAR_region: europe-west4
+      TF_VAR_zone: europe-west4-a
```
gmega, line 215: Not very useful as your pool zones are hardcoded.

gmega (changes requested):
I see a few funny choices (autoscaling, multizone pools, spot instances, and some indecision as to whether or not hardcode zones/regions) but otherwise looks fine. I think we should at least remove the vars for setting regions/zones as I don't think those will work if you change them. Either that or set the node pools from the outside (workflow) as well.

.github/release/clusters/logos-storage-rel-tests-gcp-europe-west4/main.tf
emizzle: The idea here was to leave room for beefier tests that spawn a lot of pods. However, we don't currently use tests that would require more than one or two nodes (depending on the cpu/mem allowances ofc)

.github/release/clusters/logos-storage-rel-tests-gcp-europe-west4/main.tf
emizzle: Yeah i agree it doesn't make sense to autoscale the runner. Will update.

.github/release/clusters/logos-storage-rel-tests-gcp-europe-west4/main.tf
emizzle: This was an attempt to eliminate the possibility of zonal resource exhaustion, which was one possible reason why some pods weren't being spun up correctly. I'll reduce this back down to one zone.

.github/workflows/release.yml
emizzle: Yeah agreed. An artefact of trying many different things to get this to work.

.github/workflows/release.yml
emizzle: I think this was the simplest option in terms of not needing to wire up a k8s provider in TF. Will figure out how to move it to TF so that we have the most IaC approach here. Also simplifies the release workflow which got pretty long due to all the issues I ran into while porting it to GKE.

.github/release/clusters/logos-storage-rel-tests-gcp-europe-west4/main.tf
emizzle: I've fixed the runner pool to 1 node and test pool to 2 nodes

.github/release/clusters/logos-storage-rel-tests-gcp-europe-west4/main.tf
emizzle: Updated

.github/workflows/release.yml
emizzle: Updated.

.github/release/clusters/modules/gke/variables.tf
emizzle: I tried spot instances in the past but wasn't using them in this PR. The spot toleration in the dist tests was a left over after removing affinities that never worked for me in GKE (instances were not scheduling with the affinities in place). It has been removed.

.github/release/clusters/logos-storage-rel-tests-gcp-europe-west4/main.tf
emizzle: The zone is defined as an env variable in the workflow, which maps to the TF variable. This allows everything to move in lockstep if the zone is changed (same with region btw).

.github/workflows/release.yml
emizzle: On second look, the pool zones are not hardcoded. The purpose here is TF_VAR automaps to TF variables, and these are used through the terraform config files. So changing the value in the workflow is meant to flow the changes through the entire cluster and all its resources.

.github/release/clusters/modules/gke/variables.tf
gmega: OK, we can go over affinities together. They Should Work (tm)

.github/workflows/release.yml
gmega: Hmmm `tests_pool_zones` is hardcoded in `main.tf` as far as I can tell, and it won't be overridden by a TF_VAR. So while region/zone can indeed change, `tests_pool_zones` will remain always set to `["europe-west4-a", "europe-west4-b", "europe-west4-c"]`

gmega: OK I see you got rid of it so all good

.github/release/clusters/logos-storage-rel-tests-gcp-europe-west4/rbac.tf
```diff
@@ -0,0 +1,83 @@
...
+    name      = "release-tests-runner-token"
+    namespace = "default"
+    annotations = {
+      "kubernetes.io/service-account.name" = kubernetes_service_account.release_tests_runner.metadata[0].name
+    }
+  }
+
+  type                           = "kubernetes.io/service-account-token"
+  wait_for_service_account_token = true
+}
```
gmega: I think you can probably avoid the dance with service account tokens by using workload identity federation.

.github/workflows/release.yml
```diff
@@ -173,17 +179,257 @@ jobs:
...
+      - name: Check job status
+        run: |
+          # kubectl logs may have exited early (API server closed the stream).
+          # Wait for the job to reach a terminal state before checking the result.
+          kubectl wait job/$NAMEPREFIX -n default \
+            --for=condition=Complete \
+            --timeout=300s \
+          || kubectl wait job/$NAMEPREFIX -n default \
+            --for=condition=Failed \
+            --timeout=0s
```
gmega:
```suggestion
          || kubectl wait job/$NAMEPREFIX -n default \
            --for=condition=Failed \
            --timeout=300s
```

Otherwise this won't wait at all?

```text
--timeout=30s:
	The length of time to wait before giving up.  Zero means check once
	and don't wait, negative means wait for a week.
```

gmega (approved):
LGTM, except for what appears to be a bug in your pod termination sync command in `Check job status`.

.github/workflows/release.yml
emizzle: It's just doing a check of the terminal state after the first condition times out, in which case the step continues on to check the job status to ultimately add a log line in CI "Job status: Failed". I've updated it so that we toss out the exit code entirely, hopefully making it a bit more clear that we continue to the job status check always.

.github/release/clusters/logos-storage-rel-tests-gcp-europe-west4/rbac.tf
emizzle: This has been simplified a lot by removing an explicit SA token for kubeconfig and instead using a SA + cluster role binding, but it doesn't use WIF (the former modality has been deprecated). WIF is used in the workflow to authenticate as the SA to create the cluster and run terraform commands.

PR #1440, opened 2026-06-03
chore: bump libp2p

2-towns:
## Summary

Bumps `nim-libp2p` and related vendored deps, with the fixes needed to keep the build working.
## Fixes (forced by the bump)

- `Rng` re-exported from libp2p crypto
- `import pkg/libp2p except NATConfig` (collides with storage's `NatConfig`)
- block-exchange semaphore moved to `chronos/asyncsync`
- explicit `import std/options` where libp2p no longer re-exports it
- new test helper `newStandardSwitch` on `SwitchBuilder` (dropped from libp2p API)

Blocked by: 

https://github.com/logos-storage/logos-storage-nim-dht/pull/116
~~https://github.com/status-im/nim-websock/pull/196~~

### DIFF
--- a/storage/blockexchange/network/network.nim
+++ b/storage/blockexchange/network/network.nim
@@ -318,9 +318,8 @@ proc new*(
     getConn: connProvider,
     inflightSema: newAsyncSemaphore(maxInflight),
     maxInflight: maxInflight,
+    maxIncomingStreamsTotal: Opt.some(maxInflight),
   )
-
-  #self.maxIncomingStreams = self.maxInflight
 
   proc sendWantList(
       id: PeerId,
--- a/storage/blockexchange/network/network.nim
+++ b/storage/blockexchange/network/network.nim
@@ -13,7 +13,7 @@ import std/sequtils
 import pkg/chronos
 
 import pkg/libp2p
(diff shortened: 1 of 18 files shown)


storage/blockexchange/network/network.nim
```diff
@@ -107,13 +107,17 @@ proc send*(
...
 
     await b.inflightSema.acquire()
-    await peer.send(msg)
+    try:
+      await peer.send(msg)
+    finally:
+      try:
+        b.inflightSema.release()
+      except AsyncSemaphoreError as err:
+        error "Failed to release semaphore", msg = err.msg
```
emizzle: It feels like this should do more than log an error. This errors when release was called without acquire. It's clearly called above, so in this case it should probably raise a defect with `raiseAssert`.

emizzle (approved):
Nice. One change I think might be useful, but will approve to not be a bottleneck.

storage/blockexchange/network/network.nim
2-towns: Thanks, updated.

2-towns:
Closing in favour of #1443

PR #1442, opened 2026-06-05
chore: include cancelled futures in failures

2-towns:
`allFinishedFailed` used `f.failed`, which is false for cancelled futures, so they were counted as successes. A cancelled future never belongs in `success`.

Switch to `f.completed`: cancelled futures now land in `failure`.

Example: in `updateExpiry`, a cancelled `ensureExpiry` future meant the block's expiry was not updated, yet the proc returned success.

### DIFF
--- a/storage/errors.nim
+++ b/storage/errors.nim
@@ -86,10 +86,10 @@ proc allFinishedFailed*[T](
   var res: FinishedFailed[T] = (@[], @[])
   await allFutures(futs)
   for f in futs:
-    if f.failed:
-      res.failure.add f
-    else:
+    if f.completed:
       res.success.add f
+    else:
+      res.failure.add f
 
   return res
 


cnanakos (approved):
LGTM

emizzle:
Yikes, sorry I missed this. 

I don't think cancellations should be considered failures either because we are basically redefining how chronos defines a failure to include cancellations. Imo, it should be up to the caller to inspect the successes, failures, and cancellations. Otherwise, this proc is making a decision for them by lumping together failures and cancellations.

2-towns:
> I don't think cancellations should be considered failures either because we are basically redefining how chronos defines a failure to include cancellations. Imo, it should be up to the caller to inspect the successes, failures, and cancellations. Otherwise, this proc is making a decision for them by lumping together failures and cancellations.

The problem is we don't check that, maybe because cancelled futures were implicitly added to success, take this code for example: 

```nim
let ensuringFutures = Iter[int].new(0 ..< manifest.blocksCount).mapIt(
)

let res = await allFinishedFailed[?!void](ensuringFutures)
if res.failure.len > 0:
  trace "Some blocks failed to update expiry", len = res.failure.len
  return failure("Some blocks failed to update expiry (" & $res.failure.len & " )")
```

If there is a cancellation, that means a `ensureExpiry` was cancelled so we should treat `ensuringFutures` as failure I suppose. 

To extent what you mentioned, we could return explicitly the cancelled futures so as you said: 

>  it should be up to the caller to inspect the successes, failures, and cancellations.

emizzle:
> If there is a cancellation, that means a ensureExpiry was cancelled so we should treat ensuringFutures as failure I suppose.

I think that `allFinishedFailed` as a utility has been pigeon-holed into being specific about what it treats as a failure, and therefore it's usefulness is limited to cases that want the return value treated as such. For a more general-purpose utility, the API should not make decisions for the caller, unless they are very clear (what is there is not clear imo).

In my comment, I was implying that we should not treat cancellation as failures, but as cancellations. The function could return a type like:
```nim
type FinishedFutures[T] =
  successes: openArray[Future[T]]
  failures: openArray[Future[T]]
  cancellations: openArray[Future[T]]

# borrowing ideas from marcin's impl of allFinishedValues
proc allFinished*[T](futs: auto): Future[FinishedFutures[T]] {.async: (raises: []).} =
  # wait for all futures to be either completed, failed or canceled
  await allFutures(futs)
  return FinishedFutures(
    failures: futs.countIt(it.failed)
    successes: futs.countIt(it.success)
    cancellations: futs.countIt(it.cancelled)
  )
```

So the example would change to:
```nim
let ensuringFutures = Iter[int].new(0 ..< manifest.blocksCount).mapIt(
)

let res = await allFinished[?!void](ensuringFutures)
if res.failures.len > 0:
  trace "Some blocks failed to update expiry", len = res.failure.len
  return failure("Some blocks failed to update expiry (" & $res.failure.len & " )")
if res.cancellations.len > 0:
  trace "Block expiry update was cancelled in some blocks", len = res.cancellations.len
  return failure("Block expiry update was cancelled in some blocks (" & $res.cancellations.len & " )")
```
But of course, if the caller wanted to treat cancellations as failures, the usage would change to:
```nim
let res = await allFinished[?!void](ensuringFutures)
if res.failures.len > 0 or res.cancellations.len > 0:
  trace "Some blocks failed to update expiry", len = res.failure.len
  return failure("Some blocks failed to update expiry (" & $res.failure.len & " )")
```

emizzle:
If you really wanted to, you could then make a *module-local utility* (not a general purpose utility) that does the comparison condition for you, eg: 

```nim
#  in the blockexchange module

import ../storage/errors

proc someFailed*[T](futs: auto): Future[bool] {.async: (raises: []).} =
  let res = await futs.allFinished
  return res.failures.len > 0 || res.cancellations.len > 0

proc allFinishedFailed*[T](futs: auto): Future[FinishedFutures[T]] {.async: (raises: []).} =
  let res = await futs.allFinished
  return res.failures & res.cancellations
```

2-towns:
> > If there is a cancellation, that means a ensureExpiry was cancelled so we should treat ensuringFutures as failure I suppose.
> 
> I think that `allFinishedFailed` as a utility has been pigeon-holed into being specific about what it treats as a failure, and therefore it's usefulness is limited to cases that want the return value treated as such. For a more general-purpose utility, the API should not make decisions for the caller, unless they are very clear (what is there is not clear imo).
> 
> In my comment, I was implying that we should not treat cancellation as failures, but as cancellations. The function could return a type like:
> 
> ```nim
> type FinishedFutures[T] =
>   successes: openArray[Future[T]]
>   failures: openArray[Future[T]]
>   cancellations: openArray[Future[T]]
> 
> # borrowing ideas from marcin's impl of allFinishedValues
> proc allFinished*[T](futs: auto): Future[FinishedFutures[T]] {.async: (raises: []).} =
>   # wait for all futures to be either completed, failed or canceled
>   await allFutures(futs)
>   return FinishedFutures(
>     failures: futs.countIt(it.failed)
>     successes: futs.countIt(it.success)
>     cancellations: futs.countIt(it.cancelled)
>   )
> ```
> 
> So the example would change to:
> 
> ```nim
> let ensuringFutures = Iter[int].new(0 ..< manifest.blocksCount).mapIt(
>   self.networkStore.localStore.ensureExpiry(manifest.treeCid, it, expiry)
> )
> 
> let res = await allFinished[?!void](ensuringFutures)
> if res.failures.len > 0:
>   trace "Some blocks failed to update expiry", len = res.failure.len
>   return failure("Some blocks failed to update expiry (" & $res.failure.len & " )")
> if res.cancellations.len > 0:
>   trace "Block expiry update was cancelled in some blocks", len = res.cancellations.len
>   return failure("Block expiry update was cancelled in some blocks (" & $res.cancellations.len & " )")
> ```
> 
> But of course, if the caller wanted to treat cancellations as failures, the usage would change to:
> 
> ```nim
> let res = await allFinished[?!void](ensuringFutures)
> if res.failures.len > 0 or res.cancellations.len > 0:
>   trace "Some blocks failed to update expiry", len = res.failure.len
>   return failure("Some blocks failed to update expiry (" & $res.failure.len & " )")
> ```

Yes exactly. That would be a better solution. Sorry I didn't get that you were implying that.

PR #1444, opened 2026-06-10
refactor(release): `emptyDir` volumes are now used over PVCs, so no need to clean up

emizzle:
PVCs were not needed for the release dist tests. The PVCs were dropped in exchange for ephemeral `emptyDir` volumes that are tied to the lifetime of the pods. As such, we no longer need to clean up orphaned PVCs at the end of a Release workflow run.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -388,26 +388,6 @@ jobs:
         working-directory: ${{ env.TF_DIR }}
         run: terraform destroy -auto-approve
 
-      - name: Delete orphaned GCE disks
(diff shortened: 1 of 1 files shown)


gmega (approved):
LGTM

gmega:
Btw I suppose you could have multiple runs on the same cluster but this looks easier/simpler.

emizzle:
> Btw I suppose you could have multiple runs on the same cluster but this looks easier/simpler.

Yes, that is definitely possible, but this was much easier/faster

PR #1446, opened 2026-06-15
fix: memory issues in libstorage

2-towns:
This PR fixes several memory issues in libstorage:

1- `$req[]` was read after `deallocShared(req)`: get the error message before it is freed.
2- `deallocShared(req)` was used after the request was sent to the thread, which might lead to double-free issues: remove `deallocShared(req)` and
let the thread free it.
3- `reqContent` wasn't freed when the request wasn't sent to the thread: add a `destroy` method to free it.
4- `chunk` leak: the chunk is now a pointer and freed in `destroyShared`.
5- `peerAddresses` leak: it is now a pointer and freed in `destroyShared`.
6- context not destroyed when `storage_new` fails: destroy it on the error path.

Note that 4 and 5 leak on normal upload/connect paths, while the others are on error paths that are rarely triggered.


library/storage_thread_requests/requests/node_p2p_request.nim
```diff
@@ -32,11 +34,23 @@ proc createShared*(
...
+    for i in 0 ..< peerAddresses.len:
+      ret[].peerAddresses[i] = peerAddresses[i]
+
   return ret
 
-proc destroyShared(self: ptr NodeP2PRequest) =
+proc destroyShared*(self: ptr NodeP2PRequest) =
   deallocShared(self[].peerId)
+
+  if self[].peerAddresses != nil:
```
emizzle, line 54: Should this check `self[]` for a null reference first?

```suggestion
  if self[] != nil and self[].peerAddresses != nil:
```

emizzle (approved):
Nice, LGTM except possibly one extra null check

library/storage_thread_requests/requests/node_p2p_request.nim
2-towns: Nice !

gmega:
CI doesn't seem very happy

library/storage_thread_requests/requests/node_p2p_request.nim
2-towns: Actually this doesn't compile. If we want to guard better to use 

```nim
if self == nil
```

2-towns:
> CI doesn't seem very happy

All good now

library/storage_thread_requests/requests/node_p2p_request.nim
```diff
@@ -32,11 +34,26 @@ proc createShared*(
...
-  ret[].peerAddresses = peerAddresses
+  ret[].peerAddressesLen = peerAddresses.len
+
+  if peerAddresses.len > 0:
+    ret[].peerAddresses = cast[ptr UncheckedArray[cstring]](allocShared(
+      sizeof(cstring) * peerAddresses.len
+    ))
+    for i in 0 ..< peerAddresses.len:
+      ret[].peerAddresses[i] = peerAddresses[i]
+
```
gmega, line 45: Oh man we were sharing a sequence across threads? That's indeed no bueno :-)

gmega: Shouldn't you explicitly copy those strings? What is assignment gonna do here?

library/storage_thread_requests/requests/node_p2p_request.nim
```diff
@@ -88,7 +105,13 @@ proc process*(
 
   case self.operation
   of NodeP2PMsgType.CONNECT:
-    let res = (await connect(storage, self.peerId, self.peerAddresses))
+    # connect() takes seq[cstring]; self.peerAddresses is a raw shared array,
+    # so copy the pointers into a seq.
+    var peerAddresses = newSeq[cstring](self.peerAddressesLen)
+    for i in 0 ..< self.peerAddressesLen:
+      peerAddresses[i] = self.peerAddresses[i]
```
gmega, line 114: Again, shouldn't you explicitly copy those strings?

library/storage_thread_requests/requests/node_upload_request.nim
```diff
@@ -68,21 +70,30 @@ proc createShared*(
     op: NodeUploadMsgType,
     sessionId: cstring = "",
     filepath: cstring = "",
-    chunk: seq[byte] = @[],
+    chunkData: ptr byte = nil,
+    chunkLen: int = 0,
```
gmega, line 74: Yeah this is good, but I'm still worried about cstring...

gmega (changes requested):
This is definitely an improvement, though I still worry about the correctness of some of it. I'd in particular check if our use of `cstring` and assignments of `cstring`s spanning shared and local memory objects is safe.

My understanding is that under ORC this could be OK (no per-thread heap), but under refc I don't think it is.

2-towns:
> This is definitely an improvement, though I still worry about the correctness of some of it. I'd in particular check if our use of `cstring` and assignments of `cstring`s spanning shared and local memory objects is safe.
> 
> My understanding is that under ORC this could be OK (no per-thread heap), but under refc I don't think it is.

Okay I will look into it after our work is done to meet the dealdline.

library/storage_thread_requests/requests/node_p2p_request.nim
2-towns: Yes this is a mistake. I fixed it by calling `alloc` on each.

library/storage_thread_requests/requests/node_p2p_request.nim
2-towns: I don't think we should do it for this case, this is different from the code above. 

Above, the `peerAddresses` comes from a shared context so we need to copy it. 

In this code, `peerAddresses` is used internally. We need to loop over it to convert `ptr UncheckedArray[cstring]` to `seq[cstring]`.

library/storage_thread_requests/requests/node_upload_request.nim
2-towns: I think it is okay because we use `filepath.alloc()` to copy the `cstring`. 

From what I understand, `cstring` is like a `ptr char` with: 
1. `len` function 
2. string conversion using `$`

library/storage_thread_requests/requests/node_p2p_request.nim
gmega: OK, tracked down the code and figured it's fine. `alloc` is not a great name btw as it seems to be just allocating something, but this is actually cloning a cstring into shared memory

library/storage_thread_requests/requests/node_p2p_request.nim
gmega: Hm... `self.peerAddresses` is coming from shared memory though. You're still doing a plain assignment to a local object (which you then pass to a call) from an object that lives in shared memory. If this is not a copying assignment, you're basically pointing to memory that might get deallocated under a potentially different schedule.

I think in this case it should be fine because the shared object outlives the local object you're copying into, and because I believe this will be a copying assignment, but I'd always be wary of this.

library/storage_thread_requests/requests/node_upload_request.nim
gmega: OK, alloc is basically creating a shared-memory copy of this so it should be fine.

gmega (approved):
OK I think this is fine.

PR #1447, opened 2026-06-15
fix(blockexchange): bounds per-message work

emizzle:
Fixes: https://github.com/logos-co/audit-reports/issues/100

Accomplishes two goals:
1. Bounds cpu/disk work when handling messages by capping the number of WantList and BlockPresence entries to be decoded per message. Does not bound the size of the list that can cross the wire, just the number that can be decoded.
2. Ensures WantLists per peer is capped at 1, dropping subsequent WantList handling if one is already in flight for that peer.


cnanakos (approved):
LGTM

PR #1449, opened 2026-06-16
chore: update libp2p to 2.1.4

2-towns:


config.nims
```diff
@@ -140,7 +140,7 @@ switch("warning", "ObservableStores:off")
 # Too many false positives for "Warning: method has lock level <unknown>, but another method has 0 [LockLevel]"
 switch("warning", "LockLevel:off")
 
-switch("define", "libp2p_pki_schemes=secp256k1")
+switch("define", "libp2p_pki_schemes=secp256k1,rsa")
```
gmega: Hm... why are we enabling RSA and, most importantly, why didn't we have RSA enabled before?

storage/blockexchange/network/network.nim
```diff
@@ -314,8 +314,15 @@ proc new*(
   ## Create a new BlockExcNetwork instance
   ##
 
+  # libp2p now requires a non-nil handler at construction; the real one is set
+  # by self.init() below. This placeholder only exists until then.
+  proc placeholder(
```
gmega: This is a [null object](https://en.wikipedia.org/wiki/Null_object_pattern), so I'd probably call it `nullHandler`. But that's nitpicking.

gmega (commented):
This seems fine, I'm only concerned about adding RSA to libp2p_pki_schemes. Why do we need to do it and, most importantly, why were we not doing it before already?

gmega:
I'd also run this by @cnanakos - are we OK with upgrading to 2.1.0? Does it impact your end of things in any way?

cnanakos:
> I'd also run this by @cnanakos - are we OK with upgrading to 2.1.0? Does it impact your end of things in any way?

I don't think so, but I can't confirm until nim-libp2p-mix lifts its libp2p pin and gets validated. Once that's done and this PR lands, I can validate on my side.

gmega:
Btw @AlbertoSoutullo told me 2.1.0 is not cleared by DST yet (they found actual issues), so we have to hold off merging this (and #1417) until it is.

config.nims
2-towns: I removed it with 2.1.4

storage/blockexchange/network/network.nim
2-towns: Okay I did it

2-towns:
> Btw @AlbertoSoutullo told me 2.1.0 is not cleared by DST yet (they found actual issues), so we have to hold off merging this (and #1417) until it is.

2.1.4 was release and merged into `nim-libp2p-mix`

PR #1451, opened 2026-06-17
fix: `allFinishedFailed` -> `allDone` to retain cancellations

emizzle:
This PR replaces the `allFinishedFailed` proc with `allDone`, exposing in the return all the futures and their final state. This allows the caller to make a decision on what to do with the resulting futures. It does not make a decision as to what a "failed" or "successful" future looks like. 

The [recent change in the `allFinishedFailed` util](https://github.com/logos-storage/logos-storage-nim/commit/85d8822704cb517df18a3817ff7fb3dcc008b222), where cancelled futures were changed from being "successes" to "failures", means that cancellations in the underlying futures went from being swallowed to exposed as a failure. 

### Higher potential impact changes
During block expiry update, if one of the block expiry updates is cancelled, the `CancelledError` is now propagated, and `updateExpiry` will be cancelled, where previously (before 85d8822), a cancellation in an underlying future would have been swallowed. After 85d8822, this would have returned a failed result. However, raising and propagating a `CancelledError` seems more appropriate for the case, however I'd like a second opinion on it please. 

### Lower impact changes
`StorageServer.close()` and `StorageServer.stop()` both called `allFuturesFailed`. Previously (before 85d8822), this cancellation would be swallowed. After 85d8822, the cancellation would be considered "failed" and it would raise a `StorageError` exception. Now, cancellations in the underlying futures will raise a `CancelledError`.

> [!NOTE]
> `allDone` was used instead of `allFinished` to avoid clashing with nim-chronos. Happy to change it to something else if desired, eg `allFinishedFutures`.

### DIFF
--- a/storage/node.nim
+++ b/storage/node.nim
@@ -120,10 +120,15 @@ proc updateExpiry*(
         self.networkStore.localStore.ensureExpiry(manifest.treeCid, it, expiry)
       )
 
-    let res = await allFinishedFailed[?!void](ensuringFutures)
-    if res.failure.len > 0:
-      trace "Some blocks failed to update expiry", len = res.failure.len
-      return failure("Some blocks failed to update expiry (" & $res.failure.len & " )")
+    let res = await allDone[?!void](ensuringFutures)
+    let failed = res.completed.countIt(it.value.isErr) + res.failed.len
+    if failed > 0:
+      trace "Some blocks failed to update expiry", len = failed
+      return failure("Some blocks failed to update expiry (" & $failed & " )")
+    if res.cancelled.len > 0:
+      trace "Block expiry update was cancelled in some blocks",
+        len = res.cancelled.len
+      raise newException(CancelledError, "Block expiry update was cancelled in some blocks")
   except CancelledError as exc:
     raise exc
   except CatchableError as exc:
--- a/storage/node.nim
+++ b/storage/node.nim
@@ -125,9 +125,9 @@ proc updateExpiry*(
       trace "Some blocks failed to update expiry", len = res.failed.len
       return failure("Some blocks failed to update expiry (" & $res.failed.len & " )")
     if res.cancelled.len > 0:
-      trace "Block expiry update was cancelled in some blocks",
-        len = res.cancelled.len
-      raise newException(CancelledError, "Block expiry update was cancelled in some blocks")
+      trace "Block expiry update was cancelled in some blocks", len = res.cancelled.len
+      raise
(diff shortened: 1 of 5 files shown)


storage/node.nim
```diff
@@ -120,10 +120,15 @@ proc updateExpiry*(
         self.networkStore.localStore.ensureExpiry(manifest.treeCid, it, expiry)
       )
 
-    let res = await allFinishedFailed[?!void](ensuringFutures)
-    if res.failure.len > 0:
-      trace "Some blocks failed to update expiry", len = res.failure.len
-      return failure("Some blocks failed to update expiry (" & $res.failure.len & " )")
+    let res = await allDone[?!void](ensuringFutures)
+    let failed = res.completed.countIt(it.value.isErr) + res.failed.len
```
2-towns: Why do you count the it.value.isErr here and not in the other if of this PR ?

2-towns (approved):
LGTM

2-towns:
> Due to the [recent change in the `allFinishedFailed` util](https://github.com/logos-storage/logos-storage-nim/commit/85d8822704cb517df18a3817ff7fb3dcc008b222), where cancelled futures were changed from being "successes" to "failures", means that cancellations in the underlying futures would be swallowed.

Cancelled futures were already swallowed before the PR right  ?

emizzle:
> > Due to the [recent change in the `allFinishedFailed` util](https://github.com/logos-storage/logos-storage-nim/commit/85d8822704cb517df18a3817ff7fb3dcc008b222), where cancelled futures were changed from being "successes" to "failures", means that cancellations in the underlying futures would be swallowed.
> 
> Cancelled futures were already swallowed before the PR right ?

Actually, yes that is correct. I've updated the PR description.

tests/storage/testerrors.nim
```diff
@@ -0,0 +1,104 @@
...
+    fut1.complete()
+    fut2.complete()
+    fut3.complete()
+    let res = await allDone[void](@[fut1, fut2, fut3])
+    check res.completed.len == 3
+    check res.completed[0] == fut1
+    check res.completed[1] == fut2
+    check res.completed[2] == fut3
+
+  test "cancelling allDone does not cancel the awaited futures":
```
gmega, line 81: Is this desirable? If you're blocked on an `allDone` and you get cancelled, won't you want to propagate this to children?

gmega (approved):
Apart from the cancellation behavior for `allDone` which I'm not entirely convinced is the best, LGTM.

PR #1452, opened 2026-06-17
feat: run DHT queries over Mix

gmega:


storage/blockexchange/network/network.nim
```diff
@@ -258,9 +260,15 @@ proc dropPeer*(
   except CatchableError as error:
     warn "Error attempting to disconnect from peer", peer = peer, error = error.msg
 
+proc excludeRelays*(self: BlockExcNetwork, peers: openArray[PeerId]) =
+  for p in peers:
+    self.excludedPeers.incl(p)
+
```
gmega, line 266: Why do you exclude relays from the block protocol?

gmega: OK we had a chat and I understand it: we don't want to be sending block presence messages to relay nodes, and we currently do somewhat indiscriminately broadcast those.

storage/dht_proxy/protocol.nim
```diff
@@ -0,0 +1,133 @@
...
+    lo = 1
+    hi = providers.len
+  while lo < hi:
+    let
+      mid = (lo + hi + 1) div 2
+      test = LookupResponse(status: ResponseStatus.Ok, providers: providers[0 ..< mid])
+    if test.encode().len <= budget_bytes:
+      lo = mid
+    else:
+      hi = mid - 1
```
gmega, line 131: OK I guess this is quite generic and would work even with differently-sized records. Assuming fixed-sized records you can probably replace it with an integer division :-)

storage/rest/api.nim
```diff
@@ -574,6 +574,11 @@ proc initDebugApi(node: StorageNodeRef, conf: StorageConf, router: var RestRoute
         "repo": $conf.dataDir,
         "spr":
           if node.discovery.dhtRecord.isSome: node.discovery.dhtRecord.get.toURI else: "",
+        "announceSpr":
```
gmega: I think `announceSpr` might end up being ambiguous as the UDP SPRs are also "announced" in some sense. Maybe we just call it `providerRecord`?

gmega (commented):
LGTM but I can't approve it as I was the one opening the PR :-P

storage/rest/api.nim
cnanakos: Yes much better.

cnanakos (approved):
LGTM :)

PR #1460, opened 2026-06-19
feat(monitoring): daily heatlh check for bootstrap nodes

emizzle:
## Network presets
The network presets flag (`--network`) in the storage node binary now statically reads the fleet SPR values from `network_presets.json` in the root of this project. 
1. `network_presets.json` is statically read so that it will be embedded into the binary so there's no need to distribute this file alongside the binary.
2. The daily bootstrap health check workflow can read `network_presets.json` at runtime

> [!IMPORTANT]
> There is one caveat to this design: if there is a PR merged to master that updates `network_presets.json`, then the daily health check workflow will check the updated SPRs, while storage nodes out in the wild will still be using the old SPRs. This means the bootstrap health check may not reflect the current state of the latest release. This can probably be fixed with a branching strategy that hosts a bootstrap node health check running per maintained release.

## `check_spr` binary
This PR creates a `check_spr` binary that reads an SPR and reports if that SPR is diallable (for tcp) or pingable (for udp). The `check_spr` binary takes several args and options: 
```bash
Usage:
  check_spr [options]
  check_spr <spr-uri> [--timeout <secs>]

Options:
  --source <file>    Config file to read SPRs from (default: network_presets.json).
  --network <name>   Only probe the network preset with this name.
  --timeout <secs>   Per-node probe timeout in seconds (default: 10).
  --format <fmt>     Output format: "text" (default) or "json". "text" is the
                     human-readable table; "json" is a pretty-printed summary.
  --out <file>       Write the output to <file> instead of stdout. The content
                     is whichever --format is selected (text or json).
  --help, -h         Show this help and exit.

Arguments:
  <spr-uri>          A single "spr:" URI to probe instead of reading the config
                     file. Prints ALIVE/DEAD and exits with the matching status;
                     this mode ignores --format and --out.
```
## Daily health check workflow
A new workflow was created that runs daily at 06:00 UTC. It runs `make CI=true bootstrapHealthCheck` which builds the `check_spr` binary, then runs `check_spr --network logos.dev --network logos.test --format json --out build/bootstrap-health-report.json`. 

### Failed health check
If any bootstrap nodes (in logos.dev or logos.test fleets) fail to ping, the `check_spr` process exits with a failed exit code, and a GH issue is opened detailing the failed health check.

### Health check results
Health check results are parsed from the json output in `build/bootstrap-health-report.json` and written to a table in the workflow summary.
image

## Local testing
You can run this health check locally by running
```bash
make bootstrapHealthCheck
``` 
This will build the `check_spr` binary and run it with the default options (table format and print to stdout). Output will be similar to the following:
image

However, if you'd like to have more control over some of the `check_spr` options, you can build the `check_spr` binary first, then run it directly:
```bash
# build the binary
make checkSpr

# run the binary -- only check logos.dev fleet and output json to stdout
build/check_spr --network logos.dev --format json
```
Which will output something similar to this:
```bash
[
  {
    "network": "logos.dev",
    "peerId": "16Uiu2HAmDBdAio414PhsgaHYCZvw8gw6abfs6Aut18u8hjtbf1KF",
    "address": "/ip4/24.144.78.200/udp/9090",
    "alive": true,
    "reason": "discv5 pong received"
  },
  {
    "network": "logos.dev",
    "peerId": "16Uiu2HAmS5dmewGpS5MEenQpTHEdiN4hXhMazzEMvf5HRM5hgfFh",
    "address": "/ip4/34.42.230.59/udp/9090",
    "alive": true,
    "reason": "discv5 pong received"
  },
  {
    "network": "logos.dev",
    "peerId": "16Uiu2HAmV2BBvpNFZvHNuXL7S4zZByAc1c4JghVyL78cHqJmZ59Y",
    "address": "/ip4/47.238.229.111/udp/9090",
    "alive": true,
    "reason": "discv5 pong received"
  }
]
```


storage/presets.nim
```diff
@@ -52,70 +53,40 @@ proc `bootstrapNodes`*(self: NetworkPreset): seq[SignedPeerRecord] =
...
-        "BNlEb4J23tRHY3s86jPrTpkQj8_8wQYaCwoJBAXfEfiRAnVOGgsKCQQF3xH4kQJ1TipGMEQCIGWJMsF57N1iIEQgTH7I" &
-        "rVOgEgv0J2P2v3jvQr5Cjy-RAiAy4aiZ8QtyDvCfl_K_w6SyZ9csFGkRNTpirq_M_QNgKw",
-    ],
-  ),
-]
+# Bootstrap node SPRs live in a single source-of-truth config file at the repo
+# root. staticRead embeds it into the binary at build time, so the node remains
+# self-contained (the file is not needed alongside the binary at runtime). The
+# same file is read at runtime by the bootstrap liveness checker (tools/check_spr).
+const networkPresetsJson = staticRead("../network_presets.json")
```
gmega, line 60: Hm... OK I guess. JSON is definitely easier to edit than those multiline strings...

gmega (approved):
Excellent, LGTM.

PR #1462, opened 2026-06-19
fix(discovery): togglePrivateQueries refuses enable when mix not set up

cnanakos:


tests/cbindings/storage.c
```diff
@@ -869,9 +869,9 @@ int check_toggle_private_queries(void *storage_ctx)
     }
 
     int ret = is_resp_ok(r, &res);
-    if (res == NULL || strcmp(res, "false") != 0)
+    if (ret == RET_OK)
```
gmega, line 872: Fair, free(null) should be a no-op.

gmega (approved):
Great!

PR #1466, opened 2026-06-22
fix: harden linux release builds against GCC's -march=native bug on Zen4

emizzle:
Similar to https://github.com/status-im/nimbus-eth2/pull/7228/changes.

The linux-amd64 release build was intermittently failing with:

    'asm' operand has impossible constraints

in secp256k1's hand-written x86_64 assembly (scalar_4x64_impl.h, USE_ASM_X86_64 path). Root cause, confirmed against upstream (bitcoin-core/secp256k1#1623): on GCC 11 (Ubuntu 22.04's default), -march=native's CPU autodetection misidentifies early AMD Zen4 chips as znver3 while still enabling -mavx512f via CPUID feature probing. -march=znver3 -mavx512f is an invalid combination matching no real CPU -- GCC 11 predates Zen4 entirely -- and its register allocator cannot satisfy the asm's constraints under it. Upstream confirmed this is fixed by a GCC version that knows about znver4 (13+), and closed the report as a GCC autodetection bug, not a secp256k1 defect. This explains the apparent intermittency: it only reproduces on whichever physical Zen4 host GitHub's runner fleet happens to assign to a given job.

Two independent changes address this:

- Bump the linux-amd64 and linux-arm64 release builders from ubuntu-22.04(-arm) to ubuntu-24.04(-arm), kept on the same Ubuntu version across both architectures. Ubuntu 24.04 ships GCC 13 by default (GCC isn't installed separately in this workflow -- the runner image's preinstalled toolchain comes from its Ubuntu base), which correctly recognizes znver4 and produces a valid -march=native expansion.

- Also pass -d:disableMarchNative on the linux-amd64 build (matrix- scoped; arm64/macOS don't compile the affected x86_64 asm path). This is kept even after the GCC bump because -march=native is independently unsound for a publicly distributed binary: it bakes in whatever instruction set extensions the build runner's CPU happens to have (e.g. AVX-512) into the shipped binary, which would crash with an illegal instruction on end-user machines lacking that extension -- a separate, permanent concern from the GCC 11 bug.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -104,16 +104,22 @@ jobs:
           echo "storage_binary=${storage_binary}" >>$GITHUB_ENV
           echo "c_bindings_lib=${c_bindings_lib}" >>$GITHUB_ENV
 
-      - name: Build Logos Storage binary
+      - name: Compute platform Nim flags
+        id: platform-nim-flags
+        shell: bash
         run: |
-          nim_flags="${{ env.nim_flags }}"
+          platform_nim_flags=""
           # secp256k1's x86_64 inline asm can hit a GCC register-allocation
           # failure when -march=native expands against this runner's CPU, so
           # pin a fixed baseline instead. arm64/macOS don't use that asm path.
           if [[ "${{ matrix.os }}" == "linux" && "${{ matrix.cpu }}" == "amd64" ]]; then
-            nim_flags="${nim_flags} -d:disableMarchNative"
+            platform_nim_flags="-d:disableMarchNative"
           fi
-          make NIMFLAGS="--out:${{ env.build_dir }}/${{ env.storage_binary }} ${nim_flags}"
+          echo "nim_flags=${platform_nim_flags}" >> "$GITHUB_OUTPUT"
+
+      - name: Build Logos Storage binary
+        run: |
+          make NIMFLAGS="--out:${{ env.build_dir }}/${{ env.storage_binary }} ${{ env.nim_flags }} ${{ steps.platform-nim-flags.outputs.nim_flags }}"
 
       - name: Package ${{ env.storage_binary_base }} Linux (compress and preserve perms)
         if: matrix.os == 'linux'
@@ -160,20 +166,20 @@ jobs:
         if: matrix.os == 'linux'
         run: |
           make -j${ncpu} update
-          make -j${ncpu} NIMFLAGS="-d:disableMarchNative" libstorage
+          make -j${ncpu} NIMFLAGS="${{ steps.platform-nim-flags.outputs.nim_flags }}" libstorage
 
       - name: Build ${{ env.c_bindings_lib_base }} (MacOS)
         if: matrix.os == 'macos'
         run: |
           make -j${ncpu} update
-          STORAGE_LIB_PARAMS="--passL:\"-Wl,-install_name,@rpath/${{ env.c_bindings_lib_base }}.dylib\"" make -j${ncpu} NIMFLAGS="-d:disableMarchNative" libstorage
+          STORAGE_LIB_PARAMS="--passL:\"-Wl,-install_name,@rpath/${{ env.c_bindings_lib_base }}.dylib\"" make -j${ncpu} NIMFLAGS="${{ steps.platform-nim-flags.outputs.nim_flags }}" libstorage
 
       - name: Build ${{ env.c_bindings_lib_base }} (Windows)
         if: matrix.os == 'windows'
         shell: msys2 {0}
         run: |
           make -j${ncpu} update
-          make -j${ncpu} NIMFLAGS="-d:disableMarchNative" libstorage
+          make -j${ncpu} NIMFLAGS="${{ steps.platform-nim-flags.outputs.nim_flags }}" libstorage
 
       - name: Package ${{ env.c_bindings_lib_base }} Linux
         if: matrix.os == 'linux'
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -40,15 +40,15 @@ jobs:
     outputs:
       matrix: ${{ steps.matrix.outputs.matrix }}
     steps:
-      - name: Compute matrix
-        id: matrix
-        uses: fabiocaccamo/create-matrix-action@v5
-        with:
-          matrix: |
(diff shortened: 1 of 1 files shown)


marcinczenko (approved):
Nice. This appears to also fix the issues that I had on my local machine.

My machine is:

```
Laptop 13 running Arch Linux kernel 7.0.9 on an AMD Ryzen AI 9 HX 370 / Radeon 890M, GCC 16.1.1, 64 GiB RAM.
```

and on that machine when building libstorage, I normally get the following error:

```bash
In file included from /home/mc2/code/logos-storage/logos-storage-nim/vendor/nim-secp256k1/secp256k1/../vendor/secp256k1/src/scalar_impl.h:20,
                 from /home/mc2/code/logos-storage/logos-storage-nim/vendor/nim-secp256k1/secp256k1/../vendor/secp256k1/src/secp256k1.c:28:
In function ‘secp256k1_scalar_reduce_512’,
    inlined from ‘secp256k1_scalar_mul’ at /home/mc2/code/logos-storage/logos-storage-nim/vendor/nim-secp256k1/secp256k1/../vendor/secp256k1/src/scalar_4x64_impl.h:863:5:
/home/mc2/code/logos-storage/logos-storage-nim/vendor/nim-secp256k1/secp256k1/../vendor/secp256k1/src/scalar_4x64_impl.h:356:5: error: ‘asm’ operand has impossible constraints or there are not enough registers
  356 |     __asm__ __volatile__(
      |     ^~~~~~~
/home/mc2/code/logos-storage/logos-storage-nim/vendor/nim-secp256k1/secp256k1/../vendor/secp256k1/src/scalar_4x64_impl.h:470:5: error: ‘asm’ operand has impossible constraints or there are not enough registers
  470 |     __asm__ __volatile__(
      |     ^~~~~~~
```

My quick and dirty fix was to manually adjust `vendor/nim-secp256k1/secp256k1/abi.nim` so that it uses `-DUSE_SCALAR_GENERIC`  disabling that asm path entirely:

```nim
when defined(amd64) and (defined(gcc) or defined(clang)):
  # const asmFlags = " -DUSE_ASM_X86_64"
  # Disable x86_64 inline assembly on modern AMD (Zen 4+/Ryzen AI)
  const asmFlags = " -DUSE_SCALAR_GENERIC"
else:
  const asmFlags = ""
```

The fix proposed in this PR (`disableMarchNative`) appears to fix my problem better.

The question is though: should we not also apply the same fix to the libstorage?
Actually, on my machine I hit the failing build only on `libstorage` target and the main build target somehow survives.

We can consider what follows:

1) Adjust the following in the same file:

```
- name: Build ${{ env.c_bindings_lib_base }} (Linux)
  if: matrix.os == 'linux'
  run: |
    make -j${ncpu} update
    nim_flags="${{ env.nim_flags }}"
    if [[ "${{ matrix.cpu }}" == "amd64" ]]; then
      nim_flags="${nim_flags} -d:disableMarchNative"
    fi
    make -j${ncpu} libstorage NIMFLAGS="${nim_flags}"
``` 

Since that step already has `if: matrix.os == 'linux'`, checking only `matrix.cpu == "amd64"` inside it is enough.
A cleaner variant would be to compute the adjusted Nim flags once before both build steps and reuse them.

(2) OR, perhaps we should also adjust the repo's build instructions to tell the contributors with the affected CPUs to use `NIMFLAGS="-d:disableMarchNative"` when building locally, e.g.:

```bash
make -j24 NIMFLAGS="-d:disableMarchNative"
make -j24 libstorage NIMFLAGS="-d:disableMarchNative"
```

We could also consider fixing it at the config.nims level or in the `Makefile` and then maybe we do not have to change CI at all?

If this makes sense, we can apply the fix in this PR, or I can create a separate PR (but we should probably discuss which option is better in the end: (1) CI, (2) config.nims or Makefile).

emizzle:
Good point @marcinczenko, I'll add the -d:disableMarchNative to the libstorage build 👍

emizzle:
@marcinczenko I pushed a commit for libstorage. Could you check if it works for you, please?

.github/workflows/release.yml
```diff
@@ -153,20 +160,20 @@ jobs:
...
           make -j${ncpu} update
-          make -j${ncpu} libstorage
+          make -j${ncpu} NIMFLAGS="-d:disableMarchNative" libstorage
 
       - name: Build ${{ env.c_bindings_lib_base }} (MacOS)
         if: matrix.os == 'macos'
         run: |
           make -j${ncpu} update
-          STORAGE_LIB_PARAMS="--passL:\"-Wl,-install_name,@rpath/${{ env.c_bindings_lib_base }}.dylib\"" make -j${ncpu} libstorage
+          STORAGE_LIB_PARAMS="--passL:\"-Wl,-install_name,@rpath/${{ env.c_bindings_lib_base }}.dylib\"" make -j${ncpu} NIMFLAGS="-d:disableMarchNative" libstorage
```
marcinczenko: I think that applying NIMFLAGS="-d:disableMarchNative" unconditionally to all libstorage builds is not ideal.

Why:
- The original binary fix is scoped to matrix.os == "linux" && matrix.cpu == "amd64".
- The failure is specifically -march=native + secp256k1 x86_64 asm on Linux amd64.
- macOS arm64 does not use that x86_64 asm path.
- Windows already avoids -march=native via config.nims and uses -march=x86-64; passing -d:disableMarchNative changes that to -mssse3, which is unnecessary and changes the baseline.
- Linux arm64 also does not need it (I think).

I created a PR #1467 on top of your PR, with the suggested changes.

marcinczenko (changes requested):
I think that applying NIMFLAGS="-d:disableMarchNative" unconditionally to all libstorage builds is not ideal.

Why:
- The original binary fix is scoped to matrix.os == "linux" && matrix.cpu == "amd64".
- The failure is specifically -march=native + secp256k1 x86_64 asm on Linux amd64.
- macOS arm64 does not use that x86_64 asm path.
- Windows already avoids -march=native via config.nims and uses -march=x86-64; passing -d:disableMarchNative changes that to -mssse3, which is unnecessary and changes the baseline.
- Linux arm64 also does not need it (I think).

I created a PR #1467 on top of your PR, with the suggested changes.

.github/workflows/release.yml
marcinczenko: Merged.

marcinczenko (approved):
#1467 resolves my requested changes.

PR #1474, opened 2026-06-25
ci: publish openapi.yaml as a release asset

emizzle:
Attach openapi.yaml to tagged GitHub releases so external consumers (the Python SDK, docs site, third parties) have a stable per-version URL for the API spec.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -424,6 +424,9 @@ jobs:
     needs: [build, release-tests]
     if: needs.build.result == 'success' && needs.release-tests.result == 'success'
     steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+
       - name: Set conditional env variables
         shell: bash
         run: |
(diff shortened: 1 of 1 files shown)


2-towns (approved):

gmega:
I spoke to @emizzle yesterday and we will probably not be merging this.

emizzle:
I believe there's still validity in publishing an openapi spec as part of a release process. If there are downstream applications that consume the spec and build clients on it, for example, they may want to pin their codebase to a specific spec version. But at this point, the only downstream applications are the dist tests and the python sdk (which is more generally out of date). For the former, we agreed to keep the openapi spec in the codebase, effectively pinning it by copy. For the python SDK, the release workflow should keeping this in sync (but it needs some love and is not on the critical path atm).

emizzle:
Closing this as it's not needed now. This change can easily resurrected in the future if needed.

PR #1480, opened 2026-07-01
chore: require bash 4.0 for storage-config.sh

emizzle:
Macos ships with bash 3.2 due to bash licensing that changed with 4.0 (GPLv3). This script uses features available in bash 4+ (eg associative arrays). When bash <4 is used, the following error is reported:
```bash
 ./tools/scripts/storage-config.sh: line 8: declare: -A: invalid option
  declare: usage: declare [-afFirtx] [-p] [name[=value] ...]
```

This PR adds an explicit `require bash 4` in the script, which produces an error if the dependency version is not met:
```bash
[storage_config] Error: bash version 4+ is required, but found version 3
```

### DIFF
--- a/tools/scripts/storage-config.sh
+++ b/tools/scripts/storage-config.sh
@@ -4,11 +4,6 @@
 
 set -euo pipefail
 
-# Valid networks
-declare -A _networks
-#_networks=([test]="Logos testnet" [dev]="Logos devnet")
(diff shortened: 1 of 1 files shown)


gmega:
I ended up implementing the simple version guard as part of #1479. I also dropped `require` altogether, as it started looking like overkill for checking a single command.

PR #1481, opened 2026-07-01
feat: write network presets to disk

emizzle:
Modifies the `storage-config.sh` bash script `presets` command to take an optional flag, `--out`, that writes the network presets to `$root/network-config.sh`, preventing copy and paste mistakes:
```bash
./storage-config.sh presets --out
```
If `--out` is omitted, the network presets are only written to stdout. With `--out`, the presets are written to both stdout and the `network-presets.json` file to keep convention with the other commands.

This could be used in the release process if desired.

- ensures the json is validated and pretty printed
- updates `network-presets.json` to the latest version


gmega:
I don't feel this is necessary. You can just do:

```bash
storage-config.bash > output.json
```

which is btw what the Makefile does. No copy-pasting involved.

PR #1485, opened 2026-07-03
refactor(protobuf): migrate to protobuf serde

cnanakos:


arnetheduck:
~~nevermind, misread the PR~~

storage/blockexchange/protocol/message.nim
```diff
@@ -1,193 +1,108 @@
...
-    sublist: seq[seq[byte]]
-  if ?pb.getField(1, ipb):
-    value.wantList = ?WantList.decode(ipb)
-  if ?pb.getRepeatedField(4, sublist):
-    if sublist.len > MaxBlockPresenceEntries:
-      return err(ProtoError.BufferOverflow)
-    for item in sublist:
-      value.blockPresences.add(?BlockPresence.decode(initProtoBuffer(item)))
-  ok(value)
+  Range* {.proto2.} = object
```
emizzle: Although there is no clash due to the [identifier equality rules](https://nim-lang.org/docs/manual.html#lexical-analysis-identifier-equality) that require case sensitivity for the first letter of an identifier, `Range` is very close to system.nim's built-in `range` type. It might be a good idea to use something a bit more specific, like `WindowRange` instead.

storage/blockexchange/engine/downloadcontext.nim
```diff
@@ -156,7 +155,7 @@ proc addPendingRange(
 ) =
   case t.policy
   of spRandomWindow:
-    t.pendingRanges.add(range)
+    t.pendingRanges.add(Range(start: range.start, count: range.count))
```
emizzle: Why not pass a `Range` type into this function instead? So then you'd have:
```nim
proc addPendingRange(
    t: var BroadcastAvailabilityTracker, r: Range
) =
  case t.policy
  of spRandomWindow:
    t.pendingRanges.add(r)
```

storage/blockexchange/protocol/message.nim
```diff
@@ -1,193 +1,108 @@
...
+    )
+  var val = default(BlockPresence)
+  if stream.readFieldInto(val, header, ProtoType):
+    value.add move(val)
+    true
+  else:
+    false
+
+Protobuf.serializerForResult([Message])
+
```
gmega: I was a bit confused by this. Why generate encode/decode and then wrap those with protobufEncode/protobufDecode manually? Why not just generate the latter directly?

At first I was looking at this code and thinking you were calling nim-protobuf-serialization's `decode`, but then I realized your macro generates a custom helper which already returns result.

storage/manifest/coders.nim
```diff
@@ -7,144 +7,81 @@
...
+##   Message Manifest {
+##     required uint32 manifestVersion = 1; # manifest format version
+##     optional bytes treeCid = 2;          # cid (root) of the tree
+##     optional uint32 blockSize = 3;       # size of a single block
+##     optional uint64 datasetSize = 4;     # size of the dataset
+##     optional codec: MultiCodec = 5;      # Dataset codec
+##     optional hcodec: MultiCodec = 6;     # Multihash codec
+##     optional version: CidVersion = 7;    # Cid version
+##     optional filename: string = 8;       # original filename
+##     optional mimetype: string = 9;       # original mimetype
```
gmega, line 22: Why does it say that those are optional? I think the only really optional ones are filename and mimetype?

storage/manifest/coders.nim
```diff
@@ -7,144 +7,81 @@
...
+proc decodeManifestFields(
+    fields: seq[byte]
+): Manifest {.raises: [SerializationError].} =
+  decode(Protobuf, fields, Manifest)
 
-  return pbNode.buffer.success
+proc encode*(manifest: Manifest): ?!seq[byte] =
+  var output = memoryOutput()
+  try:
+    writeField(output, 1, encodeManifestFields(manifest), pbytes, false)
```
gmega: Hm... why do we do this?

storage/manifest/coders.nim
```diff
@@ -7,144 +7,81 @@
...
-  ##
-
+  try:
+    var input = memoryInput(data)
+    if not input.readable():
+      return failure("Empty manifest input")
+
+    let outer = input.readHeader()
+    if outer.number() != 1:
+      return failure("Unexpected field number: " & $outer.number())
```
gmega: OK this indeed looks weird - seems like we're wrapping the Manifest into a single-field wrapper type and serializing/de-serializing it "manually" (without ever explicitly declaring the wrapper type) for no good reason.

storage/utils/protobuf/option.nim
```diff
@@ -0,0 +1,59 @@
...
+
+{.push raises: [].}
+
+import std/options
+
+import pkg/faststreams
+import pkg/protobuf_serialization
+import pkg/protobuf_serialization/codec
+
+Protobuf.extensionDefaults(Option[string])
```
gmega, line 18: Ugh OK... you had to define this yourself.

storage/utils/protobuf/refobject.nim
```diff
@@ -0,0 +1,25 @@
...
+##  * MIT license ([LICENSE-MIT](LICENSE-MIT))
+## at your option.
+## This file may not be copied, modified, or distributed except according to
+## those terms.
+
+{.push raises: [].}
+
+import pkg/protobuf_serialization
+
+proc readValue*[T: ref object](
```
gmega, line 14: I'm surprised those are not supported out-of-the-box. I think some of these things might even make sense to upstream?

storage/utils/protobuf/serializer.nim
```diff
@@ -0,0 +1,39 @@
...
+type ProtobufDecodeError* = object
+  msg*: string
+
+proc `$`*(e: ProtobufDecodeError): string =
+  e.msg
+
+macro serializerForResult*(_: type Protobuf, Types: untyped): untyped =
+  var stmts = newStmtList()
+  for T in Types:
+    let helperName = ident("decode" & $T)
```
gmega: Maybe use `genSym` to avoid clashes?

tests/storage/blockexchange/protocol/testwireformat.nim
```diff
@@ -0,0 +1,634 @@
...
+    ipb.finish()
+    pb.write(field, ipb)
+
+  proc protobufEncode*(value: Message): seq[byte] =
+    var ipb = initProtoBuffer()
+    ipb.write(1, value.wantList)
+    for v in value.blockPresences:
+      ipb.write(4, v)
+    ipb.finish()
+    ipb.buffer
```
gmega, line 636: Nice, so the idea is that this works as a type of compatibility record for the wire format. I suppose that updating those records has to be done manually?

gmega (approved):
OK got a few questions here and there but overall LGTM.

storage/manifest/coders.nim
cnanakos: I didn't touch this comment, I just moved it upwards in the file.

cnanakos: OK synced to reality.

emizzle (commented):
Nice job! Just a few cosmetic comments...

storage/manifest/coders.nim
cnanakos: This is the legacy dag-pb envelope and we keep the exact bytes for backward compatibility, and it's hand-rolled because Manifest is a ref object the lib won't serialize as nested field.

cnanakos: To make it more clear we can wrap it to a manifest envelope and drop the hand-rolled approach.

storage/blockexchange/protocol/message.nim
cnanakos: Yes I suppose we can remove the shims now.

storage/utils/protobuf/serializer.nim
cnanakos: Nice catch, done!

tests/storage/blockexchange/protocol/testwireformat.nim
cnanakos: Yes correct, the byte sequences are updated manually.

storage/blockexchange/engine/downloadcontext.nim
cnanakos: The `Range` type exists only because the protobuf library can't annotate anonymous nested tuples — it was introduced for serialization. Unifying the engine's `tuple[start, count]` onto it is a fair idea, but it's an orthogonal refactor: it breaks tuple destructuring across scheduler/engine/activedownload/downloadmanager and their tests, and it would couple the engine's internals to the wire schema. Let's think about it and take care of it on another PR please.

storage/blockexchange/protocol/message.nim
cnanakos: Good catch, let's use the more generic `IndexRange` which also fits the engine's internals.

PR #1486, opened 2026-07-07
fix: crash on upnp restart

2-towns:
The PR fixes a crash when restarting using UPnP (enabled on router).

It was reproduced by Giuliano during the tests and was [reported recently](https://github.com/logos-co/logos-storage-module/issues/71) but we didn't know the root cause. 

Steps: 

1. Storage Module creates a context which lives inside a Nim thread. 
2. When UPnP is available, `getExternalIp` will be called and save inside `extIp`. During that call, it initialise the `upnp` variable which is a `threadvar`. Here is the definition from Nim manual: 

> A variable can be marked with the threadvar pragma, which makes it a thread-local variable; Additionally, this implies all the effects of the global pragma.

3. When `stop` then `destroy` are called in the Storage Module, the thread is destroyed, so from the definition above, the `upnp` variable isn't initialised anymore because it was living into the thread. 
4. When calling  `init` and `start` again, `getExternalIp`  wasn't call anymore because `extIp` still hold the previous ip value. The result is `upnp` is not initalised, it has no value, So when `doPortMapping` is called, it tries to call `addPortMapping` on a nil object which lead to a segfault. 

This PR fixes this issue by calling `getExternalIp` every time `setupNap` is called. Additionnaly, it adds cleanup step on node `stop` function. 

> [!NOTE]  
> This is is gonna to be fixed by [NAT Traversal PR](https://github.com/logos-storage/logos-storage-nim/pull/1417) because nim-libplum is introduced and threads a managed internally by libplum.

Fixes https://github.com/logos-co/logos-storage-module/issues/71

### DIFF
--- a/storage/nat.nim
+++ b/storage/nat.nim
@@ -9,7 +9,7 @@
 {.push raises: [].}
 
 import
-  std/[options, os, times, net, atomics, exitprocs],
+  std/[options, os, times, net, atomics],
   nat_traversal/[miniupnpc, natpmp],
   json_serialization/std/net,
   results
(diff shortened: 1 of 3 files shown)


cnanakos (approved):
LGTM

gmega:
OK this explains why I was the only one seeing this: I guess I'm the only one who used UPnP. :-)

gmega (approved):
Yet another instance of a buggy implementation of restarts because of global vars...

PR #1487, opened 2026-07-09
chore: better shutdown

2-towns:
This PR introduces a better shutdown process.

### Context

1. In `storage.nim`, we call [`build`](https://github.com/logos-storage/logos-storage-nim/blob/1302a4c0a35e2a212a6b65b4acc6cb1ee462378e/storage/storage.nim#L229-L241) on the `SwitchBuilder` when creating a new `StorageServer`.
2. The [`connManager` is initialised in this `build` function](https://github.com/vacp2p/nim-libp2p/blob/c43199378f46d0aaf61be1cad1ee1d63e8f665d6/libp2p/builders.nim#L469).
3. When we stop the switch, the [`connManager`](https://github.com/vacp2p/nim-libp2p/blob/c43199378f46d0aaf61be1cad1ee1d63e8f665d6/libp2p/switch.nim#L327) is closed.
4. But it is never restarted in the [`start` function](https://github.com/vacp2p/nim-libp2p/blob/c43199378f46d0aaf61be1cad1ee1d63e8f665d6/libp2p/switch.nim#L343-L376).

Moreover, looking into the tests, I didn't find any usage of a switch being restarted. So we cannot support a `restart` process for our node.

### Changes

For the reason given above, this PR introduces a `shutdown` function that stops and closes the resources. It is unified into a single function so as not to suggest that the node can be restarted.

On the `libstorage` side, the `close` function is deprecated and is now a `no-op`, and the `stop` operation calls `shutdown`. Additionally, a new function `boot` is introduced as a shortcut for `create` + `start`.

This PR introduces no breaking changes.

### DIFF
--- a/library/README.md
+++ b/library/README.md
@@ -105,10 +105,9 @@ Typical usage:
 
 ---
 
-> **Restart is not supported.**.
-> `storage_stop` to stop the node and free resource.
-> Use `storage_boot` to start a new instance with a clean state or `storage_new`
-> to create a new instance.
+> **Restart is not supported.** Use `storage_stop` to stop the node and free its
+> resources. To run again, create a fresh instance with `storage_boot` (or
+> `storage_new` + `storage_start`).
 
 ### `storage_start`
 
--- a/library/README.md
+++ b/library/README.md
@@ -143,7 +143,7 @@ int storage_stop(void *ctx, StorageCallback callback, void *userData);
 
 ### `storage_close`
 
-**Deprecated — no-op.** Does nothing, kept for ABI compatibility. Use `storage_stop` to tear
+**Deprecated — no-op.** Does nothing, kept for compatibility. Use `storage_stop` to tear
 the node down before `storage_destroy`.
 
 ```c
--- a/library/README.md
+++ b/library/README.md
@@ -2,10 +2,10 @@
 
 Logos Storage exposes a C binding that serves as a stable contract, making it straightforward to integrate Logos Storage into other languages such as Go.
 
-The implementation was inspired by [nim-library-template](https://github.com/logos-co/nim-library-template)  
+The implementation was inspired by [nim-library-template](https://github.com/logos-co/nim-library-template)
 and by the [nwaku](https://github.com/waku-org/nwaku/tree/master/library) library.
 
-The source code contains detailed comments to explain the threading and callback flow.  
+The source code contains detailed comments to explain the threading and callback flow.
 The diagram below summarizes the lifecycle: context creation, request execution, and shutdown.
 
 ```mermaid
@@ -26,7 +26,7 @@ sequenceDiagram
     Thr->>Ctx: dequeue request
     Thr-->>Ctx: ACK
     Ctx-->>C: forward ACK
-    C-->>Go: RET OK 
+    C-->>Go: RET OK
     Go->>App: Unblock
     Thr->>Eng: execute (async)
     Eng-->>Thr: result ready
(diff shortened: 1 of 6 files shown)


gmega (approved):
I think it's perfectly acceptable that we just recreate the node from scratch on restart. If we ever need to do it differently, we can come up with a different factorization for the initialization procedure that preserves what needs to be preserved, and throws away/rebuilds what doesn't.

Maybe `storage_boot` could be called `storage_new_start` to reflect more readily it's just a combination of new + start instead of an entirely different thing.

emizzle (commented):
My suggestion here would be to have two paths for public-facing APIs. Each path uses antonym verbs (create/destroy, start/stop, boot/shutdown). One path is the "automatic" path that creates and destroys context for the user and is not a reusable instance. This is probably the most likely use case. Eg
```
storage_boot > storage_shutdown
```
The "manual" path would create the context, but allow the user to start and stop a reusable instance, then finally destroy the context:
```
storage_create > storage_start > storage_stop > storage_destroy
```
Note: "create" is used here instead of "new", as there is not an obvious antonym for "new", whereas "destroy" is an obvious antonym of "create".

Alternatively, as is inline with @gmega's suggestion, use the "automatic" public-facing APIs only, removing the "manual" path. In that case, since the manual methods are no longer exposed, we can use whatever antonym pairs we'd like, eg start/stop, create/destroy, boot/shutdown.

Wdyt?

2-towns:
> My suggestion here would be to have two paths for public-facing APIs. Each path uses antonym verbs (create/destroy, start/stop, boot/shutdown). One path is the "automatic" path that creates and destroys context for the user and is not a reusable instance. This is probably the most likely use case. Eg
> 
> ```
> storage_boot > storage_shutdown
> ```
> 
> The "manual" path would create the context, but allow the user to start and stop a reusable instance, then finally destroy the context:
> 
> ```
> storage_create > storage_start > storage_stop > storage_destroy
> ```
> 
> Note: "create" is used here instead of "new", as there is not an obvious antonym for "new", whereas "destroy" is an obvious antonym of "create".
> 
> Alternatively, as is inline with @gmega's suggestion, use the "automatic" public-facing APIs only, removing the "manual" path. In that case, since the manual methods are no longer exposed, we can use whatever antonym pairs we'd like, eg start/stop, create/destroy, boot/shutdown.
> 
> Wdyt?

This is difficult to have a `shutdown` only method for 2 reasons: 

1. `stop` is event based (non blocking) 
2. You need to destroy the context in the same call 

If you make `stop` blocking, it is already much easier. 

It is also much easier to provide this function in the Logos Storage Module or any bindings: the `shutdown` function would call stop + destroy and this is transparent for the consumer of the bindings. So I even doubting now that we need to introduce a `boot` or `new_start` function here, while this is easier in the bindings and `libstorage` keeps providing primitives only.

emizzle:
> If you make stop blocking, it is already much easier.

What are the trade offs for making `stop` blocking?

> while this is easier in the bindings and libstorage keeps providing primitives only

Yeah, i agree this makes sense. In that case, we could still use this PR to align verbage, eg:
```
storage_create > storage_start > storage_stop > storage_destroy
```
But also, I understand that may introduces breaking changes for little (public-facing) gain.

2-towns:
> What are the trade offs for making stop blocking?

In the context of the UI, if you make it blocking, that might freeze the UI as well, because it will wait for the return of the call to `stop`. 

> Yeah, i agree this makes sense. In that case, we could still use this PR to align verbage, eg:
> storage_create > storage_start > storage_stop > storage_destroy
> But also, I understand that may introduces breaking changes for little (public-facing) gain.

Maybe we can introduce alias so it doesn't break anything.

gmega:
Any reason we should not merge this @2-towns @emizzle?

emizzle:
@gmega I thought maybe this was an unneeded change due to @2-towns [comment](https://github.com/logos-storage/logos-storage-nim/pull/1487#issuecomment-4979569975): "So I even doubting now that we need to introduce a boot or new_start function here, while this is easier in the bindings and libstorage keeps providing primitives only."

2-towns:
Yes, I think we can revisit this later. For now it's low priority because of our other ongoing work.

PR #1488, opened 2026-07-10
chore: add static libstorage to release artifacts

2-towns:
This PR add static libstorage artifacts to release artififacts. 
It is useful for Rust bindings and could replace [this](https://github.com/nipsysdev/logos-storage-nim-bin).


emizzle (approved):
LGTM

PR #1490, opened 2026-07-23
feat: add dataset usage by type

2-towns:
## Description

This PR enhances the `space` API to add data usage by type. This can be seen as an experimental feature and it is needed for the Storage UI (Disk widget).

## Problem

The problem we currently have with the repo store is that we cannot easily access this information: data is stored as raw blocks, and the repo store keeps no per-type breakdown. The type lives in the manifest, so the usage has to be reconstructed from there.

## Changes

A new method `datasetUsageByType` is introduced. It loops over the manifests and checks the first and last blocks. If they are present, the data is considered part of the repo storage and added to the data usage returned by the function.

We don't want to loop over all the blocks because the API could be very costly in terms of performance.

It has some trade-offs. For example, an aborted dataset may have its first and last block in the repo store while missing blocks in the middle, so it would still be counted in the usage.

But this seems good enough as an `experimental` feature and showcases the space usage by type in the UI.

### DIFF
--- a/storage/rest/api.nim
+++ b/storage/rest/api.nim
@@ -12,7 +12,6 @@
 import std/sequtils
 import std/mimetypes
 import std/os
-import std/tables
 
 import pkg/questionable
 import pkg/questionable/results
--- a/storage/rest/api.nim
+++ b/storage/rest/api.nim
@@ -12,6 +12,7 @@
 import std/sequtils
 import std/mimetypes
 import std/os
+import std/tables
 
 import pkg/questionable
 import pkg/questionable/results
@@ -469,11 +470,23 @@ proc initDataApi(node: StorageNodeRef, repoStore: RepoStore, router: var RestRou
     return RestApiResponse.response($json, contentType = "application/json")
 
   router.api(MethodGet, "/api/storage/v1/space") do() -> RestApiResponse:
+    without usage =? (await node.datasetUsageByType()), err:
+      return RestApiResponse.error(
+        Http500, "Error when trying to get dataset usage: " & err.msg
+      )
+
+    var usageList: seq[RestDatasetTypeUsage]
+    for mimetype, entry in usage:
+      usageList.add RestDatasetTypeUsage(
+        mimetype: mimetype, bytesLocal: entry.bytesLocal, count: entry.count
+      )
+
     let json = %RestRepoStore(
       totalBlocks: repoStore.totalBlocks,
       quotaMaxBytes: repoStore.quotaMaxBytes,
       quotaUsedBytes: repoStore.quotaUsedBytes,
       quotaReservedBytes: repoStore.quotaReservedBytes,
+      usage: usageList,
     )
     return RestApiResponse.response($json, contentType = "application/json")
 
--- a/library/libstorage.h
+++ b/library/libstorage.h
@@ -321,7 +321,10 @@ extern "C"
     //  "totalBlocks": 100000,
     //  "quotaMaxBytes": 0,
     //  "quotaUsedBytes": 0,
-    //  "quotaReservedBytes": 0
+    //  "quotaReservedBytes": 0,
+    //  "usage": [
+    //   { "mimetype": "video/mp4", "bytesLocal": 1536, "count": 2 }
+    //  ]
     // }
     int storage_space(
         void *ctx,
--- a/library/storage_thread_requests/requests/node_storage_request.nim
+++ b/library/storage_thread_requests/requests/node_storage_request.nim
@@ -8,8 +8,6 @@
 ## - SPACE: get the amount of space used by the local node.
 ## - EXISTS: check the existence of a cid in a node (local store).
 
-import std/tables
-
 import chronos
 import chronicles
 import serde/json as serde
--- a/library/storage_thread_requests/requests/node_storage_request.nim
+++ b/library/storage_thread_requests/requests/node_storage_request.nim
@@ -8,6 +8,8 @@
 ## - SPACE: get the amount of space used by the local node.
 ## - EXISTS: check the existence of a cid in a node (local store).
 
+import std/tables
+
 import chronos
 import chronicles
 import serde/json as serde
@@ -18,7 +20,8 @@ import ../../../storage/stores/repostore
 
 from ../../../storage/storage import StorageServer, node, repoStore
 from ../../../storage/node import
-  iterateManifests, fetchManifest, fetchDatasetAsyncTask, delete, hasLocalBlock
+  iterateManifests, fetchManifest, fetchDatasetAsyncTask, delete, hasLocalBlock,
+  datasetUsageByType, DatasetTypeUsage
 from libp2p import Cid, init, `$`
 
 logScope:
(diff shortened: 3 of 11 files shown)


storage/rest/api.nim
```diff
@@ -469,11 +469,23 @@ proc initDataApi(node: StorageNodeRef, repoStore: RepoStore, router: var RestRou
...
+      return RestApiResponse.error(
+        Http500, "Error when trying to get dataset usage: " & err.msg
+      )
+
+    var usageList: seq[RestDatasetTypeUsage]
+    for mimetype, entry in usage:
+      usageList.add RestDatasetTypeUsage(
+        mimetype: mimetype, bytesLocal: entry.bytesLocal, count: entry.count
+      )
+
```
cnanakos, line 482: Main thing here is complexity. This turns `/space` from an `O(1)` read of `quotaUsedBytes` into an `O(D)` scan over every dataset: `listBlocks(Manifest)` walks the full manifest keyspace, and each dataset costs a `getBlock` + protobuf `decode` + two `hasBlock` calls. The first/last block check avoids `O(total blocks)` but not `O(datasets)`, and a node is meant to hold many of these. Since it backs the UI it'll likely be polled, which makes it `O(D)` per request with no cache. I'd suggest splitting it off `/space` into its own endpoint, e.g. `/space/usage-by-type`, so the hot `O(1)` path stays untouched and the breakdown is opt-in.

It also reports the wrong number for occupied storage. It sums the logical size, but blocks are deduplicated and refcounted, so shared datasets overcount and the total won't match real occupied storage (`quotaUsedBytes`).

Given it's inaccurate, it's worth questioning whether this breakdown really belongs on the storage node at all. If the UI can derive it from data it already fetches, that keeps `/space` untouched; if it does need to be served by the endpoint, a per-type usage counter updated at the dataset lifecycle — upload/download/delete, where the mimetype is actually known — would make it an `O(1)` read instead of a per-request scan.

One more thing on `datasetUsageByType`: it returns a `Table[string, DatasetTypeUsage]` keyed directly on the raw `manifest.mimetype`. That's validated against Nim's known MIME set so it's bounded, but still potentially a few hundred entries, a long list for a UI widget I suppose. Bucketing on the top-level media type collapses it to a small fixed set of groups.

2-towns: > Main thing here is complexity. This turns /space from an O(1) read of quotaUsedBytes into an O(D) scan over every dataset: listBlocks(Manifest) walks the full manifest keyspace, and each dataset costs a getBlock + protobuf decode + two hasBlock calls. The first/last block check avoids O(total blocks) but not O(datasets), and a node is meant to hold many of these. Since it backs the UI it'll likely be polled, which makes it O(D) per request with no cache. I'd suggest splitting it off /space into its own endpoint, e.g. /space/usage-by-type, so the hot O(1) path stays untouched and the breakdown is opt-in.

Yes I agree I will do it. I was hesitating to do that actually... 

> It also reports the wrong number for occupied storage. It sums the logical size, but blocks are deduplicated and refcounted, so shared datasets overcount and the total won't match real occupied storage (quotaUsedBytes).

Yes but it depends how do you see it. Because `usage` could be be the data representation that you have on your local storage. Yes with de-duplication optimisation, it is not the exact real bytes stored, but still, it gives you a representation of the bytes that a file should carry. 

> Given it's inaccurate, it's worth questioning whether this breakdown really belongs on the storage node at all. If the UI can derive it from data it already fetches, that keeps /space untouched; if it does need to be served by the endpoint, a per-type usage counter updated at the dataset lifecycle — upload/download/delete, where the mimetype is actually known — would make it an O(1) read instead of a per-request scan.

I am more than open if you have a simpler solution to suggest. Can you explain more what you are suggesting please ? "a per-type usage counter updated at the dataset lifecycle — upload/download/delete, where the mimetype is actually known — would make it an O(1) read instead of a per-request scan."  ? 

> One more thing on datasetUsageByType: it returns a Table[string, DatasetTypeUsage] keyed directly on the raw manifest.mimetype. That's validated against Nim's known MIME set so it's bounded, but still potentially a few hundred entries, a long list for a UI widget I suppose. Bucketing on the top-level media type collapses it to a small fixed set of groups.

That's not an issue. The UI is not displaying the raw data, it creates categories based on the mimetype.

cnanakos: > > Main thing here is complexity. This turns /space from an O(1) read of quotaUsedBytes into an O(D) scan over every dataset: listBlocks(Manifest) walks the full manifest keyspace, and each dataset costs a getBlock + protobuf decode + two hasBlock calls. The first/last block check avoids O(total blocks) but not O(datasets), and a node is meant to hold many of these. Since it backs the UI it'll likely be polled, which makes it O(D) per request with no cache. I'd suggest splitting it off /space into its own endpoint, e.g. /space/usage-by-type, so the hot O(1) path stays untouched and the breakdown is opt-in.
> 
> Yes I agree I will do it. I was hesitating to do that actually...
> 
> > It also reports the wrong number for occupied storage. It sums the logical size, but blocks are deduplicated and refcounted, so shared datasets overcount and the total won't match real occupied storage (quotaUsedBytes).
> 
> Yes but it depends how do you see it. Because `usage` could be be the data representation that you have on your local storage. Yes with de-duplication optimisation, it is not the exact real bytes stored, but still, it gives you a representation of the bytes that a file should carry.

A user sees Total: 300MB and a per-type breakdown summing to 1.5GB for example. With no concept of deduplication, which they shouldn't need of course, parts larger than the whole just reads as broken, and it undermines in my opinion the entire application, including the numbers that are right.

Whatever "usage" means, a breakdown can't credibly exceed the total shown next to it.

> 
> > Given it's inaccurate, it's worth questioning whether this breakdown really belongs on the storage node at all. If the UI can derive it from data it already fetches, that keeps /space untouched; if it does need to be served by the endpoint, a per-type usage counter updated at the dataset lifecycle — upload/download/delete, where the mimetype is actually known — would make it an O(1) read instead of a per-request scan.
> 
> I am more than open if you have a simpler solution to suggest. Can you explain more what you are suggesting please ? "a per-type usage counter updated at the dataset lifecycle — upload/download/delete, where the mimetype is actually known — would make it an O(1) read instead of a per-request scan." ?

The clean fix is a small persisted map alongside the existing counter like `quotaUsedBytes`, updated at the dataset lifecycle, add on complete and subtract on delete. By the time it will accumulate the real stored bytes at those points it fixes the total vs per-type mismatch.

The downside is the moving parts as I can see, you'd have to hook every path that adds or removes a dataset or a block so the counter can't drift, plus a one time backfill for existing datasets if any.

> 
> > One more thing on datasetUsageByType: it returns a Table[string, DatasetTypeUsage] keyed directly on the raw manifest.mimetype. That's validated against Nim's known MIME set so it's bounded, but still potentially a few hundred entries, a long list for a UI widget I suppose. Bucketing on the top-level media type collapses it to a small fixed set of groups.
> 
> That's not an issue. The UI is not displaying the raw data, it creates categories based on the mimetype.

Fair enough if the UI builds categories, the raw list isn't a display problem. The only thing I'd note is that the node still constructs and sends the full per-mimetype table, so bucketing on the top-level type server-side would trim the response and the transient map to fixed entries, and keep the collapse in one canonical place instead of each client re-implementing it.

2-towns: > A user sees Total: 300MB and a per-type breakdown summing to 1.5GB for example. With no concept of deduplication, which they shouldn't need of course, parts larger than the whole just reads as broken, and it undermines in my opinion the entire application, including the numbers that are right.

> Whatever "usage" means, a breakdown can't credibly exceed the total shown next to it.

Again, it depends how it is presented to the user. I will have a meeting with Aqeel to discuss about this issue and he is very sensible to the aspect of "how to present something that make sense to the user" btw. But for example, 300 MB could be a "optimised version of the disk" or whatever reflected on the UI, not necessary the raw space taken by the datasets.

2-towns: > and it undermines in my opinion the entire application

If we present it as an optimisation processed by Storage, it becomes the opposite, a cool feature.

2-towns:
I am closing this as it was an experiment to match features provide by Aqeel on UI but this became a low low priority.

PR #1491, opened 2026-07-28
chore: bump nixpkgs to 25.11

igor-sirotin:
Pins nixpkgs to a commit from the 25.11 release — the same commit logos-delivery uses.

Needed for the status-go upgrade to Go 1.26 (status-im/status-go#7651): its flake moves to nixpkgs 25.11, and this lets `logos-storage-nim` follow the parent nixpkgs again instead of carrying its own.

Fallout from 24.11 → 25.11: `nodejs_18` and `clang16Stdenv` no longer exist, replaced with `nodejs_20` and `clang18Stdenv`.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -2,17 +2,17 @@
   "nodes": {
     "nixpkgs": {
       "locked": {
-        "lastModified": 1751274312,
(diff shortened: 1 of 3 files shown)


gmega (commented):
LGTM, looks harmless. Conventional commit linter is complaining, I suspect might be because this coming from an external fork.

gmega (approved):

PR #1493, opened 2026-07-29
fix: re-advertise provider records when addresses change

cnanakos:


storage/discovery.nim
```diff
@@ -222,6 +223,7 @@ proc announceDirectAddrs*(
 ) =
   # UDP addresses are derived from TCP addresses by remapping protocol and port.
   let tcpAddrs = @providerAddrs
+  let addrsChanged = tcpAddrs.len > 0 and tcpAddrs != d.providerAddrs
```
gmega: This will trigger even if you have a subset of the addresses, or if the order changes, but OK.

gmega (approved):
Fine.

storage/discovery.nim
cnanakos: Good catch, let's sort them then.

PR #1494, opened 2026-07-29
fix: crash when config is invalid

2-towns:
This PR ensures that `libstorage` doesn't crash when we provide an invalid config.

The `catch Exception` was removed from `tests/integration/storageconfig.nim` because `load` seems to raise only `ConfigurationError`.

`quitOnFailure` is an option coming from `nim-confutils` : 

```nim
  template fail(args: varargs[untyped]): untyped =
    if quitOnFailure:
      errorOutput args
      errorOutput "\p"
      suggestCallingHelp()
    else:
      # TODO: populate this string
      raise newException(ConfigurationError, "")
```

### DIFF
--- a/library/storage_thread_requests/requests/node_lifecycle_request.nim
+++ b/library/storage_thread_requests/requests/node_lifecycle_request.nim
@@ -113,7 +113,8 @@ proc createStorage(
       ,
     )
   except ConfigurationError as e:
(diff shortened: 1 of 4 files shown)


gmega (approved):
LGTM, quitting indeed doesn't seem like a good default for a library, particularly since Logos will interpret the non-zero exit value as a crash.

PR #1495, opened 2026-07-29
chore: raise instead of quitting on an invalid config value

2-towns:
This PR raise an error instead of using `QuitFailure` when reading the configuration variables. 

This seems to be a better practise.


gmega (dismissed):
Contract in [nim-confutil](https://github.com/status-im/nim-confutils#custom-argument-parsing) says we should raise `ConfigurationError` on problems, so that's probably what we should do.

gmega:
OK so I've modified this so that `parseCmdArg` raises ConfigurationError (as per [the nim-confutil docs](https://github.com/status-im/nim-confutils#custom-argument-parsing)) and readers use `raiseTomlErr` so that input context (line, column) is kept on errors.

PR #1497, opened 2026-07-31
fix: add guards to avoid crash when node is not created

2-towns:
This PR adds multiple guards to avoid crashes when the node is not created.

### DIFF
--- a/library/storage_thread_requests/requests/node_download_request.nim
+++ b/library/storage_thread_requests/requests/node_download_request.nim
@@ -299,6 +299,11 @@ proc process*(
   defer:
     destroyShared(self)
 
+  if storage[].isNil:
+    const msg = "Failed to process: the node is not created."
+    error msg
+    return err(msg)
+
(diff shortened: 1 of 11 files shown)


library/storage_thread_requests/requests/node_download_request.nim
```diff
@@ -299,6 +299,11 @@ proc process*(
   defer:
     destroyShared(self)
 
+  if storage[].isNil:
+    const msg = "Failed to process: the node is not created."
+    error msg
+    return err(msg)
```
gmega, line 305: Guess you could've had a template for this, e.g.:

```nim
template ensure_created(node: ptr StorageServer, msg: const str = "Failed to process: the node is not created.") =
  if storage[].nil:
    error msg
    return err(msg)
```

Then you do:

```nim
ensure_created(storage)
```

gmega (approved):
LGTM, best to be defensive.

PR #1498, opened 2026-07-31
fix: mix setup with nat traversal

2-towns:
This PR introduces a placeholder for `mixAddr` under `nat:auto`, and a guard in `lookupProviders`: as long as `mixAddr` is the placeholder, Mix is not ready and the lookup fails.

It relies on `setLocalMultiAddr` introduced by `libp2p-mix `so libp2p had to be bumped to `2.2.0`. 

It also restrict `pickMixCompatibleMultiAddr` for IPV4 and allow circuit relay. 

Blocked by https://github.com/logos-storage/logos-storage-nim-dht/pull/116.

### DIFF
--- a/storage/dht_proxy/client.nim
+++ b/storage/dht_proxy/client.nim
@@ -62,6 +62,13 @@ proc requestLookup(
   except CatchableError as exc:
     return failure("Client error: " & exc.msg)
 
+proc isReady(mixProto: MixProtocol): bool =
+  ## Returns true if the MixProtocol is ready to be used for lookups.
+  ## mixUnsetMultiAddr() is set in the setup for nat:auto, and restored whenever
+  ## the announce no longer carries a usable address. The Mix lookup cannot be
+  ## performed until Autonat provides a Reachable address or a Relay address.
+  return $mixProto.localMixPubInfo.multiAddr != $mixUnsetMultiAddr()
+
 proc lookupProviders*(
     mixProto: MixProtocol, proxy: PeerRecord, cid: Cid
 ): Future[?!seq[SignedPeerRecord]] {.async: (raises: [CancelledError]).} =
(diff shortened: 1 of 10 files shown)


cnanakos (approved):
LGTM

storage/dht_proxy/client.nim
```diff
@@ -62,6 +62,14 @@ proc requestLookup(
   except CatchableError as exc:
     return failure("Client error: " & exc.msg)
 
+proc isReady(mixProto: MixProtocol): bool =
+  ## Returns true if the MixProtocol is ready to be used for lookups.
+  ## mixUnsetMultiAddr() is set as a placeholder in the setup and updated whenever
+  ## the announce no longer carries a usable address.
+  ## The Mix lookup cannot be performed until Autonat provides a Reachable
+  ## address or a Relay address.
+  return $mixProto.localMixPubInfo.multiAddr != $mixUnsetMultiAddr()
```
gmega, line 71: OK two questions:

1. why use the string representation of the address, and not comparing them directly;
2. why is `mixUnsetMultiAddr` a proc instead of a constant?

gmega: OK on (2) I see now: globals and chronos don't play nice.

storage/discovery.nim
```diff
@@ -244,6 +264,8 @@ proc announceDirectAddrs*(
       .expect("Should construct signed record").some
     d.protocol.updateRecord(spr).expect("Should update SPR")
 
+  d.updateLocalMultiAddr()
```
gmega, line 267: This reads a bit deceiving, you're updating the Mix protocol address not the peer's local address. Maybe "updateMixProtoAddr"? And why read implicit state and not pass this as an argument?

tests/storage/utils/testmixidentity.nim
```diff
@@ -72,3 +73,64 @@ suite "mixidentity / loadRelayPubInfoTableFromJson":
...
+
+    check picked.isSome
+
+    if picked.isSome:
+      check $picked.get == $tcpAddr
+
+  test "skips an IPv6 address":
+    check pickMixCompatibleMultiAddr(@[ip6Addr]).isNone
+
+  test "picks a circuit-relay address":
```
gmega, line 130: OK good :-)

gmega (approved):
LGTM!

PR #1500, opened 2026-08-04
fix: disable port mapping to avoid crash

2-towns:


tests/integration/nat/pcp/testpcp.nim
```diff
@@ -13,6 +13,9 @@ proc announcesDirectAddr(info: JsonNode): bool =
   ## A reachable node announces at least one direct (non-circuit) address.
   info{"providerAddresses"}.getElems.anyIt("p2p-circuit" notin it.getStr)
 
+# Port mapping is disabled in storage/nat.nim
+const skipScenario = true
+
```
gmega, line 18: I guess you could've just commented out the whole suite but fine too

gmega (approved):
I think this is an OK tradeoff for now, LGTM.

PR #1507, opened 2026-08-10
chore: add integration/container test for Mix + NAT

gmega:
This PR adds a test for the Mix + NAT case. It also:

* refactors the node entrypoint script as a flexible configuration builder;
* tightened/justified the iptables rules in the router node.


tests/integration/nat/node-entrypoint.sh
```diff
@@ -1,30 +1,184 @@
...
+    return 1
+  fi
+
+  add_bootstrap_options "$bootstrap_api_url"
+
+  _config_opts+=(
+    --nat-num-peers-to-ask=1
+    --nat-max-queue-size=1
+    --nat-min-confidence=1.0
+    --nat-schedule-interval=30s
```
2-towns: There params can also accelerate the Autonat detection for public nodes when using `_opt_uses_identify`.

tests/integration/nat/router-common.sh
```diff
@@ -9,17 +9,80 @@ set -euo pipefail
...
+  -m conntrack --ctstate NEW,ESTABLISHED,RELATED \
+  -j ACCEPT
+
+# 3. For WAN-to-LAN traffic, we only allow packets that are part of an
+#    established connection - this is what allows us to receive traffic from
+#    the internet.
+iptables -A FORWARD\
+  -d "$LAN_SUBNET"\
+  -i "$wanif"\
+  -m conntrack --ctstate ESTABLISHED,RELATED \
```
2-towns, line 41: Should we add `DNAT` for UPnP and PCP ?

tests/integration/nat/nat-and-mix/testnatandmix.nim
```diff
@@ -0,0 +1,62 @@
...
+        info.announcesCircuitAddr(),
+    )
+    check eventuallyInfo(
+      leecherClient,
+      info{"nat"}{"reachability"}.getStr == "NotReachable" and
+        info.announcesCircuitAddr(),
+    )
+
+    # Both nodes route their DHT provider queries through Mix.
+    check eventuallyInfo(seederClient, info{"privateQueries"}.getBool)
```
2-towns, line 56: Okay here you check that `privateQueries` is enabled from the settings. I would add a check that the log contains `Selected mix node` to ensure that the Mix protocol is actually applied.

2-towns (approved):
LGTM. 

We could add an extra check by looking into the logs. 

And I think we will need `DNAT` rule for iptables rules with PCP and UPnP but both tests are currently skipped.

tests/integration/nat/nat-and-mix/testnatandmix.nim
gmega: Yes, that's actually a great idea. I wanted to do this, but didn't know there was support for it. I wouldn't have bothered adding the `privateQueries` flag to the debug endpoint if I knew, though I guess that ends up being useful anyhow.

tests/integration/nat/node-entrypoint.sh
gmega: Honestly I think we could use a lower check interval for all cases - nodes take forever to switch state. I wanted to ask you about this before changing it but forgot.

tests/integration/nat/router-common.sh
gmega: Yeah I haven't though too much about it as I thought minupnpd would handle the iptables rules itself. I'll test it once we re-enable things.

PR #1512, opened 2026-08-12
fix: replace peerInfo.addrs with the extip when using nat:extip

2-towns:
## Problem

Relay servers use `switch.peerInfo.addrs` (see [here](https://github.com/vacp2p/nim-libp2p/blob/0253e455f4da4230ac3101105bcc51b21c037121/libp2p/protocols/connectivity/relay/relay.nim#L87)) to create a reservation.
This is problematic when `listen-ip` is `0.0.0.0` and the public IP is not on any network interface.
Result: the relay server creates a reservation with private addresses, which will not be announced because of our public address filter.

## Solution

This PR fixes the issue by setting `peerInfo.announcedAddrs` when `extip` is used. [libp2p uses these addresses when the peerInfo is updated](https://github.com/vacp2p/nim-libp2p/blob/0253e455f4da4230ac3101105bcc51b21c037121/libp2p/peerinfo.nim#L90-L91).
Result: `switch.peerInfo.addrs` contains the public IP only, and the reservation is announced.

## Note

We could use `withAnnouncedAddresses` when creating the switch but that would require to set the `listen-port` with `nat:extip`.

### DIFF
--- a/storage/storage.nim
+++ b/storage/storage.nim
@@ -91,6 +91,29 @@ proc start*(s: StorageServer) {.async.} =
 
   await s.storageNode.switch.start()
 
+  var realPort = Port(0)
+
+  for listenAddr in s.storageNode.switch.peerInfo.listenAddrs:
+    let maybePort = getTcpPort(listenAddr)
+    if maybePort.isSome:
+      realPort = maybePort.get
+      break
+
+  if realPort == Port(0):
+    raise newException(StorageError, "Failed to determine the real TCP port")
+
+  if s.config.nat.hasExtIp:
+    # extip means that we assume the IP is reachable.
+    let extIpAddr = getMultiAddrWithIpAndTcpPort(s.config.nat.extIp, realPort)
+
+    # Feed switch.peerInfo.addrs with extIp value only.
+    # For example, we want to make sure that relay servers will
+    # contain only the extIp, no private addresses.
+    let peerInfo = s.storageNode.switch.peerInfo
+    peerInfo.announcedAddrs = @[extIpAddr]
+    # We force the update to take in consideration the announcedAddrs
+    await peerInfo.update()
+
   # Activate SO_REUSEPORT for hole punching in tcptransport.nim.
   # Without that, hole punching would use an ephemeral port assigned by the OS.
   # NotReachable has nothing to do with AutoNAT Reachability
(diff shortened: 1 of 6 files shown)


gmega:
On your questions:

> In the PR, peerInfo.update is done after the Mix setup. Is it an issue?

It shouldn't. Right after the update, you're calling `announceDirectAddrs` which calls `updateLocalMultiAddr`, this should be enough to set the correct address for Mix (which would only affect a relay anyway if they wanted to mint a SURB).

> What is the purpose of [IdentifyPusher set to false](https://github.com/logos-storage/logos-storage-nim/blob/14a1a2df29f69d6e6e7c3ae4dc497ab84412d754/storage/storage.nim#L339)?

Identify apparently [has a push counterpart](https://github.com/libp2p/specs/tree/master/identify#identifypush) which will actively broadcast changes to `PeerInfo` to connected peers which advertise support for it. The `withIdentifyPusher` was introduced to switchbuilder [in a recent PR](https://github.com/vacp2p/nim-libp2p/pull/2440/changes) as part of a refactor (previously you had to set a flag) but it defaulted to false, so `withIdentifyPusher(false)` is behavior-preserving (i.e. makes no difference for us).

2-towns:
> On your questions:
> 
> > In the PR, peerInfo.update is done after the Mix setup. Is it an issue?
> 
> It shouldn't. Right after the update, you're calling `announceDirectAddrs` which calls `updateLocalMultiAddr`, this should be enough to set the correct address for Mix (which would only affect a relay anyway if they wanted to mint a SURB).
> 
> > What is the purpose of [IdentifyPusher set to false](https://github.com/logos-storage/logos-storage-nim/blob/14a1a2df29f69d6e6e7c3ae4dc497ab84412d754/storage/storage.nim#L339)?
> 
> Identify apparently [has a push counterpart](https://github.com/libp2p/specs/tree/master/identify#identifypush) which will actively broadcast changes to `PeerInfo` to connected peers which advertise support for it. The `withIdentifyPusher` was introduced to switchbuilder [in a recent PR](https://github.com/vacp2p/nim-libp2p/pull/2440/changes) as part of a refactor (previously you had to set a flag) but it defaulted to false, so `withIdentifyPusher(false)` is behavior-preserving (i.e. makes no difference for us).

Okay thanks for all these information. 

I moved the block that updates the `peerInfo` on top, just by precaution, I think it is better to do it right after the switch is started,

gmega (approved):
LGTM

PR #1514, opened 2026-08-13
feat(openapi): dispatch dist tests workflow on openapi update

emizzle:
Add a workflow that sends a repository_dispatch event of type 'openapi-update' to logos-storage/logos-storage-nim-cs-dist-tests whenever openapi.yaml changes on master or a release is published. Also allows a manual run with a custom openapi_url.

The client_payload includes the URL of the openapi.yaml so the dist-tests repo can download the exact spec.

### DIFF
--- a/.github/workflows/openapi-update.yml
+++ b/.github/workflows/openapi-update.yml
@@ -1,0 +1,35 @@
+name: OpenAPI Update
+
+on:
+  push:
+    branches: [master]
+    paths: [openapi.yaml]
+  release:
+    types: [published]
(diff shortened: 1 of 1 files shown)


gmega:
Hm... the only potential problem I see here: this triggers on a release, which is probably too late to do the update (the workflow for the release is already spinning tests with the wrong version of the API at that point). Still, I suppose having the update run automatically and a PR created for you and then doing a re-run is better than having to do it all manually and then doing the re-run anyway.

PR #1515, opened 2026-08-13
chore: add libplum to static dependencies

2-towns:


gmega (approved):
Yes, probably a good idea...

PR #1516, opened 2026-08-15
feat: cross-compile libstorage for windows

dlipicar:
Summary updated (Arnaud): 

This PR introduces cross compilation for Windows using Nix. Several dependencies
were updated to achieve that:

1. `nim-boringssl`: fix path issue
2. `nim-leveldb`: fix path issue
3. `nim-libplum`: bump to a version that re-raises the exception and no longer needs the cmake build process
4. `nim-libbacktrace`: bump to a version that no longer needs the cmake build process

A smoke test has been added on CI to make sure the build process was done correctly.

`nixpkgs.url` and `nixpkgs-windows.url` now point to release branches. `26.05` brings Nim 2.2.4, so version 2.2.10 is overridden manually.

=======================

Adds `packages.x86_64-windows.libstorage` — a real PE, built on an ordinary Linux builder. **Rebased onto `v0.4.4`.**

`x86_64-windows` is a **pseudo-system**: nixpkgs has no native Windows stdenv, so the key means "cross-compiled to x86_64-w64-mingw32", and the derivations under it carry `system = x86_64-linux`.

**Native outputs are untouched.** Every change is gated on `hostPlatform.isWindows`, and platform tests now go through `hostPlatform` rather than `pkgs.stdenv.is*` — the latter answers for the *build* platform, so the existing `if pkgs.stdenv.isLinux then gcc13Stdenv` would have replaced the cross stdenv with a native one and silently retargeted the whole build.

## Approach

`USE_SYSTEM_NIM=1` with `pkgs.buildPackages.nim-2_2`, rather than bypassing the Makefile. nimbus-build-system's own bootstrap compiles the Nim *compiler* with `$CC`, which under cross produces a PE that cannot run on the builder. The nixpkgs wrapper bakes os/cpu into `nim.cfg`, so nimscript's `defined(windows)` is true and `build.nims` names the output `libstorage.dll` by itself. Note this means Nim 2.2.4, not the pinned v2.2.10.

A **second nixpkgs, scoped to the Windows target only**, matching the rev logos-nix pins for *its* Windows target. Not a compiler requirement — this repo's own pin (gcc 14.3) cross-builds the library just as well, and both were tried. The constraint is agreement with the consumer: `libstorage.dll` and the C++ plugin that loads it end up in one directory and both need `libstdc++-6.dll` / `libgcc_s_seh-1.dll` / `libwinpthread-1.dll` by **filename**. Two revs put two builds of each in contention, last staged wins. If logos-nix ever moves to this repo's rev, the input can be dropped.

**UCRT, not the legacy MSVCRT.** `library/libstorage.nim` allocates the strings it hands out with `<stdlib.h>` malloc and C++ consumers call `free()` on them; the two CRTs keep separate heaps, so a mismatch is heap corruption rather than a warning. UCRT64 is also what this repo's MSYS2 CI uses.

**The import library.** A Windows shared library is two artifacts: consumers *link* the import lib and *ship* the `.dll`. nim emits only the `.dll`, and `find_library` will not return a bare `.dll`, so a consumer silently falls through to `libstorage.a` and fails on every leveldb/setjmp symbol. Now emitted, and the install phase fails loudly if missing. The `.dll` installs to `bin/` and the import lib to `lib/`, because nixpkgs' `win-dll-link` hook only ever walks `$prefix/bin` — a `.dll` in `lib/` ships with no libgcc/libstdc++/libwinpthread beside it and fails to load at runtime with no diagnostic.

## Vendored defects, all invisible on MSYS2

1. **`os.parentDir` normalises to the target's DirSep** — `\` once `os=Windows` — and rewrites the whole path, not just the component being split. `nim-leveldbstatic/leveldbstatic/{raw,leveldbstatic}.nim` and `nim-boringssl/boringssl.nim` all derive paths that way. For boringssl, fixing `baseDir` is necessary but **not sufficient**: `/` re-mangles at each use, so the two `staticRead`s feeding nasm's prefix includes need rewriting too — that file already wraps its *other* joins in `normalizePath(dirSep = '/')`. The `currentSourcePath.rsplit({DirSep, AltSep}, 1)[0]` idiom is unaffected, and is what boringssl's `baseDir` becomes.

6. **`nim-leveldbstatic` runs cmake at compile time** asking for `-G"MSYS Makefiles"`. Correcting the generator is not enough: leveldb's CMakeLists then runs `try_compile`/`try_run` probes that would have to execute a PE on the builder. The step is vestigial anyway — its only output is `port/port_config.h` in the *build* dir, which no `{.passc: "-I"...}` line puts on the include path, so `port_stdcxx.h`'s `__has_include` never finds it on Linux or macOS either.

7. **nimbus-build-system's nat-libs targets branch on `$(OS)`** — the variable cmd.exe sets, empty on a Linux builder — so cross takes their POSIX branch: no `-D*_STATICLIB` (each archive then calls its own symbols through `__imp_` stubs), a pointless `-fPIC`, and `libminiupnpc.a` under `build/` when nim-nat-traversal's Windows branch wants it at the miniupnpc root. Built correctly in `preBuild`; `make deps` then finds everything newer than its sources and no-ops.

8. **`nim-libplum` (new in v0.4.4)** is configured by the Makefile with a bare `cmake -B build`, which takes the compiler from `$CC` but leaves `CMAKE_SYSTEM_NAME` as the *build* system. libplum gates both `add_definitions(-DWIN32_LEAN_AND_MEAN)` and its `target_link_libraries(... ws2_32 iphlpapi)` on CMake's `WIN32`, so neither applies and its example executable fails to link on `__imp_closesocket` (the archive itself is fine — a static lib records no imports). Configured here with the system name it should have had and with `PLUM_NO_EXAMPLE=ON`; CMake caches both, so the Makefile's own re-configure keeps them.

9. **`nim-libplum/libplum/plum.nim`** has a bare `raise` re-raising inside `except CancelledError`, in a proc declared `{.async: (raises: [CancelledError]).}`. Nim infers a bare re-raise conservatively and under `--os:windows` widens it to `Exception`, failing with `Error: Exception can raise an unlisted exception: Exception`. Naming the caught exception and re-raising it by name pins the type to the one already declared — identical behaviour, platform-neutral, and a candidate to send upstream. (Measured: not caused by the `-d:debug` that `USE_LIBBACKTRACE=0` implies — adding `-d:release` changes nothing.)

Plus `-lstdc++` (nim-leveldbstatic compiles LevelDB's C++ through `{.compile.}` but nim drives the link through gcc), `-lwinpthread` (winpthreads' `pthread_time.h` inlines `clock_gettime` as a call to `clock_gettime64`), and an `ar` PATH shim, since `--app:staticlib` shells out to a bare `ar`.

Each sed asserts it applied — a silently-unapplied patch otherwise resurfaces hours later as an unrelated "cannot open file" deep in the compile.

## Verification

Built and run on real Windows (10.0.26200, x86-64): `libstorage.dll` loads with `LOAD_WITH_ALTERED_SEARCH_PATH`, exports `storage_new` / `storage_version` / `storage_start` / `storage_destroy`, and `storage_version()` returns a version string.

End to end, this library built from **v0.4.4** drives `storage_module` + `storage_ui` 2.1.2 inside Logos Basecamp on that machine: the node starts, reports its 20 GB quota, and reaches 7 peers in good standing.

## Known limitation

`USE_LIBBACKTRACE=0` is upstream-coupled to `-d:debug` rather than `-d:release` (Makefile:125-129), so the Windows library carries runtime checks; `--opt:size` still applies. Decoupling means cross-building nim-libbacktrace's vendored autotools library, left for a follow-up.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -22,10 +22,64 @@ let
 
   revision = lib.substring 0 8 (src.rev or "dirty");
 
+  # `pkgs.stdenv.isLinux` answers for the BUILD platform, which is Linux even
+  # when cross-compiling to Windows -- so every platform test here goes through
+  # hostPlatform, and every if/elif chain tests Windows FIRST. Under mingw
+  # cross isDarwin and isAarch64 are both false and x86_64 still matches, so a
+  # Windows target otherwise takes the Linux branch in silence.
+  hostPlatform = pkgs.stdenv.hostPlatform;
+  isWindows = hostPlatform.isWindows;
+
   tools = callPackage ./tools.nix {};
 
-  # Pin GCC/CLang versions
-  stdenv = if pkgs.stdenv.isLinux then pkgs.gcc13Stdenv else pkgs.clang18Stdenv;
+  # Pin GCC/Clang versions -- but only where we own the toolchain. On a cross
+  # build pkgs.stdenv IS the mingw stdenv; replacing it with gcc13Stdenv would
+  # silently retarget the whole build at the build platform.
+  stdenv =
+    if isWindows then pkgs.stdenv
+    else if hostPlatform.isLinux then pkgs.gcc13Stdenv
+    else pkgs.clang18Stdenv;
+
+  # Win32 imports that the Nim runtime, chronos and the vendored C libraries
+  # need, plus two that are specific to this closure:
+  #   -lstdc++      nim-leveldbstatic compiles LevelDB's C++ sources through
+  #                 nim's {.compile.} pragma, but nim drives the LINK through
+  #                 gcc, not g++, so nothing else pulls in the C++ runtime or
+  #                 __gxx_personality_seh0.
+  #   -lwinpthread  winpthreads' pthread_time.h inlines clock_gettime as a call
+  #                 to clock_gettime64, which lives in libwinpthread; having the
+  #                 headers on the include path is not enough.
+  # Passed as separate --passL: flags rather than one quoted string so nothing
+  # has to survive make's word splitting on the way to the nim command line.
+  windowsLinkFlags = [
+    "-lws2_32" "-lbcrypt" "-liphlpapi" "-luserenv" "-lntdll" "-ldbghelp"
+    "-lwinpthread" "-lstdc++"
+  ];
+
+  # A Windows shared library is TWO artifacts: consumers LINK against the import
+  # library and SHIP the .dll. nim emits only the .dll, and CMake's find_library
+  # will not return a bare .dll -- so a consumer silently falls through to
+  # libstorage.a and tries to link the whole Nim runtime statically, which then
+  # fails on every leveldb/setjmp symbol. Emitting the import lib is what makes
+  # `-lstorage` mean the DLL.
+  windowsNimFlags =
+    map (f: "--passL:${f}") windowsLinkFlags
+    ++ [ "--passL:-Wl,--out-implib,build/libstorage.dll.a" ];
+
+  # Windows splits a shared library in two: the import half is a link-time
+  # artifact and belongs in lib/, but the .dll is a RUNTIME artifact and belongs
+  # in bin/ -- CMake's own RUNTIME destination, and what openssl and every
+  # autotools port in this closure already do. Following it is not cosmetic:
+  # nixpkgs' win-dll-link hook stages a PE's dependency DLLs automatically, but
+  # its fixup only ever walks $prefix/bin, so a .dll in lib/ ships with none of
+  # libgcc_s_seh-1 / libstdc++-6 / libwinpthread-1 beside it and fails to load
+  # on Windows with no diagnostic.
+  dllDir = if isWindows then "bin" else "lib";
+
+  libExt =
+    if isWindows then "dll"
+    else if hostPlatform.isDarwin then "dylib"
+    else "so";
 
 in stdenv.mkDerivation rec {
   pname = "storage";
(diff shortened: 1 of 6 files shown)


dlipicar:
### For whoever owns `nim-libplum`: item 5 belongs upstream, not here

Of the vendored patches in this PR, the `plum.nim` one is the only change that is **correct regardless of target** — Windows just happens to be where the compiler notices. Carrying it as a `sed` in this repo's `nix/default.nix` is a workaround; the two-line edit would be better in nim-libplum itself, after which the sed here can be deleted.

**File:** `libplum/plum.nim`, in the proc declared at ~L246 as `{.async: (raises: [CancelledError]).}`

```diff
-  except CancelledError:
+  except CancelledError as exc:
     # The chronos thread was cancelled.
     # ...
     await destroyAndReclaim(id, handle)
-    raise
+    raise exc
```

**Why it fails only under `--os:windows`:** Nim infers a *bare* `raise` conservatively, and on this target that inference widens to `Exception`, so the proc no longer satisfies its own `raises: [CancelledError]`:

```
libplum/plum.nim(283, 5) Error: Exception can raise an unlisted exception: Exception
```

Naming the caught exception and re-raising it by name pins the type to the one already declared. Semantics are unchanged — `raise exc` inside an `except` branch re-raises the same object — and nothing about the edit is platform-specific.

**Two things already ruled out, so nobody repeats them:**

- **Not the build mode.** The Windows build here runs with `-d:debug` (upstream couples that to `USE_LIBBACKTRACE=0`). Adding `-d:release` changes nothing — measured, same error, same line.
- **Not a downstream misuse.** The failure is inside libplum's own translation unit; the only thing this repo does differently is target Windows.

Happy to open the PR against nim-libplum if that is easier — say the word. Once it lands and the submodule is bumped, the corresponding `sed` block in `nix/default.nix` here should go away (its `grep` guard will fail loudly if the pattern disappears from under it, so it cannot rot silently).

nix/default.nix
```diff
@@ -21,11 +21,38 @@ let
...
   tools = callPackage ./tools.nix {};
 
-  # Pin GCC/CLang versions
-  stdenv = if pkgs.stdenv.isLinux then pkgs.gcc13Stdenv else pkgs.clang18Stdenv;
+  # Determine the compiler.
+  # pkgs.stdenv is the mingw compiler for windows.
+  stdenv =
+    if isWindows then pkgs.stdenv
+    else if hostPlatform.isLinux then pkgs.gcc13Stdenv
+    else pkgs.clang18Stdenv;
```
gmega, line 34: Hm... is "else" MacOS? Or is it really "everything else"? If it's the former, we should probably state it, and error out on the else clause.

nix/default.nix
```diff
@@ -38,57 +65,111 @@ in stdenv.mkDerivation rec {
...
     fakeGit
-  ] ++ lib.optionals stdenv.isLinux [
+  ] ++ lib.optionals hostPlatform.isLinux [
     lsb-release
-  ] ++ lib.optionals stdenv.isDarwin [
+  ] ++ lib.optionals hostPlatform.isDarwin [
     darwin.cctools
+  ] ++ lib.optionals isWindows [
+    # Paired with USE_SYSTEM_NIM=1 below: nimbus-build-system would
+    # otherwise build the Nim compiler itself as a Windows binary.
```
gmega, line 87: `isWindows` means we're building on windows right? Why wouldn't we want the Nim binary to be a windows binary?

nix/default.nix
```diff
@@ -38,57 +65,111 @@ in stdenv.mkDerivation rec {
...
+    [ "-d:disableMarchNative" "-d:git_revision_override=${revision}" ]
+    ++ lib.optionals isWindows windowsNimFlags
+  );
 
   makeFlags = targets ++ [
     "V=${toString verbosity}"
     "QUICK_AND_DIRTY_COMPILER=${if quickAndDirty then "1" else "0"}"
     "QUICK_AND_DIRTY_NIMBLE=${if quickAndDirty then "1" else "0"}"
+  ] ++ lib.optionals isWindows [
+    "USE_SYSTEM_NIM=1"
```
gmega, line 106: OK I suppose because we're using a nix package for Nim

nix/default.nix
```diff
@@ -38,57 +65,111 @@ in stdenv.mkDerivation rec {
...
     pushd vendor/nimbus-build-system/vendor/Nim
     mkdir dist
     cp -r ${callPackage ./nimble.nix {}}    dist/nimble
     cp -r ${callPackage ./checksums.nix {}} dist/checksums
     cp -r ${callPackage ./csources.nix {}}  csources_v3
     chmod 777 -R dist/nimble csources_v3
     popd
+  '' + lib.optionalString isWindows ''
+    # For --app:staticlib nim runs a hardcoded `ar` (extccomp.nim:86), with no
+    # config key to override it, and a cross stdenv only has $AR.
```
gmega, line 137: WTF is ar? 😂

gmega (approved):

PR #1519, opened 2026-08-28
feat(library): generate the C bindings with nim-ffi

gmelodie:
This PR replaces the hand-written FFI runtime in `library/` (worker thread, request objects, `libstorage.h`) with nim-ffi. Each entry point is now an async `{.ffi.}` proc, and the build generates the C header from the Nim code, so the header cannot drift. The ABI changes: `storage_close` is gone, events replace `RET_PROGRESS`, and `storage_version` takes no context. The Go and Rust wrappers need a regeneration.


gmega:
Nice, I'm onto this next @gmelodie

PR #1520, opened 2026-09-02
feat: expose api enabled in the config

2-towns:
This PR uses `ApiBindAddress` to detect if the HTTP API is enabled instead of relying on a configuration flag.

### DIFF
--- a/storage/storage.nim
+++ b/storage/storage.nim
@@ -578,7 +578,7 @@ proc new*(
   # REST server
   var restServer: RestServerRef = nil
 
-  if config.apiEnabled and config.apiBindAddress.isSome:
+  if config.apiBindAddress.isSome:
     restServer = RestServerRef
       .new(
         storageNode.initRestApi(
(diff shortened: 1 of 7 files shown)


storage/storage.nim
```diff
@@ -578,7 +578,7 @@ proc new*(
   # REST server
   var restServer: RestServerRef = nil
 
-  if config.apiBindAddress.isSome:
+  if config.apiEnabled and config.apiBindAddress.isSome:
```
gmega: Huh... now that I realize it, we could've just used the absence of apiBindAddress as the indicator to not start the API, and make that option hidden. If we decide to keep this though, I think passing `apiEnabled` and not passing `apiBindAddress` should be an error.

gmega (approved):
LGTM, though I realize now it'd have probably been easier to just use `apiBindAddress` being set the enablement flag.

2-towns:
> LGTM, though I realize now it'd have probably been easier to just use `apiBindAddress` being set the enablement flag.

Yes you're right I updated the PR.

PR #1522, opened 2026-09-03
feat: migrate discovery from discv5 to the libp2p Kademlia DHT

cnanakos:


storage/blockexchange/engine/advertiser.nim
```diff
@@ -68,17 +70,26 @@ proc advertiseBlock(b: Advertiser, cid: Cid) {.async: (raises: [CancelledError])
   except CatchableError as e:
     error "failed to advertise block", cid, error = e.msgDetail
 
+proc reachable(b: Advertiser): bool =
+  b.peerInfo.addrs.len > 0
```
gmega, line 74: Hm... are we sure this will NEVER contain temporary addresses e.g. during node startup? I need to make sure...

storage/rest/api.nim
```diff
@@ -489,14 +488,17 @@ proc initNodeApi(node: StorageNodeRef, conf: StorageConf, router: var RestRouter
...
 
       if $preferredContentType().get() == "text/plain":
         return RestApiResponse.response(
-          spr.toURI, contentType = "text/plain", headers = headers
+          spr.get(), contentType = "text/plain", headers = headers
         )
       else:
         return RestApiResponse.response(
-          $ %*{"spr": spr.toURI}, contentType = "application/json", headers = headers
+          $ %*{"spr": spr.get()}, contentType = "application/json", headers = headers
```
gmega:
```suggestion
      let spr = node.discovery.getSpr().valueOr:
        trace "Unable to build the signed peer record", err = error
        return RestApiResponse.error(Http500, headers = headers)

      if $preferredContentType().get() == "text/plain":
        return RestApiResponse.response(
          spr, contentType = "text/plain", headers = headers
        )
      else:
        return RestApiResponse.response(
          $ %*{"spr": spr}, contentType = "application/json", headers = headers
        )
```

storage/utils/spr.nim
```diff
@@ -0,0 +1,52 @@
...
+
+proc parse*(T: type SignedPeerRecord, spr: string): ?!SignedPeerRecord =
+  if not spr.startsWith(SprPrefix):
+    return failure("Missing '" & SprPrefix & "' prefix")
+
+  let bytes = Base64Url.decode(spr[SprPrefix.len .. ^1]).catch
+  if bytes.isErr:
+    return failure("Invalid base64 in signed peer record: " & bytes.error.msg)
+
+  let record = SignedPeerRecord.decode(bytes.get)
```
gmega, line 45: OK I checked and this does seem to validate the signature on decode, so great!

storage/discovery.nim
```diff
@@ -39,53 +37,63 @@ export discv5
...
 
-proc toNodeId*(host: ca.Address): NodeId =
-  ## Eth address to discovery id
-  ##
+    records.add(PeerRecord.init(peerId, provider.addrs, seqNo = 0))
+  records
 
-  readUintBE[256](keccak256.digest(host.toArray).data)
+proc storedPeerRecord(d: Discovery, peerId: PeerId): ?PeerRecord =
+  let sprBook = d.switch.peerStore[SPRBook]
```
gmega, line 72: SPRBook? Why do we have this? And why are we using SPRs here?

storage/discovery.nim
```diff
@@ -39,53 +37,63 @@ export discv5
...
+      return PeerRecord.none
 
-    return
-      if node.isSome():
-        node.get().record.data.some
-      else:
-        PeerRecord.none
+    let stored = d.storedPeerRecord(info.peerId)
+    if stored.isSome:
+      return stored
```
gmega, line 93: Not sure I understand why we use this...

storage/storage.nim
```diff
@@ -131,7 +130,7 @@ proc start*(s: StorageServer) {.async.} =
           StorageError, "Failed to load or generate Mix keys: " & error.msg
         )
       # We define a placeholder here.
-      # For nat:extip, the address will be updated when calling `announceDirectAddrs` below.
+      # For nat:extip, the address will be updated by `updateLocalMultiAddr` on node start.
```
gmega, line 133: I'm **guessing** that this will happen via your address observer because the switch was already started when we called peerInfo.update in line 121?

storage/storage.nim
```diff
@@ -177,42 +176,21 @@ proc start*(s: StorageServer) {.async.} =
...
-    # For other nodes, wait for AutoNat to announce addresses and update SPR.
-    # Announced here and not above because Mix needs discovery.mixProto to be set.
-    s.storageNode.discovery.announceDirectAddrs(
-      @[extIpAddr], udpPort = s.config.discoveryPort
-    )
-  else:
-    # Other nodes wait for AutoNAT to announce addresses and update SPR.
-    # They start in client mode to avoid polluting DHT with NotReachable records;
-    # it will be flipped off once AutoNAT confirms reachability.
-    s.storageNode.discovery.protocol.clientMode = true
```
gmega, line 203: Why did we need this before? Could we have used an address observer as you did?

gmega (approved):
I was a bit confused with the changes in the NAT update address handling, so I'd suggest @2-towns to double-check this. Otherwise looks good to me.

storage/nat.nim
```diff
@@ -184,21 +165,11 @@ method handleNatStatus*(
         await autoRelayService.stop(switch)
         debug "AutoRelayService stopped"
 
-      discovery.protocol.clientMode = false
-
-      # Here we don't rely on the port mapping because we consider
-      # that port mapped is the same as the discovery port.
-      # This can be wrong for PCP but it is an accepted limitation
-      discovery.announceDirectAddrs(@[dialBackAddr.get], udpPort = discoveryPort)
+      await discovery.setServerMode(isServer = true)
```
2-towns, line 168: I think it is okay but I want to say something.

`libp2p` has 2 different addresses related to the Identify protocol:

1. `guessDialableAddr`: observed IP + listen port
2. `getMostObservedProtosAndPorts`: observed IP + observed port

`dialBackAddr`, which is the `guessDialableAddr` in our case, was introduced to announce this address explicitly and not the observed IP and observed port. Removing it lets libp2p handle this announcement.

Finding this address is a bit hard in libp2p code because there is a combination of policies, mappers, etc.

Anyway there is an address mapper here (https://github.com/vacp2p/nim-libp2p/blob/391e403cc274ab8623fb1106f2c2827e09c6c89f/libp2p/protocols/connectivity/autonatv2/service.nim#L225-L237) which uses the `guessDialableAddr` when reachability is Reachable and the listen address is not public. This address mapper is used in the peerInfo update and should update the right address, in combination with `enableDialableCandidates`.

Trying to summarise:

Identify -> Register observed IP -> Autonat check -> Dial -> Callback -> Update peerInfo -> Apply address Mapper -> Get the most observed IP -> Apply address policy -> Notify observers

@gmega

storage/blockexchange/engine/advertiser.nim
cnanakos: Checked again and no temporary addresses get in. The switch's `dialableAddressPolicy` filters them out on every `peerInfo.update()`, including the first one at `switch.start()`.

storage/discovery.nim
cnanakos: `SPRBook` is libp2p's peer store not ours. 

We read it in `findPeer` trying to preserve the old behaviour. Under `discv5` the function returned a genuine signed record. Now Kad doesn't support this and `findPeer` returns `(peerId, addrs)` tuples, if libp2p already holds a signed peer record we use it, otherwise we are falling back and we build one synthesized from the results Kad returned with seqNo equal to 0. 

This matters for `/debug/peer` which serializes the seqNo.

storage/discovery.nim
cnanakos: Same reason as the `SPRBook` above. The policy is that Kad results are the fallback when we don't have one.

storage/storage.nim
cnanakos: No, not yet, I double checked. At line 121 the observer doesn't exist yet, it's registered later, in `StorageNodeRef.start()`, so for `extip` it's the direct call in node start, not the observer.

storage/storage.nim
cnanakos: The problem here is that an unreachable node must not act as a DHT server, so it started in client mode until AutoNAT says otherwise. The observer doesn't do this, it only reports address changes, and not whether someone can dial in.

Discv5 kept its own address state so pushing addresses in explicitly was the only way.

storage/rest/api.nim
cnanakos: Good catch, applied.

storage/blockexchange/engine/advertiser.nim
gmega: Hm... dialableAddressPolicy seems like a weak filter in some sense:

```nim
proc dialableAddressPolicy*(ma: MultiAddress): bool {.gcsafe, raises: [].} =
  # Use with switchBuilder.withAddressPolicy.
  # Filter the peerInfo.addrs updated by libp2p without
  # declaring another address mapper.
  if ma.isCircuitRelayMA():
    ma.hasPublicRelayTransport()
  else:
    ma.isPublicMA()
```

You might well have a public address which is not reachable, and your final address will be a circuit address. I know we discussed this today but dialableAddressPolicy alone won't guarantee this.

storage/discovery.nim
gmega: OK, I can't see what this is buying us though - maybe we have a stored SPR, maybe we return seqNo equals to zero and rely on an unsigned record instead. Also not clear to me when we'll have an SPR available, so I don't understand when which of those happen.

storage/discovery.nim
gmega: Yeah I think the part that bothers me is that I personally don't understand when we're expected to have one.

gmega:
OK some parts of this are still murky to me, but since both of you are saying it works I'm merging it.

PR #1523, opened 2026-09-04
feat: option for advertising content

2-towns:
### Downloading without announcing

A node that downloads a CID becomes a provider for it in the DHT. That is what spreads a package across the network. 

However, there are a lot of concerns regarding this usage in `logosctl` and the Basecamp app. While the first one contains a daemon that is supposed to be always running, the second one starts the Storage Module during the startup of the app, it doesn't seem long running enough to be a reliable provider. As soon as the CLI commands are done and the daemon is stopped or the app is closed, the node stops being a provider for the downloaded packages and the DHT will contain stale provider records which might seriously affect the network.

We need to introduce a new option for the Storage Module, which allows to opt in for announcing the downloaded packages to the DHT and becoming a provider for them. This option will be disabled by default for non storage nodes, and the user can enable it in the settings of the app or in `logosctl` if they want to contribute to the network.  

The node exposes it as `--advertise-content` (`advertiseContent` in the module JSON configuration), enabled by default so a storage node keeps announcing everything it holds. When it is disabled the node announces nothing: neither the packages it downloads nor the files it uploads. `logosctl` and the Basecamp app set it to `false` and offer a setting to turn it back on.

Note that this is particularly useful for anonymity over Mix, if someone wants to go for full privacy and not announce himself as provider. This has an impact on the network and decentralisation, we would prefer having more nodes participating in the network, but this seems a reasonable compromise for now.

### DIFF
--- a/storage/blockexchange/engine/advertiser.nim
+++ b/storage/blockexchange/engine/advertiser.nim
@@ -160,6 +160,7 @@ proc stop*(b: Advertiser) {.async: (raises: []).} =
   ##
 
   trace "Advertiser stop"
+
   if not b.advertiseContent:
     return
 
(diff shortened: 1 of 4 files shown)


cnanakos (approved):
LGTM.

PR #1524, opened 2026-09-07
Using new Mix Transport for the Block Exchange and Manifest protocols

marcinczenko:
TBD... we are experimenting....


storage/blockexchange/engine/activedownload.nim
```diff
@@ -363,15 +367,20 @@ proc updatePeerAvailability*(
...
+  ## whether admission succeeded. Existing complete peers need no query.
   let existingPeer = download.ctx.swarm.getPeer(peerId)
   if existingPeer.isSome:
     # peer already tracked, skip if bakComplete
     return existingPeer.get().availability.kind != bakComplete
 
-  discard download.ctx.swarm.addPeer(peerId, availability)
-  return true # new peer added, send WantHave
+  let admitted = download.ctx.swarm.addPeer(peerId, availability)
+  return queryPolicy == PresenceQueryPolicy.QuerySelectedPeers or admitted
```
gmega, line 383: Not sure I understand what information this is adding. It produces no behavioral change, and basically evaluates to either:

* `true or admitted`, if you set queryPolicy to `PresenceQueryPolicy.QuerySelectedPeers`, or;
* `false or admitted` if you set queryPolicy to `PresenceQueryPolicy.QueryAdmittedPeers`.

You might as well just return the value of `addPeer` and then decide at the caller what to do?

storage/blockexchange/engine/discovery.nim
```diff
@@ -32,16 +34,21 @@ const
...
 type DiscoveryEngine* = ref object of RootObj
   localStore*: BlockStore # Local block store for this instance
   peers*: PeerContextStore # Peer context store
-  network*: BlockExcNetwork # Network interface
+  networks*: BlockExcNetworks # Protocol instances available for provider dialing
   discovery*: Discovery # Discovery interface
   discEngineRunning*: bool # Indicates if discovery is running
   concurrentDiscReqs: int # Concurrent discovery requests
-  discoveryQueue*: AsyncQueue[Cid] # Discovery queue
+  discoveryQueue*: AsyncQueue[DiscoveryKey]
```
gmega, line 46: I honestly wonder why we need to keep this discovery queue nonsense and not just queue requests using plain chronos.

storage/blockexchange/engine/discovery.nim
```diff
@@ -50,25 +57,33 @@ proc discoveryTaskLoop(b: DiscoveryEngine) {.async: (raises: []).} =
...
           peers =? (await request).catch:
-        let dialed = await allFinished(peers.mapIt(b.network.dialPeer(it)))
+        let network = b.networks.networkFor(key.transport)
+        if network.isNil:
+          trace "Skipping providers because selected transport is unavailable",
+            transport = key.transport
+          continue
+        let dialed = await allFinished(peers.mapIt(network.dialPeer(it)))
+        if not b.onProviders.isNil:
+          b.onProviders(cid, key.transport, peers)
```
gmega, line 86: If we cut the discovery queue nonsense, we could for example get rid of those callbacks because you could ask for more peers and await directly. Not saying we should do that NOW, but I'm definitely keen on doing it soon.

storage/blockexchange/engine/discovery.nim
```diff
@@ -50,25 +57,33 @@ proc discoveryTaskLoop(b: DiscoveryEngine) {.async: (raises: []).} =
...
         storage_inflight_discovery.set(b.inFlightDiscReqs.len.int64)
 
       if (await request.withTimeout(DefaultDiscoveryTimeout)) and
           peers =? (await request).catch:
-        let dialed = await allFinished(peers.mapIt(b.network.dialPeer(it)))
+        let network = b.networks.networkFor(key.transport)
+        if network.isNil:
+          trace "Skipping providers because selected transport is unavailable",
+            transport = key.transport
+          continue
```
gmega, line 83: Caller should probably know that the transport they're asking isn't available - this is arguably a configuration error. But since this whole thing is unsynchronized, this just gets lost here and the caller will probably retry - again and again, producing useless traffic on the DHT - cause they can't tell this error from a lookup failure.

storage/blockexchange/engine/engine.nim
```diff
@@ -72,14 +74,18 @@ const
 type
   BlockExcEngine* = ref object of RootObj
     localStore*: BlockStore # Local block store for this instance
-    network*: BlockExcNetwork
+    networks*: BlockExcNetworks
     peers*: PeerContextStore # Peers we're currently actively exchanging with
+    mixPeers*: PeerContextStore
+    mixPeerTracker: PeerInFlightTracker
```
gmega, line 80: Hm... this is under the download manager for direct peers. Why did we decide to place the one for mix here?

storage/blockexchange/engine/engine.nim
```diff
@@ -72,14 +74,18 @@ const
 type
   BlockExcEngine* = ref object of RootObj
     localStore*: BlockStore # Local block store for this instance
-    network*: BlockExcNetwork
+    networks*: BlockExcNetworks
     peers*: PeerContextStore # Peers we're currently actively exchanging with
```
gmega, line 78: Well not really, that's just the set of peers we're connected to. DHT neighbors are probably ending up in this bucket at this point.

storage/blockexchange/engine/engine.nim
```diff
@@ -160,29 +181,29 @@ proc stop*(self: BlockExcEngine) {.async: (raises: []).} =
 
   trace "NetworkStore stopped"
 
-proc searchForNewPeers(self: BlockExcEngine, cid: Cid) =
-  if self.lastDiscRequest + DiscoveryRateLimit < Moment.now():
+proc searchForNewPeers(self: BlockExcEngine, cid: Cid, transport: DownloadTransport) =
+  if self.lastDiscRequest[transport] + DiscoveryRateLimit < Moment.now():
```
gmega, line 185: As far as I understand, we want to either:

1. rate limit DHT queries per download/swarm, or;
2. rate limit DHT queries globally, i.e., for the entire engine (which is what I guess we had).

Splitting this by transport doesn't seem to make sense for either case.

storage/blockexchange/engine/engine.nim
```diff
@@ -445,13 +466,21 @@ proc broadcastWantHave(
...
   for peerCtx in peers:
-    if not download.addPeerIfAbsent(peerCtx.id, BlockAvailability.unknown()):
-      # skip presence request for peer with Complete availability
+    if not download.addPeerIfAbsent(
+      peerCtx.id,
+      BlockAvailability.unknown(),
+      self.presenceQueryPolicies[download.ctx.transport],
+    ):
+      # Skip presence request for peer with Complete availability, or when
+      # QueryAdmittedPeers is selected and swarm admission failed.
```
gmega, line 479: If QueryAdmittedPeers is selected and swarm admission fails, this will evaluate to true (the condition is `queryPolicy == PresenceQueryPolicy.QuerySelectedPeers or admitted`), so in that case you will query. Which I guess makes more sense - "query selected peers" means you'll query regardless of admission.

storage/blockexchange/engine/engine.nim
```diff
@@ -911,33 +960,39 @@ proc blockPresenceHandler*(
...
                 BlockAvailability.fromRanges(presence.ranges)
               else:
                 BlockAvailability.unknown()
             of BlockPresenceType.DontHave:
+              trace "peer doesn't have anything", peer = peer, treeCid = treeCid
               BlockAvailability.unknown()
 
           downloadOpt.get().updatePeerAvailability(peer, availability)
 
           # try to propagate peer availability to other downloads for the same tree CID
```
gmega, line 981: We should not even bother supporting this IMHO. Allow one active download per CID and that's it.

storage/blockexchange/engine/engine.nim
```diff
@@ -1059,9 +1114,12 @@ proc wantListHandler*(
           )
 
     if presence.len > 0:
+      let network = self.networks.networkFor(transport)
+      if network.isNil:
```
gmega, line 1118: This is an error - it should never be the case that you ask to send a want list on a nonexisting transport. I'd go as far as saying that the right action here would be raising `Defect`, not just returning.

storage/blockexchange/engine/engine.nim
```diff
@@ -1090,11 +1150,19 @@ proc localLookup(
...
+    self: BlockExcEngine,
+    peer: PeerId,
+    blockRange: BlockRange,
+    transport: DownloadTransport = DownloadTransport.Direct,
 ): Future[WantBlocksResult[seq[BlockDeliveryView]]] {.
     async: (raises: [CancelledError])
 .} =
-  let response = ?await self.network.sendWantBlocksRequest(peer, blockRange)
+  let network = self.networks.networkFor(transport)
+  if network.isNil:
```
gmega, line 1161: This is better, but I'd argue this is an assertion failure - you should not be allowed to start a mix transport download if mix is not configured in the first place, let alone get to `requestWantBlocks`. Maybe `networkFor` can raise that assertion failure itself so you don't  even have to check for nil anymore.

storage/blockexchange/engine/engine.nim
```diff
@@ -1211,4 +1260,78 @@ proc new*(
...
+
+proc disableMixNetwork*(self: BlockExcEngine) {.async: (raises: []).} =
+  ## Remove BlockExchange's Mix instance after MixTransport has stopped, or
+  ## when startup fails. This does not stop the shared transport service.
+  let network = self.networks.mix
+  self.networks.mix = nil
+  if not network.isNil:
+    await network.stop()
+  self.mixPeers = PeerContextStore.new()
+  self.mixPeerTracker = PeerInFlightTracker.new()
```
gmega, line 1282: What's the case for supporting dynamically enabling/disabling Mix? This is not supported at the node level. I'd just pass MixTransport as an Option in the constructor.

storage/blockexchange/engine/engine.nim
```diff
@@ -1211,4 +1260,78 @@ proc new*(
...
+      mixPeerSelectionPolicy.needsProviderTracking:
+    discovery.onProviders = proc(
+        cid: Cid, transport: DownloadTransport, providers: seq[PeerRecord]
+    ) {.gcsafe, raises: [].} =
+      if not self.peerSelectionPolicies[transport].needsProviderTracking:
+        return
+      for downloads in self.downloadManager.downloads.values:
+        for download in downloads.values:
+          if download.manifestCid == cid and download.ctx.transport == transport:
+            download.ctx.providerPeers.clear()
```
gmega, line 1332: This is the weird one: we clear providers every time, so basically you're getting the result for your last DHT query here while ignoring all the peers you may already know. I need to dig deeper but as it is the policy that uses providers looks broken.

storage/blockexchange/engine/peerselection.nim
```diff
@@ -0,0 +1,68 @@
...
+method selectPresencePeers*(
+    policy: ProviderPriorityPolicy, peers: PeerContextStore, providers: HashSet[PeerId]
+): seq[PeerContext] {.gcsafe, raises: [].} =
+  for peer in peers:
+    if peer.id in providers:
+      result.add(peer)
+  if not policy.providersOnly:
+    for peer in peers:
+      if peer.id notin providers:
+        result.add(peer)
```
gmega, line 50: This is a weird policy. It says that it prioritizes providers, but:

1. in the absence of `providersOnly` it's just returning all peers in `PeerContextStore`;
2. when `providersOnly` is set, I believe we're only ever gonna scan the peers in our last DHT query (see how `providers` is maintained at the engine). This means everyone else remains stale, forever (unless they turn up in a DHT query).

Plus, the code looks also weird. :-) Suggestion:
```suggestion
  let selected = if policy.providersOnly:
    peers.filterIt(it.id in providers)
  else:
    peers.mapIt(it)

  selected.toSeq
```

storage/blockexchange/network/network.nim
```diff
@@ -202,13 +230,20 @@ proc getOrCreatePeer(self: BlockExcNetwork, peer: PeerId): NetworkPeer =
...
+        return nil
+      return stream
+    else:
+      try:
+        trace "Getting new connection stream", peer
+        return await self.switch.dial(peer, Codec)
+      except CancelledError as error:
+        raise error
+      except CatchableError as exc:
+        trace "Unable to connect to blockexc peer", exc = exc.msg
```
gmega, line 246:
```suggestion
  var getConn: ConnProvider = if self.isMixDownload:
    proc(): Future[Connection] {.async: (raises: [CancelledError]).} =
      trace "Opening block exchange stream via MixTransport", peer
      (await self.mixTransport.dial(peer, Codec)).valueOr:
        trace "Unable to open MixTransport block exchange stream", peer, error
        nil
  else:
    proc(): Future[Connection] {.async: (raises: [CancelledError]).} =
      try:
        trace "Getting new connection stream", peer
        await self.switch.dial(peer, Codec)
      except CancelledError as error:
        raise error
      except CatchableError as exc:
        trace "Unable to connect to blockexc peer", exc = exc.msg

File: storage/blockexchange/network/network.nim [hunk 16]
```diff
@@ -281,10 +360,37 @@ proc handlePeerDeparted*(
...
   if peer in self.excludedPeers:
     return
-  trace "Cleaning up departed peer", peer
-  self.peers.del(peer)
-  if not self.handlers.onPeerDeparted.isNil:
-    await self.handlers.onPeerDeparted(peer)
+  await self.unregisterPeer(peer)
+
+proc subscribeMixSessions(self: BlockExcNetwork) =
+  ## Use MixTransport sessions, rather than physical Switch connections, as
```
gmega, line 366: libp2p switch connections are about as physical as mix transport connections :-D

storage/blockexchange/network/network.nim
```diff
@@ -365,4 +474,32 @@ proc new*(
...
+    let network = if conn of TransportStream: self.mix else: self.direct
+    if network.isNil:
+      await conn.close()
+      return
+    await network.handleConnection(conn)
+
+  # Both transports use this one mounted codec and its incoming-stream quota.
+  self.dispatchProtocol = lp_protocol.new(
+    LPProtocol, @[Codec], dispatch, maxIncomingStreamsTotal = direct.maxInflight
+  )
```
gmega, line 499: OK I think I see how this works - this is what gets mounted on the switch, and it's also the protocol that MixTransport dispatches to.

storage/stores/networkstore.nim
```diff
@@ -31,6 +31,9 @@ logScope:
 type NetworkStore* = ref object of BlockStore
   engine*: BlockExcEngine # blockexc decision engine
   localStore*: BlockStore # local block store
+  downloadId: Option[uint64]
+    # A streaming download must wait on its own handles, not another download
+    # of the same tree that may use a different transport.
```
gmega, line 36: I don't understand this - we don't need to support multiple transports at the same time, or multiple concurrent downloads for the same CID with the same or different transports (after my PR, in fact, libstorage will blocks this at the API). It should be perfectly fine to identify a download by CID alone.

gmega (approved):
Disclaimer: I did not review the tests.

My main concerns were with the peer selection policies, but we're not using the ones that look funny to me, and a couple of places where we seem to be swallowing errors. This looks otherwise mostly behavior preserving, so I guess we're good to go.

storage/blockexchange/engine/activedownload.nim
marcinczenko: 1. `QuerySelectedPeers` - the default for both Direct and Mix
Query the selected /peer even if swarm admission fails—for example, because the swarm is full. This preserves the original behaviour.

2. `QueryAdmittedPeers` — an opt-in alternative
For a peer not already in the swarm, send a presence query only if admission succeeds. This avoids spending traffic  querying peers we could not admit.

As you see, previously we were just discarding the return value of `addPeer`, which means we would be querying the selected peer even if `addPeer` fails for it.

Agent found discarding the return value unacceptable. Then I asked to preserve the original behavior and move the new behavior to an optional `QueryAdmittedPeers` policy.

I think we should keep it and later review and decided what to do with all the policies.

storage/blockexchange/engine/discovery.nim
marcinczenko: Ok, let's then record it and get back to it later.

storage/blockexchange/engine/engine.nim
marcinczenko: Is your comment about the code comment next to `peers`? This PR did not add that.

storage/blockexchange/engine/engine.nim
marcinczenko: I agree. Unless we want to keep strict separation between discovery policies for Mix and Direct download, this is not  necessary. I will use shared `lastDiscRequest`. The idea was to keep strict separation in the first release as this is all under development... I hope we are not doing pre-mature optimization.

storage/blockexchange/engine/engine.nim
marcinczenko: Correct. On master we are querying regardless of admission. Thus the default policy is `QuerySelectedPeers`.

storage/blockexchange/engine/engine.nim
marcinczenko: Allowing multiple downloads per CID indeed makes thing more complex. We should review it together with deciding if we want to allow simultaneous direct and mix download (probably not, but let's talk about it maybe during off-site).

storage/blockexchange/engine/engine.nim
marcinczenko: I am not so sure about it. For direct downloads, pretty much yes, but for Mix we still have:

```nim
proc disableMixNetwork*(self: BlockExcEngine) {.async: (raises: []).} =
    let network = self.networks.mix
    self.networks.mix = nil
    if not network.isNil:
      await network.stop()
```

I think `network.stop()` will cancel and await the BlockExchange tasks, but then in `wantListHandler` we have:

```nim
let have =
  try:
    await address in self.localStore
  except CatchableError:
    false
```

and:

```nim
let have =
  try:
    await e.address in self.localStore
  except CatchableError:
    false
```

Both catch `CancelledError` as well, converting cancellation into `have = false`.

Instead we probably should do:

```nim
let have =
  try:
    await address in self.localStore
  except CancelledError:
    raise
  except CatchableError:
    false
```

and let cancellation reach the existing outer cancellation handler:

```nim
except CancelledError as exc:
    warn "Want list handling cancelled", error = exc.msg
```

Without this, I am afraid you would get some surprising crashes during shutdown....

I will try to implement this fix.

storage/blockexchange/engine/engine.nim
marcinczenko: We already reject starting a Mix download when Mix is unavailable:

```nim
if transport == DownloadTransport.Mix and not self.networks.isMixEnabled:
  return failure("Mix transport is not enabled")
```

So an ordinary download should not reach `requestWantBlocks` with Mix never configured.

So, here it seems we should be able to use assert here.

storage/stores/networkstore.nim
marcinczenko: This was already subject of another comment. I would not change this now as a last minute fix. Master already allows multiple concurrent downloads for the same CID. We should probably disallow both scenarios.

storage/blockexchange/network/network.nim
marcinczenko: Yes,  dispatch protocol is sort of a proxy.

storage/blockexchange/network/network.nim
marcinczenko: Kind of sweat distinction, isn't it?

storage/blockexchange/engine/engine.nim
marcinczenko: We are not supporting dynamic enable/disable of mix. `disableMixNetwork` is called only during shutdown and failed startup.

storage/blockexchange/engine/engine.nim
marcinczenko: Yes, but optional right now. Let's keep it for now so that the surrounding policy functionality stays in place, so that we can later experiment with different swarm selection strategies.

storage/blockexchange/engine/peerselection.nim
marcinczenko: I like your code more than the agent's. Now we know you are human...

storage/blockexchange/network/network.nim
marcinczenko: I think it is nice idea to use separate procs for Mix and Direct downloads. Will apply it.

storage/blockexchange/engine/discovery.nim
marcinczenko: I suppose we will get to the queuing problems in discovery on first occasion after we merge this PR.

storage/blockexchange/engine/engine.nim
marcinczenko: Actually, these awaits on localStore seem to be just returning complete futures. It does not have a single await in it:

```nim
method has*(self: LevelDbDatastore, key: Key):
    Future[?!bool] {.async: (raises: [CancelledError]).} =
  try:
    let str = self.db.get($key)
    return success(str.isSome)
  except LevelDbException as e:
    return failure("LevelDbDatastore.has exception: " & e.msg)
```

So the API is misleading and should probably be implemented as sync calls... Still, for overall cleanses it is better to handle these cancellations in case the implementation fulfills the contract one day accidentally.

storage/blockexchange/engine/peerselection.nim
marcinczenko: He he: `peers.mapIt(it)` fails compilation because PeerContextStore has no integer indexing. My agent would never suggest change that does not compile. Typical human behavior... And it brakes the tests....

marcinczenko: I will get closer to your style here, but I will maintain the current semantics for now. Otherwise the tests have to change as well....

storage/blockexchange/engine/activedownload.nim
gmega: Agreed on the return value, but the proc itself isn't implementing the query policy, so passing the argument to me looks like noise.  But agreed on discussing this further down the line.

storage/blockexchange/engine/engine.nim
gmega: Yeah sorry I ended up looking at the whole code. Not for you to worry about it.

storage/blockexchange/engine/peerselection.nim
gmega: Oh that detail had actually escaped me (the ordering of the list, I guess that's what prioritizing means in this context). Human 0 x 1 AI

storage/blockexchange/engine/engine.nim
gmega: OK, hadn't considered the second order effects.

storage/blockexchange/engine/engine.nim
gmega: Oh. I guess the fact that this was marked as public (and not locally called by shutdown/close only) misled me.

PR #1525, opened 2026-09-08
feat: mix transport multiaddress

gmega:
This PR adds a mix-transport multiaddress which allows mix public information to be published as a provider record. This allows one to use an arbitrary provider as a mix exit node when performing downloads over mix.


gmega:
I'm gonna add this to libp2p-mix-transport instead of here.

PR #1529, opened 2026-09-11
fix(advertiser): gate advertising on AutoNAT reachability

cnanakos:
`addrs.len > 0` tested address shape, not dialability, so a firewalled public IP passes `isPublicMA()`, and the node kept advertising an address nobody could reach.

The gate now requires AutoNAT to report `Reachable`, or a circuit through a public relay.

The AutoNAT status handler also wakes the advertiser, since `peerInfo.update()` notifies observers only when the address list changes and an `Unknown` to `Reachable` flip on an already-public address changes nothing.


gmega (approved):
Yeah, definitely better than before. The fact we can't work on this at the address mapper level though means that we might still publish unusable addresses if `peerInfo.addrs` happens to contain those - Kad will happily take those without informing us, and this filter won't protect us from it.

Ideally, Kad should allow us to specify a filter that is not the address mapper filter path. We'll need to talk to the libp2p guys about it.

PR #1530, opened 2026-09-16
fix: keep the log file writer alive across GC cycles

dlipicar:
## Problem

With `log-file` set, libstorage stops writing its log file after a large upload, and
`storage_stop` then crashes the process. It was found through logos-storage-module:
`uploadUrl` of a 1 GiB file, then `stop()`, took the module host down after about 5500
lines of `failed to write to log file errorCode=9`.

## Cause

`setupLogging` installs a closure that captures the log file handle. libstorage builds
with `--mm:refc`. Under refc, only the GC of the thread that ran `NimMain` marks
globals (`markGlobals` checks `gcThreadId == 0`). The closure's environment is
allocated on the storage thread and referenced only from chronicles' global
`outputs`, so that thread's first full collection frees it. The collection runs once
the thread holds 4 MiB, which takes about 8192 uploaded blocks (512 MiB at the default
block size).

The writer then reads freed memory. Lines are dropped silently, or written to whatever
fd number the reused memory holds. The failed write was reported with `error`, which
went through the same writer and recursed until the stack overflowed.

The `storage` binary is not affected, because it calls `setupLogging` on the main
thread.

## Fix

- The file writer reads the handle from a module-level `Option[IoHandle]`, a plain
  value the GC never frees. The `storage_enable_log_counter` writer had the same
  closure problem and is fixed the same way.
- A failed write is reported with chronicles' `logLoggingFailure` (stderr), not
  through the logger.

## Test

`tests/cbindings/storage.c` adds `check_log_file_after_many_blocks`. An offline node
with a log file uploads 16384 blocks (16 MiB of zeros in 1 KiB blocks, about 3 s),
then stops and closes. The log must still contain the lines written after the upload.

| `check_log_file_after_many_blocks` | master e3225940 | this branch |
|---|---|---|
| macOS arm64 | SIGBUS, 3/3 | pass, 3/3 |
| Linux x86_64 | SIGSEGV, 3/3 | pass, 3/3 |
| Linux x86_64, inside the nix sandbox | fails: lines missing | pass |
| Windows x86_64 (nix cross build, run on Windows) | access violation, 3/3 | pass, 3/3 |

The whole `tests/cbindings/storage.c` suite passes on macOS and Linux (25/25).

Also checked by hand, on macOS:
- Upload sizes from 512 MiB to 1 GiB at 64 KiB blocks, and 32 or 64 MiB at 4 KiB,
  crash or lose lines on master. On this branch every line is written.
- If the log fd is made unwritable right after init, master overflows the stack
  during `start`. This branch prints "Log message not delivered" to stderr and
  keeps running.
- A build with `-d:storage_enable_log_counter` compiles, numbers its lines, and
  survives a 1 GiB upload.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


dlipicar:
replaced by https://github.com/logos-storage/logos-storage-nim/pull/1536

PR #1531, opened 2026-09-18
refactor: drop discovery routing table reseed, Kad reseeds itself

cnanakos:


gmega (approved):
Deleting always feels good! :-D

PR #1532, opened 2026-09-18
feat: enable mix per-download

gmega:
This PR implements the plan [described here](https://hackmd.io/@codex-storage/r1TWMfsKfe); i.e., it:

* gets read of the global privacy toggle for DHT queries;
* bundles private downloads with private DHT queries, and exposes this as a single "private download" flag which must be set when the download session is created, and is defined on a per-download-session basis.


library/storage_thread_requests/requests/node_download_request.nim
```diff
@@ -78,7 +83,11 @@ proc destroyShared*(self: ptr NodeDownloadRequest) =
   deallocShared(self)
 
 proc init(
-    storage: ptr StorageServer, cCid: cstring = "", chunkSize: csize_t = 0, local: bool
+    storage: ptr StorageServer,
+    cCid: cstring = "",
+    chunkSize: csize_t = 0,
+    local: bool,
+    isPrivate: bool,
 ): Future[Result[string, string]] {.async: (raises: []).} =
```
marcinczenko, line 97: The comment block at the beginning of `init` is outdated.

library/storage_thread_requests/requests/node_download_request.nim
```diff
@@ -209,6 +222,7 @@ proc stream(
     cCid: cstring,
     chunkSize: csize_t,
     local: bool,
+    isPrivate: bool,
```
marcinczenko: Do we really need `isPrivate` on `stream`?  `stream` requires an existing session and reads its existing stream. It does not select a transport or create another download. Its `isPrivate` argument serves only as an assertion of what the caller expects. Is that the intention?

marcinczenko (approved):
Just one quick comment for now about passing `isPrivate` to `stream`. Will look at the rest in a moment...

tests/cbindings/storage.c
```diff
@@ -758,6 +758,95 @@ int check_download_manifest(void *storage_ctx, const char *cid)
...
+        return RET_ERR;
+    }
+    if (is_resp_ok(r, NULL) != RET_OK)
+    {
+        return RET_ERR;
+    }
+
+    // Reusing a direct session must not silently accept a private request.
+    r = alloc_resp();
+    if (storage_download_init(storage_ctx, cid, 0, false, true, (StorageCallback)callback, r) != RET_OK)
```
marcinczenko, line 826: I would maybe used named arguments, or at least use variables for the argument that is under test here.

library/storage_thread_requests/requests/node_download_request.nim
gmega: Hm, lemme double-check this

marcinczenko (approved):
The rest looks ok to me. I did not study every tiny details and I am sure copilot or other agent will find it. On high level I agree - only the previously mentioned a bit defensive `isPrivate` would be something I would double check, but otherwise, if it passes the tests and works, we can change anything later.

library/storage_thread_requests/requests/node_download_request.nim
gmega: Yeah I was misled because `local` was there. Makes no sense. Let's drop it.

